PR #2, opened 2022-11-02
Update and apply scalafmt

mdedetrich:
There is a poll on how to approach this here https://github.com/mdedetrich/akka-apache-project/discussions/26.

For original discussion please go to https://github.com/mdedetrich/akka-apache/pull/2

This PR updates and replaces scalafmt in the codebase since Akka happens to be using a really old version of .scalafmt and this is the best time to do it. Note due to the large number of changes that have happened in Scalafmt over time from what I can tell its essentially impossible to provide the exact same configuration that was in `2.1.0`.

In addition to just updating and applying Scalafmt, the PR also adds a github action which on every PR checks that Akka is formatted. The github action uses scalafmt-native which is ultra fast and can format a codebase in seconds (for one that is the size of akka it should be ~1 minute). Since its a separate action it also runs concurrently to the main build. The PR also has been split into 3 commits with the `Apply scalafmt` commit containing all of the changes. This means if there is feedback for updating the scalafmt config its easy to rebase the PR with the updates.

Below is a notable list of configurations/explanations.

* With all of the different dialects of Scala that have come out, the new version of Scalafmt allows you to apply different dialects to different source directories using globs
* New version of scalafmt applies newline breaks for implicits consistently where as the old scalafmt version did not.
* New version of scalafmt adds a single space before every comment if there isn't already one.
* Vertical align tokens have been added for `!`/`?` and `~>` since these are idiomatic operators. While `!` and `?` didn't have much effect in the codebase, `~>`  is used in akka-stream custom graphs and greatly improves the readability of said graphs.
* New version of scalafmt consistently removes paranthesis where as the old scalafmt version did not.
* Formatting rules for hexcodes have been applied so they are consistent across codebase
* Line endings are now enforced to be unix (removal of `line.endings = preserve` resets default to unix). This was changed at https://github.com/akka/akka/pull/31192 due to Windows line endings issues however since we are formatting entire codebase it shouldn't be a problem, @He-Pin can you confirm that this works under Windows? (i.e. if you run scalafmt on this branch in Windows that it doesn't mess with line endings). Note that almost all modern Windows editors/IDE's support Unix line endings (including Windows default notepad!)

In general, the intention behind the updating of Scalafmt is to reduce the diff as much as possible while doing some easy wins (i.e. akka operators such as `!` and consistent hex code formatting). Another approach to updating scalafmt would essentially be YOLO and use more strict/consistent scalafmt options that the community agrees with. For example options such as `newlines.source = keep` and `newlines.beforeMultiline = keep` are there to minimize the diff as much as possible  so if you remove them the codebase is more consistent however the diff ends up being much larger.

If a lot of people agree that we don't care that much about keeping the current codebase and would prefer a more consistent code style I can remove the `newlines` options mentioned earlier. Since we are dealing with style formatting here, I would implore to avoid bikeshedding as much as possible, be reasonable in your feedback! This is the only chance we have of doing such a change since the idea is to do this immediately after Pekko gets approved into Apache incubator (if it does so). The PR is a draft currently since there is no intention of merging it yet.

EDIT: Once the PR gets enough approvals we will add an ignore so that the formatting doesn't show up in git blame, see https://akrabat.com/ignoring-revisions-with-git-blame/. This will either be done in the same PR or in a following one.


akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -744,7 +744,7 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter)
           endpoint ! s
         case Some(Gated(timeOfRelease)) =>
           if (timeOfRelease.isOverdue()) createAndRegisterWritingEndpoint() ! s
-          else extendedSystem.deadLetters ! s
+          else extendedSystem.deadLetters                                   ! s
```
danischroeter, line 747: this alignment seems odd - is this on purpose?

mdedetrich: Yes, its a change which aligns the `!` operator which is used in actors.

nvollmar: I don't think this is an improvement. In such a case it would probably be better to manually reformat the if/else to separate lines to keep things more in line

mdedetrich: Yeah now that I see the effect I agree, will change it later. The only other operator that comes to mind which makes sense to align this way is `->` for setting values in `Map`'s

.scalafmt.conf
```diff
@@ -1,21 +1,52 @@
...
-spaces.inImportCurlyBraces = true
-rewrite.neverInfix.excludeFilters = [
+version                                  = 3.6.1
+runner.dialect                           = scala213
+project.git                              = true
+style                                    = defaultWithAlign
+docstrings.style                         = Asterisk
+docstrings.wrap                          = false
+indentOperator.preset                    = spray
+maxColumn                                = 120
```
nvollmar, line 8: With todays wide/ultra-wide monitors I think we could keep this at 180 (old setting didn't seem to apply that strictly)

mdedetrich: Do you feel strongly about this? From what I have seem most modern Scala OS code uses 120, not 180 and I would like to keep the formatting as standard as possible.

nvollmar: Not strongly, just saw some occurrence where the re-formatting split things into a new line I didn't necessarily think improved readability.

akka-actor-testkit-typed/src/test/scala/docs/akka/actor/testkit/typed/scaladsl/AsyncTestingExampleSpec.scala
```diff
@@ -27,7 +27,7 @@ import scala.util.Success
 import scala.util.Try
 
 object AsyncTestingExampleSpec {
-  //#under-test
+  // #under-test
```
nvollmar, line 30: those comments are referenced by the documentation to import code, need to check this doesn't break with the reformatting

.scalafmt.conf
alexandru: I'm not working on a ultra-wide monitor, I work on a 16-inch MacBook, and before that, I worked on a 13-inch MacBook. I also keep my font at 16, because when reading code, the presentation quality is better than quantity.  180 chars would not fit on any screen I worked on in the past 10 years 🙂

Also, I'd argue that having to move your eyes from left to right for 180 chars can be terrible for readability. The consensus is that continuous lines have to be under 80 chars for the best results, a claim somewhat corroborated by at least one study I saw. Of course, in programming we also have indentation, but if you have that much indentation to need more than 120 chars, maybe that code is a little too complex in terms of branching.

If you have a wide-monitor, a better use of that space is to open 2 files side by side (e.g., in diffs), which is also good for PR reviews.

Personally I'd go with 100. The only reasons for why I don't like that is because Scala's function signatures can get awkwardly long, and formatting those signatures becomes a chore. I'd wish, however, for Scalafmt to have different settings depending on context — e.g., leave the function signatures on one line, but anything after the `=` hard-wrap it at 100.

Going with something more than 120, for me, it's like not having any limits at all. Which is fine, but then I personally complain about code not being readable in code reviews, and we desperately want to avoid that. Which is the value that Scalafmt brings.

.scalafmt.conf
```diff
@@ -1,21 +1,52 @@
...
-align.openParenCallSite    = false
-optIn.breakChainOnFirstMethodDot = false
-optIn.configStyleArguments = false
-danglingParentheses = false
-spaces.inImportCurlyBraces = true
-rewrite.neverInfix.excludeFilters = [
+version                                  = 3.6.1
+runner.dialect                           = scala213
+project.git                              = true
+style                                    = defaultWithAlign
```
alexandru, line 4: I'm not very fond of vertical alignment, because it generates whitespace diffs, and because code can get awkward when the length of the values / identifiers differs.

It's safer if we don't do vertical alignment. Although, personally, I can live with it.

.scalafmt.conf
nivox: I second @alexandru point. 

My preferred coding layout is 2 pane side by side, so a maxColumn of 180 would not fit even on my widescreen monitor (in VSCode I get about 160 chars per pane). 

Furthermore my brain tends to parse better shorter multi-line expressions than longer single line ones, so personally I'd stick with 120.

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
```diff
@@ -26,7 +25,7 @@ private[akka] trait AbstractProps {
     if (Modifier.isAbstract(clazz.getModifiers)) {
       throw new IllegalArgumentException(s"Actor class [${clazz.getName}] must not be abstract")
     } else if (!classOf[Actor].isAssignableFrom(clazz) &&
-               !classOf[IndirectActorProducer].isAssignableFrom(clazz)) {
+      !classOf[IndirectActorProducer].isAssignableFrom(clazz)) {
```
nvollmar, line 28: Keeping if condition terms on new lines indented would make it more readable

akka-actor/src/main/scala/akka/actor/ActorCell.scala
```diff
@@ -521,8 +521,8 @@ private[akka] class ActorCell(
           case NoMessage                                                    => // only here to suppress warning
         }
       } catch handleNonFatalOrInterruptedException { e =>
-        handleInvokeFailure(Nil, e)
-      }
+          handleInvokeFailure(Nil, e)
+        }
```
nvollmar, line 525: Curious why the last bracket is indented

.scalafmt.conf
mdedetrich: > I'm not very fond of vertical alignment, because it generates whitespace diffs, and because code can get awkward when the length of the values / identifiers differs. We align <- in for-comprehensions at $work, and I often find myself renaming identifiers in order for the for expression to look good (that, or starting code blocks, to make the declarations multi-line).

> It's safer if we don't do vertical alignment. Although, personally, I can live with it.

So I am going to remove the `!` vertical alignment because it visually doesn't look well, but I would opt for leaving operators such as `~>` since it does greatly improve clarity (`~>` is used in streams with graphs and since we are dealing with graphs the alignment helps in seeing the port flow).

The point about whitespace and diff is a valid one however github now supports viewing diffs while ignoring whitespace which does help immensely https://github.blog/2018-05-01-ignore-white-space-in-code-review/.

Also note the other operators are there because the old scalafmt also had them so the goal here is to minimize the diff.

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
mdedetrich: Yeah I am not sure what is causing this, maybe the alignment of `!` is messing with it and removing it would help

akka-actor/src/main/scala/akka/actor/ActorCell.scala
mdedetrich: This is a result of the `binPack` features, specifically `binPack.indentCallSiteOnce`. This was a feature that was added just now in scalafmt to fix other diffs, see discussion at https://github.com/mdedetrich/akka-apache/pull/2#discussion_r1002684029

akka-actor-testkit-typed/src/test/scala/docs/akka/actor/testkit/typed/scaladsl/AsyncTestingExampleSpec.scala
mdedetrich: If you are talking about sbt-paradox, it doesn't break anything. I have done the same changes to other upstream Akka projects and sbt-paradox still worked fine.

.scalafmt.conf
mdedetrich: So judging by the conversation here I would just leave it at 120, this is what the community generally converges too.

akka-remote/src/main/scala/akka/remote/Remoting.scala
mdedetrich: I have just force pushed the PR with the `!` and `?` alignment operators removed.

.scalafmt.conf
spangaer: I've also switched from 120 to 100 over the past decade for side-by-side diff.

But if it was 120, probably no need to make the formatting diff go all crazy.

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
mdedetrich: So I just removed the  vertical alignment `!` operator however it didn't actually update this so it must be something else.

.scalafmt.conf
alexandru: > Also note the other operators are there because the old scalafmt also had them so the goal here is to minimize the diff.

Good argument 👍

mdedetrich: Cool I am going to resolve this, feel free to reopen + bump if there is something else.

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
alexandru: I rather prefer the change, because I don't like branches with fruits danging on the right side 🙂 but YMMV.

.scalafmt.conf
mdedetrich: Yeah setting it to 120 would explode the diff and honestly in a way where its no longer trivial to compare it visually. At least right now if there is a difference its easy to make out.

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
mdedetrich: In any case I honestly don't see how I can fix this, I mean I can try experimenting with scalafmt options but it takes a large amount of time and I think what we have now is a good approximation.

.scalafmt.conf
mdedetrich: On final thought I will resolve this conversation, changing the value from 120 is massively detrimental to the PR's goals of trying to minimize the diff.

If we are going to change this, it would be on the alternative approach where instead of minimizing the diff we instead just try to use a more standard scalafmt configuration accepting that the diff would be massive, i.e. https://github.com/mdedetrich/akka-apache/pull/2#issuecomment-1291044281 .

If people want to entertain this then we can do a github discussions poll (once its enabled) to gauge whether the community at large would prefer that and then one can always create a formal apache vote for it.

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
nvollmar: In that case I'm fine with it.

danischroeter (approved):

alexandru (approved):

pjfanning (approved):
LGTM

mdedetrich:
So I am thinking of letting this PR sit for a couple of days (lets say Friday night CEST) and if there aren't any objections then I will merge it

alexandru:
If you get the required number of approvals, just merge it 🤷‍♂️ since it's a cosmetic change, we can fix issues with it later 😛

... because it impacts many files, and other PRs may have a harder time with the merge, so might as well merge sooner rather than later.

mdedetrich:
> If you get the required number of approvals, just merge it 🤷‍♂️ since it's a cosmetic change, we can fix issues with it later 😛

> ... because it impacts many files, and other PRs may have a harder time with the merge, so might as well merge sooner rather than later.

Okay sure, I will merge it tonight then if there aren't any complaints.

mdedetrich:
So I just realized that we have problems whereby the scalafmt action is not being run due to the github action not being whitelisted. We are currently trying to fix this (see https://issues.apache.org/jira/browse/INFRA-23846), but it makes sense to merge this PR after the action triggers otherwise there is a risk of someone merging unformatted code.

If the infra ticket ends up taking too long then I can undo the `Add github workflow to check scala code is formatted` commit since the `build-test-prValidation.yml` is running right now.

mdedetrich:
So I decided to merge the PR as is seeing as PR's are starting to open up. I have dropped the commit which replaces the scalafmt check with github action, I will add it into a future PR once https://issues.apache.org/jira/browse/INFRA-23846 is ready

spangaer:
If you want it, I don't think adding it to one of the other sbt based runs, e.g. the test run will really have a dramatic impact.

PR #5, opened 2022-11-03
Update description of project

mdedetrich:
While not so relevant now, the historical direction of a lot of Scala based projects (which Pekko is one, the Java API is technically a separate layer, i.e. none of the core is implemented in Java) is that at some point they end up adding support for runtimes aside of JVM, i.e. graalvm/scala-native, even Scala.js (or anything else that props up).

At some point Pekko might explore such things, in which case running on the JVM isn't the most accurate description.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,7 +1,7 @@
 # https://cwiki.apache.org/confluence/display/INFRA/.asf.yaml+features+for+git+repositories
 
 github:
-  description: "Build highly concurrent, distributed, and resilient message-driven applications on the JVM"
+  description: "Build highly concurrent, distributed, and resilient message-driven applications using Java/Scala"
   homepage: https://pekko.apache.org/
   labels:
     - pekko


nvollmar (approved):

an-tex:
LGTM. On a side note, there's already a working port of akka on scala.js https://github.com/akka-js/akka.js (by @andreaTP who's also a committer here)

mdedetrich:
> LGTM. On a side note, there's already a working port of akka on scala.js https://github.com/akka-js/akka.js (by @andreaTP who's also a committer here)

Indeed I am aware, I had a crazy idea of at some point suggesting to explore this in the future (i.e. integrating it into Pekko core) but that is a massive shift and right now is not the best time for obvious reasons.

RyanSkraba (approved):

danischroeter (approved):

an-tex (approved):

pjfanning (approved):
This looks like something we can go ahead and merge - if there are no objections, I will merge this later today.

andreaTP:
A few considerations:

> suggesting to explore this in the future (i.e. integrating it into Pekko core)

this is definitely possible and will simplify the support of e.g. alternative runtimes (e.g. native) in the long term; the benefits would be multiple, I believe, for both projects (we used to detect patterns unusable on single thread executor e.g.).

Regarding this change, currently, Akka is consumable (through the Java API at least) from any language on the JVM (Kotlin, Clojure etc.) and one of the core driving factors was to be extremely fast on the JVM runtime.
I would encourage you @mdedetrich to elaborate a little further on this subject, maybe adding a couple of lines in a readme file if this shift is deliberate and intended.

mdedetrich:
> Regarding this change, currently, Akka is consumable (through the Java API at least) from any language on the JVM (Kotlin, Clojure etc.) and one of the core driving factors was to be extremely fast on the JVM runtime.
I would encourage you @mdedetrich to elaborate a little further on this subject, maybe adding a couple of lines in a readme file if this shift is deliberate and intended.

So I cannot personally say if this shift is deliberate or intended, thats up to the community to collectively decide. What I will say however is this is a possibility, unlike with Akka under Lightbend because (I assume) there wasn't any real internal business reason for Lightbend to explore such a thing (and even if it was community contributed they likely wouldn't want to maintain it). To put it another way, if someone was to make a Pekko Improvement Proposal/s and it gets adoption than that is something that can potentially happen in the future.

I am not sure putting this into a README is necessarily the best place because that is implying that we have already decided and acted upon (which isn't the case). @andreaTP If this is something you want to drive then don't feel hesitant in creating a Pekko Improvement Proposal once we have the structure set up.

andreaTP:
Sounds like a plan 👍

PR #8, opened 2022-11-03
Only allow Squash merges and update notification email addresses

pjfanning:
* Squash merges keep the Git history clean
  * see https://medium.com/swlh/squash-and-rebase-git-basics-5cb1be1e0dac
* Add back notifications settings that were accidentally removed when I reset the last merge commit
  * aim is to not have emails go to dev mailing list every time someone comments on a PR  

Features are documented in https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -25,7 +25,7 @@ github:
   enabled_merge_buttons:
     squash:  true
     merge:   false
-    rebase:  false
+    rebase:  true
 
 notifications:
   commits:              commits@pekko.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -21,3 +21,16 @@ github:
     issues: true
     # Enable projects for project management boards
     projects: false
+
+  enabled_merge_buttons:
+    squash:  true
+    merge:   false
+    rebase:  false
+
+notifications:
+  commits:              commits@pekko.apache.org
+  issues:               dev@pekko.apache.org
+  pullrequests_status:  dev@pekko.apache.org
+  pullrequests_comment: commits@pekko.apache.org
+  discussions:          dev@pekko.apache.org
+


.asf.yaml
```diff
@@ -21,3 +21,16 @@ github:
     issues: true
     # Enable projects for project management boards
     projects: false
+
+  enabled_merge_buttons:
+    squash:  true
+    merge:   false
+    rebase:  false
```
mdedetrich, line 28: Rebase should also be true, there are valid reasons to do a rebase (i.e. commit history is clean and it makes more sense to have multiple commits).

mdedetrich (changes requested):

.asf.yaml
pjfanning: made that change

mdedetrich (approved):
LGTM

.asf.yaml
spangaer: I can't help at wondering why generally rebase seems to be favorered over merge, given that it destroys commit signatures.

If you want to respect commit signatures by default, merge should be allowed and rebase turned off in the UI.

If needed rebasing can still be done locally.

mdedetrich: Squashing also destroys commit signatures because or causes github to create a new commit and it doesn't have your private key to sign it.

spangaer: Well it signs with the GH key, but there's a fundamental difference between the two, because that operation intends to destroy history, the GH actor becomes the author and GH the committer and signer. With rebase signatures are just removed.

mdedetrich: So we can start a general discussion on this, to me having signed commits in `main` isn't particularly useful, the more important thing is that the commits in the PR are signed (assuming we wan't to start caring about signed commits).

The main goal of signed commits is to confirm that a commit comes from the legitimate author, once its been merged into main its kind of irrelevan. Thats why Github has the option to enforce signed commits, but its only relevant on PR's. In practical terms there is definite use in having the rebase option becomes sometimes in a PR the individual commits definitely do matter (as mentioned before, https://github.com/apache/incubator-pekko/pull/2 is a perfect case of this). While one can argue you should split the PR, this isn't always ideal.

jrudolph: It took me long years to accept squash commits as the best default in most situations (especially, compared to accidentally including WIP commits). A clean history has benefits especially compared to merging single commits. In some way it emphasizes the final history of the target project over the snapshot state a contributor worked on. I can see how this can become a problem in a big project like Linux but in smaller projects, simple histories seems worthwhile.

On the other hand, there are situations (like the scala-3 merge in pekko-http) where we do want a merge. We do not want to squash self-contained commits by multiple contributors, and rebasing is also not an option in that case. So, in general, I prefer to have all options available, but I can also see how that can lead to occasional slipups where merge is used instead of squash.

mdedetrich: Strongly agreed. Also wrt to the Linux example, one has to remember that development of Linux is not the same as github/gitlab projects. With Linux people have to send patches on the mailing list and then those patches are manually applied so having a merge commit here brings value because otherwise I would imagine it would be hard to track who merged a specific feature.

On the other hand with projects using github/gitlab its quite easy to find out who merged a specific PR so this is more of a non issue. On the other hand having a clean git history is so much nicer than having to deal with merge commits, especially when you have multiple release branches and when dealing with things like cherry pick.

I also prefer having flexibility in squash or rebase in cases of non typical PR's. For me whats most important is that the git history is clean in context of the change, sometimes thats multiple commits per PR and sometimes thats just a single commit.

jrudolph: > I can't help at wondering why generally rebase seems to be favorered over merge, given that it destroys commit signatures.

I find that argument also compelling (despite of what I said before). Only with proper commit signatures you can trust that the change in a commit is in the state that the original author proposed it. Otherwise, accidentally or by purpose with good or bad intentions, commits will be rewritten, changed, augmented, or adapted. So, you give up some level of trust in the author tags by enforcing the simple history. That means that the committer takes over more responsibility from the author.

In general, I think it is still an okish trade-off since the situation that a commit author wants to deny ownership later on seems quite theoretical. Also, there's still some level of fallback with the original commits often still being preserved by the PRs in Github.

spangaer: Well I observe this drive for linear history, with GH even nudging towards it. From where I stand, that is at most an opinion (from the same house that promotes mutable tags...). One that doesn't really match the everything-is-a-branch spirit which made me love Git in 2008.

At $work, author tag is documented to be at most attribution and committer+signatures is the true responsible.

That works for squash (squasher responsible) and merge in Git UI, but only locally for rebase.

But that indeed brings the bigger question to the table, should commit signatures not be used?

pjfanning: I think it would be better to discuss this on the dev list thread that is open for this discussion - https://lists.apache.org/thread/bfhr8mt99xnqx8jcnhg45fp7kb3mlnc3

spangaer: Anyone on a computer that can copy the thread so far there? 

Doesn't play very nice with phone select 😏

PR #13, opened 2022-11-04
remove mima checks

pjfanning:
* when we change the package names, then the mima binary compatibility checks will fail anyway
* we are also likely to start releasing as v1.0.0 meaning that the old Akka version numbers will be out of synch with Pekko's


nvollmar (approved):
lgtm

gmethvin (approved):

danischroeter (approved):

pjfanning:
thanks everyone for the reviews - I'll merge this and start looking at the other repos

PR #17, opened 2022-11-05
Add .git-blame-ignore-revs file containing format commits

mdedetrich:
Adds a `.git-blame-ignore-revs` file which can be setup on a per project basis so that git blame ignores scalafmt commits (see https://akrabat.com/ignoring-revisions-with-git-blame/). I have added the commit hashes for all scalafmt commits in the history of the project.

Note that the steps in https://akrabat.com/ignoring-revisions-with-git-blame/ still need to be added in the documentation.


pjfanning (approved):
lgtm

PR #18, opened 2022-11-05
Add github workflow to check scala code is formatted

mdedetrich:
This PR is a follow up from https://github.com/apache/incubator-pekko/pull/2 but only contains the replacement of sbt check with github actions. It also implements the git ratcheting system so that it only performs a diff against files that have actually been changed against `origin/main` meaning the format step should take ~10-20 seconds on average since it doesn't format the whole codebase as well as being initiated immediately rather than having to wait for the main test PR validation step to happen. Note that currently when one formats the code via `sbt-scalafmt` it takes roughly 1 minute.

Here is an example of a test run that I made deliberately to fail (i.e. I committed a single file that wasn't formatted). It took 4 seconds (ignoring the git branch checkout), 20 seconds full. https://github.com/apache/incubator-pekko/actions/runs/3400549218/jobs/5655050347.

Here is an example of a test run against no files that need to be formatted (i.e. not editing sources), it took <1 seconds (ignoring the git branch checkout), 16 seconds full. https://github.com/apache/incubator-pekko/actions/runs/3400546569/jobs/5655045456.

### DIFF
--- a/.github/workflows/build-test-prValidation.yml
+++ b/.github/workflows/build-test-prValidation.yml
@@ -9,8 +9,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check-code-style:
-    name: Check / Code Style
+  check-headers:
+    name: Check / Headers
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #19, opened 2022-11-05
Fix nightly build action

seglo:
Despite its name this build doesn't actually publish anything (except for test reports). It just runs some more verbose testing on different JDKs, cluster, artery, and some other things. We'll need to add the `marcospereira/action-surefire-report@v1` action to an allow list, or find an alternative, to get test report artifacts published.

Build on my fork: https://github.com/seglo/incubator-pekko/actions/runs/3400931121

### DIFF
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -7,8 +7,8 @@ on:
 
 jobs:
 
-  akka-cluster-metrics-sigar:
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm - INFRA team have whitelisted a lot of the actions that we need - so there's a good chance this action is available to use

seglo:
For reference the action is `marcospereira/action-surefire-report@v1` (Marcos is a former akka team member)

PR #31, opened 2022-11-08
=str Fix `CompletionStage#toCompletableFuture` may throw Exception

He-Pin:
As methioned at https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html#toCompletableFuture--

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/Unfold.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/Unfold.scala
@@ -21,6 +21,7 @@ import pekko.stream.impl.Stages.DefaultAttributes
 import pekko.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
 
 import java.util.Optional
+import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
 import scala.concurrent.Future
 import scala.util.{ Failure, Success, Try }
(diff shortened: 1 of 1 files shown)


akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -113,20 +114,21 @@ import scala.util.{ Failure, Success, Try }
       }
 
       def onPull(): Unit = {
-        val future = f.apply(state).toCompletableFuture
```
He-Pin, line 116: `.toCompletableFuture` which may throw exception, now fixed.

nvollmar (approved):
lgtm

pjfanning (approved):

spangaer:
Category not for initial release?

pjfanning:
if it is a fix that doesn't change the API - it could be worth considering for initial release - we could vote on that if is contentious

spangaer:
That seems to stray from the statement "no changes" except for those required for the rename.

It's certainly a fair bug fix and it would probably be fair to include it in 1.0.1 (or 1.0.10, assuming there will be some dangling rename bugs popping up post release or 1.1.0 if that's the first version allowing to deviate from Akka 2.6.x), but if you include it in 1.0.0 your going to have a lot of discussions and votes I fear.

I think the one door you could somewhat argue to open is 2.6.x regressions.

He-Pin:
I think this can be including in the initial release, as it not changing the semantics.

Claudenw (changes requested):
Under current rules this should be marked as not for initial release.

I agree that it should be in the initial release.
I request that a discussion be opened to get consensus that non-API changing changes may  be included in the first release.

pjfanning:
I favour adding this now but can be persuaded to delay adding it. We could vote on this type of change to see where the team are at - are small fixes ok or do we want to basically re-release Akka 2.6.20 as Pekko 1.0.0, warts and all?

mdedetrich:

mdedetrich (approved):
lgtm

PR #43, opened 2022-11-14
!str Eagerly fails flow if the future is already failed.

He-Pin:
If the future is already failed, it will fail the flow even if there is no downstream demands now.
refs: https://github.com/akka/akka/pull/31728
Which was reported first in akka/akka gitter

This is a behavior change, eagarly failing align with the `FailedSource`.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFromFutureSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFromFutureSpec.scala
@@ -11,6 +11,7 @@ import scala.concurrent.duration._
 import scala.util.control.NoStackTrace
 
 import akka.stream.testkit._
+import akka.stream.testkit.scaladsl.TestSink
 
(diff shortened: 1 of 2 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFromFutureSpec.scala
```diff
@@ -33,6 +34,13 @@ class FlowFromFutureSpec extends StreamSpec {
       c.expectSubscriptionAndError(ex)
     }
 
+    "fails flow from already failed Future even no demands" in {
+      val ex = new RuntimeException("test") with NoStackTrace
+      val sub = Source.fromFuture(Future.failed[Int](ex))
+        .runWith(TestSink.probe)
+      sub.expectSubscriptionAndError(ex)
```
He-Pin, line 41: a new test added

nvollmar (approved):
lgtm

He-Pin:
oops, will close:)

PR #44, opened 2022-11-14
stream: fix regression in JsonFraming

jrudolph:
Was introduced in 6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b where an invariant was changed what `pos` can be when the loop is run.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
@@ -117,21 +117,22 @@ import pekko.util.ByteString
     val bufSize = buffer.length
 
     skipToNextObject(bufSize)
+    val maxObjectLengthIndex = if (maximumObjectLength == Int.MaxValue) Int.MaxValue else pos + maximumObjectLength
 
-    while (pos < bufSize && pos < maximumObjectLength && !completedObject) {
+    while (pos < bufSize && pos < maxObjectLengthIndex && !completedObject) {
       val input = buffer(pos)
       proceed(input)
       pos += 1
     }
 
-    if (pos >= maximumObjectLength)
+    if (pos >= maxObjectLengthIndex)
       throw new FramingException(s"""JSON element exceeded maximumObjectLength ($maximumObjectLength bytes)!""")
 
     completedObject
   }
 
   private def skipToNextObject(bufSize: Int): Unit =
-    while (pos != -1 && pos < bufSize && pos < maximumObjectLength && depth == 0) {
+    while (pos != -1 && pos < bufSize && depth == 0) {
       val outer = outerChars(buffer(pos) & 0xFF)
       start += outer & 1
       depth = (outer & 2) >> 1
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
@@ -117,7 +117,7 @@ import pekko.util.ByteString
     val bufSize = buffer.length
 
     skipToNextObject(bufSize)
-    val maxObjectLengthIndex = if (maximumObjectLength == Int.MaxValue) Int.MaxValue else pos + maximumObjectLength
+    val maxObjectLengthIndex = if (pos + maximumObjectLength < 0) Int.MaxValue else pos + maximumObjectLength
 
     while (pos < bufSize && pos < maxObjectLengthIndex && !completedObject) {
       val input = buffer(pos)
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/JsonFramingSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/JsonFramingSpec.scala
@@ -41,7 +41,7 @@ class JsonFramingSpec extends PekkoSpec {
           |""".stripMargin // also should complete once notices end of array
 
       val result =
-        Source.single(ByteString(input)).via(JsonFraming.objectScanner(Int.MaxValue)).runFold(Seq.empty[String]) {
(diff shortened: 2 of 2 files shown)


nvollmar (approved):
lgtm

pjfanning (approved):

He-Pin:
Thanks, origin reported in https://github.com/akka/akka/issues/31569

akka-stream/src/main/scala/akka/stream/impl/JsonObjectParser.scala
```diff
@@ -107,21 +107,22 @@ import akka.util.ByteString
     val bufSize = buffer.length
 
     skipToNextObject(bufSize)
+    val maxObjectLengthIndex = if (maximumObjectLength == Int.MaxValue) Int.MaxValue else pos + maximumObjectLength
```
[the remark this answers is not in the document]
jrudolph: Good point. I wonder if the code can made to work in those cases, where leading bytes to be skipped plus object will not fit into an array?

We should definitely fix the condition here but might also want to add a guard to `offer` to avoid creating a new array with negative size and also throw the FramingException from below.

mdedetrich:
@jrudolph You need to rebase this.

stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
```diff
@@ -117,21 +117,22 @@ import pekko.util.ByteString
     val bufSize = buffer.length
 
     skipToNextObject(bufSize)
+    val maxObjectLengthIndex = if (pos + maximumObjectLength < 0) Int.MaxValue else pos + maximumObjectLength
```
jrudolph, line 120: @He-Pin what do you think about this solution? This way we catch the overflow directly. On the other hand, such a big `maximumObjectLength` is unlikely to work anyway because we cannot accommodate such a big object including any padding into our buffer... (i.e. we'll still fail in `offer`). Supporting huge objects is not economical in any case right now because of the buffer management issue.

Claudenw: Do you have a check for "huge objects"?  Perhaps placing an upper limit on the object size is a good start until a complete solution can be discovered.

pjfanning:
Dropped from 1.0.0 milestone as it is incomplete and having it in the milestone makes it look like a blocker.

mdedetrich:
> Dropped from 1.0.0 milestone as it is incomplete and having it in the milestone makes it look like a blocker.

This might be a bit rash, basically without this change json framing is broken. Should try and get this out by release but the questions raised need to be resolved.

mdedetrich:
I would also suggest that if we cannot get this fix out by release, lets revert the JSON framing change that caused the regression. Better to release something that is slightly slower rather than something that is broken

@pjfanning @jrudolph @raboof @He-Pin @Claudenw  wdyt?

pjfanning:
https://github.com/apache/incubator-pekko/commit/6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b is a pre-Pekko commit

https://github.com/akka/akka/commit/6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b is the akka link and it shows that this code is part of akka 2.6.20 and since pekko 1.0.0 is meant to be a close approximation of akka 2.6.20, you could say that this is not a blocker for 1.0.0

I'm open to having something done about this in 1.0.0 but we sort of need a quick decision.

pjfanning:
I've added it back to milestone while this issue is further discussed.

pjfanning:
> I would also suggest that if we cannot get this fix out by release, lets revert the JSON framing change that caused the regression. Better to release something that is slightly slower rather than something that is broken
> 
> @pjfanning @jrudolph @raboof @He-Pin @Claudenw wdyt?

I would be happy with reverting https://github.com/apache/incubator-pekko/commit/6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b 

A fuller fix would be appreciated but I'd like to do whatever it takes to get the 1.0.0 release train moving.

mdedetrich:
> https://github.com/akka/akka/commit/6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b is the akka link and it shows that this code is part of akka 2.6.20 and since pekko 1.0.0 is meant to be a close approximation of akka 2.6.20, you could say that this is not a blocker for 1.0.0

Right but this is an exceptional circumstance, Akka basically broke the JSON framing right before the license change happened and then fixed it after the license change. I understand that we want Pekko 1.0.x to be like Akka 2.6.0 but we have done exceptions, and one of those exceptions is with critical bugs (and I would classify this as critical since it doesn't work, and it did work ~2 months(iirc) in a previous version of Akka before the license change.

> I'm open to having something done about this in 1.0.0 but we sort of need a quick decision.

> I would be happy with reverting https://github.com/apache/incubator-pekko/commit/6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b

Likewise, I would also suggest if this is not resolved by the time of one of the final release candidates then lets just revert the commit.

pjfanning:
we could do this a 2 phase
1. revert https://github.com/apache/incubator-pekko/commit/6b30134c5c7e8deefdfc7ff9a3e74e0bbd2d433b
2. follow up with better fix

1 is for v1.0.0 release and 2 is a maybe for 1.0.0 release but we won't wait for it.

mdedetrich:
Okay, lets wait a few days to see if anyone responds and if this fix isn't landed/finished by then lets make a revert commit

He-Pin:
@pjfanning Let's see if we can get this in 1.0.0.

He-Pin (approved):
lgtm

He-Pin:
Close and reopen for retrigger the build.

He-Pin:
@mdedetrich @jrudolph I checked this code locally and can confirm the current fix is OK.

pjfanning (approved):
lgtm

pjfanning:
merged - it's not too late to object, we can always add extra changes or revert

PR #47, opened 2022-11-16
!str Eagerly fails flow if the future is already failed.

He-Pin:
If the future is already failed, it will fail the flow even if there is no downstream demands now.
refs: https://github.com/akka/akka/pull/31728
Which was reported first in akka/akka gitter

This is a behavior change, eagarly failing align with the FailedSource.

```
  /**
   * Emits a single value when the given `Future` is successfully completed and then completes the stream.
   * The stream fails if the `Future` is completed with a failure.
   */
```

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphStages.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphStages.scala
@@ -390,26 +390,24 @@ import pekko.stream.stage._
     override def initialAttributes: Attributes = DefaultAttributes.futureSource
     override def createLogic(attr: Attributes) =
       new GraphStageLogic(shape) with OutHandler {
-        def onPull(): Unit = {
+        override def preStart(): Unit = {
           future.value match {
             case Some(completed) =>
               // optimization if the future is already completed
-              onFutureCompleted(completed)
+              handle(completed)
             case None =>
-              val cb = getAsyncCallback[Try[T]](onFutureCompleted).invoke _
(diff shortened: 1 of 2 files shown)


mdedetrich:

nvollmar (approved):

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphStages.scala
```diff
@@ -381,27 +381,25 @@ import pekko.stream.stage._
     override def initialAttributes: Attributes = DefaultAttributes.futureSource
     override def createLogic(attr: Attributes) =
       new GraphStageLogic(shape) with OutHandler {
-        def onPull(): Unit = {
+        override def preStart(): Unit = {
           future.value match {
             case Some(completed) =>
               // optimization if the future is already completed
-              onFutureCompleted(completed)
+              handle(completed)
```
He-Pin, line 388: This will can break up some code, because we change the existing behavior. 
@mdedetrich 

The `FailedSource` failing the stream ASAP. but maybe not this one?

Especially when combine it with `flatmapConcat` with parallism materialization.

He-Pin: @nvollmar @pjfanning What's your option here

pjfanning: I don't see this change in behaviour to be a big issue. We'll need to release note it. I think most users would assume this is the way it was meant to work.

pjfanning (approved):
lgtm

Issue #51, opened 2022-11-16
Update Git rules to follow work pattern.


Claudenw:
The github rules should not allow merging of pull requests when there are 

1. open changes requested.
2. ongoing conversations (not resolved conversations).

I am not certain how to configure this on the Apache git system.

Claudenw:
pj assigning to you because you seem to have the experience in this realm.

pjfanning:
The documented features are at https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features - I'll look later to see if I can spot something that we can use

mdedetrich:
Especially with point 2, this to me I think this is too strict. I have been in cases where pull requests stuck open in perpetuity because someone forgets to click the "resolve" button. Either that or committers will just blindly hit resolve which at that point defeats the purpose of it.

Should we discuss in the mailing list? I think for now this is fine, we can always add stricter checks later if there is a consensus on it.

sam-byng:
@mdedetrich seems like we've got a resolution on this now, with the mandatory single reviewer . Could this be closed?

PR #52, opened 2022-11-16
require PR reviews and linear history

pjfanning:
Based on https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features

The changes may need tweaking - I won't know if the changes will be applied as expected - so I'll fix any issues where expected updates don't actually get applied

https://github.com/apache/incubator-pekko/issues/51

* requires only 1 PR review - https://lists.apache.org/thread/0hohnn9v0xhxn0qblb196p7mcq7jzz00 and https://cwiki.apache.org/confluence/display/PEKKO/Processes requires 1 review for trivial PRs
* tries to enforce that Scalafmt workflow passes - we can add more later
* linear git history

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -28,7 +28,21 @@ github:
     rebase:  true
 
   protected_branches:
-    main: { }
+    main:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: false
+        # contexts are the names of checks that must pass
+        contexts:
+          - Scalafmt
+  
+      required_pull_request_reviews:
+        dismiss_stale_reviews: false
+        require_code_owner_reviews: false
+        required_approving_review_count: 1
+       
+      # squash or rebase must be allowed in the repo for this setting to be set to true.
+      required_linear_history: true
 
 notifications:
   commits:              commits@pekko.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,5 +1,20 @@
-# https://cwiki.apache.org/confluence/display/INFRA/.asf.yaml+features+for+git+repositories
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
+# https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features
+---
 github:
   description: "Build highly concurrent, distributed, and resilient message-driven applications using Java/Scala"
   homepage: https://pekko.apache.org/
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -35,6 +35,7 @@ github:
         # contexts are the names of checks that must pass
         contexts:
           - Scalafmt
+          - Pull Requests / Check / Headers
   
       required_pull_request_reviews:
         dismiss_stale_reviews: false


.asf.yaml
```diff
@@ -28,7 +28,21 @@ github:
...
+        # strict means "Require branches to be up to date before merging".
+        strict: false
+        # contexts are the names of checks that must pass
+        contexts:
+          - Scalafmt
+  
+      required_pull_request_reviews:
+        dismiss_stale_reviews: false
+        require_code_owner_reviews: false
+        required_approving_review_count: 1
```
He-Pin, line 42: at least 2?

pjfanning: only 1 review needed for trivial changes according to https://cwiki.apache.org/confluence/display/PEKKO/Processes - there is no way to work out automatically what is trivial

mdedetrich (changes requested):
Is it possible to add the header check in conjunction with scalafmt?

.asf.yaml
mdedetrich: So this is an interesting point. Historically akka always required 2 approvals, even for trivial changes. I will say however in terms of practicality especially right now with a lot of mechanical and trivial changes needed to her the project going 1 should be fine.

We can revisit this process later, I personally think that requiring 2 reviewers once the project is setup has merit especially since at that point in time we will also start seeing non trivial functional changes.

He-Pin: Who is in the position to merge pull requests?

pjfanning: anyone with git commit rights - all Pekko committers can do this - you might need to link your Github and ASF accounts if you haven't done this

pjfanning:
> Is it possible to add the header check in conjunction with scalafmt?

Which of these workflows is the 'header check'? Do we need to add a 'header check' workflow first?

.asf.yaml
He-Pin: So the origin author and who proved the PR can merge it too?

mdedetrich:
> Which of these workflows is the 'header check'? Do we need to add a 'header check' workflow first?

its already there, its called  `Pull Requests / Check / Headers`, 

image

.asf.yaml
pjfanning: I don't know how to implement:

```
Any review that has open requests or discussion blocks the merge. (only merge when consensus is reached).
```

If there is a way, it can be implemented.

mdedetrich: > Any review that has open requests or discussion blocks the merge. (only merge when consensus is reached).

To me I think this is too strict, I have been in cases where pull requests stuck open in perpetuity because someone forgets to click the "resolve" button. Either that or committers will just blindly hit resolve which at that point defeats the purpose of it.

Should we discuss in the mailing list? I think for now this is fine, we can always add stricter checks later if there is a consensus on it.

mdedetrich: I would say lets just just merge the PR as it is now and we can also make a future one to implement these stricter checks.

.asf.yaml
```diff
@@ -28,7 +28,22 @@ github:
     rebase:  true
 
   protected_branches:
-    main: { }
+    main:
```
Claudenw, line 31: Please add:

```## Licensed under the terms of http://www.apache.org/licenses/LICENSE-2.0```

To the top of the file

**Review by Claudenw — CHANGES_REQUESTED** [review]

File: .asf.yaml (hunk 1 above)
Comment (Claudenw) [reply .asf.yaml:43]: @pjfanning I think that

```      required_conversation_resolution: true```

will require the conversation resolution.

Comment (pjfanning) [reply .asf.yaml:43]: I don't see `required_conversation_resolution` in https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features

File: .asf.yaml (hunk 2 above)
Comment (pjfanning) [reply .asf.yaml:46]: done

**Comment by pjfanning** [comment]
@Claudenw Is it possible for you to re-review this?

File: .asf.yaml [hunk 3]
```diff
@@ -1,5 +1,20 @@
-# https://cwiki.apache.org/confluence/display/INFRA/.asf.yaml+features+for+git+repositories
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
mdedetrich, line 1: @pjfanning I think you can remove this header change because its going to be done properly in https://github.com/apache/incubator-pekko/pull/50

.asf.yaml
```diff
@@ -28,7 +43,22 @@ github:
...
 
   protected_branches:
-    main: { }
+    main:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: false
+        # contexts are the names of checks that must pass
+        contexts:
+          - Scalafmt
```
mdedetrich, line 52: As per https://github.com/apache/incubator-pekko-sbt-paradox/pull/7, this needs to be changed to `- Code is formatted` rather than `- Scalafmt`

.asf.yaml
```diff
@@ -28,7 +43,22 @@ github:
...
+          - Pull Requests / Check / Headers
```
mdedetrich, line 53: I would recommend checking that this works in a separately. As per https://github.com/apache/incubator-pekko-sbt-paradox/pull/5#issuecomment-1414453790 otherwise we can freeze the main branch and we have to make an INFRA ticket to unfreeze it. The way I did this I pushed a new branch (ergo `test`), created a PR against `main` that only enables the `contexts` check against `test` and then make another PR against test to figure out if the check is satisfied.

mdedetrich (changes requested):
So I have changed my review back to request changes because there if the PR is merged in its current state it would freeze the main branch (check review comments).

pjfanning:
I'll close this as it is stale

PR #53, opened 2022-11-16
update github actions due to warnings about nodejs version

pjfanning:
See warnings on https://github.com/apache/incubator-pekko/actions/runs/3480833557

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2, olafurpg/setup-scala@v10
```

### DIFF
--- a/.github/workflows/binary-compatibility-checks.yml
+++ b/.github/workflows/binary-compatibility-checks.yml
@@ -38,14 +38,14 @@ jobs:
 
       # compile code instead of doing mima check (until we re-enable mima check)
       - name: Compile code
-        run: sbt -jvm-opts .jvmopts-ci "+~ ${{ matrix.scalaVersion }} Test/compile"
+        run: sbtx -jvm-opts .jvmopts-ci "+~ ${{ matrix.scalaVersion }} Test/compile"
 
       # disable mima check until first pekko release is done
       #- name: Report MiMa Binary Issues
-      #  run: sbt -jvm-opts .jvmopts-ci "+~ ${{ matrix.scalaVersion }} mimaReportBinaryIssues"
+      #  run: sbtx -jvm-opts .jvmopts-ci "+~ ${{ matrix.scalaVersion }} mimaReportBinaryIssues"
 
       #- name: Check correct MiMa filter directories
-      #  run: sbt -jvm-opts .jvmopts-ci checkMimaFilterDirectories
+      #  run: sbtx -jvm-opts .jvmopts-ci checkMimaFilterDirectories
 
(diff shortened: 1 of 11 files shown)


.github/workflows/binary-compatibility-checks.yml
```diff
@@ -23,13 +23,13 @@ jobs:
...
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
           fetch-depth: 0
 
       - name: Set up JDK 8
-        uses: olafurpg/setup-scala@v10
+        uses: olafurpg/setup-scala@v13
```
He-Pin, line 32: use coursier one?

pjfanning: I fixed the build issues using https://github.com/apache/incubator-pekko/pull/53/commits/ca3e39807208d10f4c085802eb8db57e11dc85a0

Looks like sbtx command is now needed to get `-sbt-opts` to work - https://github.com/olafurpg/setup-scala#setup-scala-github-action

pjfanning: @He-Pin would you be able to approve these changes if they are ok?

mdedetrich: So I am a bit confused here wrt to the usage of `sbtx`. Afaik `sbtx` shouldn't be necessary unless you have some ancient version of sbt and the standard `sbt` should work with sbt-opts fine.

Let me ping @dwijnand on this.

pjfanning: these workflows use `sbt -sbt-opts` - this seems to only work with sbtx (sbt doesn't support this option)

Maybe older versions of olafurpg/setup-scala installed sbtx as if was sbt

I wouldn't have changed to sbtx if the workflows were working

mdedetrich: > I wouldn't have changed to sbtx if the workflows were working

Understood just want to understand whats going on (seems like `sbtx` is a workaround which is fine for now but we may need to migrate how we pass in sbt options later at some point so it just works with standard `sbt`).

mdedetrich (approved):
So I will approve this for now since it is blocking the CI however we should make an issue exploring whats going on with `sbtx`.

He-Pin:
LGTM

He-Pin (approved):
LGTM

pjfanning:
thanks for the reviews - we can certainly change this again if we find a better solution to the sbt calls.

PR #55, opened 2022-11-17
=str Fix calling `onComplete` twice when downstream cancel.

He-Pin:
Try a another way for https://github.com/akka/akka/pull/31630
1. Keep the `onComplete` act as a finalizer
2. Do not call `onComplete` twice when restart
3. pump the option value downstream when restart

In fs2, resource/can be `null`.
```scala
package fs2

import cats.effect.{IO, Resource}

object Test extends App {
  val res1: Resource[IO, String] = Resource.make {
    println("make")
    IO.pure(null)
  } { r =>
    println("release:" + r)
    IO.unit
  }
  val str: Stream[IO, String] = Stream.resource(res1) ++ Stream.emit("done")
  str.compile.toList.unsafeRunAsync{
    case Left(value) => println(value)
    case Right(value) => println(value)
  }(cats.effect.unsafe.implicits.global)
  Thread.sleep(3000)
}
```
prints:
```
make
release:null
List(null, done)
```

I don't have a poweradapter now, so maybe can't update until tomorrow.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
@@ -4,6 +4,15 @@
 
 package org.apache.pekko.stream.scaladsl
 
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.annotation.nowarn
+import scala.concurrent.Await
+import scala.concurrent.Promise
+import scala.concurrent.duration.DurationInt
+import scala.util.Success
+import scala.util.control.NoStackTrace
+
 import org.apache.pekko
 import pekko.Done
 import pekko.stream.AbruptStageTerminationException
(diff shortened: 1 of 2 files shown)


He-Pin:
I think it's not a critical bug and easy to work around, so move to 1.1.0, thanks

rtar:
I wonder if this is a problem in this case. Giving a link in case you missed it: https://twitter.com/jdegoes/status/1605889036740136965

He-Pin:
@rtar I don't think that's was a problem because I write the code, And if line by line diff missing , I can adjust it, very simple anyway.

pjfanning:
I'm going to close this. There is evidence that the Akka change was looked at and that is a big problem.

We simply cannot look at Akka changes in their BSL licensed versions.

mdedetrich:
We should make a github issue to track this incase someone else wants to implement it, but that "someone" needs to be a person that hasn't looked at the Akka source at all and it should also shouldn't be someone who created this PR.

pjfanning:
At this stage, I think we will need to get someone else to work on this issue. We can't afford even to have the perception that Pekko contributors have looked at BSL Akka code. The only exception is if the PR in Akka was written by the same person who then later contributes it to Pekko.

PR #62, opened 2022-11-30
rename akka-docs dir to docs

pjfanning:
rename akka-docs dir to docs


mdedetrich:
Ill check this either today or tomorrow

mdedetrich:
@pjfanning Not sure if this is a bit munch but since we are renaming the sbt project/directory right now should we just change it to `docs` rather than `pekko-docs`? Every other pekko module just uses `docs` and its also the default directory for paradox (I believe calling it `akka-docs` initially was an oversight).

pjfanning:
> @pjfanning Not sure if this is a bit munch but since we are renaming the sbt project/directory right now should we just change it to `docs` rather than `pekko-docs`? Every other pekko module just uses `docs` and its also the default directory for paradox (I believe calling it `akka-docs` initially was an oversight).



> @pjfanning Not sure if this is a bit munch but since we are renaming the sbt project/directory right now should we just change it to `docs` rather than `pekko-docs`? Every other pekko module just uses `docs` and its also the default directory for paradox (I believe calling it `akka-docs` initially was an oversight).

renamed to 'docs' instead

mdedetrich (approved):
Checked out the branch locally, also ran `docs/paradoxMarkdownToHtml` to make sure everything looks okay, good work lgtm.

gmethvin (approved):
LGTM

PR #58, opened 2022-11-20
Rename akka package to org.apache.pekko

mdedetrich:
Current in progress PR

PR #63, opened 2022-12-01
Rename akka to pekko in configuration

### LINKED ISSUE #54: Should use `pekko` as HOCON config prefix rather than `akka`
We should change our config loading and `reference.conf` to use `pekko` everywhere rather than `akka`. It looks like the config loading is mostly in the [`Settings`](https://github.com/apache/incubator-pekko/blob/9edbc602928c45822a695789589cd3371b3f9c6d/akka-actor/src/main/scala/akka/actor/ActorSystem.scala#L374) class, but that also exposes the raw `Config` instance, which is used directly to load `akka.*` configs in other classes.

It makes sense to have a way to support the `akka` configuration prefix instead of or in addition to the `pekko` prefix, but I know there was some question about whether to do it for 1.0. Given that config loading is mostly centralized it should be relatively simple to do.

gmethvin:
Resolves #54

Rename `akka` in configuration to `pekko`, and update the code to use the new keys.

This changes a very large number of files, so like with #58, you probably want to check out the branch and look at it in a good diff tool.

This was mostly done as a bunch of different find/replace operations to match things that look like config paths.

This does not attempt to address issues for users migrating from akka to pekko config. We can follow up with any compatibility changes in another PR.


mdedetrich:
Thanks for the PR, will try to get this checked out by the weekend

pjfanning (changes requested):
Could you check project/AkkaBuild.scala? The initialCommands (line 172) needs updating. 

Could you change this?
```
    // all system properties passed to sbt prefixed with "akka." or "aeron." will be passed on to the forked jvms as is
    Test / javaOptions := {
      val base = (Test / javaOptions).value
      val knownPrefix = Set("akka.", "aeron.")
      val akkaSysProps: Seq[String] =
        sys.props.iterator.collect {
          case (key, value) if knownPrefix.exists(pre => key.startsWith(pre)) => s"-D$key=$value"
        }.toList

      base ++ akkaSysProps
    },
```
to
```
    // all system properties passed to sbt prefixed with "pekko." or "aeron." will be passed on to the forked jvms as is
    Test / javaOptions := {
      val base = (Test / javaOptions).value
      val knownPrefix = Set("pekko.", "aeron.")
      val pekkoSysProps: Seq[String] =
        sys.props.iterator.collect {
          case (key, value) if knownPrefix.exists(pre => key.startsWith(pre)) => s"-D$key=$value"
        }.toList

      base ++ pekkoSysProps
    },
```

Is this also something that needs pekko prefix instead of akka?
```
val runningOnCi: CliOption[Boolean] = CliOption("akka.ci-server", sys.env.contains("CI"))
```

I suspect there could be other changes needed in this file too.

pjfanning:
Also project/MultiNode.scala - `sys.props.getOrElse("akka.test.multi-in-test", "true")` needs to change

project/VersionGenerator.scala
```diff
@@ -8,15 +8,15 @@ import sbt._
 import sbt.Keys._
 
 /**
- * Generate version.conf and akka/Version.scala files based on the version setting.
+ * Generate version.conf and Version.scala files based on the version setting.
```
pjfanning, line 11: should this file be generated as org/apache/pekko/Version.scala?

gmethvin: I think only the package name really matters in this case, right? This is just generated code.

gmethvin:
@pjfanning thanks, I may have missed a few things in the build. I'd like this PR to include both system properties and HOCON configuration, so those are all valid changes.

project/VersionGenerator.scala
pjfanning: other unrelated jars on a project's classpath could have Version classes as well - I think it is better to put this class in a  pekko specific package - the akka Version class is in the akka package

gmethvin: I'll just change it

pjfanning:
> @pjfanning thanks, I may have missed a few things in the build. I'd like this PR to include both system properties and HOCON configuration, so those are all valid changes.

I've hit hit similar issues in some of my PRs. IntelliJ seems to ignore the project dir when doing global replaces.

pjfanning (approved):
lgtm - 1 or 2 missed items might pop up later but this is a very good start

mdedetrich:
@gmethvin The CI is failing

pjfanning:
> @gmethvin The CI is failing

failed mima binary checks - I thought they were disabled in main branch

mdedetrich:
@gmethvin The PR at #65 has been merged into main, can you rebase this PR to see if CI passes?

gmethvin:
@mdedetrich CI is passing now

pjfanning:
@gmethvin I merged my changes. I wasn't aware how many docs files you had changed. Could you have a look to see if this PR can be easily fixed up? If it is too problematic, I could look at reverting my changes.

nvollmar (approved):

gmethvin:
I'm going to go ahead and merge this. If I missed some things we can work them out later.

PR #75, opened 2022-12-16
System.getProperty("file.separator") replaced by "/" to avoid build p…

sdeigm:
Fixes the build problems on Windows cause by a backslash file separator


pjfanning (approved):
lgtm - I'll merge when the CI build completes - looks like it hasn't broken anything

Issue #74, opened 2022-12-16
Need to put LICENSE and NOTICE file in every pekko jar META-INF/ dir


pjfanning:
If you look at any ASF jar, you will find these 2 files.

spangaer:
Any known ASF SBT built projects? (hence prior art)

pjfanning:
I did a quick search around Apache projects and didn't find any using sbt. I'm sure other projects use sbt, just not the obvious Scala ones like Spark, Kafka, Linkis, Sedona, Flink.

spangaer:
Bummer.

If this hasn't moved by second half of January, I might have stab at it. When the project started I already thought this was a wonderfully boring task, that someone would have to do eventually 😋 

If someone else want to take it up before, I really don't mind 😉 . No dibs here, I just won't get to it sooner.


Unless it should just be a copy of the files sitting at the repo root? Because in that case 2 symlinks can do wonders. 🤷

pjfanning:
#77 is merged

PR #77, opened 2022-12-23
try to add license and notice to jars

pjfanning:
relates to https://github.com/apache/incubator-pekko/issues/74


NOTICE
```diff
@@ -0,0 +1,13 @@
...
+This product includes software developed at
+The Apache Software Foundation (https://www.apache.org/).
+
+This product contains significant parts that were originally based on software from Lightbend (Akka <https://akka.io/>).
+Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
+
+Apache Pekko is a derived from Akka 2.6.x, the last version that was distributed under the
+Apache License, Version 2.0 License.
+
+`pekko-protobuf` includes protobuf v2.5.0 code developed by Google Inc.
```
justinmclean, line 13: There is no need for the pekko-protobuf line

pjfanning: @justinmclean I was basing this on other ASF project NOTICE files that often mention code from other projects that has been included in that project. Examples:
* https://github.com/apache/spark/blob/master/NOTICE
* https://github.com/apache/poi/blob/trunk/legal/NOTICE
* https://github.com/apache/xmlbeans/blob/trunk/NOTICE.txt

For Drill, the NOTICE says the other licenses are in the main LICENSE file
* https://github.com/apache/drill/blob/master/LICENSE

So I guess, there are differences in approach. Could you suggest if we need to acknowledge the Google copyright at all? We are planning to keep the code headers that already acknowledge this.

justinmclean: They should only mention that when the original code copyright statement has been removed. You only add the copyright statement to the NOTICE file when it's been relocated in this way.

justinmclean: Another instance where you add content to NOTICE is when you bundle some software that is Apache licensed and that software has a NOICE file. In that case, you need to look at the bundled software NOTICE file and copy what is required from it your NOTICE file.

justinmclean: So in the case of Google 1. Does the software have a NOTICE file? If so (and Google software generally doesn't) then you need to add it content to NOTICE 2. Has tech copyright header been removed (with permission)? If so then yes you need to add the copyright to NOTICE, this however seems unlikely. It's very likely that Google doesn't need to mentioned in the NOTICE file.

pjfanning: thanks @justinmclean - I have already removed the Google protobuf lines from the NOTICE. I didn't find anything in the protobuf 2.5.0 that requires that we update our NOTICE file.

https://github.com/protocolbuffers/protobuf/releases?q=v2.5.0&expanded=true

NOTICE
```diff
@@ -0,0 +1,11 @@
+Apache Pekko
+Copyright 2022, 2023 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (https://www.apache.org/).
```
seglo, line 5: Is it worth mentioning https://pekko.apache.org/ ?

seglo (approved):

NOTICE
pjfanning: NOTICE files are generally added for ASF projects. I've never seen them include project specific URLs. An example is the Apache Spark NOTICE file.

https://github.com/apache/spark/blob/master/NOTICE

Issue #59, opened 2022-11-23
Don't use sbtx in the CI


mdedetrich:
In this PR https://github.com/apache/incubator-pekko/pull/53#discussion_r1024985382 to get CI running we ended up having to use sbtx instead of plain sbt due to having to upgrade a github action. It turns out that the change to sbtx was that in the older versions of the github action, it was aliasing `sbt` to `sbtx` (see https://discord.com/channels/632150470000902164/922600050989875282/1042858091387043872). We should investigate into using pure sbt rather than sbtx since sbtx isn't really supported that much anymore

pjfanning:
sbtx is just a script to run sbt commands. It still respects the projects/build.properties and the actual sbt runtime that is needed to run the sbt version are still downloaded as they would be with the standard sbt script.

mdedetrich:
This I understand, the reason behind this issue is that sbtx is an ancient script (originally created by paulp) which was mainly created because in the old days of sbt there was a lot of ceremony in setting up sbt, something that is now largely irrelevant because a lot of that functionality was moved into official sbt. If you read the linked discord ticket you will also gets similar sentiment regarding the usage of sbtx.

Due to this (and hence the main motivator for this ticket), sbtx is not installed anymore on distros/Windows/OSX. Everyone just downloads/installs/sets up pure SBT which comes to the final point which I want to get to, which is simulating the CI on your local machine (a problem that I experienced when I worked on non trivial PR's).

Note that this ticket is not high priority, but its definitely ideal to use `sbt` rather than `sbtx` for these reasons.

PR #81, opened 2023-01-02
Remove sbtx usage

jchapuis:
Fix for #59 

Replace by prior copy of `.jvmopts-ci` into `.jvmopts` which is the file sbt natively supports

Naive approach, just trying to help, not sure if that's what you expected? 😄  It seems sbt doesn't support a custom filename for jvm opts

### DIFF
--- a/.github/workflows/binary-compatibility-checks.yml
+++ b/.github/workflows/binary-compatibility-checks.yml
@@ -36,16 +36,19 @@ jobs:
       - name: Cache Coursier cache
         uses: coursier/cache-action@v6.4.0
 
+      - name: Enable jvm-opts
+        run: cp .jvmopts-ci .jvmopts
+
       # compile code instead of doing mima check (until we re-enable mima check)
       - name: Compile code
-        run: sbtx -jvm-opts .jvmopts-ci "+~ ${{ matrix.scalaVersion }} Test/compile"
+        run: sbt "+~ ${{ matrix.scalaVersion }} Test/compile"
 
       # disable mima check until first pekko release is done
       #- name: Report MiMa Binary Issues
-      #  run: sbtx -jvm-opts .jvmopts-ci "+~ ${{ matrix.scalaVersion }} mimaReportBinaryIssues"
+      #  run: sbt "+~ ${{ matrix.scalaVersion }} mimaReportBinaryIssues"
 
       #- name: Check correct MiMa filter directories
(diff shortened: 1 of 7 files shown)


pjfanning (approved):
lgtm

mdedetrich (changes requested):
Is there a reason why we are doing a `cp .jvmopts-ci .jvmopts` instead of just renaming the `.jvmopts-ci` file to `.jvmopts`?

I just did a quick grep and `.jvmopts-ci` seems to only be used by sbtx in ci so it seems to be pretty safe?

jchapuis:
@mdedetrich yes indeed, we could rename, that's the alternative. However, naturally, this means these options would also be picked up when starting sbt locally, which is maybe not what we want 🤔 ? 
Do we actually know if these options are necessary at all? some of them might no longer be relevant, e.g. `UseCompressedOops` seems to be a default nowadays

mdedetrich:
> yes indeed, we could rename, that's the alternative. However, naturally, this means these options would also be picked up when starting sbt locally, which is maybe not what we want 🤔 ?

So the underlying point of the issue and this PR is that `sbtx` is actually not used in any distribution if you install sbt the standard way. A lot of functionality over time has moved from `sbtx` to `sbt` and today there is no reason to use sbt. This is why @SethTisue was annoyed with `setup-scala` aliasing `sbt` to `sbtx` because its misleading.

The tl;dr is that nowadays if you have sbt on Linux/Mac/Windows its just using standard sbt (not `sbtx`) and its ignoring those cli options anyways

> Do we actually know if these options are necessary at all? some of them might no longer be relevant, e.g. UseCompressedOops seems to be a default nowadays

That is a fair point but I would fix this in a separate PR just to make things clear.

jchapuis:
@mdedetrich thanks yes I think I understand the context of this issue. My only concern was that by renaming `.jvmopts-ci` to `.jvmopts` in the repo, when you start `sbt` (not `sbtx`, just the standard `sbt`), it will actually pickup these options for your local development session (which could be counter-intuitive, as it's "hidden"). Compared to sbtx, it just doesn't allow specifying the file - but the functionality is there with the hardcoded name.  

If that's acceptable, for sure we can rename it and ditch the preliminary copying step. 

But I was guessing here that it had been kept separate to make it clear that these options were only relevant for the CI, and so I thought it relevant to keep this intention 🤷

mdedetrich (approved):
LGTM, questions answered

PR #83, opened 2023-01-05
Add sbt UsefulTask for publishLocal

mdedetrich:
This is particularly helpful if you want to try local pekko changes in another project

### DIFF
--- a/project/PekkoBuild.scala
+++ b/project/PekkoBuild.scala
@@ -280,6 +280,7 @@ object PekkoBuild {
         UsefulTask("", "compile", "Compile the current project"),
         UsefulTask("", "test", "Run all the tests "),
         UsefulTask("", "testOnly *.AnySpec", "Only run a selected test"),
+        UsefulTask("", "publishLocal", "Publish current snapshot version to local ~/.ivy2 repo"),
         UsefulTask("", "verifyCodeStyle", "Verify code style"),
         UsefulTask("", "applyCodeStyle", "Apply code style"),
         UsefulTask("", "sortImports", "Sort the imports"),


project/PekkoBuild.scala
```diff
@@ -280,6 +280,7 @@ object PekkoBuild {
         UsefulTask("", "compile", "Compile the current project"),
         UsefulTask("", "test", "Run all the tests "),
         UsefulTask("", "testOnly *.AnySpec", "Only run a selected test"),
+        UsefulTask("", "publishLocal", "Publish current snapshot version to local ~/.ivy2 repo"),
```
pjfanning, line 283: how about Coursier (latest sbt defaults to this) and maven?

mdedetrich: Are you asking if the latest sbt picks up this local repo?

pjfanning: sort of 2 questions?
1. does publishLocal only write to local Ivy repo?
2. regardless of whether publishLocal only published to to local Ivy repo, other sbt builds can still use the local jars (given that you set the right version to use in the upstream builds)?

mdedetrich: > does publishLocal only write to local Ivy repo?

Yes and this is intentional

> regardless of whether publishLocal only published to to local Ivy repo, other sbt builds can still use the local jars (given that you set the right version to use in the upstream builds)?

Yes, there is a default resolver called `Resolver.defaultLocal` which allows you to easily resolve projects published locally in `~/.ivy2/local` folder. I think this may already be included by default in SBT, if its not you can just do

```
resolvers += Resolver.defaultLocal
```

I do this workflow all the time, i.e. testing if current changes being applied to akka work in some other project.

pjfanning (approved):
lgtm

Issue #56, opened 2022-11-17
Submit pekko to scala-steward?


He-Pin:
I think only Akka is there, but not pekko.

jxnu-liguobin:
Why did you become a member? 😅

He-Pin:
> Why did you become a member? 😅 

MAKE YOUR PROGRESS BROS

mdedetrich:
I was also thinking of this, nothing against it. Just need to find the right time to do it.

Only thing that comes to mind is that we need to be quite cautious at least when it comes to pekko core using scala steward due to strict binary compatibility guarantees however for repos like pekko-connectors it makes definite sense.

pjfanning:
For v1.0.0 release, we probably only want to upgrade dependency jar versions where there is a security issue or serious bug in the dependency jar.

mdedetrich:
> For v1.0.0 release, we probably only want to upgrade dependency jar versions where there is a security issue or serious bug in the dependency jar.

Actually for that same reason I think it might make sense to just add scala-steward to the `1.1.x` branch when its setup. If there is some CVE/serious bug in a dependency we would likely know about it aside from scala-steward.

pjfanning:
https://github.com/apache/incubator-pekko/pull/86 might be a useful first step - allowing users to run `sbt dependencyUpdates`.

The output is currently:
```
[info] Found 1 dependency update for pekko-root
[info]   org.scala-lang:scala-library : 2.13.8 -> 2.13.10
[info] Found 1 dependency update for pekko-protobuf-v3
[info]   com.google.protobuf:protobuf-java:optional;provided : 3.16.1 -> 3.16.3 -> 3.21.12
[info] Found 2 dependency updates for pekko-actor
[info]   org.scala-lang.modules:scala-java8-compat : 1.0.0  -> 1.0.2  
[info]   org.scala-lang:scala-library              : 2.13.8 -> 2.13.10
[info] Found 4 dependency updates for pekko-pki
[info]   com.hierynomus:asn-one       : 0.5.0             -> 0.6.0          
[info]   org.scala-lang:scala-library : 2.13.8 -> 2.13.10                   
[info]   org.scalatest:scalatest:test : 3.1.4             -> 3.2.14         
[info]   org.slf4j:slf4j-api          : 1.7.36                      -> 2.0.6
[info] Found 4 dependency updates for pekko-osgi
[info]   ch.qos.logback:logback-classic:test : 1.2.11            -> 1.4.5                    
[...]

PR #86, opened 2023-01-05
add sbt-updates plugin

pjfanning:


mdedetrich:
What are your thoughts on using scala-steward instead? Its a fully configurable bot that automatically creates PR's with dependency updates that is used throughout Scala OS ecosystem (scala/scala center projects themselves even use it).

pjfanning:
the PR is based on https://github.com/apache/incubator-pekko/issues/56#issuecomment-1329221043 and the following comment with the version upgrades

it would probably be very noisy to have our PR queue filled up by scala steward when there are zero of those PRs that will be accepted right now

with just having the plugin, at least we can check manually to see if there are new updates

mdedetrich:
Was editing the post while you replied, on my point here

> There is also an argument that this should be a global sbt plugin (which you would place in ~/.sbt/1.0/plugins.sbt for example) with the same reasoning as to having global git settings.

How do you feel about adding a section in `CONTRIBUTING.md` on how to add such a global sbt plugin instead? I have also done this in the past with sbt-dependency and similar plugins.

pjfanning:
sure - I guess this is the sort of plugin that works well as a global one

PR #89, opened 2023-01-06
Record fork base commit and version in README

kw217:
It's useful for potential users of Pekko to know what version of Akka it corresponds to. This PR adds that info to the README. It also adds the precise commit of the fork, since that's likely to be of interest to us and to others in the future.

I'd like to propose tagging that commit as well, as akka-fork-base or similar - do you agree this would be a good idea?

See also https://github.com/apache/incubator-pekko-http/pull/20 which does the same for pekko-http.


pjfanning:
I have just tagged the repo at the commit - see https://github.com/apache/incubator-pekko/tags

pjfanning (changes requested):
I would be -1 on this PR - I think git tags are a better solution - and Github provides good UI support for tags and releases

Issue #32, opened 2022-11-08
Replace akka-paradox plugin with paradox


softinio:
As a quick way to remove the akka branding, lets replace [akka-paradox](https://github.com/akka/akka-paradox) plugin with [paradox](https://github.com/lightbend/paradox) basic theme. We can make better with more Pekko branding later.

As discussed in https://github.com/apache/incubator-pekko/issues/28 with @nvollmar

softinio:
If no objection, I will start working on this tomorrow.

mdedetrich:
@seglo @softinio @pjfanning So I tried to do just simply replace the akka paradox theme with the standard one and this made me fall down a rabbit hole. Here are the problems

* Due to the current Pekko sbt build only loading on an ancient JDK 1.8 version (it actually refuses to load otherwise) we are forced to downgrade to a version of parboiled that is bytecode compatible with JDK 1.8. I have managed to do this locally
* Although initially it might make sense to just replace `addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.44")` with the various sbt paradox plugins initially, this will have to be repeated for every single pekko project. Furthermore at some point we would actually like to create a pekko theme, and in order to not repeat boilerplate in the 10+ repos we maintain it makes sense to create a `sbt-paradox-pekko` sbt plugin akin to [sbt-paradox-akka](https://github.com/akka/akka-paradox). Initially it can be blank and essentially just contain the following dependencies
  ```sbt
  addSbtPlugin("com.lightbend.paradox" % "sbt-paradox" % 0.10.3")
  addSbtPlugin("com.lightbend.paradox" % "sbt-paradox-apidoc" % "1.0.0")
  addSbtPlugin("com.lightbend.paradox" % "sbt-paradox-project-info" % "3.0.0")
  ```
  but later on an actual css/js theme can be added. I already have a sample `sbt-paradox-pekko` locally .

softinio:
@mdedetrich thanks for pushing this over the finish line. Last quarter I ended having unplanned bandwidth shortage. Happy New Year. Looking forward to contributing this quarter.

PR #90, opened 2023-01-07
Remove readiness levels

mdedetrich:
In trying to replace the akka paradox theme with just the basic paradox theme, one issue I found is that akka/lightbend projects use the concept of readiness levels which comes from sbt-paradox-project-info https://github.com/lightbend/sbt-paradox-project-info#readiness-levels.

While it is possible to provide our own readiness-levels, this doesn't make too much sense for Apache projects. Also in doing so we would have to create our own sbt-paradox-pekko-project-info (akin to https://github.com/lightbend/sbt-paradox-lightbend-project-info) which is too much overhead for now.

If we do end up deciding on readiness levels this should be done via an Apache process and we can always add it in later.

References: https://github.com/apache/incubator-pekko/issues/32

### DIFF
--- a/project/project-info.conf
+++ b/project/project-info.conf
@@ -32,13 +32,6 @@ project-info {
   actor: ${project-info.shared-info} {
     title: "Pekko Actors (classic)"
     jpms-name: "pekko.actor"
-    levels: [
-      {
-        readiness: Supported
-        since: "2012-03-06"
-        since-version: "2.0"
-      }
-    ]
     api-docs: [
       {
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
these readiness levels are not currently useful for Pekko (we may need them in future but we can add them back at that point)

mdedetrich:
> these readiness levels are not currently useful for Pekko (we may need them in future but we can add them back at that point)

Yeah the only levels I see making sense are `Supported` and `Not Supported` (with `Supported` implying `Community`). This is however something we have to agree on as a community

PR #93, opened 2023-01-07
Fix paradox project info artifact names

mdedetrich:
When looking at the docs I noticed that the `artifact` fields in `project-info` still had the `akka-` prefix.

### DIFF
--- a/docs/src/main/paradox/typed/guide/modules.md
+++ b/docs/src/main/paradox/typed/guide/modules.md
@@ -40,7 +40,7 @@ This page does not list all available modules, but overviews the main functional
   symbol1=PekkoVersion
   value1="$pekko.version$"
   group=org.apache.pekko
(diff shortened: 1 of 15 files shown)


docs/src/main/paradox/typed/guide/modules.md
```diff
@@ -233,7 +233,7 @@ Distributed Data is intended to solve the following challenges:
   symbol1=PekkoVersion
   value1="$pekko.version$"
   group=org.apache.pekko
-  artifact=akka-stream-typed_$scala.binary.version$
```
mdedetrich, line 236: There is no such thing as akka-stream-typed, this was incorrect even before all of the pekko related changes.

pjfanning (approved):
lgtm

gmethvin (approved):

PR #94, opened 2023-01-09
Adds whitelist for custom serializer of classes in pekko package

nvollmar:
We implemented this in our internal Akka fork until we can migrate to Pekko.
The query view implementation need access to certain Actor internals and therefore has to reside within the akka/pekko package. As we want to use our persistence serializer for snapshots, this normally generates a warning which we want to ignore.

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -742,7 +742,7 @@ pekko {
     warn-on-no-serialization-verification = on
 
     # list of fqcn of classes that may use non-pekko serializer within pekko package without warn log
-    warn-non-pekko-serializer-whitelist = ${?akka.actor.warn-non-pekko-serializer-whitelist} []
+    warn-non-pekko-serializer-allow-list = ${?pekko.actor.warn-non-pekko-serializer-allow-list} []
 
     # Entries for pluggable serializers and their bindings.
     serializers {
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -741,6 +741,9 @@ pekko {
     # to reduce noise.
     warn-on-no-serialization-verification = on
 
+    # list of fqcn of classes that may use non-pekko serializer within pekko package without warn log
+    warn-non-pekko-serializer-whitelist = ${?akka.actor.warn-non-pekko-serializer-whitelist} []
+
     # Entries for pluggable serializers and their bindings.
     serializers {
       java = "org.apache.pekko.serialization.JavaSerializer"
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -741,7 +741,7 @@ pekko {
     # to reduce noise.
     warn-on-no-serialization-verification = on
 
-    # list of fqcn of classes that may use non-pekko serializer within pekko package without warn log
+    # list of fqcn of classes that may use non-pekko serializer within pekko package without warn log (added in Pekko 1.1.0)
     warn-non-pekko-serializer-allow-list = ${?pekko.actor.warn-non-pekko-serializer-allow-list} []
 
     # Entries for pluggable serializers and their bindings.
--- a/actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
+++ b/actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
@@ -464,7 +464,7 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
 
   @nowarn("msg=deprecated")
   private def warnUnexpectedNonPekkoSerializer(clazz: Class[_], ser: Serializer): Boolean = {
-    import scala.collection.JavaConverters._ // switch to scala.jdk.CollectionConverters once Scala 2.12 support is dropped
+    import pekko.util.ccompat.JavaConverters._
     if (clazz.getName.startsWith("org.apache.pekko.") && !ser.getClass.getName.startsWith("org.apache.pekko.") &&
       !system.settings.config.getStringList("pekko.actor.warn-non-pekko-serializer-allow-list").asScala.toSet.contains(
         clazz.getName)) {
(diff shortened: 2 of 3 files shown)


nvollmar:
I can open an issue matching the PR

nvollmar:
refs #95

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
```diff
@@ -457,8 +454,12 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
     }
   }
 
+  @nowarn("msg=deprecated")
   private def warnUnexpectedNonAkkaSerializer(clazz: Class[_], ser: Serializer): Boolean = {
```
pjfanning, line 458: Could we change the function name to use 'NonPekko'?

actor/src/main/resources/reference.conf
```diff
@@ -739,6 +739,9 @@ pekko {
     # to reduce noise.
     warn-on-no-serialization-verification = on
 
+    # list of fqcn of classes that may use non-pekko serializer within pekko package without warn log
+    warn-non-pekko-serializer-whitelist = ${?akka.actor.warn-non-pekko-serializer-whitelist} []
```
pjfanning, line 743: Could we call this an `allow-list` instead of `whitelist` - to avoid the connotations about colors?

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
nvollmar: Makes sense

actor/src/main/resources/reference.conf
nvollmar: sure...

pjfanning (commented):
would it be possible to add a unit test?

nvollmar:
> would it be possible to add a unit test?

I can look into that, just the feature to write that log doesn't have a test in the first place. Do we expect tests for logs?

pjfanning (approved):
lgtm - it is a small change away from how Akka 2.6 works but should be fine

nvollmar:
It (should) be a non-functional change as it is only affecting logging, but I'd also be fine waiting for 1.1 to merge it.

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
```diff
@@ -457,8 +453,12 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
     }
   }
 
-  private def warnUnexpectedNonAkkaSerializer(clazz: Class[_], ser: Serializer): Boolean = {
-    if (clazz.getName.startsWith("org.apache.pekko.") && !ser.getClass.getName.startsWith("org.apache.pekko.")) {
+  @nowarn("msg=deprecated")
+  private def warnUnexpectedNonPekkoSerializer(clazz: Class[_], ser: Serializer): Boolean = {
+    import scala.collection.JavaConverters._ // switch to scala.jdk.CollectionConverters once Scala 2.12 support is dropped
```
He-Pin, line 458: add a TODO here?

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
```diff
@@ -457,8 +453,12 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
...
+    if (clazz.getName.startsWith("org.apache.pekko.") && !ser.getClass.getName.startsWith("org.apache.pekko.") &&
```
He-Pin, line 459: extract `"org.apache.pekko.` to a constant field?

mdedetrich: This might make sense in a separate PR, at least when I was renaming the package to `org.apache.pekko` there were a few cases where package names was being checked `clazz.getName`?

jrudolph (commented):
It does not seem useful enough for me to include the extra complexity here (feels like a single use feature). After all it seems query views (where exactly is the problem?) are currently meant to be extended in the way you are doing. If we want to change that, we should do it there.

So there seem to be better solutions than to add something here:
 * change query views no to be `private[pekko]`
 * filter the warning in the application

nvollmar:
Just updated to latest main.

@jrudolph Didn't see your comment back then. The QueryView requires access to certain pekko private methods to implement snapshot/event recovery and so on.

actor/src/main/resources/reference.conf
```diff
@@ -741,6 +741,9 @@ pekko {
     # to reduce noise.
     warn-on-no-serialization-verification = on
 
+    # list of fqcn of classes that may use non-pekko serializer within pekko package without warn log
+    warn-non-pekko-serializer-allow-list = ${?pekko.actor.warn-non-pekko-serializer-allow-list} []
```
pjfanning, line 745: might be worth updating the comment to mention that this is new to Pekko v1.1.0

nvollmar: added

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
```diff
@@ -466,8 +462,12 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
     }
   }
 
+  @nowarn("msg=deprecated")
   private def warnUnexpectedNonPekkoSerializer(clazz: Class[_], ser: Serializer): Boolean = {
-    if (clazz.getName.startsWith("org.apache.pekko.") && !ser.getClass.getName.startsWith("org.apache.pekko.")) {
```
Roiocam, line 470: I don't think it is good idea to turn off these warnning, if have to, better using same way like `shouldWarnAboutJavaSerializer` rather than using a list struct.

actor/src/main/resources/reference.conf
He-Pin: What's the usercase for this :) I'm asking because I'm not a persistence user:)

PR #97, opened 2023-01-09
add link to CLA page

pjfanning:


kw217 (changes requested):
This is supposed to be "The TL;DR of the above very precise workflow version", so you need to add step 5 to the above lengthy version as well. That can also explain what "if necessary" means (i.e., not required for minor changes).

kw217 (approved):
LGTM, thanks!

He-Pin (approved):

PR #105, opened 2023-01-16
Publish nightlies to Apache Nexus Repository

mdedetrich:
Draft, TDB

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -48,6 +48,12 @@ jobs:
         with:
           java-version: adopt@1.11
 
+      - name: Publish to Apache Maven repo
+        env:
+          NEXUS_USER: ${{ secrets.NEXUS_USER }}
+          NEXUS_PW: ${{ secrets.NEXUS_PW }}
+        run: sbt +publish
+
       - name: Publish local
         run: |-
           sudo apt-get install graphviz
--- a/project/Publish.scala
+++ b/project/Publish.scala
@@ -15,20 +15,17 @@ package org.apache.pekko
 
 import sbt._
 import sbt.Keys._
-import java.io.File
 import com.lightbend.sbt.publishrsync.PublishRsyncPlugin.autoImport.publishRsyncHost
 import xerial.sbt.Sonatype.autoImport.sonatypeProfileName
 
 object Publish extends AutoPlugin {
 
-  val defaultPublishTo = settingKey[File]("Default publish directory")
-
   override def trigger = allRequirements
 
   override lazy val projectSettings = Seq(
-    publishTo := Some(akkaPublishTo.value),
+    publishTo := Some(pekkoPublishTo.value),
     publishRsyncHost := "akkarepo@gustav.akka.io",
-    credentials ++= akkaCredentials,
+    credentials ++= apacheNexusCredentials,
     organizationName := "Apache Software Foundation",
     organizationHomepage := Some(url("https://www.apache.org")),
     sonatypeProfileName := "org.apache.pekko",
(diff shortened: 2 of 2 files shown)


project/Publish.scala
```diff
@@ -21,42 +21,39 @@ import xerial.sbt.Sonatype.autoImport.sonatypeProfileName
...
-    publishTo := Some(akkaPublishTo.value),
+    publishTo := Some(pekkoPublishTo.value),
     publishRsyncHost := "akkarepo@gustav.akka.io",
-    credentials ++= akkaCredentials,
-    organizationName := "Lightbend Inc.",
-    organizationHomepage := Some(url("https://www.lightbend.com")),
-    sonatypeProfileName := "com.typesafe",
-    startYear := Some(2009),
+    credentials ++= pekkoCredentials,
+    organizationName := "Apache Inc.",
```
mdedetrich, line 30: @pjfanning Do all of these settings look correct to you?

.github/workflows/publish-nightly.yml
```diff
@@ -51,6 +51,12 @@ jobs:
         with:
           java-version: adopt@1.11
 
+      - name: Publish to Apache Maven repo
+        env:
+          NEXUS_USER: ${{ secrets.NEXUS_USER }}
+          NEXUS_PW: ${{ secrets.NEXUS_PW }}
+        run: sbt +publish
+
```
seglo, line 59: I think we should only publish milestone release artifacts to Apache's maven repo, otherwise the nightlies deployment is redundant. Milestone releases should be human-initiated, probably using a release-by-tag pattern like Akka currently does.

project/Publish.scala
pjfanning: Apache Software Foundation - I think Apache Inc is a private and unrelated company

.github/workflows/publish-nightly.yml
seglo: > For me the main intention to publish to apache's maven repo is so I can resolve it as a standard sbt dependency

You can resolve releases published to nightlies as a standard sbt dependency if you add it to your list of resolvers (`resolvers += "Apache Pekko Snapshots" at "https://nightlies.apache.org/pekko/snapshots"`). 

> which I want to do with pekko-http in order to rename the package to org.apache.pekko

In lieu of a milestone release you could temporarily add a reference to a nightly snapshot. That would unblock getting the satellite projects to build. Whether a snapshot is published to nightlies or Apache's maven it doesn't really make a difference because you would still need to add a resolver for sbt to resolve the artifact.

I was thinking that we just treat nightlies as the source for snapshots and then milestone releases go through the usual maven publishing workflow.

pjfanning: I tend to agree with @seglo - that the nightlies are enough

mdedetrich: So after manually checking each apache project in https://nightlies.apache.org/, whats sticking out as a sore thumb for me is that we are the **only** Apache project that is using the nightlies folder as a pseudo Nexus repo. The content that I found in the Apache nightly projects all seem to be one of the following

* Binaires/fatjar's with `Main` for the Apache projects that happen to be applications rather than libraries (which makes sense since nightlies just being a remote directory is an ideal place to put binaries for distribution so people can test it)
* Generated documentation
* Some sought of microsite

On the other hand if you look at the Apache maven snapshots directory at https://repository.apache.org/content/groups/snapshots/org/apache/ you can can clearly see that it is the primary mechanism being used to deliver snapshots (which to me are the fulfil the same intent as nightlies if we are talking about libraries, the main difference between the two is largely etymological due to history and nightlies being a subset of snapshots depending on what usage pattern you want to use). Furthermore just by quickly counting to get a rough idea of how many projects thee are, you can also see that nightlies are rarely used where as the Apache nexus snapshots is being used by what appears to be almost every Apache JVM project. Of course different projects have different strategies on what constitutes a snapshot and how far in history we go (the only thing that is uncertain to me is whether its possible for Apache to configure automatic `-SNAPSHOT` expiry for artifacts (this is possible in Nexus, just need to see if its possible for Aiven to configure it).

To a degree it also pains me to say this because of the time spent on it, but I am getting the nagging feeling that we spent a lot of effort in emulating a standard nexus deployment (because we are now "deploying" using rsync) rather than just setting up repository in sbt taking 4-5 lines (as shown in the PR) which would also have the added bonus advantage of integrating much more fluently with tools like sbt-release/sbt-ci-release, a strategy that pretty much every other Scala OS project uses for publishing (including snapshots).

Maybe the most considerate approach here is to leave the current nightlies approach as it is and come up with another strategy for snapshots that is considered useful (also with some beforehand research for other projects to see what they do)?

> You can resolve releases published to nightlies as a standard sbt dependency if you add it to your list of resolvers (resolvers += "Apache Pekko Snapshots" at "https://nightlies.apache.org/pekko/snapshots").

You are right, I believe this should work fine but I just realized that in order to integrate this in pekko without some hacky workaround I would need to solve https://github.com/apache/incubator-pekko/issues/107 because sbt's `%%` dependency will resolve the same version against all Scala versions which fails if the versions differ as they do now.

mdedetrich: After thinking about this, I will make a thread on the dev mailing list to discuss because to me we need to agree on some process on how to handle snapshot and snapshot vs nightly.

mdedetrich: Discussion posted in mailing list at https://lists.apache.org/thread/p5s8ysypyd2l2slb3o9f2v5vrf8dgsx8

mdedetrich: Resolving, already answered in mailing list

mdedetrich:
PR now ready to review. Note that there is a chance that it may not work (i.e. some credential or repository setting is wrong), but we will find out when the nightly job runs and then if there is a problem I will fix it in a follow up PR.

pjfanning (approved):
lgtm

PR #106, opened 2023-01-16
Publish nightly once a day

seglo:
I think this makes more sense otherwise snapshots will quickly become truncated or if we changed the eviction strategy to actually be time based we will end up with a lot of artifacts on nightly.

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -28,9 +28,6 @@ on:
         default: 30
   schedule:
     - cron: "0 0 * * *"
-  push:
-    branches:
-      - 'main'
 
 permissions:
   contents: read


mdedetrich:
To clarify, the effect on this change is that instead of publishing nightly only when something is pushed onto `main` (i.e. a PR is merged) that it occurs on any branch?

I think I am misunderstanding something here.

seglo:
No, the opposite. Only publish once per day.

mdedetrich (approved):
Oh I see now, previously the workflow would execute every single time a PR would be merged into master in addition to cron execution, and the removal of

```
  push:
    branches:
      - 'main'
```

means it only executes when cron executes. Approving

PR #128, opened 2023-01-26
Fix minor bugs and improve nightly publish-nightly.yml

mdedetrich:
There was some issues in the nightly build, for example graph viz not being available for the standard publish. Also refactoring the pipeline a bit to make it more clear


jrudolph (approved):
lgtm

Issue #103, opened 2023-01-16
Publish nightlies to official Apache Maven repository


mdedetrich:
In https://github.com/apache/incubator-pekko/pull/60 we managed to get nightly builds published to the Apache directory. While this is the official place to publish Apache nightlies, its technically not a maven repository (its just a directory where you rsync the contents).

Apache does have an official maven where you can publish snapshots too (see https://repository.apache.org/content/groups/snapshots/), so in addition to rsyncing our nightlies we should also publish this as an a snapshot to Apache's maven repo.

mdedetrich:
Apache JIRA Infra ticket created at https://issues.apache.org/jira/browse/INFRA-24087

PR #131, opened 2023-01-26
Rename name for Apache nexus repo to work with Sonatype

mdedetrich:
Since our sonatype is having a bit of an identity crisis having to look like an Apache's nexus repo, I need to change the name of the credential, see

```
[warn] versionScheme setting is empty; set `ThisBuild / versionScheme := Some("early-semver")`, `Some("semver-spec")` or `Some("pvp")`
[warn] so tooling can use it for eviction errors etc - https://www.scala-sbt.org/1.x/docs/Publishing.html
[error] Unable to find credentials for [Sonatype Nexus Repository Manager @ repository.apache.org].
[error]   Is one of these realms misspelled for host [repository.apache.org]:
[error]   * Apache Nexus Repository Manager
```

from the last run at https://github.com/apache/incubator-pekko/actions/runs/4015355806/jobs/6896983341

Related to https://github.com/apache/incubator-pekko/issues/103

### DIFF
--- a/project/Publish.scala
+++ b/project/Publish.scala
@@ -44,7 +44,7 @@ object Publish extends AutoPlugin {
   private def apacheNexusCredentials: Seq[Credentials] =
     (sys.env.get("NEXUS_USER"), sys.env.get("NEXUS_PW")) match {
       case (Some(user), Some(password)) =>
-        Seq(Credentials("Apache Nexus Repository Manager", apacheBaseRepo, user, password))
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm - Nexus is a product developed by Sonatype, that's probably why name has Sonatype instead of Apache

jrudolph (approved):
lgtm, probably the realm wasn't changed from the default config and then couldn't be changed without breaking everyone... (PR and commit title still a bit misleading ;) )

PR #135, opened 2023-01-31
docs: Apache Pekko naming

pjfanning:
We're going to have to go through docs page by page to decide when it's necessary to use 'Apache Pekko' as opposed to `Pekko`.

We should probably always start a page by referring to 'Apache Pekko' but later in the page, we might start shortening it to Pekko.

Current docs on https://pekko.apache.org highlight that we don't use 'Apache Pekko' enough.

This PR focuses on a few very visible instances that affect the table of contents.

### DIFF
--- a/docs/src/main/paradox/additional/deploying.md
+++ b/docs/src/main/paradox/additional/deploying.md
@@ -1,5 +1,5 @@
 ---
-project.description: How to deploy Pekko Cluster to Kubernetes and Docker.
+project.description: How to deploy Apache Pekko Cluster to Kubernetes and Docker.
 ---
 # Deploying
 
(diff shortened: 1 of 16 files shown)


docs/src/main/paradox/additional/deploying.md
```diff
@@ -1,5 +1,5 @@
 ---
-project.description: How to deploy Pekko Cluster to Kubernetes and Docker.
+project.description: How to deploy Apache Pekko Cluster to Kubernetes and Docker.
```
jchapuis, line 2: same here I guess? "an Apache Pekko Cluster"

jchapuis (approved):
lgtm apart from these small typos indeed

docs/src/main/paradox/additional/deploying.md
pjfanning: done

Issue #130, opened 2023-01-26
Setup the release process


jrudolph:
Apache has certain requirements for releasing a public version. We should strive for as much automation as possible. This ticket should be an overview over all steps necessary and the progress on those items. Please add items as needed.

Let's only consider public releases like RCs and GAs here (but not snapshots).

References:
 * [Apache Release Distribution Policy](https://infra.apache.org/release-distribution.html)
 * [Apache Publishing Maven Artifacts](https://infra.apache.org/publishing-maven-artifacts.html)
 * [Apache Release Signing](https://infra.apache.org/release-signing.html)

## Organizational steps
 * [ ] Create release template to keep track progress of release process per release

## Maven-related steps

 * [ ] Publish to Apache Nexus into staging
 * [ ] Initiate voting
 * [ ] Propagate to Maven Central when vote is positive
 * [ ] ...

## Apache-related steps

 * [ ] (one-time) setup signing infrastructure (how are keys created, where can a release be signed)
 * [ ] Create signed release and upload artifacts (where and when does that happen?)
 * [ ] ...

mdedetrich:
I believe one thing that we have to do under Apache-related steps is that we have to copy the sources (not the generated binaries, i.e. jars in our case) to an SVN repo. In Apache lingo I think this is called a "source package", see https://www.apache.org/legal/release-policy.html#source-packages.

In our case it would just be the contents of the git repo.

jrudolph:
> I believe one thing that we have to do under Apache-related steps is that we have to copy the sources (not the generated binaries, i.e. jars in our case) to an SVN repo.

Just creating the source archive shouldn't be a big problem.

One of the main questions is if we can trust GHA enough to make releases there and also sign those releases (it probably makes sense to do both on one machine). Also, even if Apache considers binaries only a convenience, the reality is that > 99% of all users will use the binaries from Maven Central which after all carry a much bigger risk than the distributed source files. If we are not allowed to release from GHA or decide against it, we would have to setup something that would allow to run safe and reproducible releases from the release manager's machine (e.g. using docker).

mdedetrich:
[...]

PR #143, opened 2023-02-02
Add SourceDistPlugin for creating source packages

mdedetrich:
Adds the prototype source dist plugin which generates source packages for Apache

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,9 +100,12 @@ lazy val root = Project(id = "pekko", base = file("."))
       akkaScalaNightly,
       docs,
       serialversionRemoverPlugin))
-  .settings(Compile / headerCreate / unmanagedSources := (baseDirectory.value / "project").**("*.scala").get)
+  .settings(
+    Compile / headerCreate / unmanagedSources := (baseDirectory.value / "project").**("*.scala").get,
+    sourceDistName := "incubator-pekko")
   .settings(PekkoBuild.welcomeSettings)
   .enablePlugins(CopyrightHeaderForBuild)
+  .enablePlugins(SourceDistPlugin)
 
 lazy val actor = pekkoModule("actor")
   .settings(Dependencies.actor)
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -26,4 +26,5 @@ addSbtPlugin("net.bzzt" % "sbt-reproducible-builds" % "0.30")
 
 addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.5.10")
 addSbtPlugin("com.lightbend.sbt" % "sbt-publish-rsync" % "0.2")
+addSbtPlugin("com.github.pjfanning" % "sbt-source-dist" % "0.1.2")
 addSbtPlugin("com.github.reibitto" % "sbt-welcome" % "0.2.2")


build.sbt
```diff
@@ -100,9 +100,12 @@ lazy val root = Project(id = "pekko", base = file("."))
       akkaScalaNightly,
       docs,
       serialversionRemoverPlugin))
-  .settings(Compile / headerCreate / unmanagedSources := (baseDirectory.value / "project").**("*.scala").get)
+  .settings(
+    Compile / headerCreate / unmanagedSources := (baseDirectory.value / "project").**("*.scala").get,
+    sourceDistName := "pekko")
```
pjfanning, line 105: incubator-pekko is more correct for the source distribution

mdedetrich: Fixed

pjfanning (approved):
lgtm

pjfanning (changes requested):
there was a problem with v0.1.0 of the plugin - some params passed to the generator were in wrong order (version and targetDir) - which meant the output files were getting created with wrong dir and names - I've published a v0.1.1 with a fix

mdedetrich:
@pjfanning Fixed

gmethvin (approved):

pjfanning (approved):

pjfanning:
v0.1.2 of the sbt plugin is available now

mdedetrich:
> v0.1.2 of the sbt plugin is available now

Will update

spangaer:
Sorry for the late comment, but wondering whether something like this was considered:
Query Git for an archive of the branch state: https://git-scm.com/docs/git-archive

I guess the archive can receive some post processing to added certain "generated files", like the "dist stuff" or
https://github.com/apache/incubator-pekko/issues/130#issuecomment-1421339657

pjfanning:
@spangaer since we've already written the sbt plugin, I don't see the real attraction to using this command line tool. If we hadn't written the plugin, the tool might have been an alternative.

project/plugins.sbt
```diff
@@ -26,4 +26,5 @@ addSbtPlugin("net.bzzt" % "sbt-reproducible-builds" % "0.30")
 
 addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.5.10")
 addSbtPlugin("com.lightbend.sbt" % "sbt-publish-rsync" % "0.2")
+addSbtPlugin("com.github.pjfanning" % "sbt-source-dist" % "0.1.2")
```
pjfanning, line 29: 0.1.3 is ready - it generates the sha256/sha512 digest files as well as the zip/tgz

spangaer:
Well, I'd think you'll always need a plugin to do more. I just saw things like "self handling" of .gitignore files, to which I think -> just let Git handle that. Where wrapping that command would help you.

But I fully understand I'm waaaay late with that suggestion.

jrudolph (approved):
lgtm

I compared `git archive` output with the output of this plugin and it seems to work well-enough (though being 10-100x slower than `git archive`). The most important thing here is that you start from a pristine checkout because files are included based on a blacklist and not on a whitelist so any extra file you might have in your workspace will be included.

I would have also preferred to simply use `git archive` but since the work is already done, it seems good enough.

mdedetrich:
> The most important thing here is that you start from a pristine checkout because files are included based on a blacklist and not on a whitelist so any extra file you might have in your workspace will be included.

We can add this as a check to sbt-source-dist so it won't work if the checkout is not pristine

> I would have also preferred to simply use git archive but since the work is already done, it seems good enough.

Its always possible to make sbt-source-dist just wrap `git archive` however arguably this is also an advantage of sbt-source-dist because its not reliant on git. Not relevant for pekko specifically but as a general Apache SBT plugin (which it may be in the future) this has theoretical relevance because not everyone uses git (then again, a good counter argument to this is to just detect if the machine has git and then proceed to use git archive).

pjfanning:
I agree that `git archive` would be a workable solution.

A few benefits of an sbt plugin
* the artifact names are set up in the sbt scripts so that users just call `sbt sourceDistGenerate` regardless of the repo - so they don't need to customise the commands for different repos. One call gets you the zip and tgz
* the `sourceDistGenerate` creates the SHA digests automatically
* we'll probably also get it use sbt-pgp to gen the gpg signing asc files
* checking for a clean unmodified repo is definitely needed

If the plugin proves problematic, we can look again at `git archive`.

mdedetrich:
> If the plugin proves problematic, we can look again at git archive.

Just a note on this (incase its not clear), there isn't any issue in sbt plugins just wrap existing cli's (such as `git` with `git archive`). In fact a lot of sbt plugins, even official ones do this (a good example is sbt-pgp).

jrudolph:
Just as quick info that in my checkout, the plugin somehow got into a loop and created a 50GB distribution file when I noticed and stopped it...

mdedetrich:
> Just as quick info that in my checkout, the plugin somehow got into a loop and created a 50GB distribution file when I noticed and stopped it...

Hmm, I am honestly thinking that in the medium term, let alone long term it would be easier to divert to git archive. @jrudolph Do you want to investigate https://github.com/pjfanning/sbt-source-dist/issues/14 ? 

@pjfanning Can you add @jrudolph as a maintainer of the repo as well?

mdedetrich:
We can also have a look into what https://github.com/Apache/daffodil does, its another Apache project that uses sbt

PR #148, opened 2023-02-07
replace use of unmaintained setup-scala action

pjfanning:
related to https://github.com/apache/incubator-pekko-connectors-kafka/pull/26

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,6 +1,6 @@
 name: Scalafmt
 
-permissions: read-all
+permissions: {}
 
 on:
   pull_request:
--- a/.github/workflows/binary-compatibility-checks.yml
+++ b/.github/workflows/binary-compatibility-checks.yml
@@ -9,6 +9,8 @@ on:
     tags:
       - v2.6.*
 
+permissions: {}
+
 jobs:
   check-binary-compatibility:
     name: Check / Binary Compatibility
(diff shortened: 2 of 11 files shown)


.github/workflows/format.yml
```diff
@@ -1,6 +1,6 @@
 name: Scalafmt
 
-permissions: read-all
+permissions: {}
```
pjfanning, line 3: I don't think read-all is needed. `{}` disables the jobs access to repository secrets - and this should be the go-to setting unless a job needs to read secrets

pjfanning: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

mdedetrich: Are you sure this is what you intend? `read-all` is more strict than `{}`, not less.

mdedetrich: Are you sure this is what you intend? `read-all` is more strict than `{}`, not less.

mdedetrich: Thanks for clarifying, had a misunderstanding then

mdedetrich (approved):
Lgtm

.github/workflows/format.yml
pjfanning: To be frank, the Github docs are not well written - leaving plenty of ambiguity.

I'm not sure why not setting permissions at all in the yml does not default to having the lowest available permissions.

A few months ago, there were people popping PRs on dozens of repos to set `permissions: {}` but honestly, they could have been wrong.

PR #152, opened 2023-02-09
try to fix nightly build issue

pjfanning:
I made a mistake in my last PR.

https://github.com/apache/incubator-pekko/pull/148/files#diff-e780cb03e0575bf97176af613cf9b23d9337cc6a8d309a7475704fc756cfcce3R166

I copy/pasted this from another project and inadvertently kept the wrong variable name (java-version instead of jdkVersion) but then reviewing the existing variable, I see now that setup-java requires numeric params - so I changed the matrix values and anything that used them. I renamed the param to javaVersion to match the naming style in this file.

I removed extraopts because it always appears to be set to ""

### DIFF
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -135,7 +135,7 @@ jobs:
       #      https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}
 
   jdk-nightly-build:
-    name: JDK ${{ matrix.jdkVersion }} / Scala ${{ matrix.scalaVersion }}
+    name: JDK ${{ matrix.javaVersion }} / Scala ${{ matrix.scalaVersion }}
     runs-on: ubuntu-20.04
     if: github.repository == 'apache/incubator-pekko'
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-builds.yml
```diff
@@ -145,25 +145,18 @@ jobs:
...
         scalaVersion: ["2.12", "2.13"]
-        jdkVersion: ["adopt@1.8.0", "adopt@1.11", "openjdk@1.17.0"]
-        include:
-          - jdkVersion: adopt@1.8.0
-            extraOpts: ""
-          - jdkVersion: adopt@1.11
-            extraOpts: ""
-          - jdkVersion: openjdk@1.17.0
-            extraopts: ""
+        javaVersion: [8, 11, 17]
```
jrudolph, line 148: Not sure about the numeric part, where did you get that from? E.g. see https://github.com/actions/setup-java#testing-against-different-java-versions

pjfanning: numeric values work (as do text ones) - https://github.com/apache/incubator-pekko-connectors-kafka/pull/26 is an example where I use numeric values (in a previous PR in pekko space)

jrudolph (approved):
lgtm

PR #155, opened 2023-02-09
diffs when running scalafmtAll

pjfanning:
I ran `sbt scalafmtAll javafmtAll scalafmtSbt` and got these diffs

Just thought it best to do a dedicated PR instead of having these appear in PR dedicated to fixing a particular issue (but that also needs scalafmt to be run - thus accidentally adding these unrelated diffs).


jrudolph (approved):
lgtm (How did they get through the checks? Were the checks only added after those files were last changed?)

pjfanning:
@jrudolph I don't really know why these diffs were not failing the build - I'm pretty sure the CI action is running

jrudolph:
> @jrudolph I don't really know why these diffs were not failing the build - I'm pretty sure the CI action is running

The GHA is only checking changed files, so maybe they were last checked before the check was added.

spangaer:
- Scalafmt version was updated too (which wouldn't be caught by GA as mentioned by @jrudolph )
- Sometimes a formatter needs a couple of iterations to converge
- ...
🤷

PR #159, opened 2023-02-09
acknowledge smhasher copyright but add apache license header

pjfanning:
I still need to work out if the LICENSE change is needed and if is not, then I can revert that bit.

If it is needed, I needed to change the bit that adds the LICENSE to jars because only the actor jar needs this version of the license - other jars need only the standard license.


justinmclean:
HI,

Note that MurmurHash3 is not ALv2 licensed. [1]

Justin

1. https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp

pjfanning:
@justinmclean I'm not really sure what to do with this file yet. One potential option is to remove it - because we might be able to use the class from the Scala jar instead. That would simplify life for us, I think.

I just hope that Lightbend Akka did not modify the copy we've inherited - in a way that makes using the Scala version break our tests and use cases.

pjfanning:
@justinmclean removing the Pekko MurmurHash code doesn't seem feasible. It is quite different from the Scala equivalent code.

* the reason I added the modified Apache v2 license header is that we generally agreed to put this on all Pekko source files in https://issues.apache.org/jira/browse/LEGAL-626
* The Scala LAMP/EPFL copyright header was already in the file (from the Lightbend/Akka code that Pekko started with)
* Scala itself is Apache v2 licensed - https://github.com/scala/scala
* The aappleby/smhasher code is Public Domain and the author has waved copyright

With all this in mind:
* is it ok to update the Pekko MurmurHash.scala file to add the LEGAL-626 header?
* is it ok to put the LAMP/EPFL copyright in our LICENSE file (for the source release)?
* only one of the jars that we will create from this project will contain this code, so I will adjust this PR to ensure that an appropriate LICENSE/NOTICE for that jar (which will match what we agree for the source release)

justinmclean:
Converting an algorithm from one language to another doesn't change its license. You need to respect the original license that it is under and ASF policy, and IMO this code is not under ALv2. See https://www.apache.org/legal/resolved.html#handling-public-domain-licensed-works for how to handle this.

Also, can we please fix the GitHub no-reply email on these messages as not everyone will want to use Github as it makes it impossible to reply to issues sent to the mailing list.

pjfanning:
@justinmclean the emails sent to notifications at pekko mailing list are generated by Github. I'm not aware of any feature in Github that lets me control the format of those emails and what the reply-to is on them.

We can take the conversation about the attempt to properly attribute the copyright and license for MurmurHash into a new email thread on dev at pekko mailing list, if you prefer.

My latest attempt at updating the Pekko license is to add this section to our LICENSE.

```
pekko-actor contains MurmurHash.scala which is derived from MurmurHash3,
written by Austin Appleby. He has placed his code in the public domain.
The author has disclaimed copyright to that source code.
```

I'm still quite confused about what to do about the LAMP/EPFL copyright that has been placed on [MurmurHash.scala](https://github.com/apache/incubator-pekko/blob/main/actor/src/main/scala/org/apache/pekko/routing/MurmurHash.scala).

Are you saying that we should ignore this in our LICENSE/NOTICE files? I presume that we leave the LAMP/EPFL copyright in the MurmurHash.scala file because it was already there.

justinmclean:
I've not had this problem with other ASF projects, possible something to configure in /.asf.yaml? We do have people involved at the ASF who refuse to get a GitHub account.

IMO LAMP/EPFL copyright was placed incorrectly on that MurmurHash file, but probably the best way to deal with it it to add the MurmurHash  header above the MurmurHash code as file can contain a mix of differently licensed code and to list that license in your LICENSE file as described [here](https://www.apache.org/legal/resolved.html#handling-public-domain-licensed-works).

pjfanning:
@mdedetrich I need to get more eyes on the license text but would you be able to review the technical aspects of these changes, when you have time?

* I've split out 3 separate special license files for 3 jars that have source code that does not originate from Lightbend or Pekko team
* the root LICENSE (source license) does not mention the pekko-protobuf-v3 module because the source repo does not include that source - but the build downloads the protobuf-java 3.16 jar and shades its classes - so we don't actually distribute this source per se but we do distribute classes in binary form (the pekko-protobuf-v3 jar) that derive from the Google protobuf source
* I'm happy to rename that `legal` directory if it is confusing - possibly renaming the dir to something like `jar-licenses`.

mdedetrich:
Will do, I will look at it by tonight (tomorrow latest)

mdedetrich (approved):
So I had a look at this and at least the technical aspects lgtm

jrudolph:
> LAMP/EPFL copyright was placed incorrectly on that MurmurHash file

Couldn't a port to another language be a derived work that could have its own copyright (in addition to the existing one)?

pjfanning:
> > LAMP/EPFL copyright was placed incorrectly on that MurmurHash file
> 
> Couldn't a port to another language be a derived work that could have its own copyright (in addition to the existing one)?

I've kept the LAMP/EPFL copyright header as well as adding one to acknowledge Austin Appleby's authorship of the original code.

jrudolph:
> I just hope that Lightbend Akka did not modify the copy we've inherited - in a way that makes using the Scala version break our tests and use cases.

In seems the Akka version predates the version in scala library:

2011/09/08: Scala: https://github.com/scala/scala/commit/52c1d019d63d2fffadc8f3f159d654187ac61ece#diff-d2dc3cf0fd4ebb89cc33b035534ad76f90c7f97e285b57ff4b95cc8fefbc0adf
2011/05/17: Akka: https://github.com/apache/incubator-pekko/commit/cfea06c5d50a50234249c701d7b64c8750f3455d

In any case, the hash values are not the same (but that doesn't mean that using the Scala version wouldn't work...).

pjfanning:
> > I just hope that Lightbend Akka did not modify the copy we've inherited - in a way that makes using the Scala version break our tests and use cases.
> 
> In seems the Akka version predates the version in scala library:
> 
> 2011/09/08: Scala: [scala/scala@52c1d01#diff-d2dc3cf0fd4ebb89cc33b035534ad76f90c7f97e285b57ff4b95cc8fefbc0adf](https://github.com/scala/scala/commit/52c1d019d63d2fffadc8f3f159d654187ac61ece#diff-d2dc3cf0fd4ebb89cc33b035534ad76f90c7f97e285b57ff4b95cc8fefbc0adf) 2011/05/17: Akka: [cfea06c](https://github.com/apache/incubator-pekko/commit/cfea06c5d50a50234249c701d7b64c8750f3455d)
> 
> In any case, the hash values are _not_ the same (but that doesn't mean that using the Scala version wouldn't work...).

The Scala and Akka classes are very different so I gave up on the idea of trying to drop the Akka one. So the ongoing PR is just come up with the right license and copyright text.

jrudolph:
After more digging the Akka one is based on the old version of MurmurHash which was removed after being deprecated in https://github.com/scala/scala/commit/5c7fd1bc039e0ea22b6b2d02596e7b8689c0213c for Scala 2.13.

Claudenw:
If you want a MurmurHash3 that is correctly impleented for the JVM and is under Apache License see commons-codec

https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/digest/MurmurHash3.html

jrudolph:
> If you want a MurmurHash3 that is correctly implemented for the JVM and is under Apache License see commons-codec

That link says

> This is public domain code with no copyrights.

pjfanning:
I'd prefer to proceed with this PR that keeps the existing code but tries to improve the license statements. The public domain smhasher code is not problematic. The real issue is to just acknowledge the use of the original code. The commons-codec code can be looked at if we cannot agree on the updated license text.

jrudolph (approved):
lgtm

pjfanning:
I've merged this to move things on. If anyone has any concerns, I'll be happy to make further changes or revert this.

Issue #149, opened 2023-02-08
Upload paradox/scala+java docs into nightlies folder


mdedetrich:
As discussed in https://the-asf.slack.com/archives/C0491A12EP2/p1675850262146849, at least initially there is a preliminary decision to upload generated paradox/scala+java docs into the apache nightlies (i.e. https://nightlies.apache.org/). The idea is the main website (which is separate and hosted at http://pekko.apache.org/) would point to the docs uploaded to nightlies. The reasons why to do this are

* Nightlies will automatically host any content that happens to have an `index.html` in a root folder. This makes it really trivial to host generated docs (see https://nightlies.apache.org/flink/flink-ml-docs-master/ as an example)
* Uploading to nightlies is very easy and almost all of the work has already been done, see https://github.com/apache/incubator-pekko/blob/main/.github/actions/sync-nightlies/action.yml. The only changes we would need to do is to run `sbt unidoc paradox` and change the `LOCAL_PATH` and `REMOTE_PATH`)
  * Each pekko module would have its own folder (i.e. pekko core would sit in `/pekko` folder, pekko-http would sit in `/pekko-http`)
    * Under that module folder there would be another one which would designate the version, i.e. `1.0.0` would sit under `/pekko/1.0.0`. Same deal with snapshots but they would have an expiry.
    * There will also be a `/pekko/current/<content>` folder which will always point to the latest release
* The core website at http://pekko.apache.org/ will then link to these docs

The Flink TLP project uses this method (go to https://flink.apache.org/ and on the left menu navigate to Documentation), so there is prior art for this and we can also ask them any questions. The reason why I am advocating for this solution is that its by far the simplest and easiest to get going, its quite similar to what other Scala/OS projects do (especially Scala projects that use Paradox), in fact Lightbend used almost exactly the same method with Akka (thats why Akka/Pekko contains https://github.com/lightbend/sbt-publish-rsync). Its more principled in design especially when taking into account the fact that docs generated from a release (i.e. paradox/scala+java doc) have a different life/update cycle compared to the main website at https://github.com/apache/incubator-pekko-site. The main site would be updated on the https://github.com/apache/incubator-pekko-site  git repo where as the doc updating would be handed by a github action/release manager within sbt.

[...]

PR #161, opened 2023-02-10
publish documentation during nightlies

jrudolph:
Resolves: https://github.com/apache/incubator-pekko/issues/149

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -129,12 +129,12 @@ jobs:
         run: |-
           sbt -Dpekko.genjavadoc.enabled=true docs/paradox unidoc
 
-      # as found in https://github.com/apache/flink/blob/5a4e0ea31aef51dfc50b9c401c91c7b869a154c4/.github/workflows/docs.yml
       - name: Upload documentation
-        uses: burnett01/rsync-deployments@5.2
+        uses: ./.github/actions/sync-nightlies
         with:
+          upload: true
           switches: --archive --compress --update --delete --progress
-          path: pekko-docs/target/paradox/site/main/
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-nightly.yml
```diff
@@ -124,3 +124,41 @@ jobs:
...
           remote_user: ${{ secrets.NIGHTLIES_RSYNC_USER }}
           remote_key: ${{ secrets.NIGHTLIES_RSYNC_KEY }}
+
+      - name: Build Documentation
+        run: |-
+          sbt -Dpekko.genjavadoc.enabled=true docs/paradox unidoc
+
+      # as found in https://github.com/apache/flink/blob/5a4e0ea31aef51dfc50b9c401c91c7b869a154c4/.github/workflows/docs.yml
+      - name: Upload documentation
+        uses: burnett01/rsync-deployments@5.2
```
mdedetrich, line 134: This may not work because Apache maintains a list of github actions that are allowed to run and I can't remember if its global or per project, @pjfanning might recall because he set up the INFRA tickets.

mdedetrich (approved):
lgtm. In case you don't know, you can manually toggle github actions in the actions tab so you don't have to wait till night for the job to trigger if you want to test it.

.github/workflows/publish-nightly.yml
pjfanning: just raise a JIRA issue in INFRA project. https://issues.apache.org/jira/secure/Dashboard.jspa

jrudolph: I just changed it to use our own script, seems similar enough...

pjfanning (approved):
lgtm

Issue #157, opened 2023-02-09
investigate unusual header on Bench.scala


pjfanning:
https://github.com/apache/incubator-pekko/blob/main/actor-tests/src/test/scala/org/apache/pekko/actor/Bench.scala

```
/* The Computer Language Benchmarks Game
   http://shootout.alioth.debian.org/
   contributed by Julien Gaugaz
   inspired by the version contributed by Yura Taras and modified by Isaac Gouy
 */
```

Doesn't appear to assert a copyright or a license.

We can keep it but maybe don't need to.

If it is decided that this does assert a copyright and/or a license, we may need to update the LICENSE or NOTICE for this repo.

If this is not a valid copyright and/or license - then we should remove the check in project/CopyrightHeader.scala

justinmclean:
Hi,

You do need to know how it is licensed. Looks to be this? Although the names are different.

https://salsa.debian.org/benchmarksgame-team/archive-alioth-benchmarksgame/-/blob/master/LICENSE.md <https://salsa.debian.org/benchmarksgame-team/archive-alioth-benchmarksgame/-/blob/master/LICENSE.md>

Justin

jrudolph:
If in doubt, it would probably also ok to just remove it, it's just benchmark code.

pjfanning:
it is basically https://salsa.debian.org/benchmarksgame-team/archive-alioth-benchmarksgame/-/blob/master/contributed-source-code/shootout/chameneos/chameneos.scala-2.scala

I suspect removing this benchmark test might be the best way forward.

justinmclean:
The license is compatible with the Apache license you could leave it in and add that information to LICENSE

pjfanning:
I'm running a build, it so far looks like that file is not even used.

jrudolph:
Fixed by removing the file in #171.

PR #171, opened 2023-02-12
remove test file with non Apache license

pjfanning:
relates to #157


jrudolph (approved):
lgtm, just a benchmark indeed.

PR #174, opened 2023-02-13
make sure to include resources in protobuf-v3 assembly

### LINKED ISSUE #140: fix issue with license/notice in pekko-protobuf-v3 jar
* this jar is built using sbt-assembly and it basically shades classes from protobuf-java v3
* the existing project/MetaInfLicenseNoticeCopy.scala works for jars that we build from compiled source and resource files but does not work with sbt-assembly built jars
* sbt-assembly may have an option to add artbitrary files to the output jar but I haven't found it
* one possible solution is to build a local jar as part of the pekko build that just contains the license and notice files in its META-INF dir and let sbt-assembly merge it and the shaded contents of protobuf-java v3
* a more radical solution might be to drop the pekko-protobuf-v3 jar altogether and just use the akka 2.6 one (likewise for the pekko-protobuf jar - this one is similar but built differently, we already have its source copied to our repo and build it like a normal jar)

jrudolph:
This makes sure that the binaries will actually contain the LICENSE files from the resource generators in `META-INF`.

Fixes #140.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -367,7 +367,7 @@ lazy val protobufV3 = pekkoModule("protobuf-v3")
         // https://github.com/sbt/sbt-assembly/issues/400
         .inLibrary(Dependencies.Compile.Provided.protobufRuntime)
         .inProject),
-    assembly / assemblyOption := (assembly / assemblyOption).value.withIncludeScala(false).withIncludeBin(false),
+    assembly / assemblyOption := (assembly / assemblyOption).value.withIncludeScala(false).withIncludeBin(true),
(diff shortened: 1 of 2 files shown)


jrudolph:
Can be verified by running `publishLocal` in `protobuf-v3` and looking at the contents of the published jar file.

pjfanning (approved):
lgtm

mdedetrich (approved):
Thanks a lot for the fix, means we don't have to do any changes to sbt-assembly saving a lot of time.

PR #176, opened 2023-02-13
fix nightly doc publishing to create target directory first

jrudolph:
Former plan did not work because our sync-nighlies scripts automatically adds a trailing slash to the local_path, which fails the rsync invocation...


jrudolph:
Going ahead with that one to start another run...

jrudolph:
That finally worked, documentation for main is now lying [here](https://nightlies.apache.org/pekko/pekko-docs-main/) and is regenerated (and overwritten on the server) automatically every night.

PR #178, opened 2023-02-13
fix cron scheduling of nightly docs publishing 

jrudolph:
Not 100% sure this is needed but the GHA UI is missing the
scheduling information currently...


pjfanning (approved):
lgtm - worth a try to see if it helps

jrudolph:
Seems to have worked even without this change. Closing...

Issue #183, opened 2023-02-14
pekko urls still have `akka://`


pjfanning:
* appearing in log messages but probably also affecting pekko-remote usage

pjfanning:
#184 is merged

Issue #108, opened 2023-01-17
Provide a way to handle upcoming remote wire incompatibility


mdedetrich:
As discussed in https://github.com/mdedetrich/akka-apache-project/discussions/28, pekko core will need to handle wire incompatibility issues arising from the change in the `akka://` address prefix to `pekko`. There may also be other wire incompatibility changes, for example if we decide to change the default `2552 ` port and anything else that we may have missed.

The solution will also need to play with the release strategy that we decide for pekko. Since the discussion I have done some thinking about it and at least to me the most amicable solution would be to make these fields configurable with typesafe config with different default values depend on the pekko release branch, i.e.

For pekko 1.0.x we would have

```
acceptPrefix = ["akka", "pekko"] // The prefix's to use when checking against incoming requests
sendPrefix = "pekko" // The prefix to use when making send requests
```

And for pekko 1.1.x (on the assumption that we don't want to accept migrating current Akka clusters to this Pekko branch by default)

```
acceptPrefix = ["pekko"] // The prefix's to use when checking against incoming requests
sendPrefix = ["pekko"] // The prefix to use when making send requests
```

This means that when doing a rolling upgrade from akka to pekko 1.0.x you would temporarily change `sendPrefix` to `"akka"` (so that currently existing Akka clusters don't refuse connections) and once the rolling upgrade is complete then you would change `sendPrefix` to `"pekko"`. After that point you are then free to upgrade to pekko 1.1.x without any problems. Making this configurable with typesafe config means that if users have bespoke update requirements its relatively easy for them to handle it.

nvollmar:
Can we deserialize protobuf messages sent by an Akka node with Pekko? I'm not too familiar with Protobuf, but akka-kryo-serialization defaults to FQCN for message serialization for example.

mdedetrich:
> Can we deserialize protobuf messages sent by an Akka node with Pekko? I'm not too familiar with Protobuf, but akka-kryo-serialization defaults to FQCN for message serialization for example.

From my first impression this should work because the only thing that got changed in the https://github.com/apache/incubator-pekko/pull/58 PR is the `option java_package` in the `*.proto` which isn't used by the wire protocol but rather in generating the typed sources from the `.*proto` files. This needs to be tested to be verified.

jrudolph:
> Can we deserialize protobuf messages sent by an Akka node with Pekko? I'm not too familiar with Protobuf, but akka-kryo-serialization defaults to FQCN for message serialization for example.

[...]

PR #184, opened 2023-02-14
use `pekko:` in urls and fix other branding issues

pjfanning:
* relates to #183 
* I'm not sure if it was deliberate to keep the `akka:` URL prefix but it is probably best to change it
* fixed up some other branding issues while I was looking at this


pjfanning:
I need to investigate a few test failures

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
```diff
@@ -484,7 +484,7 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
   // include "com.google.protobuf.GeneratedMessage" = proto in configured serialization-bindings.
   private def checkGoogleProtobuf(className: String): Boolean = checkClass("com.google.protobuf", className)
 
-  // akka-protobuf is now not a dependency of remote so only load if user has explicitly added it
+  // pekko-tobuf is now not a dependency of remote so only load if user has explicitly added it
```
jrudolph, line 487:
```suggestion
  // pekko-protobuf is now not a dependency of remote so only load if user has explicitly added it
```

docs/src/main/paradox/project/migration-guides.md
```diff
@@ -11,7 +11,7 @@ please [migrate to Akka 2.6](https://doc.akka.io/docs/akka/current/project/migra
 This is just stub documentation. It will be improved.
 
 * for Pekko jar dependencies, the groupId is "org.apache.pekko" instead of "com.typesafe.akka"
-* the jar names start with "pekko" instead of "akka" - e.g. pekko-actor_2.13.jar instead of akka-actor_2.13.jar 
+* the jar names start with "pekko" instead of "akka" - e.g. pekko-actor_2.13.jar instead of pekko-actor_2.13.jar
```
jrudolph, line 14:
```suggestion
* the jar names start with "pekko" instead of "akka" - e.g. pekko-actor_2.13.jar instead of akka-actor_2.13.jar 
```

docs/src/main/paradox/testing.md
```diff
@@ -310,7 +310,7 @@ wall time, not CPU time or system time.
 @@@ div { .group-scala }
 
 Ray Roestenburg has written a great article on using the TestKit:
-[https://web.archive.org/web/20180114133958/http://roestenburg.agilesquad.com/2011/02/unit-testing-akka-actors-with-testkit_12.html](https://web.archive.org/web/20180114133958/http://roestenburg.agilesquad.com/2011/02/unit-testing-akka-actors-with-testkit_12.html).
+[https://web.archive.org/web/20180114133958/http://roestenburg.agilesquad.com/2011/02/unit-testing-pekko-actors-with-testkit_12.html](https://web.archive.org/web/20180114133958/http://roestenburg.agilesquad.com/2011/02/unit-testing-pekko-actors-with-testkit_12.html).
```
jrudolph, line 313:
```suggestion
[https://web.archive.org/web/20180114133958/http://roestenburg.agilesquad.com/2011/02/unit-testing-akka-actors-with-testkit_12.html](https://web.archive.org/web/20180114133958/http://roestenburg.agilesquad.com/2011/02/unit-testing-akka-actors-with-testkit_12.html).
```

scripts/link-validator.conf
```diff
@@ -39,7 +39,7 @@ site-link-validator {
   ]
 
   ignore-missing-local-files-regex = ""
-  // e.g. "^api/alpakka/snapshot/pekko/stream/alpakka/googlecloud/storage/impl/Formats.*"
+  // e.g. "^api/connectors/snapshot/pekko/stream/connectors/googlecloud/storage/impl/Formats.*"
```
jrudolph, line 42: Is that really correct? But maybe doesn't matter right now...

scripts/release-train-issue-template.md
```diff
@@ -19,7 +19,7 @@ Release Akka $VERSION$
 ### Check availability
 
 - [ ] Check [reference](https://pekko.apache.org/docs/pekko/$VERSION$/) documentation
-- [ ] Check the release on [Maven central](https://repo1.maven.org/maven2/com/typesafe/akka/akka-actor_2.13/$VERSION$/)
+- [ ] Check the release on [Maven central](https://repo1.maven.org/maven2/com/typesafe/akka/pekko-actor_2.13/$VERSION$/)
```
jrudolph, line 22:
```suggestion
- [ ] Check the release on [Maven central](https://repo1.maven.org/maven2/org/apache/pekko/pekko-actor_2.13/$VERSION$/)
```

jrudolph (changes requested):
lgtm apart from a few accidental changes that should have not been done

jrudolph:
Regarding the `akka://` -> `pekko://` change, it will mean that Akka and Pekko clusters will now be less compatible than before. Not a big concern for me, if people need compatibility, this is a feature that should either be contributed or sponsored as it will be quite a big task.

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
jrudolph: Seems still broken?

jrudolph: Not sure, GH seems to show an old version to me...

mdedetrich:
> Regarding the akka:// -> pekko:// change, it will mean that Akka and Pekko clusters will now be less compatible than before. Not a big concern for me, if people need compatibility, this is a feature that should either be contributed or sponsored as it will be quite a big task.

If the intention is to change it in this PR then we should reference the ticket for it at https://github.com/apache/incubator-pekko/issues/108 .

We also need to make these urls configurable to help migration, not sure whether this should be done as part of this PR or not.

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
pjfanning: I think this is fixed now - I've just rebased the commits and removed the merge commits - and force pushed

scripts/link-validator.conf
pjfanning: I'll leave this open while I look at something else but I'll probably revert it as you suggest

mdedetrich:
I will look at this properly a bit later

scripts/link-validator.conf
pjfanning: I reverted this line

pjfanning:
* there are still a few broken tests - so that's my priority
* the scope of this PR has grown quite large - once I get the tests working, I may look at doing some of the doc and comment updates in a separate PR
* in terms of making the URL prefix configurable, I'd prefer if that was a separate issue and PR - my preference is that pekko:// becomes the default but wouldn't object to supporting akka:// too - but frankly I suspect that there are other changes that will break compat anyway - there are lots of fields and values that are changing from having akka in their names to having pekko in their names

mdedetrich:
> in terms of making the URL prefix configurable, I'd prefer if that was a separate issue and PR - my preference is that pekko:// becomes the default but wouldn't object to supporting akka:// too - but frankly I suspect that there are other changes that will break compat anyway - there are lots of fields and values that are changing from having akka in their names to having pekko in their names

Thats fine, I actually don't want to support `akka` by default (the idea of making it configurable is that people can tune it on their behalf)

jrudolph:
> * in terms of making the URL prefix configurable, I'd prefer if that was a separate issue and PR - my preference is that pekko:// becomes the default but wouldn't object to supporting akka:// too - but frankly I suspect that there are other changes that will break compat anyway - there are lots of fields and values that are changing from having akka in their names to having pekko in their names

Agreed. I think we shouldn't even think of making the scheme configurable as an 1.0.0 feature. I also assume that this alone will not be enough to make most clusters do a zero-downtime upgrade without an issue. The question is e.g. if we will get pekko-management released far enough to even test what would be needed for a zero-downtime upgrade for more common deployment settings.

mdedetrich:
> Agreed. I think we shouldn't even think of making the scheme configurable as an 1.0.0 feature. 

So there was a significant amount of people in the community concerned about compatibility which is why it was planned to do in 1.0.0, but lets discuss this in https://github.com/apache/incubator-pekko/issues/108 and/or the mailing list.

remote/src/test/resources/Makefile
```diff
@@ -17,7 +17,7 @@ domain.cnf:
 	echo "subjectAltName=DNS:localhost" >> domain.cnf
 
 domain.csr: domain.cnf
-	openssl req -new -newkey rsa:2048 -keyout domain.key -subj "/C=ZA/ST=web/O=Lightbend/CN=pekko-remote" -reqexts SAN -config domain.cnf -out domain.csr -passout pass:changeme
+	openssl req -new -newkey rsa:2048 -keyout domain.key -subj "/C=ZA/ST=web/O=Lightbend/CN=akka-remote" -reqexts SAN -config domain.cnf -out domain.csr -passout pass:changeme
```
jrudolph, line 20: Maybe create a ticket to regenerate the certificates with updated properties and fix the tests.

pjfanning: I raised #203 - the remote tests are now passing for this PR

A few other intermittent test failures affecting builds

The main test that is failing all the builds of this PR is RouterTest.showGroupRoutingWithConsistentHashing

I'm not yet sure what change is causing this test to fail. It appears that something in this PR has broken consistent hashing routing.

jrudolph: > The main test that is failing all the builds of this PR is RouterTest.showGroupRoutingWithConsistentHashing

Yep, I also looked into that, very mysterious...

jrudolph: > I'm not yet sure what change is causing this test to fail. It appears that something in this PR has broken consistent hashing routing.

It might just be coincidence (50/50 chance) that the two message hashed to different nodes instead of to the same... The question might be how and when the hashing itself changed...

mdedetrich: We should make a ticket on this and also confirm if the same problem occurred in the latest Akka ASL2 version

jrudolph: The node ring of the consistent hash is built using hashes over actor paths and those changed. So, now the destination actors occupy different places in the consistent hashing node ring than before. With the now different random setup, both `123` and `zh3` will use the same destination actor. I think the easiest to fix will be to change `zh3` to something else randomly until the test works again. Maybe we should start collecting some notes on migration issues people might run into if they also accidentally rely on an arbitrary (but potentially fixed) setup of actors in a routing pool.

pjfanning: thanks @jrudolph - I made a few changes to the id for the 2nd set of messages and eventually found one that gets the test to pass consistently - https://github.com/apache/incubator-pekko/pull/184/commits/014de7b9b74ee682976ab9c2095e432228c10069

mdedetrich: > The node ring of the consistent hash is built using hashes over actor paths and those changed. So, now the destination actors occupy different places in the consistent hashing node ring than before. 

Thanks for the explanation. Basically the tests worked before because we used static actor paths which due to the static paths the hashing algorithm placed actors in certain places on the ring which the test relied on.

>  I think the easiest to fix will be to change zh3 to something else randomly until the test works again. Maybe we should start collecting some notes on migration issues people might run into if they also accidentally rely on an arbitrary (but potentially fixed) setup of actors in a routing pool.

I need to look at the tests in more detail, but at a surface level it seems another improvement we can do later if feasible (and possibly now although its more work) is to redesign the test so that rather than relying on static values and their resulting hashes, instead in the test we generate actor paths such that we get the behaviour that we want to test?

actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
```diff
@@ -240,9 +240,8 @@ public void showGroupRoutingWithConsistentHashing() throws Exception {
     // #consistent-hashing
     // Then messages with equal Message.id reach the same actor
     // so the first message in each probe queue is equal to its second
-    // NB: this test can start failing if you change the actor path (eg the URL scheme) - to fix you
-    // will need
-    // to change id2 value until it starts passing again
+    // NB: this test can start failing if you change the actor path (eg the URL scheme)
```
mdedetrich, line 243: Can we update this comment to add a bit more context, i.e.

NB: this test can start failing if you change the actor path (eg the URL scheme) due to these values being hashed (and depending on the resulting hash its possible to change the distribution of actors in the ring which the test relies on)

actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
```diff
@@ -228,16 +228,21 @@ public void showGroupRoutingWithConsistentHashing() throws Exception {
...
-    router.tell(new Proxy.Message("zh3", "Text4"));
+    final String id2 = "abcdef";
+    router.tell(new Proxy.Message(id2, "Text3"));
+    router.tell(new Proxy.Message(id2, "Text4"));
     // the hash is calculated over the Proxy.Message first parameter obtained through the
     // Proxy.mapping function
     // #consistent-hashing
     // Then messages with equal Message.id reach the same actor
     // so the first message in each probe queue is equal to its second
+    // NB: this test can start failing if you change the actor path (eg the URL scheme) - to fix you
```
mdedetrich: Can we update this comment to add a bit more context, i.e.

NB: this test can start failing if you change the actor path (eg the URL scheme) due to these values being hashed (and depending on the resulting hash its possible to change the distribution of actors in the ring which the test relies on)

mdedetrich (changes requested):
From my end would like to update the documentation a bit more regarding the changed actor paths that caused a regression (my concern here is a more general one which is that anything that improves the clarity of the pekko code is greatly beneficial if we want to foster a community around it).

actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
pjfanning: done

actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
pjfanning: done

actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
mdedetrich: Did you push the changes?

pjfanning: apologies - pushed now

mdedetrich (approved):
lgtm, I will approve

pjfanning:
@jrudolph do you think this is ok to merge now?

pjfanning:
@jrudolph my ping above was at the weekend so it is possible that you missed it. Would you have time to re-review this?

jrudolph (approved):
lgtm

(Sorry for the delay, changes itself look fine. I'm still having the compatibility story at the back of my mind and this gets us further away from compatibility with Akka. I don't mind personally but, of course, it might matter for adoption and as I said elsewhere I don't expect a good compatibility story to turn up any time soon. On the other hand, doing this change is only a small step. Especially, since the most important thing for compatibility would be someone writing comprehensive multi-jvm tests exercising cross Akka-Pekko remoting to actually figure out what is currently missing to make compatibility work which is work that needs to be done regardless of any changes that we do here.)

pjfanning:
I'll merge this but I'm not against the idea of a compatibility mode. Just think we should use pekko as the default.

Issue #186, opened 2023-02-14
investigate pekko-osgi test artifact blueprint.xml


pjfanning:
```
<?xml version="1.0" encoding="UTF-8"?>
<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
           xmlns:akka="http://akka.io/xmlns/blueprint/v1.0.0">

    <akka:actor-system name="BlueprintSystem" />

    <akka:actor-system name="BlueprintSystemWithConfig">
        <akka:config>
            some.config {
              key=value
            }
        </akka:config>
    </akka:actor-system>
</blueprint>
```

Can this just be modified to use `pekko` instead of `akka`?

spangaer:
Now my xml tags are a bit rusty, but if you update the URL it seems fine
https://stackoverflow.com/a/11602530/574370

Merely a means of namespacing XML tags not to collide with those of OSGi.
(matter of updating the parsing code of course)

pjfanning:
Sometimes the URI matters because it may relate to an XML schema - but changing the URI appears to break no tests.

pjfanning:
https://github.com/apache/incubator-pekko/pull/201 was merged

PR #187, opened 2023-02-15
Update blueprint.xml to use pekko namespace

pjfanning:
relates to https://github.com/apache/incubator-pekko/issues/186

`sbt osgi/test` works ok

### DIFF
--- a/osgi/src/test/scala/docs/osgi/blueprint.xml
+++ b/osgi/src/test/scala/docs/osgi/blueprint.xml
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
-           xmlns:akka="http://akka.io/xmlns/blueprint/v1.0.0">
+           xmlns:pekko="http://pekko.apache.org/xmlns/blueprint/v1.0.0">
 
-    <akka:actor-system name="BlueprintSystem" />
+    <pekko:actor-system name="BlueprintSystem" />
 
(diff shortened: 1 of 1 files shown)


spangaer:
I was a bit baffled that you could just change this file without impact. From the looks of it it's only an example and it isn't really parsed anywhere, which is probably why you get away with that. (meaning parsing the config seems to be left to overrides of certain methods of the OSGi classes)

Did stumble on this one in the process
https://github.com/apache/incubator-pekko/blob/main/osgi/src/main/scala/org/apache/pekko/osgi/OsgiActorSystemFactory.scala#L76

pjfanning:
> 
> Did stumble on this one in the process https://github.com/apache/incubator-pekko/blob/main/osgi/src/main/scala/org/apache/pekko/osgi/OsgiActorSystemFactory.scala#L76

I've added a change for that to https://github.com/apache/incubator-pekko/pull/184

There are loads of Akka references still in the code. It's hard to track them all down because we stuck the word Akka into the new license header so almost every file has Akka in it.

jrudolph:
Seems the namespace handler was removed in e98dd13cd68d57d8b8eb188d693600630255de78 so it probably does not matter...

pjfanning:
> Seems the namespace handler was removed in [e98dd13](https://github.com/apache/incubator-pekko/commit/e98dd13cd68d57d8b8eb188d693600630255de78) so it probably does not matter...

Do you think this is ok to merge and if so, could you approve the PR?

jrudolph (approved):
lgtm (though, we can maybe also just delete the file)

pjfanning:
#201 was merged

PR #188, opened 2023-02-15
[DRAFT] Some rudimentary version of how releasing could work

jrudolph:
Here are some ideas how releasing could later work. It's far from finished. 

In this setup, a docker image provides the base installation to use for building.

GPG keys are provided from the outside of the container by providing access to the host gpg-agent. Alternatively, we could provide keys similarly as `sbt-ci-release` does it by actually providing the secret as an environment value. Both options don't seem super compelling. The use of `gpg-agent` at least provides the possibility to also use smartcards/yubikeys for signing automatically.

The setup of `gpg-agent` propagation to the docker container is right now quite brittle:
 * it requires that the user id must match between host and container
 * it assumes a certain setup of how gpg-agent is connected that works on Ubuntu/Linux but might not work on other distros

### DIFF
--- a/scripts/release/README.md
+++ b/scripts/release/README.md
@@ -1,0 +1,25 @@
+# Release build scripts
+
+The build environment can be setup using this command:
+
+```sh
+docker build pekko:build
+```
+
+This command will run the build, and provided gpg credentials by forwarding the GPG agent connection to the docker container.
+Understand the risk of exposing the gpg-agent to a container.
+
+```sh
+docker run \
+  -ti --rm \
+  -v `pwd`:/scripts:ro \
+  -v ${HOME}/.gnupg/:/home/user/.gnupg/:ro -v /run/user/$(id -u)/:/run/user/$(id -u)/:ro \
+  -e BUILD_COMMIT=origin/main \
+  -e GPG_SIGNING_KEY=8DEF770BCFC57CEC83BF0410DC20AD935AC6CEF4 \
+  pekko:build /scripts/build-release.sh
+```
+
+## Config environment settings
+
+ * `BUILD_COMMIT`: The commit to checkout and build
+ * `GPG_SIGNING_KEY`: The GPG key ID to use for signing artifacts
--- a/scripts/release/build-release.sh
+++ b/scripts/release/build-release.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -ex
+set -o pipefail
+
+# check config
+# BUILD_COMMIT
+# NEXUS_USER
+# NEXUS_PW
+# GPG_SIGNING_KEY
+
+# checkout
+git checkout $BUILD_COMMIT
+
+# generate source artifacts
+sbt sourceDistGenerate
+
+# sign source artifacts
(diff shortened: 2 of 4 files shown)


scripts/release/README.md
```diff
@@ -0,0 +1,29 @@
...
+`build-release.sh` will run the build, and provided gpg credentials by forwarding the GPG agent connection to the docker container.
+Understand the risk of exposing the gpg-agent to a container.
+
+```sh
+docker run \
+  -ti --rm \
+  -v `pwd`:/scripts:ro \
+  -v ${HOME}/.gnupg/:/home/user/.gnupg/:ro -v /run/user/$(id -u)/:/run/user/$(id -u)/:ro \
+  -e BUILD_COMMIT=origin/main \
+  -e GPG_SIGNING_KEY=8DEF770BCFC57CEC83BF0410DC20AD935AC6CEF4 \
```
seveneves, line 22: Is this a real signing key?

jrudolph: This is a real id of an example key.

scripts/release/build-release.sh
```diff
@@ -0,0 +1,30 @@
...
+# GPG_SIGNING_KEY
+
+# checkout
+git checkout $BUILD_COMMIT
+
+# generate source artifacts
+sbt sourceDistGenerate
+
+# sign source artifacts
+find target/dist -regex '.*\(tgz\|zip\)' | xargs -I{} sh -c "sha512sum {} > {}.sha512"
```
pjfanning, line 19: newer versions (that already exist, but not yet uptaken here) produce the sha digests automatically

scripts/release/build-release.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# upload source artifacts
+# TODO
+
+# publish to Apache Nexus Staging
+echo "pgpSigningKey := Some(\"$GPG_SIGNING_KEY_ID\")" > pgp-signing-key.sbt
+sbt +publishSigned
+
+# generate and upload docs
+# TODO
```
mdedetrich, line 30: nit: new line here

mdedetrich:
> GPG keys are provided from the outside of the container by providing access to the host gpg-agent. Alternatively, we could provide keys similarly as sbt-ci-release does it by actually providing the secret as an environment value. Both options don't seem super compelling. The use of gpg-agent at least provides the possibility to also use smartcards/yubikeys for signing automatically.

This is more of a long term idea, but I always imagined using sbt-pgp directly rather than sbt-ci-release since sbt-pgp just wraps over gpg so there is no reason why it should have the same problems as sbt-ci-release. In fact I think in general sbt-ci-release is not really needed for pekko projects, the only functionality we use from it is sbt-dynver considering there is broad agreement on how to do a proper release which is not going to be via pushing a git tag.

scripts/release/build-release.sh
mdedetrich: I will make a PR to update version now.

mdedetrich: So the PR has been merged (see https://github.com/apache/incubator-pekko/pull/195) so you don't need to worry about manually creating digests of the files.

PR #189, opened 2023-02-15
remove check for old implementation of Akka Discovery

pjfanning:
* pekko-discovery will not work directly with any implementation of Akka Discovery.
* we've already messed up this test with a global replace that affects the classname

### DIFF
--- a/discovery/src/main/scala/org/apache/pekko/discovery/Discovery.scala
+++ b/discovery/src/main/scala/org/apache/pekko/discovery/Discovery.scala
@@ -24,8 +24,6 @@ import pekko.annotation.InternalApi
 
 final class Discovery(implicit system: ExtendedActorSystem) extends Extension {
 
-  Discovery.checkClassPathForOldDiscovery(system)
-
   private val implementations = new ConcurrentHashMap[String, ServiceDiscovery]
   private val factory = new JFunction[String, ServiceDiscovery] {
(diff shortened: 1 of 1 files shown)


mdedetrich (approved):
Looks good to me.

On a somewhat related note, there is an argument that we should implement similar checks in pekko but actually for any akka artifacts that happen to be on the classpath (i.e. to prevent mixing akka and pekko artifats).

Should I make an issue for this?

pjfanning:
> Looks good to me.
> 
> On a somewhat related note, there is an argument that we should implement similar checks in pekko but actually for any akka artifacts that happen to be on the classpath (i.e. to prevent mixing akka and pekko artifacts).
> 
> Should I make an issue for this?

Worth tracking an issue but generally, having main Akka libs and Pekko libs on same classpath should be ok. This issue is more about Pekko checking about using an even older Discovery mechanism pre-dating akka-discovery joining the core of Akka.

mdedetrich:
> Worth tracking an issue but generally, having main Akka libs and Pekko libs on same classpath should be ok. 

Yes but usually it would indicate that you may have done something wrong

pjfanning:
> > Worth tracking an issue but generally, having main Akka libs and Pekko libs on same classpath should be ok.
> 
> Yes but usually it would indicate that you may have done something wrong

Sure - log an issue and it can be discussed. I would prefer to start by logging a warning. But if experience teaches us that having both sets of classes leads to issues, we can make it a failure condition.

PR #191, opened 2023-02-15
Fix link to snapshots repository documentation

raboof:
as a relative url either the link at https://pekko.apache.org/docs/pekko/current/actors.html or at https://pekko.apache.org/docs/pekko/current/typed/actors.html would be wrong


pjfanning (approved):
lgtm

PR #192, opened 2023-02-15
Correctly document where to get nightlies

raboof:


docs/src/main/paradox/project/links.md
```diff
@@ -43,7 +43,7 @@ Make sure that you add the repository to the Maven repositories in pom.xml:
 <repositories>
   <repository>
     <id>apache-pekko-snapshots</id>
-    <url>https://repository.apache.org/content/groups/snapshots</url>
+    <url>https://nightlies.apache.org/pekko/snapshots</url>
```
pjfanning, line 46: I'm not sure that we will continue to publish jars to nightlies site, it just duplicates what we have in snapshots on repository.apache.org.

That aside, there are still issues that this PR is addressing. I guess the only thing to decide if we want to change this line.

raboof: yeah, sorry, I got carried away there :) - so the 'correct' place is `repository.apache.org`?

pjfanning: yes - best to not mention nightlies until we decide if we want to keep that support

pjfanning (approved):
lgtm

PR #194, opened 2023-02-15
Update CONTRIBUTING.md to use pekko names

pjfanning:
Includes many akka references


jrudolph (approved):
lgtm (will probably need rework later if we decide to set our own guidelines)

PR #196, opened 2023-02-15
update some random akka refs

pjfanning:
splits out some of the changes in #184 so that can focus more on the Pekko/Akka URL protocol


jrudolph (approved):
lgtm

PR #197, opened 2023-02-15
replace akka based names in logging code

pjfanning:
splits out some of the changes in #184 so that can focus more on the Pekko/Akka URL protocol

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/serialization/SerializeSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/serialization/SerializeSpec.scala
@@ -276,7 +276,7 @@ class SerializeSpec extends PekkoSpec(SerializationTests.serializeConf) {
     }
 
     "log warning if non-Pekko serializer is configured for Pekko message" in {
-      EventFilter.warning(pattern = ".*not implemented by Apache Pekko.*", occurrences = 1).intercept {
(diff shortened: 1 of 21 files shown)


actor-tests/src/test/scala/org/apache/pekko/serialization/SerializeSpec.scala
```diff
@@ -276,7 +276,7 @@ class SerializeSpec extends PekkoSpec(SerializationTests.serializeConf) {
     }
 
     "log warning if non-Pekko serializer is configured for Pekko message" in {
-      EventFilter.warning(pattern = ".*not implemented by Apache Pekko.*", occurrences = 1).intercept {
+      EventFilter.warning(pattern = ".*not implemented by Pekko.*", occurrences = 1).intercept {
```
jrudolph, line 279: Weird, a reverse change like this is also proposed as part of #196?

jrudolph (approved):
I don't mind the change but it looks as if it would conflict with #196?

pjfanning:
> 196

I was trying to fix an issue in #196 and committed the change to this PR by mistake. I committed a revert of that commit to this PR after committing the fix to #196. I am getting no conflicts showing for this PR now that #196 is merged.

pjfanning:
The squash merge led to this commit: https://github.com/apache/incubator-pekko/commit/2824a44abeedaefffc68f4930d5f2435690d1eb2 

There appears to be no change to SerializeSpec in that commit.

PR #198, opened 2023-02-16
update Akka refs in PekkoSSLConfig.scala

pjfanning:
* this class is deprecated
* the deprecation notices were missed when I previously changed the version numbers on other deprecation notices to highlight the version number is an Akka version as opposed to a Pekko version
* renamed one akka named param
* added a further deprecation to highlight one particular function where calling it just leads to a log warning (and nothing else is done)


jrudolph (approved):
lgtm

PR #199, opened 2023-02-16
Add github PR check to make sure code is formatted

mdedetrich:
Adds a github PR status check which disallows PR's from being merged if the code is not formatted. Note that its been verified that this works in another pekko module (see https://github.com/apache/incubator-pekko-sbt-paradox/blob/main/.asf.yaml#L28-L35)

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -28,7 +28,13 @@ github:
     rebase:  true
 
   protected_branches:
-    main: { }
+    main:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: false
+        # contexts are the names of checks that must pass
+        contexts:
+          - Code is formatted
 
 notifications:
   commits:              commits@pekko.apache.org


jrudolph (approved):
lgtm

jrudolph:
In general, at least "Headers" and "Tests (pull_request)" should also be added.

mdedetrich:
> In general, at least "Headers" and "Tests (pull_request)" should also be added.

I am slightly wary in adding other checks right now because if you aren't careful you can freeze the github project (i.e. disallow any PR's from being merged) and you have to make an INFRA ticket to block it. For other checks we should test it out properly by making a dummy branch (as was done with incubator-pekko-sbt-paradox). The other issue here is if the name of the check changes unintentionally then we can get into this state.

I am adding it for scalafmt now because that has close to zero change of ever needing to be changed and I am 100% sure it works.

jrudolph:
> I am slightly wary in adding other checks right now because if you aren't careful you can freeze the github project (i.e. disallow any PR's from being merged) and you have to make an INFRA ticket to unblock it. For other checks we should test it out properly by making a dummy branch (as was done with incubator-pekko-sbt-paradox). The other issue here is if the name of the check changes unintentionally then we can get into this state.

Yes, I agree, that would be unfortunate...

PR #201, opened 2023-02-16
Delete blueprint.xml

pjfanning:
An alternative to #187 - see discussion there


pjfanning:
@jrudolph is this PR ok as an alternative to #187 ?

Issue #142, opened 2023-02-02
docs: change version numbers in docs


pjfanning:
A lot of the docs refer to versioning and use recent Akka release numbers to provide context. Pekko's release will 1.0.x so we are better off talking about 1.0.x and 1.0.y (possibly 1.1.z) instead of 2.5.x and 2.6.y.

Example: https://pekko.apache.org/docs/pekko/current/project/downstream-upgrade-strategy.html

PR #204, opened 2023-02-17
remove section about Akka 2.4 versioning change

pjfanning:
* partially related to #142 
* I'm leaving #142 open to deal with changing all the refs to Pekko 2.5 and Pekko 2.6 - it's sort of hard to come up with examples that make sense when we haven't done any releases yet - it may be better just to delete a lot of the docs about specific examples and just give a high level commitment to binary compatibility


jchapuis (approved):

He-Pin:
So we just need to providing something helping migrating  2.5.x to pekko.

pjfanning:
The Pekko docs have been updated to advise users to upgrade to Akka 2.6 before attempting to migrate to Pekko. We will document only the migration from Akka 2.6 to Pekko.

PR #209, opened 2023-02-18
change deprecated versions to use Akka in name

pjfanning:
* because these versions do not relate to Pekko directly


mdedetrich (approved):
lgtm

PR #211, opened 2023-02-21
Move header check into its own github action

mdedetrich:
Similar to https://github.com/apache/incubator-pekko-http/pull/77, moves the header check into its own github workflow. Note that there are some slight differences compared to pekko-http, but this is due to bespoke pekko core differences (which can be seen in the code removed from `build-test-prValidation.yml`).


pjfanning (approved):
Lgtm

PR #217, opened 2023-02-25
Test github actions check headers status check

mdedetrich:
The intention behind this PR is to test the "check headers" status check on a branch that is not `main`, since if its done incorrectly then you can block PR's from being merged entirely in the project until INFRA manually unblocks it.

Once this PR is merged than future test PR's will be made against `test` to see if the status check works as expected, at which point we will change the status check to the `main` branch.


pjfanning (approved):
lgtm

PR #218, opened 2023-02-26
Test pr

mdedetrich:
Test PR to see if header status check works

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -35,6 +35,13 @@ github:
         # contexts are the names of checks that must pass
         contexts:
           - Code is formatted
+    temp:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: false
+        # contexts are the names of checks that must pass
+        contexts:
+          - Check headers
 
 notifications:
   commits:              commits@pekko.apache.org
--- a/README.md
+++ b/README.md
@@ -39,4 +39,6 @@ Apache Pekko is governed by the [Apache code of conduct](https://www.apache.org/
 License
 -------
 
+Test
+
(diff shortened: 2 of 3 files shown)


.asf.yaml
```diff
@@ -35,6 +35,13 @@ github:
         # contexts are the names of checks that must pass
         contexts:
           - Code is formatted
+    temp:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: false
+        # contexts are the names of checks that must pass
+        contexts:
+          - Check headers
```
pjfanning, line 44: We have plenty of work to still do with adding headers - I'd prefer not to make the header check something that completely blocks merges. Humans need to approve PRs anyway. Can we wait till we have added headers everywhere we need them?
* probably conf files
* markdown files
* probably more file types too

mdedetrich: I understand the sentiment behind the concern but I am not exactly sure how its going to be a technical problem. If any supposed improvements to the header checks cause the check to fail then such a PR wouldn't get merged anyways unless we are going to tell people to ignore the now failing header check, and if so in such a case we would temporarily disable the entire header pipeline along with the check in a separate PR.

The alternative situation is that the header check just doesn't pick up everything (which is the current case, i.e. sbt-header is not picking headers from sbt project files in `project/`) in which case its not going to block anything.

mdedetrich: > markdown files

At least if we are talking about docs, readme afaik these do not require headers, its actually noted on ASF's legal site i.e.  from https://www.apache.org/legal/src-headers.html#faq-exceptions

> Short informational text files; for example README, INSTALL files. The expectation is that these files make it obvious which product they relate to.

pjfanning: I mean the markdown files used by paradox build

Issue #224, opened 2023-03-03
Remove the deprecation of Flow#recoverWith


He-Pin:
This was first reported to akka at https://github.com/akka/akka/issues/31670

And I think we should remove this deprecation in the first release too.

pjfanning:
seems reasonable - would you be able to so the Pekko PR?

He-Pin:
Yes, I will work on this.

PR #233, opened 2023-03-04
=str Remove the deprecation annotation on Flow#recoverWith.

He-Pin:
refs: https://github.com/apache/incubator-pekko/issues/224

Drop the deprecation annotation on `Flow#recoverWith`


pjfanning (approved):
lgtm

PR #234, opened 2023-03-04
can only set headerLicense config once

pjfanning:
same issue as https://github.com/apache/incubator-pekko-connectors/pull/33


nvollmar (approved):
lgtm

Issue #231, opened 2023-03-03
license: acknowledge use of netty code


pjfanning:
See https://github.com/apache/incubator-pekko/blob/4a36d2600e7a5cb311418a7ef98ca7899c4a8f4d/actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala#L164

pjfanning:
PR merged

PR #237, opened 2023-03-09
acknowledge use of netty derived code

pjfanning:
relates to #231


legal/pekko-actor-jar-license.txt
```diff
@@ -214,3 +214,9 @@ pekko-actor contains code from scala-collection-compat in the `org.apache.pekko.
 which has written by the Scala-Lang team under an Apache 2.0 license.
 Copyright (c) 2002-2023 EPFL
 Copyright (c) 2011-2023 Lightbend, Inc.
+
+---------------
+
+pekko-actor contains code from Netty in `org.apache.pekko.io.dns.DnsSettings.scala`
```
He-Pin, line 220: INAL, maybe we can add a header in that file?

pjfanning: The netty code is Apache licensed so our  file header covers that already. We still need to acknowledge it the main license and notice files.

He-Pin:
lgtm，let's merge this then.

Issue #243, opened 2023-03-11
migrate incubator-pekko-samples to use pekko snapshots


pjfanning:
Most samples in https://github.com/apache/incubator-pekko-samples should be already in a position to uptake pekko snapshot libs instead of akka ones.

There might be a few samples that need to wait for one or more pekko libs.

This is no reason not to start.

It would be very useful if at least some samples were migrates before the release of the main Pekko libs.

pjfanning:
The core Pekko docs reference these samples. Thus the importance of making progress. It would be great if PRs take on 1 or a small number of samples at a time. This would facilitate having multiple contributors.

sam-byng:
I'll look at directories starting akka-samples-cluster and akka-samples-cqrs today.

Note there are still locations in incubator pekko referencing akka-samples, presume these will need update as part of this. 
- https://github.com/apache/incubator-pekko/blob/main/scripts/publish_samples.sh#L87 -- probably most important, the publish samples script
- DOC: https://github.com/apache/incubator-pekko/blob/main/docs/src/main/paradox/additional/rolling-updates.md?plain=1#L148
- URL in a comment: https://github.com/apache/incubator-pekko/blob/main/project/Dependencies.scala#L108 ( URL is now NotFound)

sam-byng:
I've also made minor update to migration docs in line with my experience.

pjfanning:
-1 to any docs about specific build tools.

mdedetrich:
Closing since according to @sam-byng this should be done, if not we can just re-open

Issue #109, opened 2023-01-17
Create a migration guide to migrate from Akka to Pekko


kw217:
Create a migration guide explaining to a user of Akka how to migrate their project to Pekko.

We hope to automate as much of this as possible under #22 but first we need to determine the steps and lay them out so early adopters can follow them and gain experience.

This should be documented in `docs/src/main/paradox/project/migration-guide-akka-pekko.md`, linked from `migration-guides.md` in the same location, and also linked prominently from the GitHub homepage `README.md` and the project homepage `docs/src/main/paradox/index.md`.

Steps include:
* adapting Akka dependencies `"com.typesafe.akka"` -> `"org.apache.pekko"`
* adapting imports `import akka.actor` -> `import org.apache.pekko; import pekko.actor` - see #58 
* adapting class names `Akka*` to `Pekko*` - see #67 
* adapting packages and corresponding paths `akka` -> `org.apache.pekko`, if the codebase inserts code into the Akka namespace
* updating third-party dependencies to versions that depend on Pekko
* adapting configuration to use the `pekko` key - see #63 
* discussion of how to approach Akka remote applications

pjfanning:
* don't forget the pekko name in the jars (akka-actor becomes pekko-actor, etc.)
* the guide will go into the docs - there is a placeholder for the migration section in there (or at least once an inflight PR is merged)

kw217:
Cool, thanks - added that to the description.

Docs location - presumably you're referring to https://github.com/apache/incubator-pekko/blob/main/docs/src/main/paradox/project/migration-guides.md which just got updated? Perfect. Should we just put the migration guide in that doc, rather than a separate doc linked from there?

kw217:
Example migration (thanks @pjfanning): https://github.com/pjfanning/micrometer-akka/pull/3

pjfanning:
We have a migration guide. It's not perfect but we can adjust it. Please raise PRs if you want to suggest changes - or comment on this issue.

Removing the milestone as I want to not have open issues in it.

PR #249, opened 2023-03-14
Issue #109: Update migration guide

sam-byng:
https://github.com/apache/incubator-pekko/issues/109: Update migration guide based on experience working with issue #243


pjfanning (changes requested):
No - please do not start mentioning poms and SBT stuff. Users should educate themselves on their preferred build tools. Pekko is neutral and should not complicate our docs with pom trivialities.

docs/src/main/paradox/project/migration-guides.md
```diff
@@ -10,16 +10,29 @@ please [migrate to Akka 2.6](https://doc.akka.io/docs/akka/current/project/migra
 
 This is just stub documentation. It will be improved.
 
-* for Pekko jar dependencies, the groupId is "org.apache.pekko" instead of "com.typesafe.akka"
-* the jar names start with "pekko" instead of "akka" - e.g. pekko-actor_2.13.jar instead of akka-actor_2.13.jar 
+In your dependency management files:
+* Update dependency Akka version to the latest Pekko version.
```
mdedetrich, line 14: Rather than latest Pekko version I would instead say latest version of Pekko 1.0.x (which we claim to be broadly compatible with latest Akka BSL release).

mdedetrich (changes requested):
Just one thing I would mention

docs/src/main/paradox/project/migration-guides.md
sam-byng:
```suggestion
* Update dependency Akka version to the latest version of Pekko 1.0.x.
```

kw217:
Looks like all the comments have been addressed - can this be approved and merged?

mdedetrich:
Checking this now

mdedetrich (approved):
lgtm

pjfanning:
I must admit to not liking the categorisation of the changes and the duplication of the changes because of the artificial categorisation.

mdedetrich:
> I must admit to not liking the categorisation of the changes and the duplication of the changes because of the artificial categorisation.

Can you rephrase, looks fine to me?

docs/src/main/paradox/project/migration-guides.md
```diff
@@ -10,16 +10,29 @@ please [migrate to Akka 2.6](https://doc.akka.io/docs/akka/current/project/migra
...
+* Update dependency Akka version to the latest version of Pekko `1.0.x`.
+* for Pekko jar dependencies, the groupId is `org.apache.pekko` instead of `com.typesafe.akka` or `com.lightbend.akka.<package_name>`
+* the jar names start with "pekko" instead of "akka" - e.g. pekko-actor_2.13.jar instead of akka-actor_2.13.jar
 * Alpakka equivalent is "pekko-connectors" - e.g. pekko-connectors-kafka_2.13.jar instead of alpakka-kafka_2.13.jar
-* Pekko packages start with "org.apache.pekko" instead of "akka" - e.g. `import org.apache.pekko.actor` instead of `import akka.actor`
-* Where class names have "Akka" in the name, the Pekko ones have "Pekko" - e.g. PekkoException instead of AkkaException
-* Configs in `application.conf` use "pekko" prefix instead of "akka"
+
+In config files:
+* Configs use `pekko` prefix instead of `akka`
```
pjfanning, line 20: why remove the reference to 'application.conf' ?

mdedetrich: To be more accurate we could use `*.conf` instead since some conf files are `reference.conf` and others are `applciation.conf`.

in fact from checking the codebase it should all be `reference.conf`, after all the idea is that `reference.conf` is whats used for libraries and `application.conf` is whats used for applications.

docs/src/main/paradox/project/migration-guides.md
```diff
@@ -10,16 +10,29 @@ please [migrate to Akka 2.6](https://doc.akka.io/docs/akka/current/project/migra
...
+
+In source code files:
+* Pekko packages start with `org.apache.pekko` instead of `akka` - e.g. `import org.apache.pekko.actor` instead of `import akka.actor`
+* Where class names have "Akka" in the name, the Pekko ones have "Pekko" - e.g. `PekkoException` instead of `AkkaException`
+* Configs use `pekko` prefix instead of `akka`
+
+In Documentation:
```
pjfanning, line 27: I think this applies to migration use case. It's useful advice to people contributing code to Pekko - but it is not really something that someone migrating their company's projects needs to know about.

docs/src/main/paradox/project/migration-guides.md
```diff
@@ -10,16 +10,29 @@ please [migrate to Akka 2.6](https://doc.akka.io/docs/akka/current/project/migra
...
+* Where appropriate, text should refer to `Pekko` instead of `Akka`
+* URLs linking to files in https://doc.akka.io/docs/akka/ should instead point at https://pekko.apache.org/docs/pekko/
+
 * We have changed the default ports used by the pekko-remote module.
```
pjfanning, line 31: this port information is useful but it is now sort of orphaned by the artificial breaking up of the migration info into categories

there are not so many bullet points that we need to split it up like this - especially when we then stick some of the bullet points off in 'no category'

docs/src/main/paradox/project/migration-guides.md
```diff
@@ -10,16 +10,29 @@ please [migrate to Akka 2.6](https://doc.akka.io/docs/akka/current/project/migra
...
-* With @ref:[Classic Remoting](../remoting.md), Akka defaults to 2552, while Pekko defaults to 7355.
-* With @ref:[Artery Remoting](../remoting-artery.md), Akka defaults to 25520, while Pekko defaults to 17355.
+    * With @ref:[Classic Remoting](../remoting.md), Akka defaults to 2552, while Pekko defaults to 7355.
+    * With @ref:[Artery Remoting](../remoting-artery.md), Akka defaults to 25520, while Pekko defaults to 17355.
 
+## TBC:
 We are still investigating the effects of how the package name changes affect the @ref:[Persistence](../persistence.md)
```
pjfanning, line 36: See https://lists.apache.org/thread/v7c08l4bn66o4o2go0h7xzzl8pz0tr6p

I can update this section in an independent PR

docs/src/main/paradox/project/migration-guides.md
pjfanning: it's a migration guide for users migrating their own code bases - they will be updating their own application.conf files

some lib maintainers might have reference.conf files

I suppose, some users use '.json' files for configs as well (because typesafe config supports those too). 

I would still prefer if this section, tries to name some of the more common config files to help out users to know what to look for - imagaine if the Akka based app that is being migrated was written 5 years ago and the original writers have moved on.

mdedetrich: > it's a migration guide for users migrating their own code bases - they will be updating their own application.conf files

Right but its also relevant to libraries which integrate with Pekko, in which case they would be dealing with `reference.conf` files and not `application.conf`. Not all users are applications.

> I suppose, some users use '.json' files for configs as well (because typesafe config supports those too).

And also `*.prop` files too which afaik are compatible with HOCON format.

> I would still prefer if this section, tries to name some of the more common config files to help out users to know what to look for - imagaine if the Akka based app that is being migrated was written 5 years ago and the original writers have moved on.

In that case the easiest thing would be to search for `akka` prefix in files that are known to hold configurations which would be `*.conf`/`*.prop` and `*.json`.

Issue #241, opened 2023-03-10
use standard pekko notice in jars (except when additions needed)


pjfanning:
* already using standard Apache license in most jars
* should also use standard Pekko notice in most jars
* see https://github.com/apache/incubator-pekko-http/pull/119
* currently, we default to use the source repo NOTICE and this has a lot of additions and they only affect some jars

pjfanning:
PR merged

PR #250, opened 2023-03-15
update pekko notice used in jars

pjfanning:
Applies changes similar to ones already in https://github.com/apache/incubator-pekko-http

* at the jar level, most jars only need the vanilla Pekko Notice - not the source repo 'NOTICE' file that has a fair few amendments that are specific to just a few jars
* tidy up text about scala-collection-compat license and notice to not specifically mention Scala-Lang team
* https://github.com/apache/incubator-pekko/issues/241


nvollmar (approved):
lgtm

PR #251, opened 2023-03-17
Add mapValues for Scala 2.12 collection compat

mdedetrich:
This adds a `.mapValues` for `IterableView` which is a Scala 2.12 specific method that is necessary to get pekko-connectors-kafka to compile for 2.12. Note that the method has been taken from https://github.com/scala/scala-collection-compat/pull/220 which is where the other methods in this file come from (basically this exists to avoid adding scala-collection-compat as a dependency of Pekko).

### DIFF
--- a/actor/src/main/scala-2.12/org/apache/pekko/util/ccompat/package.scala
+++ b/actor/src/main/scala-2.12/org/apache/pekko/util/ccompat/package.scala
@@ -14,7 +14,7 @@
 package org.apache.pekko.util
 
 import scala.{ collection => c }
-import scala.collection.{ immutable => i, mutable => m, GenTraversable }
+import scala.collection.{ immutable => i, mutable => m, GenTraversable, IterableView }
 import scala.collection.generic.{ CanBuildFrom, GenericCompanion, Sorted, SortedSetFactory }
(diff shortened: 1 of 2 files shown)


pjfanning:
we've already highlighted in our license/notice files that this package has code borrowed from Scala team (EPFL and Lightbend( - so this addition is fine.

He-Pin:
Cool, how about attach a test fot it too？

mdedetrich:
> Cool, how about attach a test fot it too？

So I can add a test but the thing is its directly copied from [scala-collection-compat](https://github.com/scala/scala-collection-compat/pull/220) and there already is test there for this exact same function (see https://github.com/scala/scala-collection-compat/pull/220/files#diff-5465730243527013e5f690dcb15ca76177f7fc611c9024e3ab89405efb638563) so I don't think a test here is necessary.

Also do note that every other function in this file is also copied from scala-colection-compat and there aren't any existing tests for those other functions either. The whole purpose of this file is to just copy the source from within scala-collection-compat in order to avoid adding a direct dependency.

pjfanning (approved):
lgtm

He-Pin:
Okay, reasonable too.

PR #254, opened 2023-03-19
Add filterKeys for Scala 2.12 collection compat

mdedetrich:
Same deal as https://github.com/apache/incubator-pekko/pull/251 but with another method (`filterKeys`) which is also used by pekko-connectors-kafka.

Equivalent source is here https://github.com/scala/scala-collection-compat/blob/0737983fded4dada1c9ad20ae34d2a98b0b3083f/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala#L634-L635


He-Pin (approved):
lgtm

PR #257, opened 2023-03-21
Add Scala 2.12 equivalent Iterator apply method for ByteString

mdedetrich:
While working on https://github.com/apache/incubator-pekko-connectors/pull/74#pullrequestreview-1350324353 I noticed that while there is an `IterableOnce[Byte]` apply method in `ByteString` for Scala 2.13, there isn't the Scala 2.12 `Iterable[Byte]` equivalent apply method. Note that the already existing Scala 2.13 variant can be found at https://github.com/apache/incubator-pekko/blob/main/actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala#L1002-L1009 and https://github.com/apache/incubator-pekko/blob/main/actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala#L40-L43.

This change can be considered safe because if you look at the implementation either for the already existing Scala 2.13 or the Scala 2.12 variant being added in this PR, in both cases it internally converts to an array.

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
@@ -814,6 +814,13 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
         }
       }
 
+      "created from iterator" in {
+        check { (a: ByteString) =>
+          val asIterator = a.iterator
+          ByteString(asIterator) == a
+        }
+      }
+
       "compacting" in {
         check { (a: ByteString) =>
           val wasCompact = a.isCompact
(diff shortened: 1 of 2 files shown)


actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
```diff
@@ -814,6 +814,13 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
         }
       }
 
+      "created from iterator" in {
```
mdedetrich, line 817: This test is here to make sure that the `apply` method for Iterator works for both Scala 2.12 and Scala 2.13. Without the added `Iterator[Byte]` apply method in this PR, this test will fail to compile.

pjfanning (approved):
lgtm

Issue #6, opened 2022-11-03
Upgrade Scala 3.x versions


pjfanning:
Change to
* Scala 2.3.10 due to CVE - https://www.cvedetails.com/cve/CVE-2022-36944/
* Scala 3.2.0 -- Scala team recommend using latest Scala 3 version - https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html
* Scala 3.3.0 - the first LTS version of Scala 3 should be out soon

alexandru:
Bumping Scala 3's minor version isn't necessarily a good thing, some care is needed.

If a project is using Scala `3.1.3`, bumping the Pekko dependency to a new version that uses Scala `3.2.0`, also forces the project on Scala `3.2.0` — backwards compatibility is provided, but not forward compatibility. In that sense, Scala 3 upgrades behave like libraries. This is unlike upgrading from Scala 2.3.9 to Scala 2.3.10.

And care is needed because a compiler is unlike a library, as it may have bugs that don't manifest in the Pekko library's codebase, but that impacts downstream projects.

I think right now an upgrade to Scala 3.2.0 is totally fine, given the state of the Scala 3 adoption, and that of Pekko, but for the future, it's good to keep in mind that such upgrades basically force the new Scala version on downstream users, and that may not be good for Scala versions that have issues.

pjfanning:
@alexandru thanks for the comments. 

Picking out comments from the article linked above, just in case.

> You shouldn’t be afraid of updating it. Bumping the patch version (e.g. 3.1.2 -> 3.1.3) is a no-brainer. Bumping the minor version (e.g. 3.1.3 -> 3.2.0) requires the same considerations as changing the minor version of any other dependency, yet is generally advisable.

So, upgrading to 3.2.0 is not zero risk but it pretty low on the scale of these things.

I'm not going to do anything with this issue yet and it is probably best to discuss this item on the dev mailing list before proceeding to the PR stage.

mdedetrich:
I think the CVE may work for a Pekko release of 1.0.x  pending on what we ultimately decide goes into 1.0.x (since its a CVE) but 3.2.0 makes sense for 1.1.x

alexandru:
@pjfanning 
> upgrading to 3.2.0 is not zero risk but it pretty low on the scale of these things.

[...]

PR #258, opened 2023-03-21
Update Scala versions

mdedetrich:
This PR updates the Scala versions of Akka.

References: https://github.com/apache/incubator-pekko/issues/6

### DIFF
--- a/project/JdkOptions.scala
+++ b/project/JdkOptions.scala
@@ -57,14 +57,8 @@ object JdkOptions extends AutoPlugin {
       targetSystemJdk,
       jdk8home,
       fullJavaHomes,
-      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "-target:jvm-1.8"),
-      // '-release 8' is not enough, for some reason we need the 8 rt.jar
-      // explicitly. To test whether this has the desired effect, compile
-      // pekko-remote and check the invocation of 'ByteBuffer.clear()' in
-      // EnvelopeBuffer.class with 'javap -c': it should refer to
-      // ""java/nio/ByteBuffer.clear:()Ljava/nio/Buffer" and not
-      // "java/nio/ByteBuffer.clear:()Ljava/nio/ByteBuffer". Issue #27079
-      (java8home: File) => Seq("-release", "8", "-javabootclasspath", java8home + "/jre/lib/rt.jar"))
+      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "release:8"),
+      (java8home: File) => Seq("-release", "8"))
   def targetJdkJavacOptions(
       targetSystemJdk: Boolean,
       jdk8home: Option[File],
@@ -75,7 +69,7 @@ object JdkOptions extends AutoPlugin {
       fullJavaHomes,
       Nil,
       // '-release 8' would be a neater option here, but is currently not an
-      // option because it doesn't provide access to `sun.misc.Unsafe` #27079
+      // option because it doesn't provide access to `sun.misc.Unsafe` https://github.com/akka/akka/issues/27079
       (java8home: File) => Seq("-source", "8", "-target", "8", "-bootclasspath", java8home + "/jre/lib/rt.jar"))
 
   private def selectOptions(
--- a/project/PekkoDisciplinePlugin.scala
+++ b/project/PekkoDisciplinePlugin.scala
@@ -164,7 +164,7 @@ object PekkoDisciplinePlugin extends AutoPlugin {
     "-Ywarn-numeric-widen",
     "-Yno-adapted-args",
     "-deprecation",
-    "-Xlint",
+    "-Xlint:-infer-any",
     "-Ywarn-dead-code",
     "-Ywarn-inaccessible",
     "-Ywarn-infer-any",
--- a/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
@@ -33,7 +33,7 @@ private[pekko] trait Children { this: ActorCell =>
 
   import ChildrenContainer._
 
-  @nowarn("msg=never used")
+  @nowarn("msg=never used|is never updated")
   @volatile
   private var _childrenRefsDoNotCallMeDirectly: ChildrenContainer = EmptyChildrenContainer
 
--- a/actor/src/main/scala/org/apache/pekko/pattern/AskSupport.scala
+++ b/actor/src/main/scala/org/apache/pekko/pattern/AskSupport.scala
@@ -537,11 +537,11 @@ private[pekko] final class PromiseActorRef(
    * Stopped               => stopped, path not yet created
    */
   @volatile
-  @nowarn("msg=never used")
+  @nowarn("msg=is never updated")
   private[this] var _stateDoNotCallMeDirectly: AnyRef = _
 
   @volatile
-  @nowarn("msg=never used")
+  @nowarn("msg=is never updated")
   private[this] var _watchedByDoNotCallMeDirectly: Set[ActorRef] = ActorCell.emptyActorRefSet
 
   @nowarn private def _preventPrivateUnusedErasure = {
--- a/actor/src/main/scala/org/apache/pekko/pattern/CircuitBreaker.scala
+++ b/actor/src/main/scala/org/apache/pekko/pattern/CircuitBreaker.scala
@@ -294,12 +294,14 @@ class CircuitBreaker(
   /**
    * Holds reference to current state of CircuitBreaker - *access only via helper methods*
    */
+  @nowarn("msg=is never updated")
   @volatile
   private[this] var _currentStateDoNotCallMeDirectly: State = Closed
 
   /**
    * Holds reference to current resetTimeout of CircuitBreaker - *access only via helper methods*
    */
+  @nowarn("msg=is never updated")
   @volatile
   private[this] var _currentResetTimeoutDoNotCallMeDirectly: FiniteDuration = resetTimeout
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -50,8 +50,8 @@ object Dependencies {
     }
   }
 
-  val scala212Version = "2.12.16"
-  val scala213Version = "2.13.8"
+  val scala212Version = "2.12.17"
+  val scala213Version = "2.13.10"
   // To get the fix for https://github.com/lampepfl/dotty/issues/13106
   // and restored static forwarders
   val scala3Version = "3.1.2"
--- a/scripts/link-validator.conf
+++ b/scripts/link-validator.conf
@@ -46,10 +46,10 @@ site-link-validator {
     "https://javadoc.io/static/"
     # GitHub will block with "429 Too Many Requests"
     "https://github.com/"
-    "https://www.scala-lang.org/api/2.13.8/scala/runtime/AbstractFunction1.html"
-    "https://www.scala-lang.org/api/2.13.8/scala/runtime/AbstractFunction2.html"
-    "https://www.scala-lang.org/api/2.13.8/scala/runtime/AbstractFunction3.html"
-    "https://www.scala-lang.org/api/2.13.8/scala/runtime/AbstractPartialFunction.html"
+    "https://www.scala-lang.org/api/2.13.10/scala/runtime/AbstractFunction1.html"
+    "https://www.scala-lang.org/api/2.13.10/scala/runtime/AbstractFunction2.html"
+    "https://www.scala-lang.org/api/2.13.10/scala/runtime/AbstractFunction3.html"
+    "https://www.scala-lang.org/api/2.13.10/scala/runtime/AbstractPartialFunction.html"
   ]
 
   non-https-whitelist = [


nvollmar (approved):
lgtm

mdedetrich:
Currently blocked by https://discord.com/channels/632150470000902164/632150470000902166/1087726290561400832

mdedetrich:
So one thing to note on this PR is that the workaround with using `-target` described in https://github.com/akka/akka/issues/27079 has been removed.  because `-target` is now deprecated in the latest versions of Scala 2.12/Scala 2.13 and more importantly  `-release` has since been fixed so the workaround is not neccessary. The original problem with `-release` was that it didn't properly link the JDK 8 runtime library however this has since been solved. As stated in https://github.com/apache/incubator-pekko/blob/main/project/JdkOptions.scala#L60-L67 this can be verified, and I can at least confirm on my end that `-release` has indeed been fixed, i.e.

```
<@incubator-pekko/r/t/s/c/o/a/p/r/artery>-<⎇ update-scala-versions>-<±>-<*>-2-> javap -c EnvelopeBuffer.class | grep "java/nio/ByteBuffer.clear:()Ljava/nio/Buffer"
       9: invokevirtual #121                // Method java/nio/ByteBuffer.clear:()Ljava/nio/Buffer;
```

Works for both Scala 2.12 and Scala 2.13

project/JdkOptions.scala
```diff
@@ -57,14 +57,8 @@ object JdkOptions extends AutoPlugin {
...
-      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "-target:jvm-1.8"),
-      // '-release 8' is not enough, for some reason we need the 8 rt.jar
-      // explicitly. To test whether this has the desired effect, compile
-      // pekko-remote and check the invocation of 'ByteBuffer.clear()' in
-      // EnvelopeBuffer.class with 'javap -c': it should refer to
-      // ""java/nio/ByteBuffer.clear:()Ljava/nio/Buffer" and not
-      // "java/nio/ByteBuffer.clear:()Ljava/nio/ByteBuffer". Issue #27079
-      (java8home: File) => Seq("-release", "8", "-javabootclasspath", java8home + "/jre/lib/rt.jar"))
+      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "release:8"),
+      (java8home: File) => Seq("-release", "8"))
```
raboof, line 61: Great to finally remove this nasty workaround!

project/JdkOptions.scala
```diff
@@ -74,9 +68,7 @@ object JdkOptions extends AutoPlugin {
       jdk8home,
       fullJavaHomes,
       Nil,
-      // '-release 8' would be a neater option here, but is currently not an
-      // option because it doesn't provide access to `sun.misc.Unsafe` #27079
-      (java8home: File) => Seq("-source", "8", "-target", "8", "-bootclasspath", java8home + "/jre/lib/rt.jar"))
+      (java8home: File) => Seq("-source", "8", "-target", "8"))
```
raboof, line 71: Unfortunately, I don't think this is equivalent: if a signature would have changed between jdk8 and jdk11, `-target 8` would emit jdk8-compatible bytecode, but would still use the jdk11 signature - so the jar would not work on an actual jdk8.

On the one hand I don't find any instances of that in our current codebase, but on the other hand I'm not sure it's worth introducing the risk...

raboof (changes requested):

project/JdkOptions.scala
mdedetrich: Thanks for the pickup, indeed this problem seems like a different (but slightly unrelated one). I think for now I will revert the change and make an issue for it to check later in order to verify if its needed.

project/JdkOptions.scala
```diff
@@ -57,14 +57,8 @@ object JdkOptions extends AutoPlugin {
...
       fullJavaHomes,
-      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "-target:jvm-1.8"),
-      // '-release 8' is not enough, for some reason we need the 8 rt.jar
-      // explicitly. To test whether this has the desired effect, compile
-      // pekko-remote and check the invocation of 'ByteBuffer.clear()' in
-      // EnvelopeBuffer.class with 'javap -c': it should refer to
-      // ""java/nio/ByteBuffer.clear:()Ljava/nio/Buffer" and not
-      // "java/nio/ByteBuffer.clear:()Ljava/nio/ByteBuffer". Issue #27079
-      (java8home: File) => Seq("-release", "8", "-javabootclasspath", java8home + "/jre/lib/rt.jar"))
+      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "release:8"),
```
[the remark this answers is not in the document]
He-Pin: I have a PR related, let's me check

mdedetrich:
@He-Pin I checked the relevant PR's and it doesn't appear that I have missed anything (the Scala 3.2 one I ignored because thats a separate concern). There can be further simplification but would make sense to do this in context of Scala 3.

project/JdkOptions.scala
BalmungSan: It will not use the **JDK 11** signature because it was compiled targeting the **JDK 8**. It can't even see such a signature.
That is the whole point of the `release` flag, to ensure that both the bytecode and linked **Java** lib match the target **JDK** version.
See https://github.com/scala/scala/pull/9982

mdedetrich: So also according to https://github.com/scala/bug/issues/12643 which describes the same problem,  this workaround shouldn't be necessary.

It appears to be quite easy to bisect this to confirm this, will look tonight

raboof: > That is the whole point of the `release` flag, to ensure that both the bytecode and linked Java lib match the target JDK version.
See https://github.com/scala/scala/pull/9982

You seem to be referring to the scalac flags, but this line is about javac flags.

Also on javac, indeed that would be the point of the `--release` flag. However here we use the `-target` flag, which only affects the bytecode and not the signatures on the classpath.

The "correct" fix would be to use the `--release` flag, but that is not yet (easily) possible as that does not allow access to `sun.misc.Unsafe`, so it wouldn't be able to compile `actor/src/main/java/org/apache/pekko/util/Unsafe.java`. AFAICS that did not change with https://github.com/scala/bug/issues/12643.

He-Pin: I was checked the bytecode in https://github.com/akka/akka/pull/31652 ，what's the result with this pr？

He-Pin: > @som-snytt Since you made the PR at https://github.com/scala/scala/pull/9982 maybe you can comment on this?

you should use the -release option

mdedetrich: > I was checked the bytecode in [akka/akka#31652](https://github.com/akka/akka/pull/31652) ，what's the result with this pr？

I think this is a separate issue, I already checked the bytecode in https://github.com/apache/incubator-pekko/pull/258#issuecomment-1477908162

He-Pin: > @som-snytt Since you made the PR at https://github.com/scala/scala/pull/9982 maybe you can comment on this?

https://github.com/scala/scala/pull/10109

project/PekkoDisciplinePlugin.scala
```diff
@@ -76,7 +76,7 @@ object PekkoDisciplinePlugin extends AutoPlugin {
     "pekko-stream-tests-tck",
     "pekko-testkit")
 
-  val defaultScalaOptions = "-Wconf:cat=unused-nowarn:s,any:e"
+  val defaultScalaOptions = "-Wconf:cat=unused-nowarn:s,cat=lint-infer-any:s,any:e"
```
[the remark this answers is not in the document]
som-snytt: lints can be selectively disabled, `-Xlint:-infer-any,_`, per `-Xlint:help`. Whichever is more convenient for your build.

project/JdkOptions.scala
som-snytt: I'm not sure which issue I'm pinged about, but I remember 2.12 targets 8, making Unsafe available is a pain, and also if there is a subtle breakage in Java API then you may want to compile against rt.jar instead of using `--release` (which uses the `ct.sym` file).

project/PekkoDisciplinePlugin.scala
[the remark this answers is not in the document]
mdedetrich: Thanks, just did the changes.

project/JdkOptions.scala
mdedetrich: > I'm not sure which issue I'm pinged about, but I remember 2.12 targets 8, making Unsafe available is a pain, and also if there is a subtle breakage in Java API then you may want to compile against rt.jar instead of using `--release` (which uses the `ct.sym` file).

So it is specifically about targeting `rt.jar` with javac `-target` command because of unsafe, i.e. `(java8home: File) => Seq("-source", "8", "-target", "8", "-bootclasspath", java8home + "/jre/lib/rt.jar"))`.

It does sound like from what you are saying that we should just leave this.

som-snytt: Yes also I confirmed the deprecation can be silenced by `-Wconf`. I think I just saw on that scala ticket that I said maybe undeprecate target if it has a mainstream use case still, but no action was taken. I think `-Wconf` is not terrible.

som-snytt: Someone had a different need to compile against rt.jar, and https://github.com/scala/scala/pull/10336 has guidance if you're building on later jdk but against old rt.jar (jdk 8 by definition).

mdedetrich: > Someone had a different need to compile against rt.jar, and [scala/scala#10336](https://github.com/scala/scala/pull/10336) has guidance if you're building on later jdk but against old rt.jar (jdk 8 by definition).

Perfect thanks, I will just make an issue and reference that PR in it.

mdedetrich: I will mark this as resolved, made an issue at https://github.com/apache/incubator-pekko/issues/262 where we can investigate the removal of this workaround separately. Most critical thing for this PR specifically is to update scalac versions because of a CVC.

@raboof @BalmungSan @He-Pin @som-snytt Thanks a ton for the help/clarification

project/PekkoDisciplinePlugin.scala
```diff
@@ -164,7 +164,7 @@ object PekkoDisciplinePlugin extends AutoPlugin {
     "-Ywarn-numeric-widen",
     "-Yno-adapted-args",
     "-deprecation",
-    "-Xlint",
+    "-Xlint:-infer-any",
```
som-snytt, line 167: This enables no lints. `-Xlint -Xlint:-infer-any` or `-Xlint:-infer-any,_` with underscore to mean all. (Not sure if that was intended here.)

mdedetrich: Thanks, fixing now

project/JdkOptions.scala
```diff
@@ -57,14 +57,8 @@ object JdkOptions extends AutoPlugin {
...
       jdk8home,
       fullJavaHomes,
-      Seq(if (scalaVersion.startsWith("3.")) "-Xtarget:8" else "-target:jvm-1.8"),
-      // '-release 8' is not enough, for some reason we need the 8 rt.jar
-      // explicitly. To test whether this has the desired effect, compile
-      // pekko-remote and check the invocation of 'ByteBuffer.clear()' in
-      // EnvelopeBuffer.class with 'javap -c': it should refer to
-      // ""java/nio/ByteBuffer.clear:()Ljava/nio/Buffer" and not
-      // "java/nio/ByteBuffer.clear:()Ljava/nio/ByteBuffer". Issue #27079
-      (java8home: File) => Seq("-release", "8", "-javabootclasspath", java8home + "/jre/lib/rt.jar"))
```
jrudolph, line 67: The removal of `-javabootclasspath` breaks building with JDK 11 for me, also reproducible in Docker containers. I know I investigated before, if it could be removed and I ran into issues before.

I wonder how it works on Github CI and other machines.

raboof: interesting, seems to work on my machine - what does it break on for you?

pjfanning: @jrudolph we have a release candidate (1.0.1-RC1) at https://repository.apache.org/content/groups/staging/org/apache/pekko/

Would you be able to see if there is anything missing in the jars?

The RCs have been built on my Mac. I use sdkman to manage my JDK installs.

The CI builds all seem fine.

I have no objection to putting back the `javabootclasspath` if it makes the build work for you. In the short term, I would be worried that there is something up with RC and the 1.0.0 release.

som-snytt: I haven't looked at this conversation, but FYI there was a bug fix to prefer javabootclasspath, someone needed to compile against a specific version of rt.jar. This PR https://github.com/scala/scala/pull/10336

pjfanning: The RC1 pekko-remote jar: I ran the `javap -c` on the EnvelopeBuffer class and got

```
      9: invokevirtual #121                // Method java/nio/ByteBuffer.clear:()Ljava/nio/Buffer;
```

Which I think means that the RC1 jar is not affected by Issue #27079

pjfanning: > I haven't looked at this conversation, but FYI there was a bug fix to prefer javabootclasspath, someone needed to compile against a specific version of rt.jar. This PR [scala/scala#10336](https://github.com/scala/scala/pull/10336)

Thanks @som-snytt - the Pekko build uses Scala 2.13.11 so we might be benefitting from the https://github.com/scala/scala/pull/10336 change.

PR #268, opened 2023-03-23
=sbt Bump reactive-stream-version to 1.0.4

He-Pin:
Motivation:

Upgrade reactive stream to the latest version

Result:

Up to date

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -57,7 +57,7 @@ object Dependencies {
   val scala3Version = "3.1.2"
   val allScalaVersions = Seq(scala213Version, scala212Version, scala3Version)
 
-  val reactiveStreamsVersion = "1.0.3"
+  val reactiveStreamsVersion = "1.0.4"
 
   val sslConfigVersion = Def.setting {
     if (scalaVersion.value.startsWith("3.")) {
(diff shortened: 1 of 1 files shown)


pjfanning:
I'm not against this but could you spell out the pros and cons? In the end, we are being conservative about what changes go into v1.0.0.

He-Pin:
@pjfanning I'm trying to port my PRs to pekko. And I think this is fine as it's not changing any surface API.

Release note: https://github.com/reactive-streams/reactive-streams-jvm/blob/master/RELEASE-NOTES.md

pjfanning (approved):
I'm ok with upgrading here (in main branch)

mdedetrich:
So I just read the chaneglog and it seems nothing of significance has changed aside from TCK (technology confirmation kit) which seems to be some verification toolkit???

Pinging @raboof as well incase we miss something but otherwise change seems safe. I guess even if downstream users somehow resolve `1.0.3` it's not going to break anything (but then again if that's the case then there is an argument that we don't really need to bump it now in contrast to let's say Jackson and how problematic an old Jackson version had been)

He-Pin:
Let me merge this as the API changes nothing.

Issue #7, opened 2022-11-03
Upgrade Jackson version (2.14.2)


pjfanning:
There is a performance issue for Java users when using jackson-module-scala v2.13.x. Not too many users have reported issues but it might be best to update any v2.13 dependency to v2.14.1 (or newer version). We can keep the v2.11.4 dependency for Scala 2.x. We can upgrade to jackson v2.14.x for all Scala versions after Pekko's initial release.

https://github.com/FasterXML/jackson-module-scala/issues/576

jrudolph:
Actually, Jackson is a huge problem, since it leads to all kinds of compatibility problems in downstream projects.

https://github.com/apache/incubator-pekko-projection/pull/19 shows a problem where we probably cannot release a 1.0.0 of pekko-projection without upgrading Jackson everywhere (i.e. we can release it but it won't work with pekko-persistence-cassandra which might be one of the main combination that people are using). The reason is that the last OS version of akka-projection is built against an old version of alpakka-cassandra (2.0.1) that we don't intend to support (pekko-connectors is based on alpakka 3).

IMO we should under no circumstances stay on Jackson 2.11 which is not supported any more and has all kinds of published security issues that will be immediately flagged for users. The question would still be which version to use instead (2.12, 2.13, or 2.14). This is hard to find out because so many potential combinations of subprojects need to be considered. Given the history of problems with Jackson, I would recommend going to the latest version (2.14) because that will hold up for the longest time.

Updating the version here will mean that we might pull up Jackson for our users. I consider that a much smaller problem (since there will be all kinds of challenges users might have while updating to Pekko) than staying on the old and broken version.

pjfanning:
Can't we downgrade the cassandra driver - to one that uses Jackson 2.11?

Jackson 2.12.7.1 has no CVEs if we want to upgrade but not go all the way to Jackson 2.14. Jackson 2.15 is out soon but that introduces a [StreamReadConstraints](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.15.0-rc1/com/fasterxml/jackson/core/StreamReadConstraints.html) setting that defaults to high limits but not unlimited values so we would probably need to add settings in reference.conf that allow users to choose higher limits.

mdedetrich:
https://github.com/apache/incubator-pekko-projection/pull/19#issuecomment-1484991058 is relevant

pjfanning:
this is done

PR #270, opened 2023-03-24
Scala 3.3 support

mdedetrich:
Preliminary support for Scala 3.3 (currently based on RC3). Aside from one case (noted below), there hasn't been any non trivial change in sources. Most of the source changes were due to update in Scala 3.3 checking more cases and hence producing more warnings (which isn't that different from latest Scala 2.13 or Scala 2.12).

PR #273, opened 2023-03-27
upgrade to jackson 2.14.2

pjfanning:
https://github.com/apache/incubator-pekko/issues/7

* Jackson 2.14.2 for Scala 3 - supports only Scala 3.2 - so we'll need to upgrade Scala
* There is a deprecation warning - a Jackson method that we use that has now been deprecated.
* The deprecated method needs a large scale rewrite to do things the preferred Jackson 2.14 way
* Jackson is heading towards making ObjectMapper immutable and the preferred approach is to use `JsonMapper.builder()...` to create a configured MapperBuilder and to create JsonMapper/ObjectMapper instances from it
* In short term, we are probably bestter off not moving off the deprecated method - it won't be removed for ages

```
[error] /home/runner/work/incubator-pekko/incubator-pekko/serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala:170:45: method configure in class ObjectMapper is deprecated
[error]       case (feature, value) => objectMapper.configure(feature, value)
[error]                                             ^
[error] one error found
```

### DIFF
--- a/actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/InteractionPatternsSpec.scala
+++ b/actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/InteractionPatternsSpec.scala
@@ -36,7 +36,6 @@ import org.scalatest.wordspec.AnyWordSpecLike
 // #per-session-child
 // dummy data types just for this sample
 case class Keys()
-
 case class Wallet()
 
 // #per-session-child
--- a/actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/InteractionPatternsSpec.scala
+++ b/actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/InteractionPatternsSpec.scala
@@ -33,6 +33,14 @@ import pekko.actor.typed.scaladsl.TimerScheduler
 import pekko.pattern.StatusReply
 import org.scalatest.wordspec.AnyWordSpecLike
 
+// #per-session-child
+// dummy data types just for this sample
+case class Keys()
+
+case class Wallet()
+
+// #per-session-child
+
 class InteractionPatternsSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike with LogCapturing {
 
   "The interaction patterns docs" must {
@@ -359,12 +367,6 @@ class InteractionPatternsSpec extends ScalaTestWithActorTestKit with AnyWordSpec
   }
 
   "contain a sample for per session child" in {
-    // #per-session-child
-    // dummy data types just for this sample
-    case class Keys()
-    case class Wallet()
-
-    // #per-session-child
 
     object KeyCabinet {
       case class GetKeys(whoseKeys: String, replyTo: ActorRef[Keys])
--- a/project/PekkoDisciplinePlugin.scala
+++ b/project/PekkoDisciplinePlugin.scala
@@ -134,7 +134,7 @@ object PekkoDisciplinePlugin extends AutoPlugin {
           case Some((2, 12)) =>
             disciplineScalacOptions
           case _ =>
-            Nil
+            Seq("-Wconf:cat=deprecation:s")
         }).toSeq,
         Compile / scalacOptions --=
           (if (looseProjects.contains(name.value)) undisciplineScalacOptions.toSeq
--- a/project/PekkoDisciplinePlugin.scala
+++ b/project/PekkoDisciplinePlugin.scala
@@ -48,6 +48,8 @@ object PekkoDisciplinePlugin extends AutoPlugin {
     // references to deprecated PARSER fields in generated message formats?
     "pekko-persistence-query",
     "pekko-docs",
+    // references to deprecated Jackson methods that would involve a significant refactor to avoid
+    "pekko-serialization-jackson",
     // use varargs of `Graph` in alsoTo and etc operators
     "pekko-stream-tests")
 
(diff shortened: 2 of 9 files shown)


nvollmar (approved):
lgtm

pjfanning:
@mdedetrich any idea how to get the Scala 3 build to ignore deprecation issues? https://github.com/apache/incubator-pekko/actions/runs/4532777359/jobs/7984657231?pr=273 is failing, seems to be because it won't allow deprecated code in the build

mdedetrich:
@pjfanning 

> @mdedetrich any idea how to get the Scala 3 build to ignore deprecation issues? https://github.com/apache/incubator-pekko/actions/runs/4532777359/jobs/7984657231?pr=273 is failing, seems to be because it won't allow deprecated code in the build

How about adding `@nowarn(cat="msg=deprecated")` on the function where its being used?

mdedetrich:
Alternately you can just copy https://github.com/apache/incubator-pekko/pull/270/files#diff-676691b2e85cad6026adb107942b7c9325394e20e015b131fdcebd3b2b7794b0R137 from my Scala 3.3 PR, that will just globally disable any deprecation warnings in the Scala3 build (which we will probably have to do in the future anyways).

mdedetrich:
@pjfanning For the latest error you probably need to add `serialization-jackson` into `PekkoDisciplinePlugin.nonFatalJavaWarningsFor` (along with a comment explaining why). The same error is also occurring in Scala 2.12

mdedetrich:
@pjfanning Okay so for the new errors in https://github.com/apache/incubator-pekko/actions/runs/4533274521/jobs/7985858691?pr=273 you probably need to port some of the fixes from https://github.com/apache/incubator-pekko/pull/270, specifically the `@unchecked` ones

pjfanning:
> @pjfanning Okay so for the new errors in https://github.com/apache/incubator-pekko/actions/runs/4533274521/jobs/7985858691?pr=273 you probably need to port some of the fixes from #270, specifically the `@unchecked` ones

Thanks @mdedetrich - it looks like I needed all your code changes from your Scala 3.3 PR

mdedetrich (approved):
So I am going to approve this PR but since there was some ambivalence about upgrading the Scala version (see https://github.com/apache/incubator-pekko/issues/6 and https://lists.apache.org/thread/mk88ybzx51k7x7x102drtbzh6l68k0xm) we should probably check with community that this is broadly okay before merging (as people may already know, personally I am all for updating the Scala 3 version)

danischroeter (approved):

project/PekkoDisciplinePlugin.scala
```diff
@@ -134,7 +136,7 @@ object PekkoDisciplinePlugin extends AutoPlugin {
           case Some((2, 12)) =>
             disciplineScalacOptions
           case _ =>
-            Nil
+            Seq("-Wconf:cat=deprecation:s")
```
jrudolph, line 139: Does that mean we disable all deprecation warnings for Scala 3 for all submodules. Is that really what we want?

You can also match by other factors, so maybe we can restrict the silencing to Jackson in particular?

actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/InteractionPatternsSpec.scala
```diff
@@ -33,6 +33,14 @@ import pekko.actor.typed.scaladsl.TimerScheduler
 import pekko.pattern.StatusReply
 import org.scalatest.wordspec.AnyWordSpecLike
 
+// #per-session-child
+// dummy data types just for this sample
+case class Keys()
+
```
jrudolph, line 39:
```suggestion
```

jrudolph (commented):
In general, lgtm. I don't quite see how the general silencing and the `@nowarn` and `@unchecked` annotation changes play together here. Can we find a way to silence exactly the deprecation warnings we are concerned about and only in that submodule?

mdedetrich:
> I don't quite see how the general silencing and the @nowarn and @unchecked annotation changes play together here. Can we find a way to silence exactly the deprecation warnings we are concerned about and only in that submodule?

@jrudolph So the changes here were copied from my work in Scala 3.3 (see https://github.com/apache/incubator-pekko/pull/270) because it turns out all of the Scala 3 specific changes done in the Scala 3.3 PR due to Scala 3 changes was actually add in Scala 3.2 (which this PR is updating to).

With the deprecation warnings, there was a lot of places that Scala 3.2/3.3 was warning about deprecation so at least in my Scala 3.3 PR I made the decision to add the flag in order to reduce noise. While it was possible to use the `@nowarn` flag to silence the deprecation, these deprecation warnings was being generated in code that is being shared between Scala 2 and Scala 3 which means that if we were to use the `@nowarn` flag we would have to do changes to the Scala 2 compiler flags in order to not complain if no warnings were silenced (latest versions of Scala 2.12/2.13 will generate a warning if you use a `@nowarn` annotation that doesn't actually suppress anything, and iirc these deprecations  warnings are Scala 3.2/3.3 specific and don't occur in Scala 2.12/2.13)

Regarding `@unchecked`, this seems to be due to an improvement that landed in Scala 3.2 which improves the exhaustiveness checking for pattern matching cases. Originally in my Scala 3.3 PR I tried to find a flag which covers only these new cases that require `@unchecked` so it can be silenced but it appears that the change in Scala 3.2 is extending an already existing case of exhaustiveness checking.

pjfanning:
@mdedetrich @jrudolph are there further changes needed here or is it safe to proceed with a merge?

jrudolph (approved):
LGTM

mdedetrich:
LGTM as well

mdedetrich (approved):
Lgtm

Issue #261, opened 2023-03-21
Make BroadcastHub only start pulling after atleast one Consumer attached to it?


He-Pin:
### Discussed in https://github.com/apache/incubator-pekko/discussions/255

<div type='discussions-op-text'>

<sup>Originally posted by **He-Pin** March 20, 2023</sup>
Refs: https://github.com/akka/akka/issues/28327

I just made a PR in akka/akka, which added a `startAfterNrOfConsumers` to `BroadcastHub`. And surely I will send the PR to pekko too.

I think we should make the current `BroadcastHub.sink[T]` with `BroadcastHub.sink[T] = BroadcastHub.sink(startAfterNrOfConsumers = 1, defaultBufferSize)` in pekko by default, which is `0` now.</div>

https://stackoverflow.com/questions/58974630/broadcasthub-doesnt-wait-for-first-materialization-of-source

And when I try to reimplement the current `Sink.fromPublisher` with `BroadcastHub`, I found some other behavior need to address to unlock it.

PR #275, opened 2023-03-27
+str Add `startAfterNrOfConsumers` to BroadcastHub.

He-Pin:
Refs: https://github.com/akka/akka/issues/28327

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
@@ -305,6 +305,52 @@ class HubSpec extends StreamSpec {
       f2.futureValue should ===(1 to 10)
     }
 
+    "broadcast elements to downstream after at least one subscriber" in {
+      val broadcast = Source(1 to 10).runWith(BroadcastHub.sink[Int](1, 256))
+      val resultOne = broadcast.runWith(Sink.seq) // nothing happening yet
+
+      Await.result(resultOne, 1.second) should be(1 to 10) // fails
+    }
+
+    "broadcast all elements to all consumers" in {
+      val sourceQueue = Source.queue[Int](10) // used to block the source until we say so
+      val (queue, broadcast) = sourceQueue.toMat(BroadcastHub.sink(2, 256))(Keep.both).run()
+      val resultOne = broadcast.runWith(Sink.seq) // nothing happening yet
+      for (i <- 1 to 5) {
+        queue.offer(i)
+      }
+      val resultTwo = broadcast.runWith(Sink.seq)
+      for (i <- 6 to 10) {
+        queue.offer(i)
+      }
+      queue.complete() // only now is the source emptied
+
+      Await.result(resultOne, 1.second) should be(1 to 10)
+      Await.result(resultTwo, 1.second) should be(1 to 10)
+    }
+
+    "broadcast all elements to all consumers with hot upstream" in {
+      val broadcast = Source(1 to 10).runWith(BroadcastHub.sink[Int](2, 256))
+      val resultOne = broadcast.runWith(Sink.seq) // nothing happening yet
+      val resultTwo = broadcast.runWith(Sink.seq)
+
+      Await.result(resultOne, 1.second) should be(1 to 10)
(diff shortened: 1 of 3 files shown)


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
```diff
@@ -511,7 +539,9 @@ private[pekko] class BroadcastHub[T](bufferSize: Int)
     override def preStart(): Unit = {
       setKeepGoing(true)
       callbackPromise.success(getAsyncCallback[HubEvent](onEvent))
-      pull(in)
+      if (startAfterNrOfConsumers == 0) {
```
He-Pin, line 542: The old behavior

mdedetrich:
So I had a quick look into this and it appears to be the same as the equivalent upstream akka PR and it was merged there so there shouldn't be any problems. Only question arising from https://github.com/apache/incubator-pekko/issues/261 is whether for Pekko 1.1 `BroadcastHub.sink` should default to `startAfterNrOfConsumers = 1`.

Ill have a proper look in the next couple days.

He-Pin:
Fail is https://github.com/apache/incubator-pekko/issues/276
Not relevant.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Hub.scala
```diff
@@ -164,6 +164,34 @@ object BroadcastHub {
...
+   * cancelled are simply removed from the dynamic set of consumers.
+   *
+   * @param clazz                   Type of elements this hub emits and consumes
+   * @param startAfterNrOfConsumers Elements are buffered until this number of consumers have been connected.
+   *                                This is only used initially when the operator is starting up, i.e. it is not honored when consumers have
+   *                                been removed (canceled).
+   * @param bufferSize              Buffer size used by the producer. Gives an upper bound on how "far" from each other two
+   *                                concurrent consumers can be in terms of element. If the buffer is full, the producer
+   *                                is backpressured. Must be a power of two and less than 4096.
+   */
```
pjfanning, line 189: can you add `@since 1.1.0` to all the new functions added to the non-test source?

He-Pin: The method have tests.

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
```diff
@@ -305,6 +305,52 @@ class HubSpec extends StreamSpec {
...
+    "broadcast elements to downstream after at least one subscriber" in {
+      val broadcast = Source(1 to 10).runWith(BroadcastHub.sink[Int](1, 256))
+      val resultOne = broadcast.runWith(Sink.seq) // nothing happening yet
+
+      Await.result(resultOne, 1.second) should be(1 to 10) // fails
+    }
+
+    "broadcast all elements to all consumers" in {
+      val sourceQueue = Source.queue[Int](10) // used to block the source until we say so
+      val (queue, broadcast) = sourceQueue.toMat(BroadcastHub.sink(2, 256))(Keep.both).run()
```
He-Pin, line 317: @pjfanning This is the test.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Hub.scala
pjfanning: I wasn't asking about tests - I was asking you to add `@since 1.1.0` to the scaladoc.

pjfanning: like https://github.com/apache/incubator-pekko/blob/4169e7d476015fce84494725da2400d915f9318e/actor/src/main/scala/org/apache/pekko/actor/FSM.scala#L958

He-Pin: I see, will update.

pjfanning (approved):
lgtm - but I would appreciate if someone else has a look too

He-Pin:
@mdedetrich would you like give another  look, thanks

He-Pin:
I think this code is fine as it was merged into akka too. let me merge it to keep the pr queue shorter.

PR #278, opened 2023-04-01
=str Implement Source.never as a dedicated GraphStage.

He-Pin:
Using a more straightforward implementation will reduce the AsyncCallback in FutureSource.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphStages.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphStages.scala
@@ -458,6 +458,21 @@ import pekko.stream.stage._
   }
 
   @InternalApi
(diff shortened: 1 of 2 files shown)


jrudolph:
Is that a concrete concern or just a cleanup? Not sure if it is necessary to add code for a case that is only used in tests... (but also not strictly opposed).

He-Pin:
> Is that a concrete concern or just a cleanup? Not sure if it is necessary to add code for a case that is only used in tests... (but also not strictly opposed).

i think this new is more easy to understand.

mdedetrich (changes requested):
Would it be possible add the `NeverSource` to the pekko stream docs so people are aware of it? Also if its easy to do so, a quick test to verify that the behaviour works as expeted.

He-Pin:
@mdedetrich I checked there are `NeverSourceSpec` and `never.md` already.

mdedetrich (approved):
lgtm

PR #281, opened 2023-04-06
Add FunctionConverters, FutureConverters and ObjectConverters

mdedetrich:
The primary goal of this PR is two things

* Get rid of the `scala-java8-compat` library dependency if you are using newer versions of Scala where its not necessary (Scala 2.13 and newer specifically), see https://github.com/scala/scala-java8-compat#do-you-need-this .
* Make it easier to handle the dropping of Scala 2.12 support. This is due to the fact that since we are currently using `scala-java8-compat`, some of the syntax in `scala-java8-compat` doesn't exist in `scala.jdk` (which is the replacement for `scala-java8-compat` on Scala 2.13 and newer). In this PR, care has been taken when adding `FunctionConverters`, `FutureConverters` and `ObjectConverters` to only use methods/extension methods that exist or are not deprecated for Scala 2.13+
  * This is the reason why, for example, `[asJavaPredicate(s => Objects.isNull(s))](predicate: Predicate[State] = asJavaPredicate(s => Objects.isNull(s)))` has been replaced with `predicate = ((s: State) => Objects.isNull(s)).asJava` since 
  `.asJava` is the only method that exists in both `scala.jdk` and `scala-java8-compat`.
  * Also explains why method calls were changed from `asScala` to `toScala`, its because `asScala` is deprecated
  * `OptionConverters.toScala` doesn't exist in Scala 2.13
  
In general this was achieved by delegating `scala-java8-compat` only for Scala 2.12 (and `scala-java8-compat` is only included in pekko for Scala 2.12) and delegating to `scala.jdk` for Scala 2.13+ (`scala.jdk` is included automatically since its in the scala runtime).

This feature will have an impact on end users, specifically in these cases
* They are using Scala 2.13+ and transitively relying on `scala-java8-compat` as a dependency.
* They are using Scala 2.13+ and are explicitly relying on an older binary incompatible `scala-java8-compat` as a dependency.

Both of these cases have been documented in the migration guide.

One thing to discuss is when this PR should be merged (i.e. for Pekko 1.0.x or 1.1.x). There are arguments either way, doing it earlier in 1.0.x means that its less painful for end users when we drop support for Scala 2.12 since Scala 2.13+ will not contain `scala-java8-compat`, so end users don't have to worry about it at all. On the other hand it is technically speaking a breaking change, however it can be argued that in the cases of breaking changes you should have migrated to `scala.jdk` and sbt (via eviction warnings)/compiler will warn you in such cases. Furthermore we have already done some breaking changes in Pekko anyways (when it comes to dependencies, ergo Jackson).

Also do note that when this PR will be merged we will have to update all of the Pekko modules since the Pekko modules are cross compiled against Scala 2.12/Scala 2.13 so even if this PR is approved, a convenient time needs to be chosen for when to merge this.


mdedetrich:
> Could we consider delaying this to v1.1?

I wrote a paragraph on this, there are pros and cons to delaying this to 1.1. Personally when I ran through the cases in my head about how this could theoretically break, I couldn't come to a scenario where a user wouldn't be notified. The cases that would occur is that compile would no longer work (where as documented you would just add `scala-java8-compat` to your dependencies)

He-Pin:
i think the overall design is good but we said pekko 1.0 should be a carbon copy of 2.6.20, so defer this to 1.1.0？

mdedetrich:
> pekko 1.0 should be a carbon copy of 2.6.20, so defer this to 1.1.0？

This really isn't that true anymore, also I edited my previous point but one good argument for 1.0.x is if we decide to drop Scala 2.12 support in 1.1.x.

actor/src/main/scala/org/apache/pekko/actor/Actor.scala
```diff
@@ -100,8 +100,8 @@ final case class ActorIdentity(correlationId: Any, ref: Option[ActorRef]) {
    * not defined if no actor matched the request.
    */
   def getActorRef: Optional[ActorRef] = {
-    import scala.compat.java8.OptionConverters._
-    ref.asJava
+    import pekko.util.OptionConverters._
+    ref.toJava
```
He-Pin, line 104: so all the `as` changed to `to`?

He-Pin: And I think the title should be `OptionConverters `?

pjfanning:
I guess that I would be happy enough to merge this now. The conversions are internal - to support the Pekko Java API. And the new conversion functions are marked as internal.

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/javadsl/CommandHandlerWithReply.scala
```diff
@@ -117,7 +116,7 @@ final class CommandHandlerWithReplyBuilder[Command, Event, State]() {
    * @return A new, mutable, CommandHandlerWithReplyBuilderByState
    */
   def forNullState(): CommandHandlerWithReplyBuilderByState[Command, Event, State, State] = {
-    val predicate: Predicate[State] = asJavaPredicate(s => Objects.isNull(s))
+    val predicate = ((s: State) => Objects.isNull(s)).asJava
```
He-Pin, line 119: Hah, I see

actor/src/main/scala/org/apache/pekko/actor/Actor.scala
mdedetrich: > And I think the title should be `OptionConverters `?

Don't understand what you mean by this? Is there a typo or are you thinking of something else.

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/javadsl/CommandHandlerWithReply.scala
He-Pin: Btw, Why can this be ：
```scala
val predicate: Predicate[State] = (s: State) => Objects.isNull(s)
```

mdedetrich: > Btw, Why can this be ：
> 
> ```scala
> val predicate: Predicate[State] = (s: State) => Objects.isNull(s)
> ```

Nope, there aren't implicit conversions. `(s: State) => Objects.isNull(s)` is of type `Function1[State, Boolean]` and the .`asJava` turns that into a `Predicate[State]`).

actor/src/main/scala/org/apache/pekko/actor/Actor.scala
He-Pin: @mdedetrich I was trying to find a class named `ObjectConverters`

mdedetrich: Ah so you can find it when reviewing?

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/javadsl/CommandHandlerWithReply.scala
mdedetrich: Yes, the type needs to be `Predicate[State]` (which is actually a Java `Functional` interface)

actor/src/main/scala/org/apache/pekko/actor/Actor.scala
mdedetrich: My bad, a had a mental blip. `ObjectConverters` is meant to be `OptionConverters`. Edited original post to also reflect this.

mdedetrich:
@jrudolph @raboof Would it be possible to get your opinion on this? The decision can affect if we want to land this before 1.0.0 or  after

He-Pin:
Is there any ETA of 1.1.0？

mdedetrich:
> Is there any ETA of 1.1.0？

No idea, but it would be good to get an broad answer to my question before we start considering a 1.0.0 release because as both mentioned here and on the mailing list there are good reasons to put this change into 1.0.0.

pjfanning:
We are a long way from v1.1. Whenever v1.0.0 is released, we can split branches - 1.0 support and 1.1 support - with their own snapshots.

pjfanning:
I support adding this PR now since it has limited impact on the API.

mdedetrich:
> I support adding this PR now since it has limited impact on the API.

In that case if you are happy to do so can you review the PR? I wont merge it immediately because I do genuinely want feedback from others but if no one comments on it by the time we actually do a release I will take it as an implicit acceptance.

pjfanning (approved):
Lgtm

He-Pin:
Cool, then, i think we can provide more than a carbon copy of 2.6.20。

He-Pin (approved):
lgtm, as not affect much of the user facing api.

mdedetrich:
@pjfanning I want to go ahead and merge this today. Is it okay if we delay the MILESTONE release until we find out that this PR works with every Pekko module (I will go ahead and update the modules myself). There may be a chance that I might miss one of the collection compat methods and if that happens then I will have to go back and add some more changes into Pekko.

pjfanning:
sure - it could be a week or 2 before scala 3.3 comes out anyway

mdedetrich:
Okay merging now then.

PR #283, opened 2023-04-12
Add ScalaTestWithActorTestKitBase trait to handle multiple extends

mdedetrich:
This PR's factors out the common logic of `ScalaTestWithActorTestKit` into a trait `ScalaTestWithActorTestKitBase` while keeping `ScalaTestWithActorTestKit` as is (i.e. an `abstract class`). The context behind this change is that I was looking at flaky tests within https://github.com/apache/incubator-pekko-persistence-r2dbc and I noticed that one of the reasons behind the flakiness is a hardcoded timeout value (see https://github.com/apache/incubator-pekko-persistence-r2dbc/blob/main/.github/workflows/build-test.yml#L75).

While there may be "easier" ways to solve this problem, a much more principled way is to use testcontainers/testcontainers-scala which already has logic (from community effort) that handles these problems so in reality it just works™. 

The easiest way to integrate testcontainers-scala with scalatest/testkit is to just define your test suites within a trait and then have a the different testcontainers extend those traits (as documented here https://github.com/testcontainers/testcontainers-scala/blob/master/docs/src/main/tut/usage.md#lifecycle). Unfortunately currently this is not really possible because `ScalaTestWithActorTestKitBase` is an abstract class and there is no equivalent version that is a trait (in stark contrast to `TestKitBase` which is a trait and not an `abstract class`).

With this feature, we can (for example) change https://github.com/apache/incubator-pekko-persistence-r2dbc/blob/ab3709fec338b2641ae00350e3b07512e2da4acf/core/src/test/scala/org/apache/pekko/persistence/r2dbc/state/CurrentPersistenceIdsQuerySpec.scala#L26 to a `trait CurrentPersistenceIdsQuerySpec` and then have an `PostGresCurrentPersistenceIdsQuerySpec` and `YugabyteCurrentPersistenceIdsQuerySpec` extend `CurrentPersistenceIdsQuerySpec`). This is also useful aside from `Testcontainers`, basically as documented whenever you need to abstract over your test suite you can use this trait.

This PR won't have any effect on behaviour (it just adds a new trait) and furthermore this is also not going to impact production code because we are dealing with testkit.

### DIFF
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/scaladsl/ScalaTestWithActorTestKit.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/scaladsl/ScalaTestWithActorTestKit.scala
@@ -38,13 +38,9 @@ import pekko.actor.typed.ActorSystem
  * The application.conf of your project is not used in this case.
  * A specific configuration can be passed as constructor parameter.
  */
-abstract class ScalaTestWithActorTestKit(testKit: ActorTestKit)
+abstract class ScalaTestWithActorTestKit(override val testKit: ActorTestKit)
     extends ActorTestKitBase(testKit)
-    with TestSuite
-    with Matchers
-    with BeforeAndAfterAll
-    with ScalaFutures
-    with Eventually {
+    with ScalaTestWithActorTestKitBase {
 
   /**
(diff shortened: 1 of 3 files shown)


pjfanning (commented):
Lgtm

pjfanning (approved):

PR #285, opened 2023-04-14
update links to point to pekko sample docs

pjfanning:
some samples that are referenced in pekko docs are not in https://github.com/apache/incubator-pekko-samples - so I have set those links up to work the old Akka docs way (and link to the Akka sample)

### DIFF
--- a/project/Paradox.scala
+++ b/project/Paradox.scala
@@ -35,8 +35,8 @@ object Paradox {
       "extref.platform-guide.base_url" -> "https://developer.lightbend.com/docs/akka-platform-guide/%s",
       "extref.wikipedia.base_url" -> "https://en.wikipedia.org/wiki/%s",
       "extref.github.base_url" -> (GitHub.url(version.value) + "/%s"), // for links to our sources
-      "extref.samples.base_url" -> "https://developer.lightbend.com/start/?group=akka&amp;project=%s",
-      "extref.ecs.base_url" -> "https://example.lightbend.com/v1/download/%s",
+      "extref.samples.base_url" -> s"$pekkoBaseURL/docs/pekko-samples/current/%s",
+      "extref.akka-samples.base_url" -> "https://developer.lightbend.com/start/?group=akka&amp;project=%s",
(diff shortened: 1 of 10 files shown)


project/Paradox.scala
```diff
@@ -35,8 +35,8 @@ object Paradox {
       "extref.platform-guide.base_url" -> "https://developer.lightbend.com/docs/akka-platform-guide/%s",
       "extref.wikipedia.base_url" -> "https://en.wikipedia.org/wiki/%s",
       "extref.github.base_url" -> (GitHub.url(version.value) + "/%s"), // for links to our sources
-      "extref.samples.base_url" -> "https://developer.lightbend.com/start/?group=akka&amp;project=%s",
-      "extref.ecs.base_url" -> "https://example.lightbend.com/v1/download/%s",
+      "extref.samples.base_url" -> s"$pekkoBaseURL/docs/pekko-samples/current/%s",
+      "extref.akka-samples.base_url" -> "https://developer.lightbend.com/start/?group=akka&amp;project=%s",
       "pekko.doc.dns" -> s"$pekkoBaseURL",
```
He-Pin, line 40: https://developer.lightbend.com/start/?group=akka&project=%s

?

pjfanning: We don't have Pekko versions of all the samples. `"extref.akka-samples.base_url"` has the same value as the previous setting for `"extref.samples.base_url"`. I've changed some of the `extref` settings in the docs to use this `akka-samples` based link - so that the generated link URL remains as before - because the `samples` based links have been changed to link to Pekko URLs if and when there is an actual working Pekko sample to link to.

He-Pin: @pjfanning what's the `&amp` in the `akka-sample.base_url`?

pjfanning: `&amp;` escapes `&`. This value is not new. I've just changed the extref prefix that maps to it.

He-Pin (approved):
lgtm

PR #288, opened 2023-04-14
=act Refactory AsyncDnsResolver to fix the race condition of `nextId()`

He-Pin:
The issue was first reported in akka/akka.
The `nextId()` can be called from different threads.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
@@ -14,23 +14,24 @@
 package org.apache.pekko.io.dns.internal
 
 import java.net.{ Inet4Address, Inet6Address, InetAddress, InetSocketAddress }
+import java.util.concurrent.atomic.AtomicInteger
 
 import scala.collection.immutable
-import scala.concurrent.ExecutionContextExecutor
-import scala.concurrent.Future
-import scala.util.Try
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.{ Failure, Success, Try }
 import scala.util.control.NonFatal
 
 import org.apache.pekko
 import pekko.actor.{ Actor, ActorLogging, ActorRef, ActorRefFactory }
 import pekko.annotation.InternalApi
+import pekko.dispatch.ExecutionContexts
+import pekko.event.LoggingAdapter
 import pekko.io.SimpleDnsCache
 import pekko.io.dns._
 import pekko.io.dns.CachePolicy.{ Never, Ttl }
 import pekko.io.dns.DnsProtocol.{ Ip, RequestType, Srv }
 import pekko.io.dns.internal.DnsClient._
-import pekko.pattern.{ ask, pipe }
-import pekko.pattern.AskTimeoutException
+import pekko.pattern.{ ask, pipe, AskTimeoutException }
 import pekko.util.{ Helpers, Timeout }
 import pekko.util.PrettyDuration._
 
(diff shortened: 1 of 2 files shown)


actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -47,110 +46,149 @@ private[io] final class AsyncDnsResolver(
...
+    settings.NameServers,
     settings.SearchDomains,
     settings.NDots)
 
-  private var requestId: Short = 0
-
-  private def nextId(): Short = {
-    requestId = (requestId + 1).toShort
-    requestId
-  }
```
He-Pin, line 93: This method can be called from different threads.

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -47,110 +48,150 @@ private[io] final class AsyncDnsResolver(
...
+      name: String,
+      requestType: RequestType,
+      resolvers: List[ActorRef],
+      cache: SimpleDnsCache,
+      log: LoggingAdapter): Future[DnsProtocol.Resolved] = {
+    cache.get((name, requestType)) match {
+      case Some(resolved) =>
+        log.debug("{} cached {}", requestType, resolved)
+        Future.successful(resolved)
+      case None => resolveWithResolvers(settings, name, requestType, resolvers, cache, log)
```
He-Pin, line 118: There may have multi queries in the fight if the cache is missed as it was

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -47,110 +48,150 @@ private[io] final class AsyncDnsResolver(
...
+
+  private val requestIdGen: AtomicInteger = new AtomicInteger(0)
+
+  private def nextId(): Short = requestIdGen.updateAndGet { current =>
+    if (current == Short.MaxValue) {
+      Short.MinValue
+    } else {
+      current + 1
+    }
+  }.toShort
```
He-Pin, line 105: Make use an AtomicInteger as it not inside the actor now.

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
pjfanning: is there a test case that shows this?

I'm broadly supportive of fixing this - but I am wondering if a singleton actor instance would be a more actor based solution. Actor instances show only be processing 1 message at a time.

He-Pin: Yes, but the SimpleDnsCache is threadSafe so I think that would be a little heavy, maybe a FunctionRef is more lightweight

He-Pin:
It passed on my pc, will check.

He-Pin:
> @kerr this is looking unlikely to get into the 1.0.0 release. Can we delay this to the 1.1.0 release?

I'm fine, but this is a known issue, let's roll out the first release then.

He-Pin:
@mdedetrich What's your option, I can allocate one day this weekend to work on this. or should we defer this to the next release?

mdedetrich:
I think the committers (definitely myself) are a bit busy/anxious about getting a release out soon so if its not a big deal I would prefer if we look into this after the first release. My attention is already stretched pretty thin so after a release of Pekko I should be able to focus on issues like this a lot more.

He-Pin:
@mdedetrich Then let's defer this

pjfanning:
@He-Pin is there anything in this PR that we need to add to the code merged for #385 ?

He-Pin:
@pjfanning I think this once can be closed now.

PR #289, opened 2023-04-15
Add sbt-dependency-check

mdedetrich:
This PR adds sbt-dependency-check which scans the entire dependency tree of Pekko against OWASP to scan for any potential CVE's. This PR does a pretty basic addition of the plugin but there is future room for improvement, i.e. since this is a standard sbt plugin its possible to run `dependencyCheckAggregate` along with `sourceDistGenerate` when making a release. Another notable setting is `dependencyCheckFailBuildOnCVSS` which would cause the project to automatically fail if it finds a dependency with a certain security level.

The contents of both the generated report and the surpression file are located in the `dependency` folder (this can be renamed if needed). This also means that the report will be contained in the source dist distribution.

Note that the report has actually picked up `CRITICAL` CVE's from our dependency list which we should look into, i.e. 
image

### DIFF
--- a/project/Paradox.scala
+++ b/project/Paradox.scala
@@ -17,6 +17,7 @@ import com.lightbend.paradox.sbt.ParadoxPlugin
 import com.lightbend.paradox.sbt.ParadoxPlugin.autoImport._
 import com.lightbend.paradox.apidoc.ApidocPlugin
 import com.lightbend.sbt.publishrsync.PublishRsyncPlugin.autoImport._
+import net.vonbuchholtz.sbt.dependencycheck.DependencyCheckPlugin.autoImport._
 import org.apache.pekko.PekkoParadoxPlugin.autoImport._
 import sbt.Keys._
 import sbt._
@@ -90,6 +91,24 @@ object Paradox {
 
   val parsingSettings = Seq(Compile / paradoxParsingTimeout := 5.seconds)
 
+  val sourceGeneratorSettings = Seq(
+    Compile / paradoxMarkdownToHtml / sourceGenerators += Def.taskDyn {
+      val targetFile = (Compile / paradox / sourceManaged).value / "security" / "dependency-check-report.md"
+      val sourceFile = (LocalRootProject / dependencyCheckOutputDirectory).value.get / "dependency-check-report.html"
+
+      (LocalRootProject / dependencyCheckAggregate).map { _ =>
+        val data = IO.readLines(sourceFile)
+        IO.delete(targetFile) // Since we are appending lets clear any existing files if they exist
+        IO.writeLines(targetFile,
+          List(
+            "## Dependency Check Report",
+            "```raw"), append = true)
+        IO.writeLines(targetFile, data, append = true)
+        IO.writeLines(targetFile, List("```"), append = true)
+        List(targetFile)
+      }
+    }.taskValue)
+
   val settings =
     propertiesSettings ++
     rootsSettings ++
@@ -97,6 +116,7 @@ object Paradox {
     groupsSettings ++
     parsingSettings ++
     themeSettings ++
+    sourceGeneratorSettings ++
     Seq(
       Compile / paradox / name := "Pekko",
       resolvers += Resolver.jcenterRepo,
--- a/dependency-check/suppression.xml
+++ b/dependency-check/suppression.xml
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+</suppressions>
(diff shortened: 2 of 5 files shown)


mdedetrich:
So I just checked the netty dependency which has the `CRITICAL` level but there isn't anything that can be done about it since there isn't a new proper release that solves it (see https://mvnrepository.com/artifact/io.netty/netty).

@pjfanning I can add this netty version into `suppression.xml` since there isn't anything we can do about it, just let me know if you want me to update the PR for this.

pjfanning:
do we really need to commit the html file?

mdedetrich:
> do we really need to commit the html file?

I don't know? The primary point of the plugin is to create this file. One thing to note is that the generation of this file takes a **looong** amount of time. This is due to the fact that it downloads the entire OWASP database locally, does some processing on it and then scans all of your dependencies. You can think of it as similar to compiled game assets, i.e. theoretically speaking you shouldn't commit these into git but because they are expensive to generate and rarely change its done for pragmatic/practical reasons.

I would also presume that having this committed in git would make Apache review/release process of Pekko a lot easier (you just need to open the file) and since the report (at least currently) is in the source dist package its a nice QOL feature for Pekko users. As stated before we would make sure that the report is in sync when a release is made but this also can be automated.

Also incase you haven't opened the file, I would recommend do so because its actually interactive (i.e. its not a static text). i.e. it gives you the ability to direct to the original OWASP vulnerability and you can also use the generated html file to add suppressions, i.e.

image

Ultimately I have no problem with removing it from git but then we kind of need to ascertain what the goals are here. At least from my experience in working with companies for legal/license/CVE's, a report is always the best way to handle this and thats why these plugins generally make reports. This plugin is also just a wrapper for the well known https://github.com/jeremylong/DependencyCheck whos entire point is also just to generate reports.

He-Pin:
Or a weekly action can do too？

mdedetrich:
> Or a weekly action can do too？

We can do this, but the point is still where do we persist the file? I guess we can upload it as part of pekko docs and then rsync it/commit it to git. @pjfanning would you prefer this?

pjfanning:
We could rsync it to nightlies site.

kw217:
I agree, having the report pre-generated is really useful for consumers (esp. corporate). But we shouldn't have big generated artifacts in git (it slows down clones, for one thing). It should go on the nightlies site, or near the docs.

mdedetrich:
So I plan on changing this so it just gets included in the docs. I have similar work with sbt-license-report which I have yet to submit.

mdedetrich:
So I plan on changing this so it just gets included in the docs. I have similar work with sbt-license-report which I have yet to submit.

pjfanning:
If we remove the html file and possibly add the filename to .gitignore, I think we can merge this. We could do a 2nd PR that automates the publishing of the HTML.

mdedetrich:
In that case let me spend a few minutes today trying to automate this considering I already did it for `sbt-license-report`, I will ping when I update the PR.

mdedetrich:
So I did some work on this but due to the amount of time it takes to get the NVD database I have some concerns about how nicely it would work with github actions especially when integrated with paradox so I opened up an issue upstream at https://github.com/albuch/sbt-dependency-check/issues/305

mdedetrich:
Good news is that after finally managing to get the NVD database to download onto my system I managed to integrate sbt-dependency-check directly into our paradox docs

image

The bad news is that as you can tell, since I am inlining the html the styling doesn't look the best. Unfortunately sbt-dependency-check doesn't have markdown as a supported output format which leaves us with 2 options

* Use a much more strict/data orientated format such as JSON/XML, parse it and then convert it into markdown so that it renders better
* Try to find a css stylesheet and wrap the raw block in a div as documented here https://developer.lightbend.com/docs/paradox/current/directives/css-friendliness.html#raw-text-in-fenced-blocks so that it doesn't look so bad.

project/Paradox.scala
```diff
@@ -90,13 +91,32 @@ object Paradox {
 
   val parsingSettings = Seq(Compile / paradoxParsingTimeout := 5.seconds)
 
+  val sourceGeneratorSettings = Seq(
+    Compile / paradoxMarkdownToHtml / sourceGenerators += Def.taskDyn {
+      val targetFile = (Compile / paradox / sourceManaged).value / "security" / "dependency-check-report.md"
+      val sourceFile = (LocalRootProject / dependencyCheckOutputDirectory).value.get / "dependency-check-report.html"
+
+      (LocalRootProject / dependencyCheckAggregate).map { _ =>
+        val data = IO.readLines(sourceFile)
```
mdedetrich, line 100: There are probably better/more performant ways to do this but we can also look into this later

mdedetrich:
Closing this PR since https://github.com/apache/incubator-pekko/pull/366 is already solving this problem in a more simple way.

Issue #247, opened 2023-03-14
correct values for project-info.conf jdk-versions


pjfanning:
currently have:
```
  shared-info {
    jdk-versions: ["Adopt OpenJDK 8", "Adopt OpenJDK 11"]
```

Adopt JDK is no longer supported. Maybe Temurin or Adoptium.

Generally need to work out what we want to appear in the docs.

pjfanning:
#293 merged

PR #293, opened 2023-04-17
Update project-info JDKs

pjfanning:
We test with Temurin OpenJDK 8, 11 and 17. Adopt OpenJDK is no longer supported and Adoptium have taken over and name their JDKs as Temurin.

Relates to https://github.com/apache/incubator-pekko/issues/247

### DIFF
--- a/project/project-info.conf
+++ b/project/project-info.conf
@@ -3,7 +3,7 @@ project-info {
   scaladoc: "https://pekko.apache.org/api/pekko/"${project-info.version}"/pekko/"
   javadoc: "https://pekko.apache.org/japi/pekko/"${project-info.version}"/pekko/"
   shared-info {
-    jdk-versions: ["Adopt OpenJDK 8", "Adopt OpenJDK 11"]
+    jdk-versions: ["Temurin OpenJDK 8", "Temurin OpenJDK 11", "Temurin OpenJDK 17"]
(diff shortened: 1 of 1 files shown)


mdedetrich:
Can we just say "OpenJDK"? Adopt/Temurin are all distributions based off OpenJDK and while its true that we run Temurin in CI in the end whats important is that we are testing against OpenJDK.

pjfanning:
> Can we just say "OpenJDK"? Adopt/Temurin are all distributions based off OpenJDK and while its true that we run Temurin in CI in the end whats important is that we are testing against OpenJDK (or more accurately a distribution that has been compiled from OpenJDK sources).

changed

mdedetrich (approved):
So I am going to approve the PR, but technically speaking we don't test against JDK 11/17, we implicitly assume its support from it working with JDK 8.

Due to this there may be an argument that we should add OpenJDK 11/17 to the CI

pjfanning:
Nightly builds test JDK 11 and 17 - https://github.com/apache/incubator-pekko/actions/runs/4716114700

mdedetrich:
> Nightly builds test JDK 11 and 17 - https://github.com/apache/incubator-pekko/actions/runs/4716114700

Thanks, didn't realize that

Issue #245, opened 2023-03-13
remove or replace algolia and google analytics properties


pjfanning:
The values in project/Paradox.scala are Akka specific. We may not need the values at all. We might need developers with Lightbend background to review if they can be simply removed - or if we need to replace them with new Pekko values.

pjfanning:
removed in #295

PR #295, opened 2023-04-19
remove akka specific algolia and google analytics values

pjfanning:
see #245


project/Paradox.scala
```diff
@@ -61,10 +61,6 @@ object Paradox {
       "pekko.version" -> version.value,
       "scalatest.version" -> Dependencies.scalaTestVersion.value,
       "sigar_loader.version" -> "1.6.6-rev002",
-      "algolia.docsearch.api_key" -> "543bad5ad786495d9ccd445ed34ed082",
-      "algolia.docsearch.index_name" -> "akka_io",
-      "google.analytics.account" -> "UA-21117439-1",
-      "google.analytics.domain.name" -> "akka.io",
```
He-Pin, line 67: Any change to update with pekko's

pjfanning: ASF bans the use of Google Analytics for privacy reasons - see https://github.com/apache/incubator-pekko-sbt-paradox/pull/54

I don't know much about Algolia but
* I don't want to use Lightbend's api key
* We will have to jump through hoops to get ASF to approve us getting an Algolia account and since I don't even know what value-add it gives to the Pekko/Akka docs, I'd prefer to remove the settings and we can look at adding Pekko/ASF specific settings if and when it turns out Algolia is needed or super useful

He-Pin: It adds search ability.

pjfanning: search works anyway

PR #300, opened 2023-04-26
Improve PR Workflow: Separate the validateCompile step from validatePullRequest

sam-byng:
Currently the PR workflow 'Build / Test' job has a single step validatePullRequest that both compiles and then tests the changes. This takes 30 minutes and is very verbose -- it is difficult to see exactly where a failure is originating.

This PR separating compile and test: This will make the source of errors in steps clearer, without adding any extra pipeline time. (Steps in github jobs are dependent on one another, so the build files can be passed onto prValidation without the need for PR validation to compile again)

On the overall issue: I think next steps should be to extract test logs into their own short-lifetime artifact files and reduce the CI logs to only show failing tests. Will follow up with a PR on that if people think that's a good idea.

### DIFF
--- a/.github/workflows/build-test-prValidation.yml
+++ b/.github/workflows/build-test-prValidation.yml
@@ -33,6 +33,21 @@ jobs:
       - name: Enable jvm-opts
         run: cp .jvmopts-ci .jvmopts
 
+      - name: sbt validateCompile
+        run: |-
+          sbt \
+          -Dpekko.mima.enabled=false \
+          -Dpekko.test.multi-in-test=false \
+          -Dpekko.test.timefactor=2 \
+          -Dpekko.actor.testkit.typed.timefactor=2 \
+          -Dpekko.test.tags.exclude=gh-exclude,timing \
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm, this will improve the feedback loop

mdedetrich (approved):
Also lgtm, I would also cherry pick this into the 1.0.x branch once this gets .ergef.

@jrudolph Do you have any objections to this? To me it seems quote benign.

He-Pin:
@pjfanning What's option here

pjfanning:
merged - if there are build issues, we can fix them

He-Pin:
@sam-byng Merged, thanks.

mdedetrich:
I think its legitimate to cherry pick this onto the `1.0.x` branch, @sam-byng Do you want to do this?

mdedetrich:
Cherry pick PR against `1.0.x` made at https://github.com/apache/incubator-pekko/pull/630

Issue #203, opened 2023-02-17
regenerate the keystore/truststore used in `remote` tests


pjfanning:
* Current key has Lightbend and 'akka-remote'.
* https://github.com/apache/incubator-pekko/blob/main/remote/src/test/resources/Makefile
* will affect a few tests (if we rename 'akka-remote' to 'pekko-remote', a few expected values in tests will need to change)

sam-byng:
I will take this

PR #298, opened 2023-04-25
Fix #203: Update truststores/keystores for pekko-remote

sam-byng:
Partial fix for issue #203

1.  Regenerate the keystore/truststore for remote tests

    I've followed the guide at https://www.digicert.com/kb/ssl-support/openssl-quick-reference-guide.htm to update the makefile's domain.csr recipe. I then ran 'make' and deleted any files that were not there previously (i.e. the private key)

2. Update CN=akka-remote to pekko-remote to reflect new test cert

Other half of fix is in [another PR](https://github.com/apache/incubator-pekko/pull/301): Regenerate the keystore/truststore for stream-tests

PR #301, opened 2023-04-27
Issue 203: Update the truststore/keystore for stream-tests

sam-byng:
This and https://github.com/apache/incubator-pekko/pull/298 will close Issue #203 

Note that no SAN extension file is used in this case. Therefore the verify hostname code looks at the CN= instead of the subjectAltName field.
I've added a Makefile here to make it easier to regenerate this in future: just delete the trust/keystores and run 'make


stream-tests/src/test/resources/Makefile
```diff
@@ -0,0 +1,15 @@
+# Documents how truststore and keystore were created
```
pjfanning, line 1: do we need a new file? incubator-pekko/remote/src/test/resources/Makefile is very similar to this - can we update that instead?

sam-byng: The difference in this case is that there is no SAN extension for stream-tests. I think it's worth highlighting that. Agree this could be done in a shared file but then it would have to go in a logical location / we should use a symlink.

pjfanning: can you add a short(-ish) comment to the new file to highlight its similarity to remote/src/test/resources/Makefile ?

sam-byng: added

pjfanning (approved):
lgtm

Issue #246, opened 2023-03-13
update signature and fiddle properties in project/Paradox.scala


pjfanning:
These props are used in the Paradox doc generation.

```
      "signature.akka.base_dir" -> (ThisBuild / baseDirectory).value.getAbsolutePath,
      "fiddle.akka.base_dir" -> (ThisBuild / baseDirectory).value.getAbsolutePath,
```

They should (probably) be updated to replace 'akka' with 'pekko'.

They relate to `@@signature` and `@@fiddle` annotations in the doc markdown files.

We also have `scalafiddle.js`. It has:
```
window.scalaFiddleTemplates = {
  "Pekko": {
    pre: "// $FiddleDependency org.pekko-js %%% pekkojsactor % 2.2.6.1 \n" +
    "// $FiddleDependency org.pekko-js %%% pekkojsactorstream % 2.2.6.1 \n" +
    "// $FiddleDependency org.pekko-js %%% pekkojsactortyped % 2.2.6.1 \n",
    post: ""
  }
}
```
These values look wrong.

PR #302, opened 2023-04-27
Fixing #246: update signature and fiddle properties in project/Paradox.scala

samueleresca:
Fix issue #246 

- The `Paradox.scala` props have been updated to pekko value.
- The `scalafiddle.js` has been reverted to `org.akka-js`. Akka.js is distributed under [Scala LICENSE](https://www.scala-lang.org/license/)

### This is how is rendered after the conversion from scalafiddler to standard examples:

Screenshot 2023-04-27 161935

### DIFF
--- a/actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/IntroSpec.scala
+++ b/actor-typed-tests/src/test/scala/docs/org/apache/pekko/typed/IntroSpec.scala
@@ -109,7 +109,7 @@ object IntroSpec {
   }
   //#hello-world-main
 
-  // This is run by ScalaFiddle
+  // Entry point for the execution
   HelloWorldMain.main(Array.empty)
   //#fiddle_code
   //format: ON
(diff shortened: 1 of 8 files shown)


pjfanning:
I'm not sure how to proceed with this.

I've never used Scala Fiddle. https://scalafiddle.io/ seems to be broken. It's possible that we might be better off removing the scalafiddle stuff in the pekko docs and just presenting the sample code as something that the user needs to work out how to run.

samueleresca:
> I'm not sure how to proceed with this.
> 
> I've never used Scala Fiddle. https://scalafiddle.io/ seems to be broken. It's possible that we might be better off removing the scalafiddle stuff in the pekko docs and just presenting the sample code as something that the user needs to work out how to run.

Yes, Scalafiddle seems unresponsive (see https://github.com/scala/scala-lang/issues/1295) I can proceed by removing the snippet entirely.

samueleresca:
Also, regarding the `@@signature` placeholder:
- The only place it is used is in the `mergePrioritizedN.md` file
- The `mergePrioritizedN.md` Signature subheading is broken and pointing to the wrong Source path

I think this might be replaced with the @apidoc similar to the [mergePrioritized.md#Signature](https://github.com/apache/incubator-pekko/blob/main/docs/src/main/paradox/stream/operators/Source-or-Flow/mergePrioritized.md#signature)

pjfanning:
Thanks @samueleresca - seems like getting rid of the fiddle and signature stuff is best

pjfanning (approved):
lgtm

Issue #139, opened 2023-02-01
try to fix the multi-node CI build


pjfanning:
https://github.com/apache/incubator-pekko/blob/main/.github/workflows/multi-node.yml

This workflow is a copy of a Lightbend Akka workflow. It requires Google Cloud and a set of tokens to access it. 

We need to decide if these tests could be run on alternative infrastructure.

l2dy:
The test is based on Kubernetes, so it should be portable. However, 30+ GiB of memory (10 pods and each takes 3 GiB) is required, while GitHub-hosted runners have 7 GB only.

jrudolph:
We shouldn't make this a requirement for 1.0.0.

The multi-node tests mean that tests that are run as multi-jvm tests in the regular build are instead run distributed over multiple machines / VMs / containers to cover a more diverse set of working environments. The idea is that some race conditions might only turn up under more realistic environments than running on a single machine over localhost networking.

Since we will not change functionality for 1.0.0, there is no reason to expect any additional failures to turn up that would only show in multi-node tests (the very same tests are still run as multi-jvm tests as part of nightly builds).

They become more important when we start changing functionality and want to ensure changes that might change non-deterministic behavior are stable under a more diverse set of working conditions.

PR #303, opened 2023-04-28
Delete multi-node.yml

pjfanning:
We don't have a cluster to run these tests on and they just fail every night

relates to https://github.com/apache/incubator-pekko/issues/139


mdedetrich:
Is it a big deal to just comment the file rather than deleting it? I am in the process of trying to procure hardware at which point it would be useful.

pjfanning:
it is easy to revert this delete - there is an open issue (#139)

commenting out bits of the yml is difficult - comment the whole file and the builds will all fail - comment some parts mean that Github will run the job even if it completes after a few seconds - this is still bad for the world's environment, a complete waste of a VM, if only for a few seconds every night

mdedetrich (approved):
Im happy enough with an issue tracking this. Lgtm

PR #306, opened 2023-05-02
Add more format commits to .git-blame-ignore-revs

mdedetrich:
While browsing through the akka/pekko git history I noticed that are many more format commits than originally anticipated. This PR adds more formatting commits to `.git-blame-ignore-revs`.

Note that I have strictly only added commits that effect whitespace/formatting, there are other format commits which also fix other things (such as typos) which I haven't added since that is still useful information for git blame.


nvollmar (approved):
lgtm

PR #311, opened 2023-05-06
use github links as a 1st step away from lightbend links

pjfanning:
will later try to publish the paradox docs for the 2 quickstart projects but the github links act as a start


mdedetrich (approved):
lgtm

PR #312, opened 2023-05-08
Add toJavaPrimitive to OptionConverters

mdedetrich:
When trying to integrate the changes from https://github.com/apache/incubator-pekko/pull/281 into pekko-http I hit a stumbling block where I missed the `.toJavaPrimitive` methods so this PR adds these methods into our `OptionConverters` helper. Since the `.toJavaPrimitive` methods are not being used within Pekko core itself, `Scala212CompatTest` was added to make sure that these helper methods cross compile against all Scala versions so that we don't deal with any regressions.


He-Pin (approved):
lgtm

PR #314, opened 2023-05-08
Add OptionConverters.toJava and OptionConverters.toScala

mdedetrich:
And yet another case of finding more converters that have to be added to `OptionConverters` to get pekko-http with https://github.com/apache/incubator-pekko/pull/281 PR


He-Pin (approved):
lgtm

He-Pin:
I don't know if some methods are also being used in Playframework.

mdedetrich:
> I don't know if some methods are also being used in Playframework.

For Play it doesn't matter, if they are cross compiling for 2.12/2.13 then they can just add scala-java-compat themselves (ideally only for 2.12 but if they do it for all Scala versions it won't matter).

PR #316, opened 2023-05-09
Add OptionConverters.toScala methods for java Optional primitives

mdedetrich:
As pointed out in https://github.com/apache/incubator-pekko-http/pull/142#pullrequestreview-1417421049 we should do our best to avoid usage of the `Rich*` implicit classes within Java code especially since Java doesn't do inlining (unlike Scala).

I checked the pekko-http code and within Java sources we only use the `.toScala` methods (the Scala to Java conversion cases are done within Scala source which uses the standard extension methods).

### DIFF
--- a/actor/src/main/scala-2.12/org/apache/pekko/util/OptionConverters.scala
+++ b/actor/src/main/scala-2.12/org/apache/pekko/util/OptionConverters.scala
@@ -21,8 +21,20 @@ import java.util._
 @InternalStableApi
 private[pekko] object OptionConverters {
   import scala.compat.java8.OptionConverters.SpecializerOfOptions
+  import scala.compat.java8.OptionConverters._
 
   @inline final def toScala[A](o: Optional[A]): Option[A] = scala.compat.java8.OptionConverters.toScala(o)
+
+  // The rest of the .toScala methods that work with OptionalDouble/OptionalInt/OptionalLong have to be manually
+  // redefined because the scala.compat.java8.OptionConverters.toScala variants work with scala.lang primitive types
+  // where as scala.jdk.javaapi.OptionConverters.toScala works with java.lang primitive types. Since the primary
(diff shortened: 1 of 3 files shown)


actor/src/main/scala-2.12/org/apache/pekko/util/OptionConverters.scala
```diff
@@ -21,9 +21,21 @@ import java.util._
...
+  import scala.compat.java8.OptionConverters._
 
   @inline final def toScala[A](o: Optional[A]): Option[A] = scala.compat.java8.OptionConverters.toScala(o)
 
+  // The rest of the .toScala methods that work with OptionalDouble/OptionalInt/OptionalLong have to be manually
+  // redefined because the scala.compat.java8.OptionConverters.toScala variants work with scala.lang primitive types
+  // where as scala.jdk.javaapi.OptionConverters.toScala works with java.lang primitive types. Since the primary
+  // usecase of these functions is for calling within Java code its preferrable to return Java primitives, see
+  // https://github.com/scala/bug/issues/4214
+  def toScala(o: OptionalDouble): Option[java.lang.Double] = if (o.isPresent) Some(o.getAsDouble) else None
```
mdedetrich, line 33: I am intentionally not using `@inline` here because its the actual definition of the method being placed here, not another method call. JVM will inline these statements without issues.

mdedetrich:
@pjfanning @He-Pin @nvollmar Would you be able to review this? Its the last PR needed for https://github.com/apache/incubator-pekko-http/pull/142

actor/src/main/scala-2.12/org/apache/pekko/util/OptionConverters.scala
mdedetrich: There is an idea to enable the inliner for Scala 2.12/Scala 2.13 which will be relevant later on.

Issue #328, opened 2023-05-23
Source and Javadoc Snapshot Jars are missing LICENSE and NOTICE files


Claudenw:
LICENSE file should be included in all jars, releases and snapshots.  It is currently missing from the SNAPSHOT jars.  The change should place the LICENSE in **ALL** snapshot jars as well a the release jars when they are bult.

pjfanning:
@Claudenw can you tell us which jar. We have tooling to include LICENSE in jars but maybe this is missing from some places

pjfanning:
@Claudenw can you look at https://repository.apache.org/content/groups/snapshots/org/apache/pekko/pekko-actor_2.13/0.0.0+26660-7809c783-SNAPSHOT/ - last night's pekko-actor jars?

* Download pekko-actor_2.13-0.0.0+26660-7809c783-SNAPSHOT.jar 
* It has these files in the jar

```
      551  01-01-2000 00:00   META-INF/DISCLAIMER
    12128  01-01-2000 00:00   META-INF/LICENSE
     1732  01-01-2000 00:00   META-INF/NOTICE
```

pjfanning:
@mdedetrich we don't include LICENSE, NOTICE, DISCLAIMER in the `-sources` and `-javadoc` jars. Is that something that can be changed in sbt-apache-sonatype?

mdedetrich:
Can we clarify if we need to add them into the `-sources` and `-javadoc` jars as well? Both are highly suspicious because they are **only** meant to contain either Java/Scala sources or generated documentation.

pjfanning:
I checked some random Apache `-sources` and `-javadoc` jars and found that they had LICENSE and NOTICE files. I doubt whether it's universal but it does seem like something that many teams try to support.

* https://repo1.maven.org/maven2/org/apache/spark/spark-core_2.13/
* https://repo1.maven.org/maven2/org/apache/streampark/streampark-common_2.12/

pjfanning:
I'm not guaranteeing that sbt makes this easy - the sources and javadoc jars are basically generated using black box code and there doesn't appear to straightforward ways to manipulate them to add extra files.

mdedetrich:
> I checked some random Apache -sources and -javadoc jars and found that they had LICENSE and NOTICE files. I doubt whether it's universal but it does seem like something that many teams try to support.

I checked some other projects like Apache Daffodil and indeed it does seem to be the case. Do we also have to include `DISCLAIMER` as well?

[...]

PR #330, opened 2023-05-23
Update sbt-apache-sonatype

mdedetrich:
Updates sbt-apache-sonatype which brings in https://github.com/mdedetrich/sbt-apache-sonatype/pull/15

References: https://github.com/apache/incubator-pekko/issues/328

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -20,7 +20,7 @@ addSbtPlugin("net.bzzt" % "sbt-reproducible-builds" % "0.30")
 addSbtPlugin("com.dwijnand" % "sbt-dynver" % "4.1.1")
 addSbtPlugin("com.lightbend.sbt" % "sbt-publish-rsync" % "0.2")
 addSbtPlugin("com.github.pjfanning" % "sbt-source-dist" % "0.1.5")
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

mdedetrich:
> If there is no obvious way to fix the -javadoc jar issue, we can bring it to Incubator PMC or Legal and see what they say.

I deliberately left out `-javadoc` separately because some projects (i.e. Apache Daffodil) do not contain these files in generated documentation where as every single Apache project I could reference had it in `-sources`.

If its an issue its shouldn't be a problem for me to add it.

pjfanning:
Some projects do seem to add license/notice to all jars - even javadoc jars. 

I think the path of least resistance is to fix up the javadoc jars too.

Otherwise, we're probably going to have to get into a debate with Incubator PMC and/or Legal.

@Claudenw wdyt?

mdedetrich:
> Some projects do seem to add license/notice to all jars - even javadoc jars.

I know, but the `META-INF` folder doesn't even seem to be standard (in the general sense) for `-javadoc`, at least when packaging `-javadoc` locally.

mdedetrich:
Hmm, some tests fail but I will re-run PR just to be safe

PR #331, opened 2023-05-23
remove rsync plugin

pjfanning:
we use a github workflow to rsync


mdedetrich (approved):
Nice, I was planning to do this at some point and completely forgot about this.

Also needs to be done in the other projects

Issue #326, opened 2023-05-22
link validator check broken by license-report


pjfanning:
Not urgent. Just spotted the broken CI job.

https://github.com/apache/incubator-pekko/actions/runs/5042361658

fyi @mdedetrich

mdedetrich:
I just did a quick cursory look and aside from the link validator failing due to docs not being fully fixed (which is expected) it appears to also fail if use non https links, i.e.

```
 * Failure responses found.
 * Non-https URLs found (configure `non-https-accepted` if needed).
Error: Process completed with exit code 255.
```

mdedetrich:
Ah I see the issue now, sbt-license-report is putting links in the report that are derived from the pom fields in the transitive dependencies. Some of those fields are only http, so a low hanging fruit improvement can be enabling http.

pjfanning:
is it possible to get link-validator to skip this file altogether?

Then we can validate only https is used in our main pages.

mdedetrich:
After checking both the example [application.conf](https://github.com/ennru/site-link-validator/blob/master/core/src/main/resources/application.conf) and the [source-code](https://github.com/ennru/site-link-validator/blob/d8fa59afbe2228115868bb88d145e7578ae7e5fa/core/src/main/scala/net/runne/sitelinkvalidator/Main.scala#L68) I can't seem to find any such config.

Ill manually add the links for know, its a bit annoying but we only need to do it once and the dependencies don't get updated that often anyways.

PR #332, opened 2023-05-23
Add sbt-license-report generated links to http whitelist

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko/issues/326

The solution is pretty dumb, but there doesn't appear to be a config option that allows you to explicitly ignore a file.


scripts/link-validator.conf
```diff
@@ -46,6 +46,14 @@ site-link-validator {
     "https://javadoc.io/static/"
     # GitHub will block with "429 Too Many Requests"
     "https://github.com/"
+    # Github links generated by sbt-license-report
+    "http://github.com/FasterXML/jackson"
```
pjfanning, line 50: maybe just `http://github.com` to match the `429 Too Many Requests` section above?

mdedetrich: Fixed and pushed, lets see if this works.

PR #333, opened 2023-05-24
Document JAVA_8_HOME in error message

raboof:
When no Java 8 installation is found, mention in the error message that you can also use the JAVA_8_HOME environment variable to help the build find it.

### DIFF
--- a/project/JdkOptions.scala
+++ b/project/JdkOptions.scala
@@ -88,7 +88,7 @@ object JdkOptions extends AutoPlugin {
           jdk11options(java8home)
         case None =>
           throw new MessageOnlyException(
-            "A JDK 8 installation was not found, but is required to build Apache Pekko. To manually specify a JDK 8 installation, use the \"set every jdk8home := \\\"/path/to/jdk\\\" sbt command. If you have no JDK 8 installation, target your system JDK with the \"set every targetSystemJdk := true\" sbt command, but beware resulting artifacts will not work on JDK 8")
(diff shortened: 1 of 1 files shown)


project/JdkOptions.scala
```diff
@@ -88,7 +88,7 @@ object JdkOptions extends AutoPlugin {
           jdk11options(java8home)
         case None =>
           throw new MessageOnlyException(
-            "A JDK 8 installation was not found, but is required to build Apache Pekko. To manually specify a JDK 8 installation, use the \"set every jdk8home := \\\"/path/to/jdk\\\" sbt command. If you have no JDK 8 installation, target your system JDK with the \"set every targetSystemJdk := true\" sbt command, but beware resulting artifacts will not work on JDK 8")
+            "A JDK 8 installation was not found, but is required to build Apache Pekko. To manually specify a JDK 8 installation, set the JAVA_8_HOME environment variable to its path or use the \"set every jdk8home := \\\"/path/to/jdk\\\" sbt command. If you have no JDK 8 installation, target your system JDK with the \"set every targetSystemJdk := true\" sbt command, but beware resulting artifacts will not work on JDK 8")
```
He-Pin, line 91: Yes, this is needed to build the doc.

He-Pin: The syntax is not the same on `Windows`, and it will be `path\\to\\jdk`

pjfanning:
I think adding Windows style paths to the exception message may make the message more confusing. What we could aim to do is to work on a detailed installation guide and have the message link to that.

That change could be done in a follow up PR when the detailed installation guide is ready.

He-Pin (approved):
LGTM, and the path for Windows is easy to be inspected from the current information.

Issue #335, opened 2023-05-24
detailed build instructions


pjfanning:
As part of Pekko release, we will need the Apache Incubator PMC to vote on our source release. They will need to be able to take the source release and to build Pekko jars using it.

There is an inadequate initial stab at:

https://github.com/apache/incubator-pekko/blob/main/README.md#building-from-source

We need to also spell out that you probably need to have 2 Java Runtimes installed. See #333 for background (and https://lists.apache.org/thread/zz77bc6tjhdp68cojj6vksrmlqqvn7ln)

We probably also want to make the instructions usable by Linux and Windows users.

mdedetrich:
So we should also probably make a clear distinction between "building/compiling" and "making a release". The reason I am pointing this out is that you can compile/build the project fine on JDK 8 (I only have a single visible JDK installed, JDK 8 which is pointed to Pekko via [jenv](https://github.com/jenv/jenv)), it just won't have the optional JDK 11 features. This will also become important when inlining, i.e. https://github.com/apache/incubator-pekko/pull/305 because since inlining doesn't work with Scala's incremental compilation, you only want to enable it when doing a release build and not during development.

pjfanning:
what's at https://github.com/apache/incubator-pekko/blob/main/README.md#building-from-source will do

PR #336, opened 2023-05-24
Document JDK build prerequisites

raboof:
I think mentioning the need to install Java 8 should be sufficient here: in the vast majority of cases it should be picked up automatically, and otherwise the build error tells you how to explicitly point to it.

References (but perhaps not completely fixes) #335

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,9 @@ See https://pekko.apache.org for the documentation including the API docs. The d
 Building from Source
 --------------------
 
-- You will need to install [sbt](https://www.scala-sbt.org/) if you don't already have it installed
-- You should have a Java Runtime installed. Java 8 minimum.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,8 @@ Building from Source
 --------------------
 
 - You will need to install [sbt](https://www.scala-sbt.org/) if you don't already have it installed
-- You should have a Java Runtime installed. Java 8 minimum.
+- The Java Development Kit (JDK) used for running sbt should be Java 11 at minimum.
```
pjfanning, line 22: Thanks - this is a good start. One nit - could you put sbt bit after the Java bit? In hindsight, sbt needs Java so we list Java first.

pjfanning (approved):
lgtm

PR #327, opened 2023-05-22
Unwrap single string interpolation syntax

mdedetrich:
Unwraps single string interpolation i.e. `s"${something}"` into `s"$something"`. Make sure to do a rebase merge of this PR because I need to add its commit to `.git-blame-ignore-revs`

PR #341, opened 2023-05-24
Add unwrap single string into .git-ignore-blame-revs

mdedetrich:
Adds the changes from https://github.com/apache/incubator-pekko/pull/327 into `.git-ignore-blame-revs`.


pjfanning:
I don't think this should be done. I can agree to using this file for formatting only changes but I don't think code refactors like this should not be treated like this.

He-Pin:
I think the change in the previous pr is a refractory but not a format, maybe this is not needed?

mdedetrich:
> I don't think this should be done. I can agree to using this file for formatting only changes but I don't think code refactors like this should not be treated like this.

This is a formatting only change, it has zero impact on any business logic or code. I was quite meticulous in doing/checming this.

Unwrapping single string definitions isn't any different to lets say removing redundant parens which scalafmt does and hence the massive scafmt which we did at the start of the fork has these kinds of changes as well.

There are also other commits in `.git-ignore-blame-revs` with similar reasoning.

He-Pin (approved):
Lgtm

PR #343, opened 2023-05-25
add link for sdkman to build notes

pjfanning:


README.md
```diff
@@ -19,6 +19,7 @@ See https://pekko.apache.org for the documentation including the API docs. The d
 - Make sure you have installed a Java Development Kit (JDK) version 11 or later.
 - Make sure you have [sbt](https://www.scala-sbt.org/) installed and using this JDK.
 - Make sure you also have Java 8 installed (JDK or JRE).
+- One popular tool for installing and managing multiple versions of Java as well as sbt is [SDKMan](https://sdkman.io/).
```
raboof, line 22: People who already have a 'favorite' way of installing things (e.g. using their distribution package manager, jabba, ...) might be put off by 'yet another tool'. Starting with "One populat tool..." is a good start to make it clear that this is optional and you're welcome to use any method you prefer, but I wonder if we should make this even more explicit: maybe "If you're not sure how to install those, one popular tool for ...."?

pjfanning: fair point - I'll close this

PR #348, opened 2023-05-27
Remove redundant semicolons

mdedetrich:
Something that I quickly noticed, death to the semicolons


He-Pin:
How about using IDEA to do a whole cleanup ?

He-Pin (approved):
lgtm

mdedetrich:
> How about using IDEA to do a whole cleanup ?

Eventually I actually want to use scalafix to do this since it will always do it correctly (it uses official scalac/scalameta) but there are some blockers

PR #358, opened 2023-05-28
Add inline final to FutureConverters

mdedetrich:
While the inline isn't that important here, the most critical thing is that the addition of `final` in order to make sure that we don't get any problems down the road. Note that this change is making `FutureConverters` consistent with `OptionConverters`.


pjfanning (approved):
lgtm

PR #359, opened 2023-05-28
add license header to sbt files

pjfanning:
The CopyrightHeader.scala change to apply headers to `.sbt` files didn't actually have an effect. I then manually added the headers. I left the CopyrightHeader.scala change there just in case a future version of sbt-header checks the `.sbt` files.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,12 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


mdedetrich:
> The CopyrightHeader.scala change to apply headers to `.sbt` files didn't actually have an effect.

Yes I am aware of this issue, I think its due to a peculiarity where sbt plugins cannot see sources of the sbt build files themselves unless you add `sbt-license-header` as a plugin to your sbt plugins.

Anyways this should do for now

mdedetrich (approved):
lgtm

He-Pin (approved):
lgtm

He-Pin:
If it doesn't take effect for sbt files, then I think the resources are not included, but that can be done in the following PR.

PR #361, opened 2023-05-29
modify out of date note about java8 compat

pjfanning:
a lot of this is now inlined in Pekko instead to simplify the jar dependencies

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -533,8 +533,12 @@ Scala has proven the most viable way to do it, as long as you keep the following
 1. If the underlying Scala code requires an `ExecutionContext`, make the Java API take an `Executor` and use
    `ExecutionContext.fromExecutor(executor)` for conversion.
 
-1. Use `org.apache.pekko.util.FutureConverters` to translate Futures to `CompletionStage`s.
+1. Use `org.apache.pekko.util.FutureConverters` to translate `Future`s to `CompletionStage`s.
 
+1. Use `org.apache.pekko.util.OptionConverters` to translate `Option`s to Java `Optional`s.
+
+1. Use `org.apache.pekko.util.FunctionConverters` to translate Scala Functions to Java Functions.
+ 
 1. Make sure there are Java tests or sample code touching all parts of the API
 
 1. Do not use lower type bounds: `trait[T] { def method[U >: Something]: U }` as they do not work with Java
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -533,9 +533,7 @@ Scala has proven the most viable way to do it, as long as you keep the following
 1. If the underlying Scala code requires an `ExecutionContext`, make the Java API take an `Executor` and use
    `ExecutionContext.fromExecutor(executor)` for conversion.
 
-1. Make use of `scala-java8-compat` conversions, see [GitHub](https://github.com/scala/scala-java8-compat)
-   (eg. `scala.compat.java8.FutureConverters` to translate Futures to `CompletionStage`s).
-   Note that we cannot upgrade to a newer version scala-java8-compat because of binary compatibility issues.
+1. Use `org.apache.pekko.util.FutureConverters` to translate Futures to `CompletionStage`s.
```
He-Pin, line 536: image

It's private.

He-Pin: @mdedetrich Should we make it public?

mdedetrich: Specifically about the documentation, I think this is fine because its in `CONTRIBUTING.md`, i.e. its intended for people working on Pekko codebase not Pekko users.

mdedetrich, line 536: Can we add `ObjectConverters`/`FunctionConverters` in here as well?

pjfanning: This particular notes are for people contributing code to Pekko project. They are not about public APIs.

Users writing code that uses Pekko: if they are using Java APIs of Pekko, they should not be getting Scala Futures as return types - so they won't need to convert them.

mdedetrich (changes requested):
Should document all the added converter functions.

CONTRIBUTING.md
pjfanning: done

mdedetrich: Can we also add vice versa after these statements since the converters go both ways?

mdedetrich (approved):
One additional comment but others lgtm

CONTRIBUTING.md
pjfanning: added the 'vice versa' to each

He-Pin (approved):
lgtm

CONTRIBUTING.md
He-Pin: Ha, I forgot that, I was thinking a user guide:)

PR #363, opened 2023-05-29
=str Skip parsing when buffer size < 1

He-Pin:
I think we can do a quic returning if the buffer size is zero.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
@@ -115,20 +115,21 @@ import pekko.util.ByteString
   private def seekObject(): Boolean = {
     completedObject = false
     val bufSize = buffer.length
+    if (bufSize > 0) {
+      skipToNextObject(bufSize)
+      val maxObjectLengthIndex = if (pos + maximumObjectLength < 0) Int.MaxValue else pos + maximumObjectLength
 
-    skipToNextObject(bufSize)
-    val maxObjectLengthIndex = if (pos + maximumObjectLength < 0) Int.MaxValue else pos + maximumObjectLength
+      while (pos < bufSize && pos < maxObjectLengthIndex && !completedObject) {
+        val input = buffer(pos)
+        proceed(input)
+        pos += 1
+      }
 
-    while (pos < bufSize && pos < maxObjectLengthIndex && !completedObject) {
-      val input = buffer(pos)
-      proceed(input)
-      pos += 1
-    }
+      if (pos >= maxObjectLengthIndex)
+        throw new FramingException(s"""JSON element exceeded maximumObjectLength ($maximumObjectLength bytes)!""")
 
-    if (pos >= maxObjectLengthIndex)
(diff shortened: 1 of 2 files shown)


mdedetrich (changes requested):
This might be considered excessive but considering we have had a regression in this area of Pekko, would it be possible to add a unit test which checks that when the buffer size is 0 to check that both current Pekko and the change in this PR both return `false`?

The easiest way to do this is to make `seekObject()` package private (so you can access it in a test) and construct `JsonObjectParser` (which initialises `buffer` to empty so `buffer.length` is zero) and immediately call `seekObject()` to see if it returns `false`.

If this is too difficult then let me know and I can reconsider.

He-Pin:
It can be false if the downstream poll, but the upstream is not ready, or downstream is faster and upstream is idle.

I can add more detailed test for JsonFramming but I don't see much value to add a test for the short-circuit logic，as  it's identical to the objectCompleted.

As the chances this short-circuit can benefit, i think that maybe why it's not there in the firsr place.

mdedetrich:
tbh I am not entirely familiar with this section of Pekko, I am just being cautious/skeptical because we had a general regression here which I am trying to avoid in the future.

If you are happy with adding a more detailed `JsonFraming` test that would be great, I suspect you would do this in a separate PR? If so let me know and can review this in more detail in the next couple of days.

pjfanning:
Could we do this in v1.0.1? It's not a bug and it just doesn't feel urgent.

mdedetrich:
Agreed

He-Pin:
let's defer this.

mdedetrich:
> i want to add some random tests against JsonFraming.

Perfect thanks! Such tests should definitely be merged before 1.0.0 if they are ready in time to give us extra confidence.

mdedetrich:
@He-Pin I am adding this to 1.1.x milestone

stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
```diff
@@ -115,7 +115,9 @@ import pekko.util.ByteString
   private def seekObject(): Boolean = {
     completedObject = false
     val bufSize = buffer.length
-
+    if (bufSize == 0) {
```
pjfanning, line 118: I don't like 'return' statements. Can't you just do this ?

```
if (bufSize > 0) {
  // add everything from the rest if the function here
}
```

jrudolph: Agreed, maybe not super important in that case, but let's avoid a potential exception being thrown.

pjfanning (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
```diff
@@ -115,20 +115,21 @@ import pekko.util.ByteString
   private def seekObject(): Boolean = {
     completedObject = false
     val bufSize = buffer.length
+    if (bufSize > 1) {
```
He-Pin, line 118: a `{}` is minimal 2 bytes.

mdedetrich (approved):
lgtm

He-Pin:
The error is unrelated.

PR #364, opened 2023-05-31
Use proper type for SemVerSpec

mdedetrich:
@He-Pin Was correct, there is an actual proper type for `SemVerSpec` which is better than using a `String`.


pjfanning (approved):
lgtm

PR #365, opened 2023-06-01
Rename akka files and folders to pekko

mdedetrich:
Still found some leftover files/folders/packages which had akka in the name.


He-Pin (approved):
lgtm

Issue #357, opened 2023-05-27
rework link validation CI job


pjfanning:
* Fails very often when using curl to install coursier.
* possibly a rate limiter on git.io
* there is https://github.com/coursier/setup-action
* and this can work with https://github.com/coursier/cache-action
* so that we might be able to have the CI job avoid always downloading coursier on every run

PR #367, opened 2023-06-02
Use setup-coursier Instead downloading on every run in link-validator (#357)

ebongzzang:
- According to Issue #357, `Link validator` job often failed when downloading coursier from git.io possibly due to git.io rate limit
- To fix the issue, Instead download coursier on every run, apply [coursier/setup-action](https://github.com/coursier/setup-action) with [coursier/cache-action](https://github.com/coursier/cache-action)

### DIFF
--- a/.github/workflows/link-validator.yml
+++ b/.github/workflows/link-validator.yml
@@ -30,11 +30,11 @@ jobs:
       - name: Cache Coursier cache
         uses: coursier/cache-action@v6.4.0
 
+      - name: Setup Coursier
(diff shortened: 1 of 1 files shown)


mdedetrich (approved):
Thanks! lgtm

mdedetrich:
Turns out that this broke the link validator due to `coursier/setup-coursier` not being whitelisted by INFRA, ticket made https://issues.apache.org/jira/browse/INFRA-24664

PR #370, opened 2023-06-05
Cache snapshots to prevent overloading Apache Nexus repo

mdedetrich:
The Apache Snapshots nexus repo is sometimes giving us request refused, likely because we are overloading the repo with excessive amount of requests.  This is due to the fact that we are using `SNAPSHOT` versions which by default don't cache because they are mutable (i.e. you can reupload snapshots with different implementations as much as you like).

Since for both pekko and its project forks the snapshot version is derived from git hash, there is no disadvantage to caching the snapshot.


project/Paradox.scala
```diff
@@ -72,7 +72,6 @@ object Paradox {
 
   val themeSettings = Seq(
     // allow access to snapshots for pekko-sbt-paradox
-    resolvers += Resolver.ApacheMavenSnapshotsRepo,
```
mdedetrich, line 75: Its cleaner to put this into root `build.sbt` because that way we can easily remove it once Pekko is released.

mdedetrich: Let me redo this change, pekko core doesn't really need it since its at the top of the dependency change

mdedetrich: So this dependency shouldn't even be here, pekko-sbt-paradox is an sbt plugin which means it gets resolved in `project/plugins.sbt`.

raboof: (perhaps it is used to fetch themes?)

mdedetrich: I just checked and that doesn't appear to be the case https://github.com/apache/incubator-pekko-sbt-paradox/blob/main/build.sbt#L52

mdedetrich: Okay I was wrong, I missed `https://repo1.maven.org/maven2/org/apache/pekko/pekko-theme-paradox/0.0.0+38-68da3106-SNAPSHOT/pekko-theme-paradox-0.0.0+38-68da3106-SNAPSHOT.pom`. Reverting the change

mdedetrich: > There is a `resolverSettings ` in `PekkoBuild.scala`, would that be be better place?

This is temporary so actually I would prefer to not have it in `resolverSettings` which is also designed to be global (in our case we only want it for Paradox).

He-Pin (approved):
lgtm

PR #371, opened 2023-06-07
Generate random DNS id's

mdedetrich:
DNS id generated should be random, see https://datatracker.ietf.org/doc/html/rfc5452#section-4.3. One additional thing that needs to be confirmed is whether we need to handle collisions

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -1145,11 +1145,13 @@ pekko {
         # other platforms, will default to 1).
         ndots = default
 
-        # The policy used to generate dns transaction ids. Options are thread-local-random or secure-random.
-        # Defaults to thread-local-random similar to Netty, secure-random produces FIPS compliant random numbers but
-        # could block looking for entropy (these are short integers so are easy to bruit-force, use thread-local-random
-        # unless you really require FIPS compliant random numbers).
-        id-generator-policy = thread-local-random
+        # The policy used to generate dns transaction ids. Options are `thread-local-random`,
+        # `enhanced-double-hash-random` or `secure-random`. Defaults to `enhanced-double-hash-random` which uses an
+        # enhanced double hashing algorithm optimized for minimizing collisions with a FIPS compliant initial seed.
+        # `thread-local-random` is similar to Netty and `secure-random` produces FIPS compliant random numbers every
+        # time but could block looking for entropy (these are short integers so are easy to brute-force, use
+        # `enhanced-double-hash-random` unless you really require FIPS compliant random numbers).
+        id-generator-policy = enhanced-double-hash-random
       }
     }
   }
--- a/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
@@ -12,6 +12,7 @@ package org.apache.pekko.io.dns
 import org.apache.pekko.annotation.InternalApi
 
 import java.security.SecureRandom
+import java.util.Random
 import java.util.concurrent.ThreadLocalRandom
 
 /**
@@ -39,18 +40,13 @@ private[pekko] object IdGenerator {
   object Policy {
     case object ThreadLocalRandom extends Policy
     case object SecureRandom extends Policy
-    val Default: Policy = ThreadLocalRandom
-
-    def apply(name: String): Option[Policy] = name.toLowerCase match {
-      case "thread-local-random" => Some(ThreadLocalRandom)
-      case "secure-random"       => Some(SecureRandom)
-      case _                     => Some(ThreadLocalRandom)
-    }
+    case object EnhancedDoubleHashRandom extends Policy
   }
 
   def apply(policy: Policy): IdGenerator = policy match {
-    case Policy.ThreadLocalRandom => random(ThreadLocalRandom.current())
-    case Policy.SecureRandom      => random(new SecureRandom())
+    case Policy.ThreadLocalRandom        => random(ThreadLocalRandom.current())
+    case Policy.SecureRandom             => random(new SecureRandom())
+    case Policy.EnhancedDoubleHashRandom => new EnhancedDoubleHashGenerator(new SecureRandom())
   }
 
   def apply(): IdGenerator = random(ThreadLocalRandom.current())
@@ -60,4 +56,31 @@ private[pekko] object IdGenerator {
    */
   def random(rand: java.util.Random): IdGenerator =
     () => (rand.nextInt(UnsignedShortBound) + Short.MinValue).toShort
+
+  private[pekko] class EnhancedDoubleHashGenerator(seed: Random) extends IdGenerator {
+
+    /**
+     * An efficient thread safe generator of pseudo random shorts based on
+     * https://en.wikipedia.org/wiki/Double_hashing#Enhanced_double_hashing.
+     *
+     * Note that due to the usage of synchronized this method is optimized
+     * for the happy case (i.e. least contention) on multiple threads.
+     */
+    private var index = seed.nextLong
+    private var increment = seed.nextLong
+    private var count = 1L
+
+    override final def nextId(): Short = synchronized {
+      val result = (0xFFFFFFFF & index).asInstanceOf[Short]
+      index -= increment
+
+      // Incorporate the counter into the increment to create a
+      // tetrahedral number additional term.
+      increment -= {
+        count += 1
+        count - 1
+      }
+      result
+    }
+  }
 }
--- a/actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala
@@ -28,6 +28,7 @@ import pekko.actor.ExtendedActorSystem
 import pekko.annotation.InternalApi
 import pekko.event.Logging
 import pekko.io.dns.CachePolicy.{ CachePolicy, Forever, Never, Ttl }
+import pekko.io.dns.IdGenerator.Policy
 import pekko.io.dns.internal.{ ResolvConf, ResolvConfParser }
 import pekko.util.Helpers
 import pekko.util.Helpers.Requiring
@@ -67,10 +68,17 @@ private[dns] final class DnsSettings(system: ExtendedActorSystem, c: Config) {
   val PositiveCachePolicy: CachePolicy = getTtl("positive-ttl")
   val NegativeCachePolicy: CachePolicy = getTtl("negative-ttl")
 
-  lazy val IdGeneratorPolicy: IdGenerator.Policy = IdGenerator
-    .Policy(c.getString("id-generator-policy"))
-    .getOrElse(throw new IllegalArgumentException("id-generator-policy must be 'thread-local-random' or " +
-      s"'secure-random' value was '${c.getString("id-generator-policy")}'"))
+  lazy val IdGeneratorPolicy: IdGenerator.Policy = {
+    c.getString("id-generator-policy") match {
+      case "thread-local-random"                               => Policy.ThreadLocalRandom
+      case "secure-random"                                     => Policy.SecureRandom
+      case s if s.isEmpty | s == "enhanced-double-hash-random" => Policy.EnhancedDoubleHashRandom
+      case _ =>
+        throw new IllegalArgumentException(
+          "Invalid value for id-generator-policy, id-generator-policy must be 'thread-local-random', 'secure-random' or" +
+          s"`enhanced-double-hash-random`")
+    }
+  }
 
   private def getTtl(path: String): CachePolicy =
     c.getString(path) match {
--- a/bench-jmh/src/main/scala/org/apache/pekko/io/dns/IdGeneratorBanchmark.scala
+++ b/bench-jmh/src/main/scala/org/apache/pekko/io/dns/IdGeneratorBanchmark.scala
@@ -29,16 +29,34 @@ import java.security.SecureRandom
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
 @Warmup(iterations = 3, time = 5, timeUnit = TimeUnit.SECONDS)
 @Measurement(iterations = 3, time = 5)
-@Threads(8)
 @Fork(1)
 @State(Scope.Benchmark)
 class IdGeneratorBanchmark {
   val threadLocalRandom = IdGenerator.random(ThreadLocalRandom.current())
   val secureRandom = IdGenerator.random(new SecureRandom())
+  val enhancedDoubleHash = new IdGenerator.EnhancedDoubleHashGenerator(new SecureRandom())
 
+  @Threads(1)
   @Benchmark
   def measureThreadLocalRandom(): Short = threadLocalRandom.nextId()
 
+  @Threads(1)
   @Benchmark
   def measureSecureRandom(): Short = secureRandom.nextId()
+
+  @Threads(1)
+  @Benchmark
+  def measureEnhancedDoubleHash(): Short = enhancedDoubleHash.nextId()
+
+  @Threads(2)
+  @Benchmark
+  def multipleThreadsMeasureThreadLocalRandom(): Short = threadLocalRandom.nextId()
+
+  @Threads(2)
+  @Benchmark
+  def multipleThreadsMeasureSecureRandom(): Short = secureRandom.nextId()
+
+  @Threads(2)
+  @Benchmark
+  def multipleThreadsMeasureEnhancedDoubleHash(): Short = enhancedDoubleHash.nextId()
 }


He-Pin:
1. collisions should be handled too.
2. the call to `nextId` is not thread safe, does this pr handle this?

mdedetrich:
@Claudenw Pinging

pjfanning:
The old code wasn't thread safe either - why do we need to worry about this code being thread safe?

To make the old code thread safe, we'd need to use an AtomicInteger or something like that.

pjfanning:
I'm not an expert but any idea why the value is a `short`. If we used a type like `int` or `long`, the risk of collisions reduces.

mdedetrich:
> I'm not an expert but any idea why the value is a `short`. If we used a type like `int` or `long`, the risk of collisions reduces.

It has to be according to the standard, see https://datatracker.ietf.org/doc/html/rfc5452#section-4.3. The generated ID's fill a octet (i.e. 2 byte) range which in Java/Scala is a `Short`.

Actually this new implementation is strictly better than the current one because the current one only used 1 byte range since it just incremented from `0` and `Short.MaxValue + 1` doesn't overflow into `Short.MinValue`.

He-Pin:
The collision can happen in DnsClient.scala, if there are two flight queries with same id.

mdedetrich:
> The collision can happen in DnsClient.scala, if there are two flight queries with same id.
> 
> And i was.submited a pr in akka for this, but can be done with a fsm like thing too.

If this is regarding it not being thread safe we can solve this separately. Let's fix one problem at a time, the RC is on our doorstep.

On another note CI is failing because the tests need to be updated. I'll do this in a bit.

pjfanning:
Do we know what changing `pekko.io.dns.resolver` to use `inet-address` instead of `async-dns`?

mdedetrich:
> Do we know what happens if we change `pekko.io.dns.resolver` to use `inet-address` instead of `async-dns`?

On the top of my head this uses a completely separate implementation from Java stdlib so it shouldn't be an issue.

He-Pin:
> Do we know what happens if we change `pekko.io.dns.resolver` to use `inet-address` instead of `async-dns`?

Java lib based one may be slower and blocking？

mdedetrich:
> > Do we know what happens if we change `pekko.io.dns.resolver` to use `inet-address` instead of `async-dns`?
> 
> Java lib based one may be slower and blocking？

From what @raboof said, the reason why the `async-dns` was implemented in the first place is that there are caching issues with the java implementation.

pjfanning:
I raised https://github.com/apache/incubator-pekko/issues/373 for the lower priority issue of tracking ids

mdedetrich:
So I have just updated the tests and rebased/force pushed the changes. The reason why the tests are failing is that they assumed the `nextId` was incrementally increasing and hence checking against raw id's. I have updated the tests so that we don't test against the id's but the properties for the id, i.e. typically speaking the test expects an initial id from the `dnsClient1` i.e. 

```scala
val firstId = dnsClient1.expectMsgPF() {
  case q4: Question4 if q4.name == "cats.com" =>
    q4.id
}
```

and then that `secondId` is being reused in later on, such as

```scala
val secondId = dnsClient2.expectMsgPF() {
  case q4: Question4 if q4.name == "cats.com" && q4.id != firstId =>
    q4.id
}
```

i.e. in this specific case we can't test for hardcoded id's but we can still make sure that the `secondId` is not the same as the `firstId`.

mdedetrich:
So I just checked systemd-resolved and it does appear to handle hash collisions, it does this by tracking the state of current transactions (see https://github.com/systemd/systemd/blob/main/src/resolve/resolved-dns-transaction.c#L195-L196). Will work on this now.

pjfanning (dismissed):
lgtm - we can continue to look at fully collision-proof implementations but this PR is probably enough to unblock an RC

mdedetrich:
Lets wait a bit before merging this, I think I was wrong in what I said before about collisions and we do need to handle this.

pjfanning:
It looks like a singleton actor instance to me - so `receive` should only be processing one message at a time.

He-Pin:
> > `nextId` can be called concurrently, so I think two thread can get the same result. `idIncrement` and `idIndex` should be protected with `AtomicLong`
> 
> So having gone through the codebase, `nextId()` is only called in a `receive` block (indirectly) which means that its only going to be called in a single threaded context (this is something that actor guarantees).
> 
> Did I miss something?

IIRC, it can be called from the `recover` of the future。

pjfanning:
@mdedetrich I don't really mind how this is implemented. If exposing the map state outside of the actor makes this easier to implement, then I suggest that we do that. 

We can always change that afterwards.

He-Pin:
@mdedetrich What do you think about my PR？I think that's correct too.

mdedetrich:
> @mdedetrich What do you think about my PR？I think it's correct too.

Which one?

mdedetrich:
> @mdedetrich I don't really mind how this is implemented. If exposing the map state outside of the actor makes this easier to implement, then I suggest that we do that. 
> 
> We can always change that afterwards.

I don't think that would fly in general, I mean it breaks one of the core principles of actors and it would open a whole can of worms

He-Pin:
> > @mdedetrich I don't really mind how this is implemented. If exposing the map state outside of the actor makes this easier to implement, then I suggest that we do that. 
> > 
> > We can always change that afterwards.
> 
> I don't think that would fly in general, I mean it breaks one of the core principles of actors and it would open a whole can of worms

https://github.com/akka/akka/pull/31912

pjfanning:
@He-Pin https://github.com/akka/akka/pull/31912 looks like a viable approach. Is there any way that a unit test could be developed?

He-Pin:
@pjfanning I think that can be tested, but I will not have much time before weekend.

mdedetrich:
> @He-Pin [akka/akka#31912](https://github.com/akka/akka/pull/31912) looks like a viable approach. Is there any way that a unit test could be developed?

So I had a look at this in more detail and I don't think this would solve the specific problem we are dealing with here. Unless I misunderstood, this is solving the issue where you can have multiple in flight requests with the same id where as I am dealing with the issue where we don't want to generate multiple ids by checking in flight transactions.

While the difference is subtle its somewhat critical here, the solution in https://github.com/akka/akka/pull/31912 involves putting guards around `inflightRequests` within `DnsClient` itself where as the problem we have is that we have an external `AsyncDnsResolver` actor that needs to access the state of `DnsClient.

If `inflightTransactions` was hypothetically inside of `AsyncDnsResolver` we wouldn't even be having this issue  and the solution would be done by now and this is what I alluded to when I made this earlier statement

> Initially I thought it would have been possible to just have a currentIds collection inside of AsyncDnsResolver, adding id's is easy since nextId() is also defined inside of AsyncDnsResolver, the problematic part is removing the id's. If you check occurrences of inflightRequests -= <ID> you can see there are 3 cases where id's are removed (i.e. dropped requests, failed UDP packets etc etc) and only in one of those cases (specifically DropRequest) does AsyncDnsResolver have context of this.

A hypothetical solution to this issue would be to reverse the problem, i.e. for the other cases where `inflightRequests -= <ID>` occurs in Actors outside of `AsyncDnsResolver`, those actors could send a message to `AsyncDnsResolver` stating that an id has been dropped but the problem here is this would create a cyclic dependency because the relationship is currently the opposite way around, i.e. `AsyncDnsResolver` keeps track of the other actors (such as `DnsClient`) and not the other way around (`DnsClient` does not have any actor references to `AsyncDnsResolver`)

He-Pin:
@mdedetrich Yes, that why I had https://github.com/apache/incubator-pekko/pull/288 and https://github.com/akka/akka/pull/31912 for this. This  PR can move faster with a smaller scope I think.

$work time, sorry not makes me clear.

mdedetrich:
> @mdedetrich Yes, that why I had #288 and [akka/akka#31912](https://github.com/akka/akka/pull/31912) for this. This PR can move faster with a smaller scope I think.
> 
> $work time, sorry not makes me clear.

I still think the issue is the same, i.e. `requestIdGen` in https://github.com/apache/incubator-pekko/pull/288/files#diff-4891e059729a403d92d62966e039a3f9b6aa24fbc7562e1ed9daa280c46328d5R97 would need to access `inFlightTransactions` in `DnsClient` actor, but sure lets wait till the weekend when you have time to look at this properly.

pjfanning:
This is just my view but if we can get a solution working that cycles through a large set of possible ids without a collision (like the latest one based on UniqueRandomShortProvider) that that should be enough for Pekko 1.0.0. A solution where we keep track of all in-flight ids might be nice but it would appear to be more expensive than the UniqueRandomShortProvider solution.

In the end of the day, Akka went years without needing a fully collision proof implementation.

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -85,12 +85,15 @@ private[io] final class AsyncDnsResolver(
...
+  // it will be fast
+  shortSupplier.nextId()
 
-  private def nextId(): Short = {
-    requestId = (requestId + 1).toShort
-    requestId
-  }
+  // DNS id's need to be random, see https://datatracker.ietf.org/doc/html/rfc5452#section-4.3.
+  private def nextId(): Short =
+    shortSupplier.nextId()
```
He-Pin, line 96: Do not need to start a new line.

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -257,4 +260,5 @@ private[pekko] object AsyncDnsResolver {
     Future.successful(Answer(-1, immutable.Seq.empty[ResourceRecord], immutable.Seq.empty[ResourceRecord]))
 
   case class ResolveFailedException(msg: String) extends Exception(msg)
+
```
He-Pin, line 263: Remove this new line?

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  count = 1
+
+  private def resetSeed(): Unit = {
+    index = r.nextLong
+    increment = r.nextLong
+  }
+
+  private def resetBuffer(): Unit = {
+    System.arraycopy(Filler, 0, buffer, 0, UniqueRandomShortProvider.MaxLimit)
```
He-Pin, line 50: This seems not threadsafe

pjfanning: I think we'll need to synchronize this block.

```
      resetBuffer()
      resetSeed()
```

pjfanning: it's the recoverWith problem - the nextId() can be called by multiple threads - it seems like we need a synchronize somewhere

mdedetrich: The thread safety is a known problem and is being addressed separately, we discussed this earlier. I would recommend making the algorithm thread safe later, lets get an initial solution for now.

Using `synchronized` is usually just a heavy handed cop-out due to how slow it is, there are smarter ways to make it thread safe.

pjfanning: @kerr we're working on an improvement to this solution and the thread safety issue will probably be easier to fix with this improvement. Let's get the improvement done and we can review if we can also fix that thread safety issue.

The adjustment is to @Claudenw's ShortSupplier and involves a change to stop us from potentially returning recently used IDs when we reach the resetBuffer stage.

mdedetrich:
So just to clarify this a bit more, the current solution in this PR is not ideal and both @pjfanning, @Claudenw and myself have an improvement however I am holding off adding this due to what @IainHull posted here https://github.com/apache/incubator-pekko/pull/371#discussion_r1224490114 since it may be a waste of time.

Hopefully the legal hurdles will be overcome and we can discuss and review that solution comprehensively.

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
```diff
@@ -0,0 +1,106 @@
...
+    limit = UniqueRandomShortProvider.MaxLimit
+  }
+  def nextId(): Short = {
```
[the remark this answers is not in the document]
alexandru: Formatting is important, but PR comments on formatting (in large projects) don't scale.

Scalafmt is enabled on the project. I think it's Scalafmt's job to reformat and complain. Its configuration is never perfect, but it avoids the bike shedding in PRs. Does Scalafmt not add/require these new lines? If not, then maybe a configuration change would be needed.

I see there's also a `verifyCodeStyle` command in `build.sbt`, it would be a good idea to include this in CI to have the build fail if the code style fails.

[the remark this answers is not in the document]
pjfanning: There is CI checking - forcing a newline before a new function is just not one of the rules that our scalafmt enforces. 

I agree about keeping scalafmt conversations out of PRs when there is automated checking of the agreed formatting already. Separate conversations can be had about changing the format rules.

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
mdedetrich: > The original code doesn't seem to worry about concurrency, and given that this PR just wants to introduce randomness, it's probably better to just drop those annotations — as they also introduced confusion, e.g., is this code supposed to be thread-safe or not; future readers will also wonder.

Definitely agreed, @Claudenw (who wrote the algoritm) did state that its not thread safe so we should just drop that `volatile`'s that he added there otherwise its giving false impressions.

However lets wait for https://github.com/apache/incubator-pekko/pull/371#issuecomment-1585680543

@alexandru Regarding scalafmt, its run on every CI run via the usage of scalafmt-native, see https://github.com/apache/incubator-pekko/actions/runs/5220862460/jobs/9424448652?pr=371 and https://github.com/apache/incubator-pekko/blob/main/.github/workflows/format.yml. Its actually not possible to merge a PR if it hasn't been formatted with scalafmt.

pjfanning:
I've created a label to track all the issues and PRs related to async DNS resolution. I think we need to prioritise work on solving the Akka-related CVE and not necessarily fix everything that is wrong.

I've created 3 issues to track 3 problems and created a label to link them and the PRs.

https://github.com/apache/incubator-pekko/labels/async-dns-resolver

This PR relates to #384

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
pjfanning: @IainHull I think that the policies that you support in #385 are sufficient. You have a mechanism for tracking duplicate IDs in your PR. It's possible that there is stuff in this PR that may be a useful addition to the #385 changes but so far, I think they are rival solutions.

mdedetrich: @IainHull  Does it make sense to just make a new PR? Otherwise it seems to be a bit chaotic

pjfanning: @mdedetrich I think the idea is that #385 could be merged instead of this. Is there anything that #385 is missing that we would need for a Pekko 1.0.0 release?

He-Pin:
@mdedetrich In recent Netty PR: 

refs: https://github.com/netty/netty/pull/13439/

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
Claudenw: The volatile keyword was to keep the vars in cache as much as possible, they can be removed.  However `UniqueRandomShortProvider.nextId()` needs to be atomic for thread safety.

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+ * through all possible 65536 values before starting to re-generating
+ * the same numbers again.
+ */
+
+@InternalApi
+private[pekko] class UniqueRandomShortProvider {
+  import UniqueRandomShortProvider._
+
+  // These methods are volatile so that they stay within cache
+  @volatile private var index = 0L
```
pjfanning, line 33: I think @Claudenw agreed that the volatiles should go

mdedetrich:
So I just rebased the PR with @Claudenw 's latest algorithm and one thing we can say, according to benchmarks its pretty damn fast

```
[info] Benchmark                                        Mode  Cnt   Score    Error   Units
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash  thrpt    3  ≈ 10⁻⁴           ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom        thrpt    3   0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom   thrpt    3   2,286 ±  0,453  ops/ns
```

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
mdedetrich: So I just spoke with him and he said that he did put volatile there deliberately, its a trick to keep the cache warm (see https://github.com/apache/incubator-pekko/pull/371#discussion_r1228105897).

I will do another set of benchmarks without volatile to see if it makes a difference

pjfanning:
> So I just rebased the PR with @Claudenw 's latest algorithm and one thing we can say, according to benchmarks its pretty damn fast
> 
> ```
> [info] Benchmark                                        Mode  Cnt   Score    Error   Units
> [info] IdGeneratorBanchmark.measureEnhancedDoubleHash  thrpt    3  ≈ 10⁻⁴           ops/ns
> [info] IdGeneratorBanchmark.measureSecureRandom        thrpt    3   0,001 ±  0,001  ops/ns
> [info] IdGeneratorBanchmark.measureThreadLocalRandom   thrpt    3   2,286 ±  0,453  ops/ns
> ```

Isn't 10e-4 a small number meaning poor performance?

mdedetrich:
> Isn't 10^-4 a small number meaning poor performance?

Ah its in throughput mode, in this case something seems to be up.

actor/src/main/scala/org/apache/pekko/util/UniqueRandomShortProvider.scala
mdedetrich: I removed them as they don't seem to be making a difference

pjfanning:
If we can get the performance into the same ballpark as 'secure-random' then this generator is a useful addition. I would default to 'secure-random' though (unless the EnhancedDoubleHash was made to be faster than secure-random).

One useful aspect of EnhancedDoubleHash is that it would depend much less on the entropy settings of the device that the code was running on. SecureRandom can be very fast on some hardware but can be very slow in low entropy environments. Low entropy environments can be improved using tools like haveged.

mdedetrich:
> If we can get the performance into the same ballpark as 'secure-random' then this generator is a useful addition. I would default to 'secure-random' though (unless the EnhancedDoubleHash was made to be faster than secure-random).

So I don't know whats going on here, but I am having a suspicion that the reason for its slowness is all of the collision detection that was added ontop. When I was discussing with @Claudenw I was contemplating reverting the algorithm to the original simple version that had no buffers because even if there are collisions we can now handle them

mdedetrich:
So here is a benchmark of the version that has no buffer to keep track of collisions or rollover detection

```
[info] Benchmark                                        Mode  Cnt  Score    Error   Units
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash  thrpt    3  1,105 ±  1,203  ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom        thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom   thrpt    3  2,345 ±  0,602  ops/ns
```

So its significantly faster than `SecureRandom` and only approx twice as slow thread local random

actor/src/main/resources/reference.conf
```diff
@@ -1145,11 +1145,13 @@ pekko {
         # other platforms, will default to 1).
         ndots = default
 
-        # The policy used to generate dns transaction ids. Options are thread-local-random or secure-random.
-        # Defaults to thread-local-random similar to Netty, secure-random produces FIPS compliant random numbers but
-        # could block looking for entropy (these are short integers so are easy to bruit-force, use thread-local-random
-        # unless you really require FIPS compliant random numbers).
-        id-generator-policy = thread-local-random
+        # The policy used to generate dns transaction ids. Options are thread-local-random, enhanced-double-hash-random
```
pjfanning, line 1148: Could you put the values in quotes? This makes them stand out a bit more. 'bruit' should 'brute' as in brute force.

Could you change 'initial seed, thread-local-random' to this (with a full stop instead of a comma)?
```
initial seed. 'thread-local-random'
```

mdedetrich: Done

actor/src/main/resources/reference.conf
```diff
@@ -1145,11 +1145,13 @@ pekko {
...
-        # The policy used to generate dns transaction ids. Options are thread-local-random or secure-random.
-        # Defaults to thread-local-random similar to Netty, secure-random produces FIPS compliant random numbers but
-        # could block looking for entropy (these are short integers so are easy to bruit-force, use thread-local-random
-        # unless you really require FIPS compliant random numbers).
-        id-generator-policy = thread-local-random
+        # The policy used to generate dns transaction ids. Options are `thread-local-random`,
+        # `enhanced-double-hash-random` or `secure-random`. Defaults to `enhanced-double-hash-random` which uses an
+        # enhanced double hashing algorithm optimized for minimizing collisions with a FIPS compliant initial seed.
+        # `thread-local-random` is similar to Netty and secure-random produces FIPS compliant random numbers every time
+        # but could block looking for entropy (these are short integers so are easy to brute-force, use
```
pjfanning, line 1152: should be `secure-random` to match the others

mdedetrich: Fixed

mdedetrich:
So here are the results when using a simple faster version with the  `synchronized` block

```
[info] Benchmark                                        Mode  Cnt  Score    Error   Units
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash  thrpt    3  0,009 ±  0,012  ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom        thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom   thrpt    3  2,312 ±  0,531  ops/ns
```

While its still faster than secure random its now much slower than thread local random, i.e. performance impact was significant. Rewriting the algorithm with AtomicLong/volatile can likely mitigate this greatly

mdedetrich:
So as @pjfanning pointed out, the current benchmarks are somewhat misleading in that they are run over 8 threads which is the complete worst case scenario for `synchronized` especially when taking into account that in almost every case the calls will from the same thread, only when `recoverWith` in the `Future` block gets hit can we get multiple calls.

If we modify benchmark to just use single thread then here is the non `synchronized` version

```
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash  thrpt    3  0,441 ±  0,031  ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom        thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom   thrpt    3  0,313 ±  0,034  ops/ns
```

and synchronized version

```
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash  thrpt    3  0,380 ±  0,008  ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom        thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom   thrpt    3  0,312 ±  0,053  ops/ns
```

pjfanning:
@alexandru I suggested to @mdedetrich to use a Semaphore instead. It seems a little faster in the microbenchmark but you can set a timeout on the acquire as well.

mdedetrich:
@alexandru Thanks for the reply

So to provide some context on how, why and where the concurrency in this specific circumstance matters the code we are dealing with is entirely inside an `Actor` `receive` block which is actually normally meant to be guaranteed to be thread safe (i.e. single thread), the reason why we need to make `nextId()` thread safe is because there is a `recoverWith` on a `Future` which ends up transitively calling `nextId()` within an `ExecutionContext`. This `ExecutionContext` is configurable and hence parameterised, its using the Akka default dispatcher which iirc is a `ForkJoinPool`.

I have strong suspicion that we should be optimising for our happy case, because as you can infer from the usage of `recoverWith` this is probably not going to happen that often so while the usage of `synchronized` is typically ill advised, as you pointed out it is heavily optimized/uses assumptions that mirror the case we are dealing with.

The thing is (and this is a general statement and not strictly about this concurrency issue at hand), its hard for us to get some real data on this and there are many variables at play. For example there are actually 3 versions of @Claudenw algorithm, each with different guarantees of how we want to handle collisions. The "simple" version doesn't handle collisions at all which with initial impressions appears might be the best solution because its going to get handled by the duplicate id changes merged at https://github.com/apache/incubator-pekko/pull/385. However if we get lots of collisions/transactions then one of his variants which mitigate collisions might end up overall being faster.

And while we have a concurrency issue at hand, at the same time we are dealing with random numbers here which means we don't care about "wrong" answers, only if we get duplicates that we can't detect which might open up for some interesting optimisations.

Nonetheless we are trying to work out these concurrency issues and very much appreciate your experience here (if you have some extra time to spare let me know and we can co-ordinate).

pjfanning:
I tried a solution with an AtomicReference around the state but its performance was slower than the synchronization and Semaphore locking variants. Approximately in SecureRandom territory in the 8 threads microbenchmark.

```
  private val state = {
    val (index, increment) = {
      val r = new SecureRandom()
      (r.nextLong, r.nextLong())
    }
    new AtomicReference(Array(index, increment, 1))
  }

  @tailrec
  final def nextId(): Short = {
    val array = state.get()
    var index = array(0)
    var increment = array(1)
    var count = array(2)
    val result = (0xFFFFFFFF & index).asInstanceOf[Short]
    index -= increment

    // Incorporate the counter into the increment to create a
    // tetrahedral number additional term.
    increment -= {
      count += 1
      count - 1
    }
    if (state.compareAndSet(array, Array(index, increment, count))) {
      result
    } else {
      nextId()
    }
  }
```

mdedetrich:
> I tried a solution with an AtomicReference around the state but its performance was slower than the synchronization and Semaphore locking variants. Approximately in SecureRandom territory in the 8 threads microbenchmark.

This isn't too surprising, I suspect that if you wanted to get any tangible benefits of `AtomicLong`/`volatile` you would need to use it on the individual variables (i.e. index/counter etc etc) where its needed which is hard to do correctly

pjfanning:
> > I tried a solution with an AtomicReference around the state but its performance was slower than the synchronization and Semaphore locking variants. Approximately in SecureRandom territory in the 8 threads microbenchmark.
> 
> This isn't too surprising, I suspect that if you wanted to get any tangible benefits of `AtomicLong`/`volatile` you would need to use it on the individual variables (i.e. index/counter etc etc) where its needed which is hard to do correctly

Unfortunately, the 3 numbers in the state are all important. And they should be updated together, so I find it hard to see any solution based on AtomicLongs or AtomicReferences.

Using the Semaphore locking has the best result in my testing. Could we just go with that for RC1? We are very likely to need an RC2 so there is time for someone to find a better solution.

mdedetrich:
> Unfortunately, the 3 numbers in the state are all important. And they should be updated together, so I find it hard to see any solution based on AtomicLongs or AtomicReferences.

Ordinarily yes, but we aren't looking for complete correctness here but I can't think of any way to abuse this atm

> Using the Semaphore locking has the best result in my testing. Could we just go with that for RC1? We are very likely to need an RC2 so there is time for someone to find a better solution.

Fine with me unless there are some other real objections.

mdedetrich:
So I have just pushed the branch using the simple version of enhanced double hash with the `synchronized` keyword. Here are the benchmarks

```
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash                 thrpt    3  0,377 ±  0,032  ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom                       thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom                  thrpt    3  0,308 ±  0,057  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureEnhancedDoubleHash  thrpt    3  0,025 ±  0,013  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureSecureRandom        thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureThreadLocalRandom   thrpt    3  0,609 ±  0,014  ops/ns
```

Note that I have modified the benchmarks to be more indicative of what we are actually testing in reality, i.e. the original benchmarks designated 8 threads however its unlikely in reality that we will have more than 8 threads hitting the `nextId()` function at once with such high contention, so instead I set the default to just a single thread but added another set of benchmarks for multiple threads.

@pjfanning @IainHull Let me know if I may have missed something

actor/src/main/scala/org/apache/pekko/util/RandomShortProvider.scala
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+@InternalApi
+private[pekko] class RandomShortProvider(seed: Random) {
+  // These methods are volatile so that they stay within cache
+  private var index = seed.nextLong
+  private var increment = seed.nextLong
+  private var count = 1L
+
+  count = 1
```
pjfanning, line 31: line 31 seems unnecessary

mdedetrich: Fixed, thanks

actor/src/main/scala/org/apache/pekko/util/RandomShortProvider.scala
```diff
@@ -0,0 +1,43 @@
...
+
+@InternalApi
+private[pekko] class RandomShortProvider(seed: Random) {
+  // These methods are volatile so that they stay within cache
+  private var index = seed.nextLong
+  private var increment = seed.nextLong
+  private var count = 1L
+
+  def nextId(): Short = synchronized {
+    val result = (0xFFFFFFFF & index).asInstanceOf[Short]
```
pjfanning, line 32: would it be possible to consider https://github.com/mdedetrich/incubator-pekko/pull/1 ? @alexandru suggested that 'synchronized' is a bad solution and the semaphore can timeout (while synchronized can just hang).

mdedetrich: Sure ill look at it tomorrow, will be busy tonight

He-Pin (approved):
lgtm, and it's cpu bound and so ok to use the synchronized.

And there is already have a util, let's make use of that one.

actor/src/main/scala/org/apache/pekko/util/RandomShortProvider.scala
mdedetrich: So I have just update the version to use a semaphore as well as making the timeout configurable plus I did some other improvements to make the config reading/parsing consistent with the rest of Pekko.

I used the exact same version that you suggested (with a minor performance improvement to avoid creating a new class in the failure case of acquiring a lock) and the performance nosedived.

```
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash                 thrpt    3  ≈ 10⁻¹⁰            ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom                       thrpt    3    0,001 ±   0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom                  thrpt    3    0,311 ±   0,031  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureEnhancedDoubleHash  thrpt    3  ≈ 10⁻¹⁰            ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureSecureRandom        thrpt    3    0,001 ±   0,001  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureThreadLocalRandom   thrpt    3    0,608 ±   0,036  ops/ns
```

mdedetrich: So using a much shorter timeout the improvements are better (i.e. this is the results with 10 nanos)

```
[info] IdGeneratorBanchmark.measureEnhancedDoubleHash                 thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureSecureRandom                       thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom                  thrpt    3  0,312 ±  0,074  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureEnhancedDoubleHash  thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureSecureRandom        thrpt    3  0,001 ±  0,001  ops/ns
[info] IdGeneratorBanchmark.multipleThreadsMeasureThreadLocalRandom   thrpt    3  0,622 ±  0,018  ops/ns
```

I am not sure what problem we are trying to achieve with a semaphore though, in such a hot loop with only a single permit its actually slower than `synchronized` which because the JVM know its only single threaded (i.e. permit of 1)  is hyper optimized for this case.

pjfanning (approved):
lgtm

actor/src/main/scala/org/apache/pekko/util/RandomShortProvider.scala
mdedetrich: @pjfanning You just approved PR, did you read this?

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
```diff
@@ -58,4 +56,37 @@ private[pekko] object IdGenerator {
...
+     */
+    private val lock = new Semaphore(1)
+    private var index = seed.nextLong
+    private var increment = seed.nextLong
+    private var count = 1L
+
+    private val onAcquireFailure = IdGenerator.random(seed)
+
+    override final def nextId(): Short = {
+      if (lock.tryAcquire(semaphoreTimeout.length, semaphoreTimeout.unit)) {
```
pjfanning, line 77: revert to synchronized if you find it is working better for you

mdedetrich: Okay I will do so

pjfanning: btw I just merged a PR fixing a small issue in the negative ID fix - I have resolved the conflict in your PR/branch - so you might want to pull the latest code

mdedetrich:
> @mdedetrich this PR is showing a Merge Conflict in GitHub

Fixed

IainHull:
@mdedetrich Sorry for the delay getting back, this looks good.

mdedetrich:
@IainHull Note that I made a ticket for anyone that wants to further improve performance of double enhanced hasher https://github.com/apache/incubator-pekko/issues/411

PR #374, opened 2023-06-07
fix up refs to Akka in source code

pjfanning:


discovery/src/main/scala/org/apache/pekko/discovery/ServiceDiscovery.scala
```diff
@@ -150,7 +150,7 @@ final class Lookup(val serviceName: String, val portName: Option[String], val pr
   require(serviceName.trim.nonEmpty, "'serviceName' cannot be empty")
 
   /**
-   * Which port for a service e.g. Akka remoting or HTTP.
+   * Which port for a service e.g. Apache Pekko remoting or HTTP.
```
He-Pin, line 153: Some parts are `Pekko` but some are `Pekko`, it there any guide about this?

pjfanning: not really - some times, it seems more correct to use Apache Pekko.

At some point, we'll probably need someone to specifically go through all the docs and decide when Pekko or Apache Pekko is used.

He-Pin (approved):
lgtm

PR #379, opened 2023-06-09
=act Stash dns query if there is a collision.

He-Pin:
If there is already an onging query with same id, stash the query, otherwise send it directly.
When receive a response, unstash all and process.

refs: https://github.com/akka/akka/issues/31905

Modification:

1. stash the query if there is a collision and unstash after response received.
2. check the response references questions, if not match the previous question, just drop it.
3. response directly when possible.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
@@ -15,11 +15,10 @@ package org.apache.pekko.io.dns.internal
 
 import java.net.{ InetAddress, InetSocketAddress }
 
+import scala.annotation.nowarn
 import scala.collection.{ immutable => im }
 import scala.concurrent.duration._
 import scala.util.Try
-
-import scala.annotation.nowarn
 
 import org.apache.pekko
 import pekko.actor.{ Actor, ActorLogging, ActorRef, NoSerializationVerificationNeeded, Props, Stash }
@@ -52,14 +51,14 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
   import DnsClient._
   import context.system
 
-  val udp = IO(Udp)
-  val tcp = IO(Tcp)
+  val udp: ActorRef = IO(Udp)
+  val tcp: ActorRef = IO(Tcp)
 
   private[internal] var inflightRequests: Map[Short, (ActorRef, Message)] = Map.empty
 
   lazy val tcpDnsClient: ActorRef = createTcpClient()
 
-  override def preStart() = {
+  override def preStart(): Unit = {
     udp ! Udp.Bind(self, new InetSocketAddress(InetAddress.getByAddress(Array.ofDim(4)), 0))
   }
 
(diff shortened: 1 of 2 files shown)


pjfanning:
Generally, this looks ok to me. I'd like to have some more eyes on this change before we merge it.

pjfanning:
@He-Pin would it be possible to add a test or 2 to this PR - for regression testing but also I think a test might make it clearer what this PR is fixing.

Could you also provide the link to the Akka issue - or a Pekko issue - that describes the problem. I'm pretty sure I've seen the related Akka but I can't find it today.

He-Pin:
@pjfanning Let me add some tests

He-Pin:
> @He-Pin did you see the comment in https://github.com/apache/incubator-pekko/pull/371#discussion_r1224490114 about 'To protect DNS poisoning you should also add a check to the response handler in DnsClient to ensure the question in the response matches the question asked.' - fyi @IainHull
> 
> I don't know if that is something that should be done in addition to this change or instead of this change. It seems useful to have both.

Cool, Will update it tomorrow.

actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
```diff
@@ -121,42 +112,73 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
...
+                  questions.head, previousQuestion.head, id)
+              } else {
+                val response = if (flags.responseCode == ResponseCode.SUCCESS) {
+                  Answer(id, answerRecs, additionalRecs)
+                } else {
+                  Answer(id, Nil, Nil)
+                }
+                replyTo ! response
+                inflightRequests -= response.id
+                unstashAll()
```
He-Pin, line 148: response directly here

actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
```diff
@@ -121,42 +112,73 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
...
+            if (flags.isTruncated) {
+              log.debug("DNS response truncated, falling back to TCP")
+              tcpDnsClient ! queryMsg
+            } else {
+              // check if the question is match
+              val previousQuestion = msg.questions
+              if (questions.head != previousQuestion.head) {
+                log.warning(
+                  "Dns response referenced to different question:{} expected:{} for id:{},maybe an error.",
+                  questions.head, previousQuestion.head, id)
```
He-Pin, line 139: check the question

He-Pin:
@pjfanning @mdedetrich I have update the PR ,with additional checking,

He-Pin (commented):
Stashing DNS queries on collision to prevent query ID conflicts.

1. **Bug fix**: This addresses akka/akka#31905, which is a real issue in DNS resolution where concurrent queries with the same ID could cause incorrect responses.

2. **Implementation**: The approach of stashing queries on collision and unstashing after response is sound. The validation of response references against expected questions is a good safety measure.

3. **Testing**: Added tests in `DnsClientSpec.scala` covering the collision scenario.

4. **Staleness**: Open since June 2023 with no reviews. This is a genuine bug fix with tests. It should be rebased to ensure it still applies cleanly and the tests still pass. @He-Pin this seems like a worthwhile fix -- any blockers?

Philippus:
By now, we might as well wait a few days for akka 2.8.1 to become Apache 2.

He-Pin:
Closing as obsolete. The issue this was trying to address has since been covered on main by #385 / c56edca78f: duplicate DNS transaction ids now return DuplicateId to the resolver, which retries with a fresh id, and the client verifies the original question for UDP replies, DropRequest, and UDP command failures.

I do not think this PR should be rebased. The remaining behavioral difference here is the stash-and-replay approach for duplicate ids, but main's retry-with-new-id flow is simpler and already covered by DnsClientSpec and AsyncDnsResolverSpec. Also, this PR's response-question check is superseded by main's isSameQuestion implementation.

PR #380, opened 2023-06-09
Make 2.12 ccompat extension classes final

mdedetrich:
These classes shouldn't be extended and can also help out in potential performance


He-Pin (approved):
lgtm

Issue #384, opened 2023-06-12
Async DNS Resolver should not use predictable IDs in requests


pjfanning:
See https://akka.io/security/akka-async-dns-2023-31442.html

pjfanning:
solved by #371 and #385

PR #385, opened 2023-06-12
CVE-2023-31442 Address DNS poisoning vulnerability

IainHull:
This also addresses a concurrency bug in the sequence generator, and verifies the question asked in the response.

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -1144,6 +1144,12 @@ pekko {
         # Defaults to a system dependent lookup (on Unix like OSes, will attempt to parse /etc/resolv.conf, on
         # other platforms, will default to 1).
         ndots = default
+
+        # The policy used to generate dns transaction ids. Options are sequence, thread-local-random or secure-random.
+        # Defaults to thread-local-random similar to Netty, secure-random produces FIPS compliant random numbers but
+        # could block looking for entropy (these are short integers so are easy to bruit-force), sequence is the old
+        # behavior.
+        id-generator-policy = thread-local-random
       }
     }
   }
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -1145,10 +1145,10 @@ pekko {
         # other platforms, will default to 1).
         ndots = default
 
-        # The policy used to generate dns transaction ids. Options are sequence, thread-local-random or secure-random.
+        # The policy used to generate dns transaction ids. Options are thread-local-random or secure-random.
         # Defaults to thread-local-random similar to Netty, secure-random produces FIPS compliant random numbers but
-        # could block looking for entropy (these are short integers so are easy to bruit-force), sequence is the old
-        # behavior.
+        # could block looking for entropy (these are short integers so are easy to bruit-force, use thread-local-random
+        # unless you really require FIPS compliant random numbers).
         id-generator-policy = thread-local-random
       }
     }
--- a/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+package org.apache.pekko.io.dns
+
+import org.apache.pekko.annotation.InternalApi
+
+import java.security.SecureRandom
+import java.util.concurrent.ThreadLocalRandom
+import java.util.concurrent.atomic.AtomicInteger
+import scala.annotation.tailrec
+
+/**
+ * INTERNAL API
+ *
+ * These are called by an actor, however they are called inside composed futures so need to be
+ * nextId needs to be thread safe.
+ */
+@InternalApi
+private[pekko] trait IdGenerator {
+  def nextId(): Short
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[pekko] object IdGenerator {
+  sealed trait Policy
+
+  object Policy {
+    case object Sequence extends Policy
+    case object ThreadLocalRandom extends Policy
+    case object SecureRandom extends Policy
+    val Default: Policy = ThreadLocalRandom
+
+    def apply(name: String): Option[Policy] = name.toLowerCase match {
+      case "sequence"            => Some(Sequence)
+      case "thread-local-random" => Some(ThreadLocalRandom)
+      case "secure-random"       => Some(SecureRandom)
+      case _                     => Some(ThreadLocalRandom)
+    }
+  }
+
+  def apply(policy: Policy): IdGenerator = policy match {
+    case Policy.Sequence          => sequence()
+    case Policy.ThreadLocalRandom => random(ThreadLocalRandom.current())
+    case Policy.SecureRandom      => random(new SecureRandom())
+  }
+
+  /**
+   * @return a random sequence of ids for production
+   */
+  def random(rand: java.util.Random): IdGenerator = new IdGenerator {
+    override def nextId(): Short = rand.nextInt(Short.MaxValue).toShort
+  }
+
+  /**
+   * @return a predictable sequence of ids for tests
+   */
+  def sequence(): IdGenerator = new IdGenerator {
+    val requestId: AtomicInteger = new AtomicInteger(0)
+
+    @tailrec
+    override final def nextId(): Short = {
+      val oldId = requestId.get()
+      val newId = (oldId + 1) % Short.MaxValue
+
+      if (requestId.compareAndSet(oldId, newId.intValue())) {
+        newId.toShort
+      } else {
+        nextId()
+      }
+    }
+  }
+}
--- a/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
@@ -35,13 +35,11 @@ private[pekko] object IdGenerator {
   sealed trait Policy
 
   object Policy {
-    case object Sequence extends Policy
     case object ThreadLocalRandom extends Policy
     case object SecureRandom extends Policy
     val Default: Policy = ThreadLocalRandom
 
     def apply(name: String): Option[Policy] = name.toLowerCase match {
-      case "sequence"            => Some(Sequence)
       case "thread-local-random" => Some(ThreadLocalRandom)
       case "secure-random"       => Some(SecureRandom)
       case _                     => Some(ThreadLocalRandom)
@@ -49,10 +47,11 @@ private[pekko] object IdGenerator {
   }
 
   def apply(policy: Policy): IdGenerator = policy match {
-    case Policy.Sequence          => sequence()
     case Policy.ThreadLocalRandom => random(ThreadLocalRandom.current())
     case Policy.SecureRandom      => random(new SecureRandom())
   }
+
+  def apply(): IdGenerator = random(ThreadLocalRandom.current())
 
   /**
    * @return a random sequence of ids for production
--- a/actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
@@ -19,6 +19,8 @@ import scala.collection.immutable
 import scala.concurrent.ExecutionContextExecutor
 import scala.concurrent.Future
 import scala.util.Try
+import scala.util.Success
+import scala.util.Failure
 import scala.util.control.NonFatal
 
 import org.apache.pekko
@@ -41,9 +43,16 @@ import pekko.util.PrettyDuration._
 private[io] final class AsyncDnsResolver(
     settings: DnsSettings,
     cache: SimpleDnsCache,
-    clientFactory: (ActorRefFactory, List[InetSocketAddress]) => List[ActorRef])
+    clientFactory: (ActorRefFactory, List[InetSocketAddress]) => List[ActorRef],
+    idGenerator: IdGenerator)
     extends Actor
     with ActorLogging {
+
+  def this(
+      settings: DnsSettings,
+      cache: SimpleDnsCache,
+      clientFactory: (ActorRefFactory, List[InetSocketAddress]) => List[ActorRef]) =
+    this(settings, cache, clientFactory, IdGenerator(settings.IdGeneratorPolicy))
 
   import AsyncDnsResolver._
 
@@ -84,13 +93,6 @@ private[io] final class AsyncDnsResolver(
     nameServers,
     settings.SearchDomains,
     settings.NDots)
-
-  private var requestId: Short = 0
-
-  private def nextId(): Short = {
-    requestId = (requestId + 1).toShort
-    requestId
-  }
 
   private val resolvers: List[ActorRef] = clientFactory(context, nameServers)
 
@@ -151,11 +153,19 @@ private[io] final class AsyncDnsResolver(
     }
 
   private def sendQuestion(resolver: ActorRef, message: DnsQuestion): Future[Answer] = {
-    val result = (resolver ? message).mapTo[Answer]
-    result.failed.foreach { _ =>
-      resolver ! DropRequest(message.id)
+    (resolver ? message).transformWith {
+      case Success(result: Answer) =>
+        Future.successful(result)
+      case Success(DuplicateId(_)) =>
+        sendQuestion(resolver, message.withId(idGenerator.nextId()))
+      case Failure(t) =>
+        resolver ! DropRequest(message.id)
+        Future.failed(t)
+      case Success(a) =>
+        resolver ! DropRequest(message.id)
+        Future.failed(
+          new IllegalArgumentException("Unexpected response " + a.toString + " of type " + a.getClass.toString))
     }
-    result
   }
 
   private def resolveWithSearch(
@@ -208,13 +218,13 @@ private[io] final class AsyncDnsResolver(
       case Ip(ipv4, ipv6) =>
         val ipv4Recs: Future[Answer] =
           if (ipv4)
-            sendQuestion(resolver, Question4(nextId(), caseFoldedName))
+            sendQuestion(resolver, Question4(idGenerator.nextId(), caseFoldedName))
           else
             Empty
 
         val ipv6Recs =
           if (ipv6)
-            sendQuestion(resolver, Question6(nextId(), caseFoldedName))
+            sendQuestion(resolver, Question6(idGenerator.nextId(), caseFoldedName))
           else
             Empty
 
@@ -224,7 +234,7 @@ private[io] final class AsyncDnsResolver(
         } yield DnsProtocol.Resolved(name, ipv4.rrs ++ ipv6.rrs, ipv4.additionalRecs ++ ipv6.additionalRecs)
 
       case Srv =>
-        sendQuestion(resolver, SrvQuestion(nextId(), caseFoldedName)).map(answer => {
+        sendQuestion(resolver, SrvQuestion(idGenerator.nextId(), caseFoldedName)).map(answer => {
           DnsProtocol.Resolved(name, answer.rrs, answer.additionalRecs)
         })
     }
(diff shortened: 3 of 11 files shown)


actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
```diff
@@ -35,12 +35,21 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
 @InternalApi private[pekko] object DnsClient {
   sealed trait DnsQuestion {
     def id: Short
+    def withId(newId: Short): DnsQuestion = {
```
IainHull, line 38: Enable us to resend questions with a new id when handling duplicates

actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
```diff
@@ -92,23 +101,26 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
     case Question4(id, name) =>
       log.debug("Resolving [{}] (A)", name)
       val msg = message(name, id, RecordType.A)
-      inflightRequests += (id -> (sender() -> msg))
-      log.debug("Message [{}] to [{}]: [{}]", id, ns, msg)
-      socket ! Udp.Send(msg.write(), ns)
+      newInflightRequests(msg, sender()) {
```
IainHull, line 104: All questions are now protected with `newInflightRequests`. This ensures that transaction id is not already in use and the returns the result.

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -41,10 +43,17 @@ import pekko.util.PrettyDuration._
 private[io] final class AsyncDnsResolver(
     settings: DnsSettings,
     cache: SimpleDnsCache,
-    clientFactory: (ActorRefFactory, List[InetSocketAddress]) => List[ActorRef])
+    clientFactory: (ActorRefFactory, List[InetSocketAddress]) => List[ActorRef],
+    idGenerator: IdGenerator)
     extends Actor
     with ActorLogging {
 
+  def this(
```
IainHull, line 51: This maintains binary compatibility a new parameter with a default value does not.

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
```diff
@@ -151,11 +153,19 @@ private[io] final class AsyncDnsResolver(
...
 
   private def sendQuestion(resolver: ActorRef, message: DnsQuestion): Future[Answer] = {
-    val result = (resolver ? message).mapTo[Answer]
-    result.failed.foreach { _ =>
-      resolver ! DropRequest(message.id)
+    (resolver ? message).transformWith {
+      case Success(result: Answer) =>
+        Future.successful(result)
+      case Success(DuplicateId(_)) =>
+        sendQuestion(resolver, message.withId(idGenerator.nextId()))
```
IainHull, line 160: With random transaction ids its possible the id is already in use. If that is the case retry with a different transaction id

pjfanning:
relates to #384 - alternative to #371 

Thanks @IainHull - we'll go through this over the next few days.

pjfanning:
@IainHull minor nit - our build enforces scalafmt rules. You can use sbt commands to reformat but I find scalafmt-native the quickest and easiest solution.

pjfanning:
@IainHull my understanding is that your employer has a [Corporate CLA](https://www.apache.org/licenses/cla-corporate.txt) registered with the ASF. Alternatively, if you could fill in an [Individual CLA](https://www.apache.org/licenses/icla.pdf) that would make it easier for us to merge this if we all agree this is best way to solve this. So far, it looks like a good solution to me.

As an aside, do you know what the performance overhead of the additional code is? Of course, security is more important than raw performance.

mdedetrich:
> @IainHull minor nit - our build enforces scalafmt rules. You can use sbt commands to reformat but I find scalafmt-native the quickest and easiest solution.

`sbt scalafmtAll` should format entire codebase if you dont want to install additional tools.

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
```diff
@@ -0,0 +1,80 @@
...
+    val requestId: AtomicInteger = new AtomicInteger(0)
+
+    override def nextId(): Short = {
+      val oldId = requestId.get()
+      val newId = (oldId + 1) % Short.MaxValue
+
+      if (requestId.compareAndSet(oldId, newId.intValue())) {
+        newId.toShort
+      } else {
+        nextId()
```
pjfanning, line 76: Could this function be marked `@tailrec`?

actor/src/main/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolver.scala
pjfanning: Seems useful to maintain compatibility but we can document any places where we need to break compatibility. For me, Pekko is a new project and users already have to migrate package names, config names, some class names - so we are already not fully compatible with Akka.

IainHull:
> @pjfanning re scalafmt, I'll do it now. I thought I had did it already, but most have undone it after the organize imports changed to many files.

IainHull:
> @IainHull my understanding is that your employer has a [Corporate CLA](https://www.apache.org/licenses/cla-corporate.txt) registered with the ASF.
Yes Workday should have a corporate CLA. That would be quickest option, if I sign a personal one it will need legal approval and they will want to know why the corporate one was not used.

I can change my commute to use my corporate email address it that helps

IainHull:
> As an aside, do you know what the performance overhead of the additional code is? Of course, security is more important than raw performance.

I performed a very simple benchmark of just `IdGenerator` the default `ThreadLocalRadom` implementation is way faster with 8 threads and lots of contention. This is because the `SecureRandom` and `Sequence` use `AtomicInteger` under the hood.

```
[info] Benchmark                                       Mode  Cnt  Score   Error   Units
[info] IdGeneratorBanchmark.measureSecureRandom       thrpt    3  0.005 ± 0.008  ops/ns
[info] IdGeneratorBanchmark.measureSequence           thrpt    3  0.005 ± 0.002  ops/ns
[info] IdGeneratorBanchmark.measureThreadLocalRandom  thrpt    3  0.652 ± 0.670  ops/ns
```

I expect a more serious performance test that includes the collision detection etc would have to use a real DNS to get a full understanding of the overhead.

pjfanning:
@He-Pin @mdedetrich @alexandru do any of you see any reason not to merge this? I have one minor change request (about adding a tailrec annotation). Other than that, I think these changes tick the boxes that we need to fix the security issue and looks better from a thread safety respect than what is currently there. 

I'm sort of anxious to draw a line under this and move onto an RC. I'm expecting that we'll need at least a couple of RCs to allow the Incubator team to fully vet the licensing and other issues and it would be great to start that process this week or next week. As we draw into summer, we could find ourselves short of people to do the reviews and votes.

IainHull:
@pjfanning @He-Pin @mdedetrich @alexandru 

Thanks for all you help I have just added the tailrec and a simple jmh test for the IdGenerator.

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
```diff
@@ -0,0 +1,82 @@
...
+   * @return a random sequence of ids for production
+   */
+  def random(rand: java.util.Random): IdGenerator = new IdGenerator {
+    override def nextId(): Short = rand.nextInt(Short.MaxValue).toShort
+  }
+
+  /**
+   * @return a predictable sequence of ids for tests
+   */
+  def sequence(): IdGenerator = new IdGenerator {
```
mdedetrich, line 67: My concern here is that while this is definitely useful for tests I don't know the code being there for it to be accidentally used in production considering the security ramifications.

Its actually possible to modify the tests so it can handle random ID's and I would argue that its actually beneficial to do this since you are testing intended behaviour rather than against hardcoded id's.

Thankfully the hard work has already been done here, see https://github.com/apache/incubator-pekko/pull/371/files#diff-17ee4dc4858fd137eeb621bf70244df69a1c97eeb5f9d178699c74fbc500dd6c . You can copy that and it should just work.

@IainHull @pjfanning @He-Pin wdyt?

mdedetrich (changes requested):
So I have made one comment regarding the sequential id generator which is only used in tests.

Regarding the various configurable id algorithms being added, is the intention here to merge this PR and then add a thread safe version of @Claudenw 's algorithm after the PR. @pjfanning @He-Pin ?

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
pjfanning: I don't think it is problematic to support the old sequential style as an option as long as that we document the potential  problems. Users are looking for secure by default. This does not mean that we can't support other modes.

In terms of the tests, it may be fine to switch them to use random ids if the switch is easy.

pjfanning:
> So I have made one comment regarding the sequential id generator which is only used in tests.
> 
> Regarding the various configurable id algorithms being added, is the intention here to merge this PR and then add a thread safe version of @Claudenw 's algorithm after the PR. @pjfanning @He-Pin ?

I don't really see the new for adding the extra pseudorandom id generator. Maybe we could add it later. 

My aim is to press forward to a release and we don't really need multiple generators to be supported.

mdedetrich:
> > So I have made one comment regarding the sequential id generator which is only used in tests.
> > Regarding the various configurable id algorithms being added, is the intention here to merge this PR and then add a thread safe version of @Claudenw 's algorithm after the PR. @pjfanning @He-Pin ?
> 
> I don't really see the new for adding the extra pseudorandom id generator. Maybe we could add it later.
> 
> My aim is to press forward to a release and we don't really need multiple generators to be supported.

Well we can always add it after this PR, its a shame to have the effort spent wasted. But agreed that it shouldn't block the PR.

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
pjfanning: @IainHull the point about sequential generators being against the spec is valid. Do you think it would be feasible to drop the sequential generator and update the tests?

IainHull: Yes I should be able to get the done today, the tests are the hardest part and they are already done I will copy them in from @mdedetrich's PR.

actor/src/main/resources/reference.conf
```diff
@@ -1144,6 +1144,12 @@ pekko {
         # Defaults to a system dependent lookup (on Unix like OSes, will attempt to parse /etc/resolv.conf, on
         # other platforms, will default to 1).
         ndots = default
+
+        # The policy used to generate dns transaction ids. Options are sequence, thread-local-random or secure-random.
+        # Defaults to thread-local-random similar to Netty, secure-random produces FIPS compliant random numbers but
+        # could block looking for entropy (these are short integers so are easy to bruit-force), sequence is the old
+        # behavior.
+        id-generator-policy = thread-local-random
```
mdedetrich, line 1152: This needs to be `SecureRandom` by default or we have to use @Claudenw version as bare minimum. We should not have insecure algorithm by default

mdedetrich (changes requested):
Also the default policy should be `SecureRandom` (bare minimum random using `SecureRandom` seed).

actor/src/main/resources/reference.conf
pjfanning: If we are worried about contention on the secure random, we could add another id generator that uses 'ThreadLocal<SecureRandom>' (SecureRandom has some synchronization). It would be good to have a plain SecureRandom too because ThreadLocals can be problematic if you have lots of short lived threads.

Claudenw: If you are generating unsigned shorts (0-65535) for a total of 65536 values then then possibility of collision in the generation for the Nth number is:

 P(n) = 1- (65535/35536)^(n*(n-1)/2)

So on the 300th selection the probability of collision is 49.5588

Simple random number generation is not going to work here.

pjfanning: @Claudenw this PR handles duplicates

That said, we could add a generator based on what you were working on as an optional generator.

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
IainHull: I have just pushed the changes to remove the sequence id generator

actor/src/main/resources/reference.conf
IainHull: It could be, but it comes with a lot of downsides. Its very easy to run out of entropy when a container starts which will block the startup of your application trying to generate the transactions id for DNS queries to join the cluster.

The other thing to note is that we are generating a short, it is trivial to brute-force 32k numbers if an attacker wanted to break DNS SecureRandom would not prevent that.

IMO the only reason to enable SecureRandom would be if you require it for FIPS.

pjfanning: https://www.digitalocean.com/community/tutorials/how-to-setup-additional-entropy-for-cloud-servers-using-haveged is a way to resolve low entropy.

mdedetrich: So I just spoke with @Claudenw and his solution at https://github.com/Claudenw/IdSupplier/blob/main/src/org/example/ShortSupplier.java should do a better job at avoiding so many collision's due to 32k numbers, in other words it actually goes out if its way to minimize the amount of collisions detected by using enhanced double hashing + 2 buffers to also handle roll over

> IMO the only reason to enable SecureRandom would be if you require it for FIPS.

We have a person in our company that specializes in security who's on IETF and his recommendation was to use `SecureRandom` as a seed + a PRF such as SHA-256/HMAC. His view is that `SecureRandom` on every is excessive, however non crypto secure hashes is way too easy.

The common/trivial way to solve this issue appears to be using `SecureRandom` every time and check for collisions against in flight transactions, thats what systemd-resolved does for example.

pjfanning:
@mdedetrich what would a generator that matched 'bare minimum random using SecureRandom seed' look like?

Maybe something like?
```
      SecureRandom sr = SecureRandom.getInstance("SHA1PRNG"); // could be static?
      Random rnd = new Random(sr.nextLong());
```

Something like that could be wrapped in a ThreadLocal - one that initializes with code like the one above.

It's possible that this plain Random with a secure seed would be a reasonable compromise.

mdedetrich:
> @mdedetrich what would a generator that matched 'bare minimum random using SecureRandom seed' look like?
> 
> Maybe something like?
> 
> ```
>       SecureRandom sr = SecureRandom.getInstance("SHA1PRNG"); // could be static?
>       Random rnd = new Random(sr.nextLong());
> ```
> 
> Something like that could be wrapped in a ThreadLocal - one that initializes with code like the one above.
> 
> It's possible that this plain Random with a secure seed would be a reasonable compromise.

It would be something like

```
public static String hmacWithJava(String algorithm, String data, String key)
  throws NoSuchAlgorithmException, InvalidKeyException {
    SecretKeySpec secretKeySpec = new SecretKeySpec(key.getBytes(), algorithm);
    Mac mac = Mac.getInstance(algorithm);
    mac.init(secretKeySpec);
    return bytesToHex(mac.doFinal(data.getBytes()));
}
```

Where algorithm = `HmacSHA256`, `data` is your initial `SecureRandom.seed` and `key` is just an incrementing counter every time `nextId()` is called (which is how you would use SHA-256 as a PRNG)

mdedetrich:
@IainHull @pjfanning  @pjfanning Are there any objections to adding @Claudenw  random solution described at https://github.com/apache/incubator-pekko/pull/385#discussion_r1228053082 ? Making it thread safe would largely be trivial (i.e. adding `synchronized` to `nextId()`) and the final result shouldn't be too slow, at least compared to making `SecureRandom`/`SHA256` thread safe.

I can rebase/target my PR https://github.com/apache/incubator-pekko/pull/371 against this PR to bring in @IainHull 's changes and add another algorithm option, or we can just merge this PR as is and then I will rebase mine.

pjfanning:
> @IainHull @pjfanning @pjfanning Are there any objections to adding @Claudenw random solution described at [#385 (comment)](https://github.com/apache/incubator-pekko/pull/385#discussion_r1228053082) ? Making it thread safe would largely be trivial (i.e. adding `synchronized` to `nextId()`) and the final result shouldn't be too slow, at least compared to making `SecureRandom`/`SHA256` thread safe.
> 
> I can rebase/target my PR #371 against this PR to bring in @IainHull 's changes and add another algorithm option, or we can just merge this PR as is and then I will rebase mine.

Couldn't we merge this and add a new generator and make it the default generator if that makes sense? We won't release until all the conversations are complete.

mdedetrich:
> Couldn't we merge this and add a new generator and make it the default generator if that makes sense? We won't release until all the conversations are complete.

Sure, @IainHull PR needs to be formatted

pjfanning:
I pushed https://github.com/apache/incubator-pekko/pull/385/commits/710d0874bfa028be3a6bd1ff1cbdd26d5c3292e7 to fix the scalafmt issue

pjfanning:
@mdedetrich any objections to me merging this

mdedetrich (approved):
Nope, lgtm

IainHull:
> I pushed https://github.com/apache/incubator-pekko/pull/385/commits/710d0874bfa028be3a6bd1ff1cbdd26d5c3292e7 to fix the scalafmt issue

Thanks.

IainHull:
Sorry @pjfanning I just found a bug checking something that Norman said about the rational for the netty change. I'll get a fix ready with in the hour.

pjfanning:
> Sorry @pjfanning I just found a bug checking something that Norman said about the rational for the netty change. I'll get a fix ready with in the hour.

Thanks.

pjfanning (changes requested):
Changing review status until Iain's change is ready

actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
```diff
@@ -140,9 +157,25 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
             log.debug("Client for id {} not found. Discarding unsuccessful response.", msg.id)
         }
       } else {
-        val (recs, additionalRecs) =
-          if (msg.flags.responseCode == ResponseCode.SUCCESS) (msg.answerRecs, msg.additionalRecs) else (Nil, Nil)
-        self ! Answer(msg.id, recs, additionalRecs)
+        inflightRequests.get(msg.id) match {
+          case Some((_, orig)) if !isSameQuestion(msg.questions, orig.questions) =>
```
IainHull, line 161: There was a double bug here.  First the if expression was missing the `!`, then `isSameQuestion` was matching on a `Seq` with `List` extractors. The double negative caused the test to pass.

IainHull:
Looking at integration test failure. I wont get is solved tonight

pjfanning:
> Looking at integration test failure. I wont get is solved tonight

I ran the AsyncDnsResolverIntegrationSpec locally (using the PR branch) and the tests passed when run use `sbt actor-tests / testOnly *AsyncDns`.

I'm not sure if this is user error on my part or if the tests are failing in CI due to some lingering state in the CI run (it runs a lot more tests than I did in my run).

pjfanning:
I reran the CI build and it passed this time. So we have something that appears to kick in intermittently.

pjfanning:
@mdedetrich what do you think we should do? Should we merge this and live with intermittent failures? Do we know if the randomised IDs are what causes these failures and if so, would having a mechanism to plug in sequential ID generator for unit testing only would be a reasonable solution?

mdedetrich:
Lets get an opinion from @IainHull

pjfanning:
fyi I made a small mod - https://github.com/apache/incubator-pekko/pull/385/commits/b0500182461029683aafd0b2ddd358d3ab23f110. For some reason, GitHub is offering me the option to make code suggestions like it often does in PRs. Feel free to modify or revert this if it is incorrect.

IainHull:
> fyi I made a small mod - https://github.com/apache/incubator-pekko/commit/b0500182461029683aafd0b2ddd358d3ab23f110
Thanks PJ thats a good spot.

I have been struggling to run the docker based tests locally. They pass but they are skipped on a mac because it does not find the docker host. When I set  the `DOCKER_HOST` environment variable the tests run but don't seem to start the docker container they depend on.

I will examine the output of the test when it passed and failing to incase there is an intermittent bug instead of an intermittent test. Then I will get back to you.

pjfanning (approved):
lgtm

I think the tests failing was sheer bad luck. I don't recall seeing these tests fail before.

mdedetrich:
@pjfanning Should we go ahead and merge? Even if its not perfect we don't have to release right now and I can work ontop of the PR to add the previously mentioned algorithms

pjfanning:
merged - thanks @IainHull 

@mdedetrich are you planning on working an extra generator based on @Claudenw's work? Another option is to switch the default generator to be 'secure-random' (and users can set a config to use the thread-local-random if their organisation is happy that that generator suits their needs).

mdedetrich:
> merged - thanks @IainHull
> 
> @mdedetrich are you planning on working an extra generator based on @Claudenw's work?

Yes I will get to it tonight, tomorrow

PR #388, opened 2023-06-13
Add final to methods to prepare for future inline

mdedetrich:
Since https://github.com/apache/incubator-pekko/pull/305 is basically ready and just waiting for an upstream fix to Scala 2.13, this PR brings in some changes from that PR in preparation for the RC so that we don't potentially break users when we do add the inline.

The changes are quite trivial, i.e. it involves add `final` to some methods inside of `ByteIterator`. If you don't do this and enable the inliner then the inliner complains with the following

```
[error] /Users/mdedetrich/github/incubator-pekko/actor/src/main/scala-2.12/org/apache/pekko/util/ByteIterator.scala:227:12: org/apache/pekko/util/ByteIterator$MultiByteArrayIterator::clear()V is annotated @inline but could not be inlined:
[error] The method is not final and may be overridden.
[error]       clear()
```

While this will prevent users from overriding the methods, it doesn't make sense to override this method because the implementation relies on this implementation to work. Furthermore other similar methods are marked as `final` so I believe missing this was an oversight like due to `@inline` currently not doing anything.

### DIFF
--- a/actor/src/main/scala-2.12/org/apache/pekko/util/ByteIterator.scala
+++ b/actor/src/main/scala-2.12/org/apache/pekko/util/ByteIterator.scala
@@ -208,7 +208,7 @@ object ByteIterator {
 
     @inline private def current: ByteArrayIterator = iterators.head
     @inline private def dropCurrent(): Unit = { iterators = iterators.tail }
-    @inline def clear(): Unit = { iterators = MultiByteArrayIterator.empty.iterators }
+    @inline final def clear(): Unit = { iterators = MultiByteArrayIterator.empty.iterators }
 
     @inline final def hasNext: Boolean = current.hasNext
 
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

mdedetrich:
Let me merge this PR, there are a few more methods that I need to verify/add

mdedetrich:
Yes but I am quite sure this is final, I am going through all Pekko modules just to see that these methods are not manually overriden

mdedetrich:
So I just updated the PR to remove the `final` from `remote/src/main/scala/org/apache/pekko/remote/artery/Association.scala`. Note that `Association` is marked as private and internal ABI so we can also add it later but the `@inline` that is already there is like a mistake (its not possible to inline because `compareAndSwapObject` is from java SDK). Its like that methods calling `swapState` need to be inlined but again that is something we can do later since this is all private internal pekko stuff/

The `ByteString` methods here are more critical

Will merge PR once tests pass.

mdedetrich:
Merging, test failure is from link validator which is unrelated.

PR #390, opened 2023-06-13
upgrade to protobuf 3.16.3 (due to CVE)

pjfanning:
* https://github.com/apache/incubator-pekko/security/dependabot/59
* bumps from 3.16.1 to 3.16.3 -- so the risks are low (a patch level bump)

### DIFF
--- a/legal/pekko-protobuf-v3-jar-license.txt
+++ b/legal/pekko-protobuf-v3-jar-license.txt
@@ -202,6 +202,6 @@
 
 ---------------
 
-pekko-protobuf-v3 contains the sources of Google protobuf 3.16.1 runtime support,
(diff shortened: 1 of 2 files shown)


mdedetrich (approved):
lgtm

mdedetrich:
I think this also needs to be done in pekko-grpc so we should take a not of it.

pjfanning:
> I think this also needs to be done in pekko-grpc so we should take a not of it.

Looking at pekko-grpc, it seems like it might be using protobuf 3.21 due to io.grpc libs

pjfanning:
I've added https://github.com/apache/incubator-pekko-grpc/pull/89

He-Pin (approved):
lgtm

He-Pin:
IIRC , Akka dropped akka-protobuf in later version.

PR #391, opened 2023-06-13
Fix incorrect coursier setup action

mdedetrich:
See https://github.com/coursier/setup-action for the correct naming. https://issues.apache.org/jira/browse/INFRA-24664 is also relevant.


pjfanning (approved):
lgtm

PR #394, opened 2023-06-13
Update SSL Config version for Scala 2.12 and 2.13

mdedetrich:
This PR updates ssl config specifically for Scala 2.12/Scala 2.13 (note that the Scala 3 version is unchanged because Pekko was forced to use the latest version of ssl-config for Scala 3 support). The reasoning why this is being done now is the following

* The current version of ssl-config has an actual bug when it comes to PKCS12 (see https://github.com/lightbend/ssl-config/pull/306)
* 0.6.1 is binary compatible with 0.4.3, not that it matters since we only really care about source compatibility (see https://github.com/lightbend/ssl-config/pull/342). Thanks to our friendly resident @raboof 
* While having different versions for libraries due to different Scala versions in `Test` scope is tolerable, I think this is something we should strongly avoid when it comes to actual compile/runtime dependencies otherwise people can get interesting surprises when updating Scala versions
* Other depending libraries like Play also use ssl-config, so its best that we all agree on the latest version.


He-Pin (approved):
lgtm,0.6.1 is right

pjfanning (approved):
Lgtm

PR #395, opened 2023-06-13
Fix Scala 3.3 compile issues

mdedetrich:
Not sure how this was missed but when I locally compile Pekko for Scala 3.3 it fails on this code. We should probably investigate adding Scala 3.3 to normal CI, at least `compile`/`test:compile`.


He-Pin (approved):
lgtm

PR #396, opened 2023-06-13
Use consistent scalatest/scalatest versions

mdedetrich:
For starters `scalaTestScalaCheckVersion` as `1-14` isn't meant to run with `scalaCheckVersion` 1.15.1 and `scalaTest` `3.2.x` series does support Scala 2.12

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -48,20 +48,9 @@ object Dependencies {
 
   val sslConfigVersion = "0.6.1"
 
-  val scalaTestVersion = Def.setting {
-    if (scalaVersion.value.startsWith("3.")) {
-      "3.2.9"
-    } else {
-      "3.1.4"
-    }
-  }
-  val scalaTestScalaCheckVersion = Def.setting {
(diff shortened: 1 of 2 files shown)


He-Pin:
Then after this change, will not be able to run tests against 2.12.*？

mdedetrich:
> Then after this change, will not be able to run tests against 2.12.*？

Scala 2.12 is still fully supported, all of these artifacts versions I picked are also published for 2.12

He-Pin:
@mdedetrich Will it compile and test with scala 2.12.x?

mdedetrich:
> @mdedetrich Will it compile and test with scala 2.12.x?

Yes

mdedetrich:
Just to be clear, nothing whatsoever has changed regarding Scala 2.12 support in any way (or any other Scala version for that matter). All that the PR does is pick specific versions of ScalaTest and its various integrations that happen to be cross compiled for **all** scala versions we support (i.e. 2.12/2.13/3) rather than having bespoke versions depending on the Scala version.

pjfanning (approved):
lgtm

PR #398, opened 2023-06-14
Add occasionally failing link to validator ignore list

mdedetrich:
I have seen this failing already a couple of times which makes it unreliable enough to get added onto ignore list


pjfanning (approved):
lgtm

PR #399, opened 2023-06-15
remove sequence id generator

pjfanning:
* we can add this back in the future if there is a use case
* it's against the DNS resolver spec to use sequential IDs


mdedetrich (approved):
lgtm

PR #401, opened 2023-06-15
support negative short ids in async DNS resolver code

pjfanning:
the ID space is small enough - we need to allow negative IDs to double the ID space

fyi @IainHull

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
@@ -30,6 +30,8 @@ private[pekko] trait IdGenerator {
  */
 @InternalApi
 private[pekko] object IdGenerator {
+  private val MaxUnsignedShort = 65535
+
   sealed trait Policy
 
   object Policy {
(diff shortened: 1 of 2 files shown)


mdedetrich (approved):
lgtm

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
```diff
@@ -54,7 +56,6 @@ private[pekko] object IdGenerator {
   /**
    * @return a random sequence of ids for production
    */
-  def random(rand: java.util.Random): IdGenerator = new IdGenerator {
-    override def nextId(): Short = rand.nextInt(Short.MaxValue).toShort
-  }
+  def random(rand: java.util.Random): IdGenerator =
+    () => (rand.nextInt(MaxUnsignedShort) - Short.MinValue).toShort
```
brharrington, line 60: Since `Short.MinValue` is negative should this be `+ Short.MinValue`? I'm assuming the intent is to move the value into the permitted range for a short before casting. That said I don't think it will alter the behavior as it will overflow and use negative values when cast to short anyway.

pjfanning: good point @brharrington - I'll create a new PR

PR #402, opened 2023-06-15
fix sign in code to generate a random signed short

pjfanning:
see https://github.com/apache/incubator-pekko/pull/401#discussion_r1231609526

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
@@ -32,7 +32,7 @@ private[pekko] trait IdGenerator {
 private[pekko] object IdGenerator {
   // Random.nextInt(bound) generates a random int in the range 0 (inclusive) to bound (exclusive),
   // so add 1 to Max Unsigned Short (65535)
-  private val SignedShortBound = 65536
+  private val UnsignedShortBound = 65536
 
   sealed trait Policy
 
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
```diff
@@ -30,7 +30,9 @@ private[pekko] trait IdGenerator {
  */
 @InternalApi
 private[pekko] object IdGenerator {
-  private val MaxUnsignedShort = 65535
+  // Random.nextInt(bound) generates a random int in the range 0 (inclusive) to bound (exclusive),
+  // so add 1 to Max Unsigned Short (65535)
+  private val SignedShortBound = 65536
```
nvollmar, line 35: Wouldn't that be `UnsignedShortBound`?

pjfanning: thanks - sorted that now

actor/src/main/scala/org/apache/pekko/io/dns/IdGenerator.scala
```diff
@@ -57,5 +59,5 @@ private[pekko] object IdGenerator {
    * @return a random sequence of ids for production
    */
   def random(rand: java.util.Random): IdGenerator =
-    () => (rand.nextInt(MaxUnsignedShort) - Short.MinValue).toShort
+    () => (rand.nextInt(SignedShortBound) + Short.MinValue).toShort
```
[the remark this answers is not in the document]
pjfanning: 0 is valid. so `rand.nextInt(65536) will give us a response int between 0 and 65535 - ie 65536 possible answers. This is what I'm looking for.

65535 + Short.MIN_VALUE (-32768) is 32767 (which is Short.MAX_VALUE).

0 + Short.MIN_VALUE is -32768 (Short.MIN_VALUE).

PR #407, opened 2023-06-16
Dont publish to nightly and combine publish artifacts and docs

mdedetrich:
This PR does the following things

1. Stops publishing jars into nightlies. This was an artifact from ages ago and we decided to stop doing this since Apache has an official snapshots nexus repo
2. Adds in the coursier cache, this is normal but was missing here causing the builds to take longer than normal
3. Combined snapshot publishing and docs into a single step which should save a lot on time due to not needing to recompile multiple times (this is also typical of other Pekko projects).


pjfanning (approved):
lgtm

PR #408, opened 2023-06-17
=str Make use of ConcurrentHashMap instead of List to reduce contention.

He-Pin:
Was: https://github.com/akka/akka/pull/31262/
refs: https://github.com/akka/akka/issues/29557

Use an ConcurrentMap as set to trace the contention in onFeedbackDispatched

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
@@ -13,19 +13,22 @@
 
 package org.apache.pekko.stream.stage
 
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicReference
+
+import scala.annotation.nowarn
 import scala.annotation.tailrec
 import scala.collection.{ immutable, mutable }
 import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration.FiniteDuration
-import scala.annotation.nowarn
+
 import org.apache.pekko
 import pekko.{ Done, NotUsed }
 import pekko.actor._
 import pekko.annotation.InternalApi
 import pekko.japi.function.{ Effect, Procedure }
-import pekko.stream.Attributes.SourceLocation
 import pekko.stream._
+import pekko.stream.Attributes.SourceLocation
 import pekko.stream.impl.{ ReactiveStreamsCompliance, TraversalBuilder }
 import pekko.stream.impl.ActorSubscriberMessage
 import pekko.stream.impl.fusing.{ GraphInterpreter, GraphStageModule, SubSink, SubSource }
@@ -1227,13 +1230,11 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount: 
        * Add this promise to the owning logic, so it can be completed afterPostStop if it was never handled otherwise.
        * Returns whether the logic is still running.
        */
-      @tailrec
       def addToWaiting(): Boolean = {
-        val previous = asyncCallbacksInProgress.get()
-        if (previous != null) { // not stopped
-          val updated = promise :: previous
-          if (!asyncCallbacksInProgress.compareAndSet(previous, updated)) addToWaiting()
-          else true
+        val callbacks = asyncCallbacksInProgress.get()
+        if (callbacks ne null) { // not stopped
+          callbacks.add(promise)
+          asyncCallbacksInProgress.get ne null // logic may already stopped
         } else // logic was already stopped
           false
       }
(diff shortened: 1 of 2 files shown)


mdedetrich:
Is it possible to provide jmh benchmarks to show the difference?

He-Pin:
@mdedetrich I think this is a bug fix not a performance optimization, I can switch to `ConcurrentHashMap` i you like.

mdedetrich:
Ah I see, I will get @jrudolph to review this since he made the original issue.

He-Pin:
After read the paper of TrieMap, I think I need to change it back to ConcurrentHashMap

mdedetrich (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -1282,7 +1283,9 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
   private var callbacksWaitingForInterpreter: List[ConcurrentAsyncCallback[_]] = Nil
   // is used for two purposes: keep track of running callbacks and signal that the
   // stage has stopped to fail incoming async callback invocations by being set to null
-  private val asyncCallbacksInProgress = new AtomicReference[List[Promise[Done]]](Nil)
+  // Using ConcurrentHashMap's KeySetView as Set to track the inProgress async callbacks.
+  private val asyncCallbacksInProgress: AtomicReference[java.util.Set[Promise[Done]]] =
+    new AtomicReference(new ConcurrentHashMap[Promise[Done], NotUsed]().keySet(NotUsed))
```
pjfanning, line 1288: why use an AtomicReference on a mutable data structure like a ConcurrentHashMap? why not just mutate the map?

pjfanning (changes requested):
I'm a bit confused as to the use of `AtomicReference[mutable.Map]` instead of just plain `mutable.Map`.

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
He-Pin: The `AtomicRef` is needed as the callback can be called in a `AsyncCallback` which can be accessed from another thread.
Otherwise it will be a `@volatile asyncCallbacksInProgress` I think to indicate the stage has been stoped.

He-Pin: @pjfanning do you suggest I chang it to `var @volatile asyncCallbacksInProgress` ?

pjfanning: ConcurrentHashMaps are thread safe. I would suggest that we should look at making asyncCallbacksInProgress a val and we can do the usual Map calls on the Map - get, put, clear, etc.

He-Pin: That will not work, because an empty `ConcurrnetHashMap` can't indicate whether the stage is stopped. you can take a look at the comments upper 2~3 lines of this line.

pjfanning: > That will not work, because an empty `ConcurrnetHashMap` can't indicate whether the stage is stopped. you can take a look at the comments upper 2~3 lines of this line.

Keep the AtomicReference then but could could to create the Set using this?

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#newKeySet--

He-Pin: But the default value is BOOLEAN.TRUE, seems a little weird in this case.

pjfanning: Does it matter what the value type is? The Map needs a value type but you are only interested in the key type.

pjfanning: This saves use from having to create a map and then create a keyset - there is just one call to create a keyset.

He-Pin: Noop, it will still create a map, I will update the pr.

He-Pin:
@pjfanning I have updated it.

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
He-Pin: @pjfanning I have updated as you suggestion.

pjfanning (approved):
lgtm

jrudolph:
Yes, this was originally a performance improvement. Regardless of that fact, let's not just merge changes in performance critical code without running benchmarks or discussing arguments why they can be omitted.

It has been a long time since I worked on this / proposed the change (almost 3 years... oh my). I guess in this case, it's fine, it seems after all they ran some benchmarks in Akka to validate it does not introduce obvious performance regressions.

> I'm a bit confused as to the use of `AtomicReference[mutable.Map]` instead of just plain `mutable.Map`.

The comment at its declaration explains it. There's a race condition between scheduling an async callback (i.e. an incoming external event into the stream) and shutting down a stream. The atomic reference makes sure, that we will report feedback to the async callback reliably even if it raced against shutdown (this is basically what this whole feature is about).

PR #415, opened 2023-06-17
Add another remove fqcn to .git-blame-ignore-revs

mdedetrich:
Adds another https://github.com/apache/incubator-pekko/pull/413 to `.git-blame-ignore-revs` since it doesn't effect Scala code in any significant way.


pjfanning (approved):
lgtm

PR #421, opened 2023-06-19
add netty license header to file with significant netty input

pjfanning:
see https://lists.apache.org/thread/mo611z4dpro1373r4f5pbgtlyjmhhtz9

The original source is https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsServerAddressStreamProvider.java

we already have https://github.com/apache/incubator-pekko/pull/237/files - this just builds on those changes

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala
@@ -11,6 +11,22 @@
  * Copyright (C) 2018-2022 Lightbend Inc. <https://www.lightbend.com>
  */
 
+/*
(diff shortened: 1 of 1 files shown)


pjfanning:
> lgtm,but how do you find the origin source @pjfanning

There is a comment with a http link in the Pekko/Akka code

https://github.com/apache/incubator-pekko/blob/265f3068da09560a5593c29edef35d0d15f0f9cd/actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala#L177

It is all the comments saying 'based on' in the code that is causing us all the trouble. We are now being made to forensically examine all the code that might have been borrowed.

Issue #418, opened 2023-06-19
acknowledge use of Aeron code in Pekko


pjfanning:
See https://lists.apache.org/thread/mo611z4dpro1373r4f5pbgtlyjmhhtz9

These files:
* ./remote/src/test/java/org/apache/pekko/remote/artery/aeron/AeronStat.java
* ./remote/src/test/java/org/apache/pekko/remote/artery/RateReporter.java
* ./remote/src/main/java/org/apache/pekko/remote/artery/aeron/AeronErrorLog.java

possibly come from https://github.com/real-logic/aeron

He-Pin:
So aeron is not beening used as a lib？

pjfanning:
If you look at the 3 files in the description - they have `Copyright 2014-2017 Real Logic Ltd.`

This is what is getting us in trouble. We need to find out where these files came from and account for them in our LICENSE and NOTICE files.

Presumably, Lightbend copied these files to avoid having a jar depedendency.

pjfanning:
#422 is merged

PR #422, opened 2023-06-19
acknowledge use of aeron code

pjfanning:
see #418 

https://github.com/real-logic/aeron/ does not have a NOTICE file so we don't have to update our NOTICE files


mdedetrich (approved):
lgtm

PR #424, opened 2023-06-19
upgrade sbt-source-dist plugin

pjfanning:
* brings in an improvement where the files in the zip and tgz archives are grouped with a root directory (while preserving the other paths).
* https://github.com/pjfanning/sbt-source-dist/pull/22


mdedetrich (approved):
lgtm

PR #427, opened 2023-06-19
Delete PrettyByteString.scala

pjfanning:
Doesn't seem to be used.

And any comment that has 'based on' and a link to code is going to cause us problems in code reviews. Licensing and other concerns.


mdedetrich:
Did you check that this isn't used in all of Pekkos modules or just this one?

pjfanning:
I found one in a test in pekko-http-core

https://github.com/search?q=repo%3Aapache%2Fincubator-pekko-http+PrettyByteString&type=code

Looks fixable - it should be feasible to modify the test.

I won't do anything for next few days. I've sent a mesage to @justinmclean about this class and the fact that @jrudolph was a Lightbend staff member - so there is a string argument that we don't need special license treatment for this.

He-Pin:
It may be used in http or debugging i think

jrudolph:
Yes, can go. In this case, "based" meant basically upstreaming some debugging code from akka-http to akka. The http module has similar/same code in `LogByteStringTools.scala` which can be used in the tests instead. I'll add a PR for pekko-http.

jrudolph:
https://github.com/apache/incubator-pekko-http/pull/191

pjfanning:
https://github.com/apache/incubator-pekko-http/pull/191 is merged so this is now safe to merge.

mdedetrich (approved):
lgtm

Issue #426, opened 2023-06-19
acknowledge licensing issues in AbstractNodeQueue


pjfanning:
https://github.com/apache/incubator-pekko/blob/532b7173c175c0c379074cb022e98f5002860c22/actor/src/main/java/org/apache/pekko/dispatch/AbstractNodeQueue.java#L21-L26

pjfanning:
@jrudolph @mdedetrich  could the text in https://github.com/apache/incubator-pekko/pull/433 be adapted to include this too? AbstractNodeQueue does not have any inspiration from Andriy Plokhotnyuk's code but it is inspired by the MPSC code.

mdedetrich:
> @jrudolph @mdedetrich could the text in #433 be adapted to include this too? AbstractNodeQueue does not have any inspiration from Andriy Plokhotnyuk's code but it is inspired by the MPSC code.

So I just had a look at the referenced code at https://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue and this does look like its a similar case, i.e. the code is in a different language (C) and is minimal compared to `AbstractNodeQueue`.

mdedetrich:
Closing as completed, will re-open if more needs to be done.

pjfanning:
@mdedetrich I think it is better if we modify the text from #433 to also mention 'AbstractNodeQueue.java'.

pjfanning:
#434 merged

Issue #425, opened 2023-06-19
acknowledge licensing issues in AbstractBoundedNodeQueue


pjfanning:
https://github.com/apache/incubator-pekko/blob/532b7173c175c0c379074cb022e98f5002860c22/actor/src/main/java/org/apache/pekko/dispatch/AbstractBoundedNodeQueue.java#L18-L27

mdedetrich:
This seems reasonable to me

PR #429, opened 2023-06-19
add licenses for AbstractBoundedNodeQueue.java

pjfanning:
* see #425 and #426 
 * neither the Andriy Plokhotnyuk or Dmitriy Vyukov links contain notice files.
 * https://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue
 * https://github.com/plokhotnyuk/actors/blob/master/LICENSE
 * https://github.com/plokhotnyuk/actors/blob/2e65abb7ce4cbfcb1b29c98ee99303d6ced6b01f/src/test/scala/akka/dispatch/Mailboxes.scala does not contain a source code header

### DIFF
--- a/actor/src/main/java/org/apache/pekko/dispatch/AbstractBoundedNodeQueue.java
+++ b/actor/src/main/java/org/apache/pekko/dispatch/AbstractBoundedNodeQueue.java
@@ -11,6 +11,29 @@
  * Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
  */
 
+/*
+ * Copyright (c) 2010-2011 Dmitry Vyukov. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+ * 1. Redistributions of source code must retain the above copyright notice, this list of
+ *    conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *    of conditions and the following disclaimer in the documentation and/or other materials
+ *    provided with the distribution.
+ *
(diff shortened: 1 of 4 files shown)


actor/src/main/java/org/apache/pekko/dispatch/AbstractBoundedNodeQueue.java
```diff
@@ -11,6 +11,29 @@
  * Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
  */
 
+/*
```
jrudolph, line 14: I think the existing comment far more clearly covers what happened. There was no code copied from any of both projects but rather ideas. If anything, the code from Andriy Plokhotnuyk has methods that are somewhat similar.

pjfanning: @justinmclean do you have suggestions for this? Should we treat ideas separately from source?

jrudolph: IANAL but usually there's no copyright on the ideas themselves. The question is if enough code has been copied or literally translated to warrant a copyright header. I'd say, it isn't enough code and the existing comment faithfully represents that the code has been *based* on that code and ideas without being a literal copy. As mentioned in https://github.com/apache/incubator-pekko/issues/425#issuecomment-1598371026, we can add a small notice to the LICENSE file to clarify that (whether or not we might need a license), give credit and specify that the existing licenses were permissive.

pjfanning:
closing in favour of https://github.com/apache/incubator-pekko/pull/433

PR #434, opened 2023-06-20
acknowledge derived code in AbstractBoundedNodeQueue 

pjfanning:
and include text from BSD license with copyright declaration

relates to #426 and #425 

I think we have to include the text of the BSD license - especially the copyright part

https://www.1024cores.net/home/code-license

We will get in a lot less trouble if we include too much than too little.


mdedetrich (approved):
lgtm

Issue #442, opened 2023-06-21
include critical TLS fix from Akka 2.6.21 (pekko-stream TLSActor)


pjfanning:
include critical TLS fix from Akka 2.6.21 - Akka 2.6.21 is Apache licensed

* https://github.com/akka/akka/issues/29922
* https://github.com/akka/akka/compare/v2.6.20...v2.6.21

pjfanning:
#443 merged

PR #443, opened 2023-06-21
TLSActor: apply check to avoid an endless loop

pjfanning:
see #442

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/io/TLSActor.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/io/TLSActor.scala
@@ -71,6 +71,7 @@ import pekko.util.ByteString
 
   import TLSActor._
 
+  private var unwrapPutBackCounter: Int = 0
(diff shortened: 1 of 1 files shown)


mdedetrich (approved):
lgtm

jrudolph:
FTR this has been ongoing since years with behavior changing between different JDK versions. This change is not a fix but a mitigation of the impact by failing early instead of spinning a CPU core.

jrudolph:
So, after this change, should you ever come across the new `Stuck in unwrap loop, bailing out, last handshake status` exception, please report back, in the best case with a reproducer so we can actually work on a fix.

In a previous similar instance of the problem, I worked together with someone from OpenJDK team at Oracle to get a similar situation fixed in OpenJDK directly but the issue is so hard to reproduce.

In one of the previous scenarios the problem only occurred while the connection was torn down, in which case the fix to bail out might actually have little impact on user code (since no data was flowing over the TLS connection any more), so hopefully that is the case here again.

PR #445, opened 2023-06-21
[pekko-testkit-typed] Support for Junit5 

thmue:
I added a Junit5 Extension to support Junit5 framework integration.

The extension is activated through an annotation:
@ExtendWith(TestKitJunit5Extension.class)
on the class level.

Additionally, a testKit with a @Junit5TestKit annotation needs to be created.
A builder pattern is used to construct the Testkit instance.

@Junit5TestKit
public ActorTestKit testKit = new Junit5TestKitBuilder().build();

The extension itself uses the Junit5 AfterAllCallback to shutdown the testkit after all tests are executed.

I closed the the last PR due to some rebase issues, this PR is directly based on main. Additionally, i discarded the changes in the docs. Sorry for the inconvenience.

### DIFF
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/Junit5TestKit.java
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/Junit5TestKit.java
@@ -4,10 +4,11 @@
  *
  *   https://www.apache.org/licenses/LICENSE-2.0
  *
- * This file is part of the Apache Pekko project, derived from Akka.
+ * This file is part of the Apache Pekko project, which was derived from Akka.
  */
 
 package org.apache.pekko.actor.testkit.typed.javadsl;
+
 
 import java.lang.annotation.*;
 
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/Junit5TestKit.java
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/Junit5TestKit.java
@@ -1,0 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+package org.apache.pekko.actor.testkit.typed.javadsl;
+
+import java.lang.annotation.*;
+
+@Documented
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public  @interface Junit5TestKit {
+}
+
+
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
@@ -12,7 +12,7 @@ package org.apache.pekko.actor.testkit.typed.javadsl
 import org.junit.jupiter.api.extension.{ AfterAllCallback, BeforeTestExecutionCallback, ExtensionContext }
 import org.junit.platform.commons.support.AnnotationSupport
 
-import scala.jdk.CollectionConverters._
+import scala.jdk.CollectionConverters.CollectionHasAsScala
 
 final class TestKitJunit5Extension() extends AfterAllCallback with BeforeTestExecutionCallback {
 
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+package org.apache.pekko.actor.testkit.typed.javadsl
+
+import org.apache.pekko
+import org.junit.jupiter.api.extension.{AfterAllCallback, BeforeTestExecutionCallback, ExtensionContext}
+import org.junit.platform.commons.support.AnnotationSupport
+
+import scala.jdk.CollectionConverters._
+
+final class TestKitJunit5Extension() extends AfterAllCallback with BeforeTestExecutionCallback {
+
+  var testKit: Option[ActorTestKit] = None
+
+
+  /**
+   * Get a reference to the field annotated with @Junit5Testkit  [[pekko.actor.testkit.typed.javadsl.Junit5TestKit]]
+   *
+   */
+  override def beforeTestExecution(context: ExtensionContext): Unit =  {
+
+    context.getTestInstance.ifPresent(instance => {
+      val fielValue = AnnotationSupport.findAnnotatedFieldValues(instance, classOf[Junit5TestKit]).asScala.toList.head
+      testKit = Some(fielValue.asInstanceOf[ActorTestKit])
+    })
+  }
+
+
+  /**
+   * Shutdown testKit
+   */
+  override def afterAll(context: ExtensionContext): Unit = {
+     testKit.get.shutdownTestKit()
+  }
+
+}
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
@@ -23,9 +23,14 @@ final class TestKitJunit5Extension() extends AfterAllCallback with BeforeTestExe
    */
   override def beforeTestExecution(context: ExtensionContext): Unit = {
 
-    context.getTestInstance.ifPresent((instance: AnyRef)  => {
-      val fielValue = AnnotationSupport.findAnnotatedFieldValues(instance, classOf[Junit5TestKit]).asScala.toList.head
-      testKit = Some(fielValue.asInstanceOf[ActorTestKit])
+    context.getTestInstance.ifPresent((instance: AnyRef) => {
+      val annotations = AnnotationSupport.findAnnotatedFieldValues(instance, classOf[Junit5TestKit])
+      if (annotations.isEmpty) {
+        throw new IllegalArgumentException("Could not find field annotated with @Junit5TestKit")
+      } else {
+        val fieldValue = annotations.asScala.toList.head
+        testKit = Some(fieldValue.asInstanceOf[ActorTestKit])
+      }
     })
   }
 
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
@@ -9,9 +9,10 @@
 
 package org.apache.pekko.actor.testkit.typed.javadsl
 
-import org.junit.jupiter.api.extension.{ AfterAllCallback, BeforeTestExecutionCallback, ExtensionContext }
+import org.junit.jupiter.api.extension.{AfterAllCallback, BeforeTestExecutionCallback, ExtensionContext}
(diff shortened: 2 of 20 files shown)


project/Dependencies.scala
```diff
@@ -24,6 +24,8 @@ object Dependencies {
     .withRank(KeyRanks.Invisible) // avoid 'unused key' warning
 
   val junitVersion = "4.13.2"
+  val junit5Version = "5.10.0-M1"
```
pjfanning, line 27: can you use an actual release version instead of a pre-release milestone?

pjfanning:
thanks @thmue - this looks useful.

It does seem a little strange to not support junit5 in the legacy testkit and only in the typed testkit.

pjfanning (changes requested):
would it be possible to run scalafmt on the code? We require all PRs to be formatted correctly. The CI build will probably fail.

project/Dependencies.scala
thmue: yes, i changed it to 5.9.3

thmue:
thank you for your review, i updated the commit after running scalafmt.

Unfortunately i was not aware of the legacy testkit. In case this is not (that much) deprecated/legacy and if you find it useful to support it, i am happy to try to add support to it. 
However,  i don't see any test framework support in the subproject [pekko-testkit].

actor-testkit-typed/src/test/java/jdocs/org/apache/pekko/actor/testkit/typed/javadsl/LogCapturingExtensionExampleTest.java
```diff
@@ -0,0 +1,49 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
+ */
+
+/*
+ * Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
```
pjfanning, line 11: Is there a reason to add a Lightbend license header? If this is a copy/paste of another class, can you include a comment with a URL to that class.

Likewise for all the classes in this PR that you added Lightbend license headers to.

thmue: ah sorry, i guess i just copied the classes

pjfanning:
The CI build is strict about unused imports.

```
[06-22 09:13:36.548] [error] /home/runner/work/incubator-pekko/incubator-pekko/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala:12:19: Unused import
[06-22 09:13:36.549] [error] import org.apache.pekko
```

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
```diff
@@ -9,8 +9,7 @@
 
 package org.apache.pekko.actor.testkit.typed.javadsl
 
-import org.apache.pekko
-import org.junit.jupiter.api.extension.{ AfterAllCallback, BeforeTestExecutionCallback, ExtensionContext }
+import org.junit.jupiter.api.extension.{AfterAllCallback, BeforeTestExecutionCallback, ExtensionContext}
```
pjfanning, line 12: this is now a scalfmt issue - the removal of the spaces in this import

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/Junit5TestKit.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, derived from Akka.
```
pjfanning, line 7: Could you rebase this PR? Today, I've changed the license headers in most of our files and you should us the new header.

Line 7 here is replaced with:
```
This file is part of the Apache Pekko project, which was derived from Akka.
```

pjfanning:
the build is still failing because of missing apache headers in new source files - https://github.com/apache/incubator-pekko/actions/runs/5365713746/jobs/9754413547

use `sbt headerCreate` to add the missing headers

pjfanning:
also Scala 3 compile issue -> https://github.com/apache/incubator-pekko/actions/runs/5379239855/jobs/9760373110?pr=445

needs to be `(instance: T) =>` -- you'll need to replace the `T` with the correct type.

Don't use https://github.com/scala/scala-collection-compat. Try this

```
import org.apache.pekko
import pekko.util.ccompat._
```

ccompat is short for collection-compat.

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
```diff
@@ -23,7 +23,7 @@ final class TestKitJunit5Extension() extends AfterAllCallback with BeforeTestExe
    */
   override def beforeTestExecution(context: ExtensionContext): Unit = {
 
-    context.getTestInstance.ifPresent(instance => {
+    context.getTestInstance.ifPresent((instance: AnyRef)  => {
       val fielValue = AnnotationSupport.findAnnotatedFieldValues(instance, classOf[Junit5TestKit]).asScala.toList.head
```
pjfanning, line 27: could you fix the val name? - should be `fieldValue`

also calling .head on a list is dangerous - could you check that there is actually something in the list first?

thmue: fixed it

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/Junit5TestKit.java
```diff
@@ -0,0 +1,21 @@
...
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+package org.apache.pekko.actor.testkit.typed.javadsl;
+
+
```
pjfanning, line 12: this file should be moved to `src/java/...` - likewise with all java files, none should be in `src/scala/...`

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala
pjfanning: you don't appear to have to fixed the issue with the unguarded call to `.head`

pjfanning: actually, I see the isEmpty check in https://github.com/apache/incubator-pekko/pull/445/files#diff-a3b630cebfd1e64c3ee9e76b844d1e05a68e8f0e4712ba3e502d1088e8daec1dR28

pjfanning:
@thmue the timing of this PR is a little unfortunate. We are trying to get 1.0.0 released and this is not going in it.

We can merge this change later but I would still like to get more opinions on it. To me, supporting junit5 in just one testkit is a bit strange but other users might disagree.

One technical thing. The source code header you've added to the new files - that code is new and not Akka derived. So, the right header to use is arguably the one in https://www.apache.org/legal/src-headers.html

thmue:
@pjfanning thank you for your efforts regarding this PR and of course, the much anticipated 1.0 release is much more important than this PR.  :-)


Currently, there are two junit5 extensions, the LogCapturingExtension and the TestKitJunit5Extension.
Both extensions can be used in a junit5 test, instead of the  junit4 @Rule (TestKitJunitResource.class and LogCapturing.class) annotations. 
In the akka docs i could only find these classes for junit4.
If other junit5 integrations are necessary, it would be great if you could point me in the right direction.

project/Dependencies.scala
```diff
@@ -24,6 +24,8 @@ object Dependencies {
     .withRank(KeyRanks.Invisible) // avoid 'unused key' warning
 
   val junitVersion = "4.13.2"
+  val junit5Version = "5.9.3"
```
pjfanning, line 27: can you use junit 5.10.0?

A rebase of the PR would be useful too. We are now in a position to be able to merge v1.1.0 target PRs.

pjfanning:
@He-Pin @mdedetrich what do you think? It seems ok to add Junit5 support since there is already Junit4 support. Oddly, it appears that other pekko testkits don't have explicit junit support - even if we do want junit5 support for these other testkit modules, they can be added by later PRs.

mdedetrich (changes requested):
Can we rename all instances (i.e. public/visible classes such as `Junit5TestKitBuilder`) of `Junit5` to `JUnit5`? `JUnit5` is the correct capitlisation of the word.

pjfanning:
> Can we rename all instances (i.e. public/visible classes such as `Junit5TestKitBuilder`) of `Junit5` to `JUnit5`? `JUnit5` is the correct capitlisation of the word.

We already have https://github.com/apache/incubator-pekko/blob/85c0df12a025920658b638d8a5f02993f6620c31/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunitResource.scala so it would be more consistent to stick with `Junit`.

mdedetrich (changes requested):
Some other things I noticed

mdedetrich:
> We already have https://github.com/apache/incubator-pekko/blob/85c0df12a025920658b638d8a5f02993f6620c31/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunitResource.scala so it would be more consistent to stick with `Junit`.

Since we are on v1.1.x, I am wondering whether we should deprecate the old class name with `Junit` and create aliases for new `JUnit` classes?

pjfanning:
> > We already have https://github.com/apache/incubator-pekko/blob/85c0df12a025920658b638d8a5f02993f6620c31/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunitResource.scala so it would be more consistent to stick with `Junit`.
> 
> Since we are on v1.1.x, I am wondering whether we should deprecate the old class name with `Junit` and create aliases for new `JUnit` classes?

I don't think our users care too much about Junit vs JUnit branding - it just seems easier to me to stck with Junit since we have prior art.

pjfanning:
Code doesn't compile with Scala 3.3.0.

```
[error] -- [E007] Type Mismatch Error: /home/runner/work/incubator-pekko/incubator-pekko/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJunit5Extension.scala:33:38 
[error] 33 |    context.getTestInstance.ifPresent((instance: AnyRef) => {
[error]    |                                      ^
[error]    |                      Found:    AnyRef => Unit
[error]    |                      Required: java.util.function.Consumer[? >: Object]
```

mdedetrich:
> I don't think our users care too much about Junit vs JUnit branding - it just seems easier to me to stick with Junit since we have prior art.

Sure but its a low hanging improvement PR, but this isn't something I will block the PR over (its a separate issue anyways).

mdedetrich:
> Code doesn't compile with Scala 3.3.0.

Its probably a small type inference thing, I will look into it today.

actor-testkit-typed/src/test/java/jdocs/org/apache/pekko/actor/testkit/typed/javadsl/GreeterMain.scala
```diff
@@ -0,0 +1,61 @@
...
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+package jdocs.org.apache.pekko.actor.testkit.typed.javadsl
+
+import org.apache.pekko.actor.typed.{ ActorRef, Behavior }
```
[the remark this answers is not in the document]
thmue: changed it to a normal class

mdedetrich:
@thmue Can you rebase your the PR with the new changes coming in from `project/Dependencies.scala`. After that I will do another look and hopefully we can get this merged.

I am also still contemplating whether we should do https://github.com/apache/incubator-pekko/pull/445#issuecomment-1657198849, i.e. rename the class so it has proper naming and deprecate the old name (I have a preference for doing this since its a nice/easy improvement).

thmue:
i rebased it, unfortunately i had a lot of build problems. However these could be due to some local setup problems (intellij / scala  etc....)

PR #447, opened 2023-06-22
tidy up scripts

pjfanning:
the script files are coming up as something that might block us doing a release

https://lists.apache.org/thread/pfjf5qwhq9ydj91s3kc1055qgbzfqsv1

It's possible that we could remove every file in the https://github.com/apache/incubator-pekko/tree/main/scripts directory as I don't think we use them, in practice. That said, some of them look potentially useful and there is an argument that in most cases, we don't need to add Apache headers because Pekko team haven't made material changes to these files.

Exceptions in this PR are arbitrary.
* git-remove-history.sh has an author mentioned in it. This raises a can of worms about us having to seek out the original code to see about its licensing - I suggest removing it is easier
* github-tagging.sh - we tag manually as part of https://github.com/apache/incubator-pekko-site/wiki/Pekko-Release-Process - and this script looks like it is meant to be called from other scripts but we don't appear to have any scripts left that call this
* compile_etags_file.sh calls a `.bat` script that is missing


nvollmar (approved):
lgtm

Issue #430, opened 2023-06-20
Allow building on jdk19


raboof:
Building on jdk19 currently fails because of a newly added deprecation in `sun.misc.Unsafe`:

```
[error] actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala:350:41: method objectFieldOffset in class Unsafe is deprecated
[error]   private[this] val taskOffset = unsafe.objectFieldOffset(classOf[TaskHolder].getDeclaredField("task"))
```

Perhaps we can exclude this warning (until #417 is in place), though for some reason `-Wconf:msg=method objectFieldOffset in class Unsafe is deprecated:silent` didn't seem to be sufficient.

He-Pin:
Maybe just need to add a `@nowarn("msg=deprecated")`?

pjfanning:
So, after #467 it appears you can compile and test Pekko with JDK 20.

Paradox plugin - for building reference docs - fails with

```
[error] java.lang.RuntimeException: Error creating extended parser class: Could not determine whether class 'org.pegdown.ParserWithDirectives$$parboiled' has already been loaded
[error] 	at org.parboiled.Parboiled.createParser(Parboiled.java:58)
[error] 	at com.lightbend.paradox.markdown.Reader.<init>(Reader.scala:40)
[error] 	at com.lightbend.paradox.sbt.ParadoxPlugin$.$anonfun$baseParadoxSettings$2(ParadoxPlugin.scala:83)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[error] 	at sbt.Execute.work(Execute.scala:291)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
[...]

PR #455, opened 2023-06-29
+workflow Add JDK 20 to java nightly build matrix.

He-Pin:
Add JDK 20, which is the latest java release

refs: https://github.com/apache/incubator-pekko/issues/430

### DIFF
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -143,7 +143,7 @@ jobs:
         # binary version is required and Akka build will set the right
         # full version from it.
         scalaVersion: ["2.12", "2.13"]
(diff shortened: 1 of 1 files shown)


pjfanning:
@He-Pin Have you run this locally? We have a link on our home page about the nightly builds and I don't want them to start failing. If it is failing, we could create a separate workflow.

He-Pin:
> @He-Pin Have you run this locally? We have a link on our home page about the nightly builds and I don't want them to start failing. If it is failing, we could create a separate workflow.

Will need to add some exports.

pjfanning:
I wasn't trying to kill this PR off - just wanted the JDK 20 to be done in a separate workflow - until we get the build to work

I have created #464

PR #456, opened 2023-06-30
setup CI job to publish 1.0 docs

pjfanning:
* publishes docs for Pekko 1.0
* matches https://doc.akka.io/docs/akka/2.6/typed/index.html
  * the version number appears in HTML pages is the full x.y.z version (Akka 2.6.21; Pekko 1.0.0)
  * the version number in URL path is just x.y (Akka 2.6; Pekko 1.0)
* we need a GitHub CI job because the rsync needs secrets and I don't know them
* we can run the job off a branch after we complete the 1.0.0 release 
* we don't need to run the job every night after we are happy the docs are stable

### DIFF
--- a/.github/workflows/publish-1.0-docs.yml
+++ b/.github/workflows/publish-1.0-docs.yml
@@ -58,10 +58,10 @@ jobs:
       # Create directory structure upfront since rsync does not create intermediate directories otherwise
       - name: Create nightly directory structure
         run: |-
-          mkdir -p target/nightly-docs/docs/pekko/1.0/
-          mv docs/target/paradox/site/main/ target/nightly-docs/docs/pekko/1.0/docs
-          mv target/scala-2.13/unidoc target/nightly-docs/docs/pekko/1.0/api
-          mv target/javaunidoc target/nightly-docs/docs/pekko/1.0/japi
+          mkdir -p target/nightly-docs/docs/pekko/1.0.0/
+          mv docs/target/paradox/site/main/ target/nightly-docs/docs/pekko/1.0.0/docs
+          mv target/scala-2.13/unidoc target/nightly-docs/docs/pekko/1.0.0/api
+          mv target/javaunidoc target/nightly-docs/docs/pekko/1.0.0/japi
 
       - name: Upload nightly docs
         uses: ./.github/actions/sync-nightlies
--- a/.github/workflows/publish-1.0-docs.yml
+++ b/.github/workflows/publish-1.0-docs.yml
@@ -1,0 +1,76 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Based on Apache Arrow's java-nightly workflow
+# https://github.com/apache/arrow/blob/master/.github/workflows/java_nightly.yml
+name: Publish 1.0 documentation
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  publish-nightly:
+    name: Publish 1.0 docs
+    runs-on: ubuntu-20.04
+    if: github.repository == 'apache/incubator-pekko'
+    steps:
+      # TODO we will need to change to a 1.0 branch or a release tag
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Setup Java 11
+        uses: actions/setup-java@v3
+        with:
+          distribution: temurin
+          java-version: 11
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v6
+
+      - name: Install Graphviz
+        run: |-
+          sudo apt-get install graphviz
+
+      # TODO come up with a better way to control the version, possibly based on git tags
+      - name: Build Documentation
+        run: |-
+          sbt -Dpekko.genjavadoc.enabled=true "set ThisBuild / version := \"1.0.0\"; docs/paradox; unidoc"
+
+      # Create directory structure upfront since rsync does not create intermediate directories otherwise
+      - name: Create nightly directory structure
+        run: |-
+          mkdir -p target/nightly-docs/docs/pekko/1.0/
+          mv docs/target/paradox/site/main/ target/nightly-docs/docs/pekko/1.0/docs
+          mv target/scala-2.13/unidoc target/nightly-docs/docs/pekko/1.0/api
+          mv target/javaunidoc target/nightly-docs/docs/pekko/1.0/japi
+
+      - name: Upload nightly docs
+        uses: ./.github/actions/sync-nightlies
+        with:
+          upload: true
+          switches: --archive --compress --update --delete --progress --relative
+          local_path: target/nightly-docs/./docs/pekko/ # The intermediate dot is to show `--relative` which paths to operate on
+          remote_path: ${{ secrets.NIGHTLIES_RSYNC_PATH }}/pekko/
+          remote_host: ${{ secrets.NIGHTLIES_RSYNC_HOST }}
+          remote_port: ${{ secrets.NIGHTLIES_RSYNC_PORT }}
+          remote_user: ${{ secrets.NIGHTLIES_RSYNC_USER }}
+          remote_key: ${{ secrets.NIGHTLIES_RSYNC_KEY }}


.github/workflows/publish-1.0-docs-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Based on Apache Arrow's java-nightly workflow
+# https://github.com/apache/arrow/blob/master/.github/workflows/java_nightly.yml
+name: Publish nightly 1.0 documentation
```
mdedetrich, line 20: Is this publishing for 1.0 or nightly? Name of the job is confusing.

.github/workflows/publish-1.0-docs-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
```
mdedetrich, line 25: If we are publishing docs for 1.0 it shouldn't be done via a cron job

.github/workflows/publish-1.0-docs-nightly.yml
pjfanning: * it's nightly because it runs on a cron schedule (once a day at night time)
* I could rename it as 'daily'

.github/workflows/publish-1.0-docs-nightly.yml
pjfanning: I can live with getting rid of the cron schedule. Right now, it is more convenient to run it daily.

There would be no plan to run it daily indefinitely - ie the cron schedule could be removed later.

.github/workflows/publish-1.0-docs-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  publish-nightly:
+    name: Publish nightly
+    runs-on: ubuntu-20.04
+    if: github.repository == 'apache/incubator-pekko'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
mdedetrich, line 39: There is likely a variable here that allows you to specify a tag to checkout the repo.

jrudolph:
> This may not even be necessary, afaik the rsync credentials are just your apache ID so its likely much easier to do this locally?

I think I checked that before and it did not work.

.github/workflows/publish-1.0-docs-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+          sudo apt-get install graphviz
+
+      - name: Build Documentation
+        run: |-
+          sbt -Dpekko.genjavadoc.enabled=true "set ThisBuild / version := \"1.0.0\"; docs/paradox; unidoc"
+
+      # Create directory structure upfront since rsync does not create intermediate directories otherwise
+      - name: Create nightly directory structure
+        run: |-
+          mkdir -p target/nightly-docs/docs/pekko/1.0/
```
mdedetrich, line 61: We need to also upload the docs to `current` and also prevent the current publish doc job from pushing to current. Currently its latest snapshot pointing to `current` but it should be changed to latest release version @jrudolph can help out here.

.github/workflows/publish-1.0-docs-nightly.yml
jrudolph: We can run it on `push` for the final docs branch. In the meantime, it might be enough to run it on `workflow_dispatch`.

.github/workflows/publish-1.0-docs-nightly.yml
mdedetrich: From https://github.com/actions/checkout#usage its called `ref` so you can just do `ref: 1.0.0` and it will check out pekko at that point in time.

pjfanning:
> The design of the implementation has some issues. If we are going to publish docs for 1.0 it should be done by checking out pekko via a tag at precisely version 1.0.0. It would be even better if the version can be specified as an input (see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs) so we can reuse this workflow for later (i.e. we can manually upload any docs for any version).
> 
> This may not even be necessary, afaik the rsync credentials are just your apache ID so its likely much easier to do this locally?

* In the issue description, I suggest that this job can be changed to run off a branch after the 1.0.0 release is done.
* I intend to start a dev mailing list discussion about the branching strategy after we get the 1.0.0-RC3 testing done and the vote moves over to Incubator PMC
  * generally, I'm going to suggest that we create a 'branch-1.0' and that 'main' be used for forward dev (basically the '1.1.0' or '2.0.0' release which allows some changes in Pekko)
  * while 'branch-1.0' would just receive a small number of backports from 'main' for bug fixes and '1.0.1' release would be made from that branch
* this yml can be changed after we have decided the branching strategy
* it's useful to have 1.0 docs now even before the release
  * the links in pekko-http docs are broken because of the use of pekko 1.0.0-RC3 - pekko-http uses the pekko version number to work out the link and we have no 1.0 docs to target

mdedetrich:
I understand the point about branching (btw we should be using `1.0.x` as this is standard convention for branch names in Scala OS projects) but branches are both orthogonal and overkill for doc publishing. Publishing docs off branches is not how it should be done because branches can have unreleased changes. What we are aiming for (which is also what is standard for both Akka and Scala OS projects) is that you have a URI path reference to every version (i.e. `docs/1.0.0/`, `docs/1.0.1/` `docs/1.1.0/`) and a URI path reference for current that points to the latest released version, **NOT** branch (i.e. `docs/current/` which would point to `docs/1.1.0/` with my previous example).

If we want to publish the docs right now with minimal effort, the easiest would be to just use the `ref` variable in the checkout action to point to a tag (can be v1.0.0-RC3 or v1.0.0) with a manual github actions trigger. Creating a github input for this would be nicer as we don't have to make modifications to the git branch for every version but hardcoding the version works for now (and because its a manual trigger we can put in `v1.0.0` as a ref now and then manually trigger it once we release 1.0.0)

.github/workflows/publish-1.0-docs-nightly.yml
pjfanning: we will use .htaccess to direct current to 1.0 -- see https://github.com/apache/incubator-pekko-site/blob/main/src/main/paradox/.htaccess -- this can be modified

pjfanning:
> The only thing that needs to be figured out is how `current` is currently implemented (pun intended!), iirc @jrudolph did this.

See  https://github.com/apache/incubator-pekko-site/blob/main/src/main/paradox/.htaccess

mdedetrich:
> > The only thing that needs to be figured out is how `current` is currently implemented (pun intended!), iirc @jrudolph did this.
> 
> See https://github.com/apache/incubator-pekko-site/blob/main/src/main/paradox/.htaccess

In that case wrt `current` I am happy if we do the easiest thing which is just manually point `current` to `1.0.0` and we can come up with a better solution later

pjfanning:
We currently have a code freeze in main branch. No changes should go into main in next few days that will affect the docs in a negative way and if it does, it will need to be fixed.

This PR is just to start publishing some docs so that we can fix our pekko-http docs - with all the pekko links being broken in pekko-http docs right now.

This is not a permanent solution. This yml can very easily be changed as we go along.

Once we branch for 1.0, why we would we have any doc changes that should not be visible to users? 1.0 branch is meant to be stable - no breaking changes in it.

mdedetrich:
> Once we branch for 1.0, why we would we have any doc changes that should not be visible to users? 1.0 branch is meant to be stable - no breaking changes in it.

Lets assume we have a branch `1.0.x` and someone creates a PR against `1.0.x` for a new feature which requires additional documentation. The feature gets merged into `1.0.x` but its not released yet. If doc publishing is based off branches then we would publish docs for a feature that isn't even released yet.

Branches are also mutable and change over time which would make any kind of doc publishing github actions job brittle, unlike tags which we deliberately designed to be as much as possible to be immutable.

> We currently have a code freeze in main branch. No changes should go into main in next few days that will affect the docs in a negative way and if it does, it will need to be fixed.

Yes, and what this encourages is for people to check the docs properly when a release is being made. Its slightly unfortunate that https://github.com/apache/incubator-pekko-sbt-paradox/pull/58 didn't land before the code freeze but I am happy to make an exception for that (not sure how official this would be though).

.github/workflows/publish-1.0-docs-nightly.yml
pjfanning: I removed the cron and renamed the file/job

.github/workflows/publish-1.0-docs-nightly.yml
pjfanning: I removed the cron

pjfanning:
> > Once we branch for 1.0, why we would we have any doc changes that should not be visible to users? 1.0 branch is meant to be stable - no breaking changes in it.
> 
> Lets assume we have a branch `1.0.x` and someone creates a PR against `1.0.x` for a new feature which requires additional documentation. The feature gets merged into `1.0.x` but its not released yet (and realistically speaking its not uncommon to have unreleased features in branches for many months). If doc publishing is based off branches then we would publish docs for a feature that isn't even released yet, confusing people.
> 
> Branches are also mutable and change over time which would make any kind of doc publishing github actions job brittle, unlike tags which we deliberately designed to be as much as possible to be immutable.
> 
> > We currently have a code freeze in main branch. No changes should go into main in next few days that will affect the docs in a negative way and if it does, it will need to be fixed.
> 
> Yes, and what this encourages is for people to check the docs properly when a release is being made. Its slightly unfortunate that [apache/incubator-pekko-sbt-paradox#58](https://github.com/apache/incubator-pekko-sbt-paradox/pull/58) didn't land before the code freeze but I am happy to make an exception for that (not sure how official this would be though).

We can adjust this job when we start making merges to the 1.0 branch that include doc changes that are breaking.

When we discuss the branching strategy on the mailing list, I will be arguing against adding features in a patch release.

mdedetrich:
> When we discuss the branching strategy on the mailing list, I will be arguing against adding features in a patch release.

Its not just features (thats just an example), this holds for any kind of documentation change, and if you are at the point of trying to manually distinguish "breaking changes" at this point you may as well use git tags. This should be discussed on mailing list and not here, but I would strongly recommend to do what is standard for Akka/Scala OS (since that is what almost all of Pekko users are used to) rather than changing the status quo. As a bonus, using git tags is also far easier.

pjfanning:
I firmly believe that for the next few months, we are more likely to be fixing bugs in the docs and that these doc fixes will be something that we will want to release to the 1.0 docs immediately.

The chances that we will be adding new features to the 1.0 docs in the short term is very low.

The pragmatic solution for now is to have a branch and to build the docs off the branch. When we get to having our first doc change that we don't want to publish yet, we can adjust our strategy.

For instance, we don't yet have a 1.0-snapshot doc publish - just a main-snapshot doc publish. When we start adding new features to 1.0, we will need the 1.0-snapshot doc publish job too.

mdedetrich:
> For instance, we don't yet have a 1.0-snapshot doc publish - just a main-snapshot doc publish. When we start adding new features to 1.0, we will need the 1.0-snapshot doc publish job too.

We don't need to do this, we currently have snapshot docs being published nightly which automatically handles the version as well as `snapshot` URI path reference. This doesn't need to change at all.

pjfanning:
> > For instance, we don't yet have a 1.0-snapshot doc publish - just a main-snapshot doc publish. When we start adding new features to 1.0, we will need the 1.0-snapshot doc publish job too.
> 
> We don't need to do this, we currently have snapshot docs being published nightly which automatically handles the version as well as `snapshot` URI path reference. This doesn't need to change at all.

that job is running against the 'main' branch - you can use run manually and choose the branch though

we would probably need to add a .htaccess redirect when we have 1.0-snapshot publishes - we don't have a redirect yet

.github/workflows/publish-1.0-docs.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v6
+
+      - name: Install Graphviz
+        run: |-
+          sudo apt-get install graphviz
+
+      - name: Build Documentation
+        run: |-
+          sbt -Dpekko.genjavadoc.enabled=true "set ThisBuild / version := \"1.0.0\"; docs/paradox; unidoc"
```
mdedetrich, line 54: Can we just leave a comment here stating this is temporary. Will approve the PR once this is done.

mdedetrich:
> that job is running against the 'main' branch - you can use run manually and choose the branch though

> we would probably need to add a .htaccess redirect when we have 1.0-snapshot publishes - we don't have a redirect yet

Yes, which is how it should be. snapshots docs are expensive to maintain (each doc release is 500mb) so we should only be publishing docs for snapshot for the latest Pekko release branch which will be `main`. This is also what Akka does, see https://doc.akka.io/docs/akka/snapshot/typed/index.html which points to the latest snapshot (currently `2.8.2+35-be1670b9-SNAPSHOT`). There is no other Akka snapshot docs for any other branch being published, its just always the latest snapshot on `main`.

.github/workflows/publish-1.0-docs.yml
pjfanning: I've added some TODO comments

mdedetrich: You didn't push the commit?

pjfanning: I did - see https://github.com/pjfanning/incubator-pekko/tree/1.0.0-docs

But GitHub seems to be acting up - I'm getting error pages popping up - and so far the PR does not reflect my most recent commit

mdedetrich (approved):
lgtm

jrudolph:
My assumptions on the general plan are:

 * docs for each release (even patch releases) are kept available
 * we must be able to make changes to docs after a release for various reasons (but it might not happen all to often that this is necessary)
 * The `current` URL to docs/api points to the latest released version

One way to achieve that is by adopting a structure similar to the Akka one, where every doc version is kept in its own directory. The `current` version could be updated by changing the `.htaccess` link to the latest released version (manually or automatically).

The docs publishing could work by running on a created tag and then automatically figuring out the place where to publish to. In addition to that we can establish a branch name standard e.g. (v1.0.0-docs) that can be picked up by the job which configures the docs building so that it assumes it is version 1.0.0 and overwrites the previous version of the docs. This is somewhat dangerous but could be reverted by rerunning the same job on an older version. (We could also published named revisions for existing versions like putting it into a new directory `1.0.0-docs-v2` or so and then fix everything up using `.htaccess` or similar but that seems like overkill.

pjfanning:
@jrudolph Akka only seems to have version docs for '2.6' and not for '2.6.21'

Is this something we should follow? I think I prefer '2.6' to '2.6.21' myself.

For the rare cases where a version like '2.6.22' has something new that needs documenting, the '2.6' docs could just say that the something new (a feature or a new config, etc.) is only available since 2.6.22.

mdedetrich:
Also on the point of "never merging breaking changes in patch versions" this is generally impossible because its not just breaking changes, its actually any real change in the codebase that the docs reference. Paradox has `@@snip` directives that pull in actual source code/configuration files (i.e. `reference.conf`) so if any of those changes it will again reference things that are not released yet.

And there is actual proof/evidence of this, i.e. https://github.com/apache/incubator-pekko/pull/385. In that PR the `reference.conf` was updated which is directly copied into the docs (see https://github.com/apache/incubator-pekko/blob/59897a642503033a7ad53a7ab108da34397dd0ed/docs/src/main/paradox/general/configuration-reference.md#L17). Such a change in the future would be made in a patch release.

While there is an argument to be made about why our code is so tightly coupled to configuration/code, the benefits are huge (i.e. we don't have problems with docs being out of sync/incorrect).

jrudolph:
> @jrudolph Akka only seems to have version docs for '2.6' and not for '2.6.21'

I guess they pruned old versions but for the latest ones they always keep all, e.g. https://doc.akka.io/docs/akka/2.8.2/

mdedetrich:
> I guess they pruned old versions but for the latest ones they always keep all, e.g. https://doc.akka.io/docs/akka/2.8.2/

Correct, i.e. see https://doc.akka.io/docs/akka/2.8.1/ (it exists and even warns you that its not the latest version).

.github/workflows/publish-1.0-docs.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      # TODO come up with a better way to control the version, possibly based on git tags
+      - name: Build Documentation
+        run: |-
+          sbt -Dpekko.genjavadoc.enabled=true "set ThisBuild / version := \"1.0.0\"; docs/paradox; unidoc"
+
+      # Create directory structure upfront since rsync does not create intermediate directories otherwise
+      - name: Create nightly directory structure
+        run: |-
+          mkdir -p target/nightly-docs/docs/pekko/1.0/
```
mdedetrich, line 61: I just noticed but this needs to be `1.0.0`, not `1.0`

mdedetrich (changes requested):
Missed one more thing

pjfanning:
with #385, we could have insisted that the reference.conf change mentioned the version the change became live in

* we don't have any scripting to tell someone they are on an old copy of the docs
* having 2.8.2 and 2.8.1 docs that are subtly different - how does someone know there are diffs?
* I honestly think it is better to have just 2.8 docs and for those docs to try their best to point out when changes affect 2.8.2+ (and that the item being documented did not apply before v2.8.2)

.github/workflows/publish-1.0-docs.yml
pjfanning: that's why I brought up the 2.6 example

mdedetrich:
> with #385, we could have insisted that the reference.conf change mentioned the version the change became live in
> 
> * we don't have any scripting to tell someone they are on an old copy of the docs
> * having 2.8.2 and 2.8.1 docs that are subtly different - how does someone know there are diffs?
> * I honestly think it is better to have just 2.8 docs and for those docs to try their best to point out when changes affect 2.8.2+ (and that the item being documented did not apply before v2.8.2)

Lets just for now copy what is familiar for Akka/Scala folks (i.e. users) and we can decide later if we want to changes in the proper way (i.e. mailing list with a possible vote). There is already enough reasoning why having docs for patch versions makes sense and now is not the time to change the status quo.

jrudolph:
> Is this something we should follow? I think I prefer '2.6' to '2.6.21' myself.

It depends a bit on how we use our versioning scheme. Akka never updated from 2.x and Akka HTTP never from 10.x. We are probably in the same situation that a Pekko `2.x` is not going to happen any time soon. So, what you are suggesting is that we only keep a single version of the docs for the time being (in `1.0`). I'm not completely opposed to it, if it simplifies things.

> For the rare cases where a version like '2.6.22' has something new that needs documenting, the '2.6' docs could just say that the something new (a feature or a new config, etc.) is only available since 2.6.22.

That makes a lot of sense and follows precedence from all kinds of projects.

Assuming we only keep a single 1.0 docs folder for released versions, the question still remains how we update it between versions when necessary. We would probably still just keep a `1.0-docs` branch into which the latest version is merged during a release.

mdedetrich:
> Assuming we only keep a single 1.0 docs folder for released versions, the question still remains how we update it between versions when necessary. We would probably still just keep a 1.0-docs branch into which the latest version is merged during a release.

There is a much easier way to solve this issue and it doesn't require branches at all, just modify this github action workflow to have 2 inputs, one is a git ref and the other is the desired version and this will essentially allow you to just override any version of docs that points to any git reference (which can even just be a temporary branch, or a doc branch or a specific commit).

jrudolph:
> There is already enough reasoning why having docs for patch versions makes sense and now is not the time to change the status quo.

The status quo actually is that few people know that all the versions are available. If they know, it's likely because they landed there accidentally from a search engine (and it wasn't what they wanted...).

jrudolph:
> There is a much easier way to solve this issue and it doesn't require branches at all, just modify this github action workflow to have 2 inputs, one is a git ref and the other is the desired version and this will essentially allow you to just override any version of docs that points to any git reference (which can even just be a temporary branch, or a doc branch or a specific commit).

But why would you want to publish from an anonymous ref? That just makes things so much more confusing in case you want to figure out where changes came from.

mdedetrich:
> > There is already enough reasoning why having docs for patch versions makes sense and now is not the time to change the status quo.
> 
> The status quo actually is that few people know that all the versions are available. If they know, it's likely because they landed there accidentally from a search engine (and it wasn't what they wanted...).

All I can say is that my personal experience differs and having separately published docs for patch versions is incredibly useful because I have a guarantee that if I have akka/pekko version `x` in my project, that the docs published for that version `x` precisely only contains the info for that version `x`, nothing more and nothing less.

pjfanning:
There is a small backlog of 1.1.0 focused PRs. So when we release 1.1.0, there would be a 1.1 set of docs independent from 1.0.

mdedetrich:
> > There is a much easier way to solve this issue and it doesn't require branches at all, just modify this github action workflow to have 2 inputs, one is a git ref and the other is the desired version and this will essentially allow you to just override any version of docs that points to any git reference (which can even just be a temporary branch, or a doc branch or a specific commit).
> 
> But why would you want to publish from an anonymous ref? That just makes things so much more confusing in case you want to figure out where changes came from.

This is on the assumption that we already have docs published for a version thats already been published (lets say `v1.1.0`) and the case for republishing already published versions is very rare. If we want to override the doc publishing for `v1.1.0` one can make a branch based off of the `v1.1.0` tag, bring in any additional doc related changes and just republish based off those changes.

You could use the official `v1.1.0` release branch, but that is problematic for the reasons I stated earlier.

.github/workflows/publish-1.0-docs.yml
mdedetrich: > that's why I brought up the Akka 2.6 example

Right but thats because Akka 2.6 is not the latest release branch, 2.8 is and for the 2.8 branch Akka does publish all patch versions as stated. In the case of Pekko 1.0 is our latest release branch

> If we insist on 1.0.0, we will probably need to discuss with infra team first to see if they are happy for us to use so much disk space. 1.0 strategy will use a lot less disk space than 1.0.x strategy.

We can but it shouldn't block us if we use the same strategy that Akka does, i.e. only do the patch doc publishing for the latest one (maybe 2) release branches.

mdedetrich:
> There is a small backlog of 1.1.0 focused PRs. So when we release 1.1.0, there would be a 1.1 set of docs independent from 1.0.

True although not sure what the context of bringing this up is (i.e. in any case this is a given)?

mdedetrich (approved):
lgtm

.github/workflows/publish-1.0-docs.yml
pjfanning: changed to 1.0.0

PR #459, opened 2023-06-30
modify rsync jobs for docs to try to stop them deleting each other's work

pjfanning:
the new publish-1.0-docs CI job deleted the 'main-snapshot' dir on the nightlies site
* note the `--delete` in the rsync command
* this delete is generally useful because we might remove doc files or rename them and we don't want to leave the old copy behind

I think this change will stop the 2 jobs interfering with each other
* the nightly job that publishes main-snapshot docs
* the on demand job that publishes 1.0.0 docs


mdedetrich (approved):
lgtm

PR #461, opened 2023-07-01
Also publish API version of docs

mdedetrich:
Not the most ideal solution, but it will work for now and we can look into fixing this later. This will publish docs into both `1.0.0` and `1.0` since other Pekko modules usually don't care about linking to patch versions (i.e. pekko-http).

Update to `.htaccess` is here https://github.com/apache/incubator-pekko-site/pull/12


pjfanning (approved):
lgtm

pjfanning:
@mdedetrich the CI run failed. It looks like the `cp` commands need `-r`

mdedetrich:
Already on it, see https://github.com/apache/incubator-pekko/pull/463

raboof:
> This will publish docs into both `1.0.0` and `1.0` since other Pekko modules usually don't care about linking to patch versions (i.e. pekko-http).

Sounds good - this is partly for SEO reasons, we want people using search engines to find information about classes to hopefully find the latest version of the class, and not some historical one.

PR #463, opened 2023-07-01
Fix recursive copy

mdedetrich:
cp needs `-r` for directories


He-Pin (approved):
mv will work too

mdedetrich:
> mv will work too

I had to change from cp to mv because is destructive (it moves the directory) and I need the originating directory multiple times.

PR #466, opened 2023-07-03
workaround-removal-biased-locking

pjfanning:
In JDK 20, JVM option `-XX:-UseBiasedLocking` fails the build.

Seems like a flag we can just remove from the sbtopts file.


pjfanning:
@raboof @jrudolph @mdedetrich any objection to just removing `-XX:-UseBiasedLocking` from the sbtopt file?

raboof:
No objection

jrudolph:
Yep, just drop it, probably premature optimization anyway...

mdedetrich:
All for it as well, in fact in general I wanted to remove the more esoteric JVM options (`-J-XX:+UnlockExperimentalVMOptions -J-XX:+UseJVMCICompiler` is another one I would like to remove, I have seen it in other Pekko projects and while it may provide improvements to compile speed I have noticed on many occasions it also throws spurious exceptions/warnings).

PR #467, opened 2023-07-03
disable -Werror when using Java 20

pjfanning:
* relates to https://github.com/apache/incubator-pekko/issues/430
* Java 19 is not LTS and Java 20 is latest non-LTS release
* the compiler warnings leading to build failing (-Werror does this) is the main problem with using a new JDK to build Pekko


He-Pin (approved):
lgtm

PR #471, opened 2023-07-06
Add testQuickUntilPassed

mdedetrich:
Adds a variant of `testQuick` called `testQuickUntilPassed` which runs `testQuick` in a loop until all tests pass. This is intended to be used by the release process to account for some tests being flaky

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ See https://pekko.apache.org for the documentation including the API docs. The d
     - `sbt +compile` will compile for all supported versions of Scala
 - `sbt test` will compile the code and run the unit tests
 - `sbt testQuick` similar to test but when repeated in shell mode will only run failing tests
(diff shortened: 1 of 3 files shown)


project/TestQuickUntilPass.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  override def trigger = allRequirements
+
+  object autoImport {
+    val testQuickUntilPassed = inputKey[Unit]("runs testQuick continuously until it passes")
+  }
+
+  import autoImport._
+
+  private def testQuickRecursive(input: String): Def.Initialize[Task[Unit]] = Def.taskDyn {
```
pjfanning, line 32: could be marked `@tailrec` ?

mdedetrich: Doesn't work because of `Def.taskDyn`

pjfanning (approved):
lgtm

PR #472, opened 2023-07-06
Set default timefactor for ActorsLeakSpec so it can pass locally

mdedetrich:
After doing some extensive testing (talking about 30+ test runs) it turns out that `ActorsLeakSpec` cannot pass locally unless you set `pekko.test.timefactor = 2` which is what is set in CI (see https://github.com/apache/incubator-pekko/blob/88bf6329f193eedd45091f4f9a515943bd8ecb23/.github/workflows/nightly-builds.yml#L168-L175).

### DIFF
--- a/remote/src/test/scala/org/apache/pekko/remote/classic/ActorsLeakSpec.scala
+++ b/remote/src/test/scala/org/apache/pekko/remote/classic/ActorsLeakSpec.scala
@@ -41,6 +41,7 @@ object ActorsLeakSpec {
        pekko.remote.classic.quarantine-after-silence = 3 s
        pekko.remote.use-unsafe-remote-features-outside-cluster = on
        pekko.test.filter-leeway = 12 s
+       pekko.test.timefactor = 2
        # test is using Java serialization and not priority to rewrite
        pekko.actor.allow-java-serialization = on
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

remote/src/test/scala/org/apache/pekko/remote/classic/ActorsLeakSpec.scala
```diff
@@ -41,6 +41,7 @@ object ActorsLeakSpec {
        pekko.remote.classic.quarantine-after-silence = 3 s
        pekko.remote.use-unsafe-remote-features-outside-cluster = on
        pekko.test.filter-leeway = 12 s
+       pekko.test.timefactor = 2
```
He-Pin, line 44: Maybe this will always be needed on some box.

mdedetrich: Yes, its currently set in CI so this isn't a concern. The bigger problem was that when you ran this test locally it basically never passed when you just did `sbt test`.

He-Pin: Maybe you need increase the jvm xmx too？

PR #473, opened 2023-07-06
Properly cleanup state after SharedLeveldbJournalSpec exits

mdedetrich:
The `persistence-shared/src/test/scala/org/apache/pekko/persistence/journal/leveldb/SharedLeveldbJournalSpec.scala` test currently fails if its run more then once because state that is stored in a local temp folder is not cleaned up after the test completes.

You can replicate this locally by doing the following

* run `persistence-shared/test` locally once, it should pass
* run `persistence-shared/test` second time, it should fail and continue to fail

If you apply this change then the tests continue to pass.


pjfanning (approved):
lgtm

Issue #353, opened 2023-05-27
lots of scaladoc warnings in build


pjfanning:
Check out git repo and run `sbt unidoc` and you will lots of warnings like `Could not find any member to link for` (and others).

It would be great if anyone who want to help can look at these and submit some PRs.

ydash:
@pjfanning 
May I work on it?

pjfanning:
@ydash - this is still an issue. If you have time to contribute some PRs that would be great.

We are looking to do a release at the moment so we would like to minimise disruption. Could you keep any PRs small? Not one change in each PR but a relatively limited number of changes in each PR.

ydash:
It seems to appear two types of warn log, `Could not find any member to link` and `The link target is ambiguous`.

One of `Could not find any member to link` logs is following:
```
`[warn] \path\incubator-pekko\cluster-sharding-typed\src\main\scala\org\apache\pekko\cluster\sharding\typed\testkit\javadsl\EntityRef.scala:22:1: Could not find any member to link for "EntityRef".`
```

This problem can be resolved by explicitly writing a full path of an entity:

```diff
-  * For testing purposes this `EntityRef` can be used in place of a real [[EntityRef]]. 
+  * For testing purposes this `EntityRef` can be used in place of a real [[org.apache.pekko.cluster.sharding.typed.javadsl.EntityRef EntityRef]].
```

https://github.com/apache/incubator-pekko/blob/966204814ecac09b9d7cc6e866fe2a71a660ab05/cluster-sharding-typed/src/main/scala/org/apache/pekko/cluster/sharding/typed/testkit/javadsl/EntityRef.scala#L23

Following is an example of `The link target is ambiguous` logs:
```
[warn] \path\incubator-pekko\actor-testkit-typed\src\main\scala\org\apache\pekko\actor\testkit\typed\javadsl\TestProbe.scala:70:1: The link target "ActorTestKit#createTestProbe" is ambiguous. Several members fit the target:
[warn] [M](name: String, clazz: Class[M]): org.apache.pekko.actor.testkit.typed.javadsl.TestProbe[M] in class ActorTestKit [chosen]
[warn] [M](name: String): org.apache.pekko.actor.testkit.typed.javadsl.TestProbe[M] in class ActorTestKit
[warn] [M](clazz: Class[M]): org.apache.pekko.actor.testkit.typed.javadsl.TestProbe[M] in class ActorTestKit
[...]

PR #477, opened 2023-07-10
Make unidoc generate external javadoc links

ydash:
Refs #353 

I added a mapping from a claspath to an external javadoc url to solve warning logs which scaladoc's references to slf4j library causes.


pjfanning (approved):
lgtm

PR #480, opened 2023-07-13
add release notes

pjfanning:
* copied from https://cwiki.apache.org/confluence/display/PEKKO/Pekko+Core+Release+Notes
* pekko-http docs have release notes - so this is more consistent
  * https://pekko.apache.org/docs/pekko-http/current/release-notes/index.html

### DIFF
--- a/docs/src/main/paradox/index.md
+++ b/docs/src/main/paradox/index.md
@@ -5,6 +5,7 @@
 @@@ index
 
 * [security/index](security/index.md)
+* [release-notes/index](release-notes/index.md)
 * [guide/index](typed/guide/index.md)
 * [general/index](general/index.md)
 * [index-actors](typed/index.md)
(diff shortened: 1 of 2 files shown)


pjfanning:
the build issue is not related to this change

this URL is not reliable - https://dspace.jaist.ac.jp/dspace/bitstream/10119/4784/1/IS-RR-2004-010.pdf

mdedetrich (approved):
lgtm although I don't like the fact we are duplicating release notes in multiple places. We should probably investigate whether its fine for us to have release notes on the doc website itself and just link to it from the wiki (if we even need to).

pjfanning:
> lgtm although I don't like the fact we are duplicating release notes in multiple places. We should probably investigate whether its fine for us to have release notes on the doc website itself and just link to it from the wiki (if we even need to).

the plan is to make these docs the source of reference - the cwiki page is/was there for drafting the notes - not meant to be the permanent home

PR #481, opened 2023-07-13
enable mima check

pjfanning:
* Copied some excludes files from Akka 2.6
* It appears that the mima check does not factor in some build logic that adds the Java9+ classes to pekko jars
* so mima check incorrectly thinks the classes built for Java9+ are missing from the locally built modules

### DIFF
--- a/.github/workflows/binary-compatibility-checks.yml
+++ b/.github/workflows/binary-compatibility-checks.yml
@@ -30,11 +30,11 @@ jobs:
           # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
           fetch-depth: 0
 
-      - name: Setup Java 8
+      - name: Setup Java 11
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 8
+          java-version: 11
 
       - name: Cache Coursier cache
         uses: coursier/cache-action@v6.4.0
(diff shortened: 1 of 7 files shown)


mdedetrich:
> Copied some excludes files from Akka 2.6

Can you comment on which filters this is the case (i.e. copied from Akka) versus the ones you added to deal with the JDK 9 problem to make it easier to review?

I will look into this in a bit more detail, @raboof and @jrudolph might be able to provide some more context on the custom JDK 9 build stuff.

pjfanning:
> > Copied some excludes files from Akka 2.6
> 
> Can you comment on which filters this is the case (i.e. copied from Akka) versus the ones you added to deal with the JDK 9 problem to make it easier to review?
> 
> I will look into this in a bit more detail, @raboof and @jrudolph might be able to provide some more context on the custom JDK 9 build stuff.

All the excludes files in this PR are copied for the Java 9 build reason.

mdedetrich:
> All the excludes files in this PR are copied for the Java 9 build reason.

You mean that these files were copied from Akka to fix the Java 9 build reason? The context behind my statement is I want to know if anything bespoke (i.e. not from Akka) was done regarding MIMA filters in this PR (if so I have to pay extra attention).

pjfanning:
The `.excludes` files are direct copies of files from Akka 2.6. The build yml and build scala files changes are mine.

mdedetrich:
> The `.excludes` files are direct copies of files from Akka 2.6. The build yml and build scala files changes are mine.

Perfect, thanks for clarification. Will try and review this today but given this it shouldn't take long.

mdedetrich (approved):
lgtm

Issue #483, opened 2023-07-14
support creating ForkJoinPools with maximumPoolSize


pjfanning:
ForkJoinPool has extra contructors in JDK9+ that allow maximumPoolSize to be set.

Relates to https://github.com/apache/incubator-pekko/discussions/482

Pekko still supports JDK8 so implementing this is not straightforward.

Options include:
* adding a new dispatcher executor type of `fork-join-executor-jdk9` which is similar to `fork-join-executor` but supports the extra `maximumPoolSize` config (the class for this could be added to jdk9+ build (we support having a scala-jdk9 source directory)
* have `fork-join-executor` support the `maximumPoolSize` but ignore it or fail if Java 8 runtime is used. We might need to use Java reflection to do this.

mdedetrich:
I think it makes sense to solve this when we get multi-jar-release working so we can target jdk 9+ in a principled fashion. For this reason I think removing it from the 1.1.0-M2 milestone makes sense as I seriously doubt that we will get multi-jar-release working by then, @He-Pin @pjfanning do you agree?

He-Pin:
IIRC, the multi-jar-release has bad IDE support, and @Glavo once told me about that too, for me the current implementation seems good, so does netty work this way, another library like reactor-core is using the multi-jar-release mechanism.

pjfanning:
I don't believe in dropping useful features because one implementation that can't be done is preferred over one that is ready to go.

He-Pin:
@mdedetrich Netty's `PlatformDependent0` is doing the same, I don't think multi-jar-release should be a blocker for this, What about android? multi-jar-release can not ensure the code will work on Android too.

mdedetrich:
> multi-jar-release can not ensure the code will work on Android too.

Android supports JDK 1.8+ which means it will ignore java 9 compiled classes in the multi-release-jar (its the java-9 compiled classes which will be a non standard location).

In other words it will work fine

> I don't think multi-jar-release should be a blocker for this

I am not saying that its a blocker, just that multi-jar-release may already be done before the feature is complete

He-Pin:
Some methods are missing in Android, AFAIK, we will see.

He-Pin:
@mdedetrich It will be contagious when user packaging shadow jars, @glavo just told me。

pjfanning:
> @mdedetrich It will be contagious when user packaging shadow jars, @Glavo just told me。

I don't think this is a good reason not to use MR jars. Good shading tools support MR jars.

[...]

PR #485, opened 2023-07-14
support jdk9 forkjoinpool maximum-pool-size

pjfanning:
* relates to #483 
* needs a lot more work
* open for debate on how best to proceed

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -482,6 +482,10 @@ pekko {
         # Setting to "FIFO" to use queue like peeking mode which "poll" or "LIFO" to use stack
         # like peeking mode which "pop".
         task-peeking-mode = "FIFO"
+
+        # This config is new in Pekko v1.1.0 and only has an effect if you are running with JDK 9 and above.
+        # Read the documentation on `java.util.concurrent.ForkJoinPool` to find out more. Default in hex is 0x7fff.
+        maximum-pool-size = 32767
       }
 
       # This will be used if you have set "executor = "thread-pool-executor""
--- a/actor/src/main/scala-jdk-9/org/apache/pekko/dispatch/PekkoJdk9ForkJoinPool.scala
+++ b/actor/src/main/scala-jdk-9/org/apache/pekko/dispatch/PekkoJdk9ForkJoinPool.scala
@@ -17,8 +17,9 @@
 
 package org.apache.pekko.dispatch
 
-import org.apache.pekko.annotation.InternalApi
-import org.apache.pekko.dispatch.ForkJoinExecutorConfigurator.PekkoForkJoinTask
+import org.apache.pekko
+import pekko.annotation.InternalApi
+import pekko.dispatch.ForkJoinExecutorConfigurator.PekkoForkJoinTask
 
 import java.util.concurrent.{ ForkJoinPool, ForkJoinTask, TimeUnit }
 
--- a/actor/src/main/scala-jdk-9/org/apache/pekko/dispatch/PekkoJdk9ForkJoinPool.scala
+++ b/actor/src/main/scala-jdk-9/org/apache/pekko/dispatch/PekkoJdk9ForkJoinPool.scala
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.dispatch
+
+import org.apache.pekko.annotation.InternalApi
+import org.apache.pekko.dispatch.ForkJoinExecutorConfigurator.PekkoForkJoinTask
+
+import java.util.concurrent.{ ForkJoinPool, ForkJoinTask, TimeUnit }
+
+/**
+ * INTERNAL PEKKO USAGE ONLY
+ *
+ * An alternative version of [[ForkJoinExecutorConfigurator.PekkoForkJoinPool]]
+ * that supports the `maximumPoolSize` feature available in [[java.util.concurrent.ForkJoinPool]] in JDK9+.
+ */
+@InternalApi
+private[dispatch] final class PekkoJdk9ForkJoinPool(
+    parallelism: Int,
+    threadFactory: ForkJoinPool.ForkJoinWorkerThreadFactory,
+    maximumPoolSize: Int,
+    unhandledExceptionHandler: Thread.UncaughtExceptionHandler,
+    asyncMode: Boolean)
+    extends ForkJoinPool(parallelism, threadFactory, unhandledExceptionHandler, asyncMode,
+      0, maximumPoolSize, 1, null, ForkJoinPoolConstants.DefaultKeepAliveMillis, TimeUnit.MILLISECONDS)
+    with LoadMetrics {
+
+  override def execute(r: Runnable): Unit =
+    if (r ne null)
+      super.execute(
+        (if (r.isInstanceOf[ForkJoinTask[_]]) r else new PekkoForkJoinTask(r)).asInstanceOf[ForkJoinTask[Any]])
+    else
+      throw new NullPointerException("Runnable was null")
+
+  def atFullThrottle(): Boolean = this.getActiveThreadCount() >= this.getParallelism()
+}
--- a/actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala
@@ -13,9 +13,13 @@
 
 package org.apache.pekko.dispatch
 
-import java.util.concurrent.{ ExecutorService, ForkJoinPool, ForkJoinTask, ThreadFactory }
-
 import com.typesafe.config.Config
+
+import java.lang.invoke.{ MethodHandle, MethodHandles, MethodType }
+import java.util.concurrent.{ ExecutorService, ForkJoinPool, ForkJoinTask, ThreadFactory }
+import scala.util.Try
+
+import org.apache.pekko.util.JavaVersion
 
 object ForkJoinExecutorConfigurator {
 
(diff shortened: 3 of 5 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinPoolConstants.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.dispatch
+
+private[dispatch] object ForkJoinPoolConstants {
+  val MaxCap: Int = 0x7FFF // 32767
+  val DefaultKeepAlive: Long = 60000
```
He-Pin, line 22: Java code style for the `Constants`. And why we have such a big value? the default fork join pool is expected to not run any blocking things, and blocking code should be managed.

pjfanning: these are the values that java.util.concurrent.ForkJoinPool use by default. They are not public - that's why In need my own constants.

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala
```diff
@@ -84,12 +85,28 @@ class ForkJoinExecutorConfigurator(config: Config, prerequisites: DispatcherPrer
...
+      val asyncMode: Boolean,
+      val maxPoolSize: Int)
       extends ExecutorServiceFactory {
+
+    def this(threadFactory: ForkJoinPool.ForkJoinWorkerThreadFactory,
+        parallelism: Int,
+        asyncMode: Boolean) = this(threadFactory, parallelism, asyncMode, ForkJoinPoolConstants.MaxCap)
+
+    private lazy val pekkoJdk9ForkJoinPoolClassOpt: Option[Class[_]] =
+      Try(Class.forName("org.apache.pekko.dispatch.PekkoJdk9ForkJoinPool")).toOption
```
He-Pin, line 97: Maybe add a util to read the current jdk version instead.

pjfanning: because of the way Pekko builds the Java9+ classes, even the unit tests seem not to have this class available. It does eventually get built and appears in the jar.

actor/src/main/scala-jdk-9/org/apache/pekko/dispatch/PekkoJdk9ForkJoinPool.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.concurrent.{ ForkJoinPool, ForkJoinTask, TimeUnit }
+
+/**
+ * INTERNAL PEKKO USAGE ONLY
+ *
+ * An alternative version of [[ForkJoinExecutorConfigurator.PekkoForkJoinPool]]
+ * that supports the `maximumPoolSize` feature available in [[java.util.concurrent.ForkJoinPool]] in JDK9+.
+ */
+private[dispatch] final class PekkoJdk9ForkJoinPool(
```
He-Pin, line 30: `@InternalApi`

actor/src/main/resources/reference.conf
```diff
@@ -482,6 +482,10 @@ pekko {
         # Setting to "FIFO" to use queue like peeking mode which "poll" or "LIFO" to use stack
         # like peeking mode which "pop".
         task-peeking-mode = "FIFO"
+
+        # This config is new in Pekko v1.1.0 and only has an effect if you are running with JDK 9 and above.
+        # Read the documentation on `java.util.concurrent.ForkJoinPool` to find out more. Default in hex is 0x7fff.
+        maximum-pool-size = 32767
```
He-Pin, line 488: These are huge too, will cause OOM on common 4core 8GB box

pjfanning: please read the java.util.concurrent.ForkJoinPool code in Java 9+

this is the value hat it uses by default

migesok: May be it makes sense to make to turn around the configuration for FJP and make it similar to how the default global FJP in JDK is configured:
- "parallelism" calculation could stay the same, be derived from runtime CPU count
- add a new parameter called something like `max-spare-threads` (similar to "java.util.concurrent.ForkJoinPool.common.maximumSpares" system property) with 256 as the default value (again the same as the default for the global FJP)
- the resulting maximum pool size will be calculated as parallelism + max-spare-threads

Additionally I think another parameter needs to be exposed - `reject-blocking-if-pool-saturated` or something similar which governs the `saturate` predicate in the new FJP constructor - the user should have a choice between failing on blocking task once the pool limit is reached or keep piling the tasks in memory.

Does it make sense?

pjfanning: I'm not changing how the existing configs work - I worry far more about breaking behaviour for existing users than adding new features that only a small number of people want.

If we want new behaviour, we need new configs and to deprecate the old ones.

Please don't complicate this PR by adding extra requirements to it. Raise a new issue if you don't like the behaviour of the existing configs.

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinPoolConstants.scala
mdedetrich: > Java code style for the Constants

If this is a Scala source file we should be using Scala convention for constants, which would be `MaxCap` rather than `MAX_CAP`

actor/src/main/resources/reference.conf
migesok: Thank you! I see your point now about changing the existing behaviour. I just want to suggest that instead of exposing `maximum-pool-size` directly, it might be better to expose the spare thread count as JDK devs did for the global pool. The default value could be similar to the current default (effectively no limit at all).
Given that parallelism is calculated programmatically in the current Akka FJP configurator, combining it with a `maximum-pool-size` when the values are close might be inconvenient. Usually when you look at it, you think in the terms of "how many parallel blocking operations I can afford, additionally to the normal parallelism level".

I can contribute exposing `saturate` config in a separate PR later if this one goes through.

He-Pin:
We are using Pekko now, and all of our nodes are running on Java 21, I would like to see this feature be shipped in 1.1.x.

mdedetrich:
> We are using Pekko now, and all of our nodes are running on Java 21, I would like to see this feature be shipped in 1.1.x.

I don't think this is possible to do properly without multi-release-jar which is [hitting problems](https://github.com/sbt/sbt-multi-release-jar/issues/24#issuecomment-1891156070)

He-Pin:
> > We are using Pekko now, and all of our nodes are running on Java 21, I would like to see this feature be shipped in 1.1.x.
> 
> I don't think this is possible to do properly without multi-release-jar which is [hitting problems](https://github.com/sbt/sbt-multi-release-jar/issues/24#issuecomment-1891156070)

Why? Netty can support multi jdk just with some features tags.

If that method is not called, no problem, we can built a release with jdk 11.

pjfanning:
This code can be written using Java reflection. The PR as it stands uses Java reflection.

He-Pin:
I think we should go with methodHanfle, this can be done whiteout the multi release jar support.

He-Pin:
> This code can be written using Java reflection. The PR as it stands uses Java reflection.

Better with method handle, on Java 17+, we need an addition add opens, the less reflection the better,we should move on.

pjfanning:
> > This code can be written using Java reflection. The PR as it stands uses Java reflection.
> 
> Better with method handle, on Java 17+, we need an addition add opens, the less reflections the better,we should move on prepare for future Java releases.

I've changed the PR to use a MethodHandle.

pjfanning:
@mdedetrich He Pin wants this in v1.1.0. I'm neutral on this. Is it ok to use MethodHandles like this and rewrite later when we have Multi-Release jar support?

He-Pin:
@pjfanning I want this because it will dress some issue especially when user is blocking the actors.
image

We have something like this in our production code with Virtual Thread too, If you want to defer this to 1.2.x/ 2.0.x that's fine, but this will be a killer feature in 1.1.x.

actor/src/main/resources/reference.conf
He-Pin: The virtual thread is setting to `256` by default if the `maxPoolSizeValue` is not set, but yes, we can not change it and user can set to a limit match their systeam.

pjfanning: I'm not sure what these comments mean. If someone wants changes, can they provide actual code. I can interpret code more easily than abstract comments.

I have read and re-read the ForkJoinPool code and if we use the old Java constructor (as opposed to the new one added in Java 9) - that the old constructor calls the new constructor and passes the value of MAX_CAP for the maxPoolSizeValue. MAX_CAP=0x7fff which is 32767.

I will make changes if I am wrong but please use code and code links to show me where I am going wrong.

He-Pin: 1. I mean your current code is right, 
2. we can not change it to a smaller number,eg 256
3. user can change it  if they like.

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinPoolConstants.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.dispatch
+
+private[dispatch] object ForkJoinPoolConstants {
+  final val MaxCap: Int = 0x7FFF // 32767
+  final val DefaultKeepAlive: Long = 60000
```
He-Pin, line 22: DefaultKeepAliveInMiilis

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala
He-Pin:
```scala
private lazy val pekkoJdk9ForkJoinPoolHandleOpt: Option[MethodHandle] = {
     if (JavaVersion.majorVersion >= 9) {
      pekkoJdk9ForkJoinPoolClassOpt.map { clz =>
        val methodHandleLookup = MethodHandles.lookup()
        val mt = MethodType.methodType(classOf[Unit], classOf[Int],
          classOf[ForkJoinPool.ForkJoinWorkerThreadFactory],
          classOf[Int], classOf[Thread.UncaughtExceptionHandler], classOf[Boolean])
        methodHandleLookup.findConstructor(clz, mt)
      }} else None
    }
```

He-Pin (approved):
lgtm

pjfanning:
> lgtm at the first round , just a little nitpick , I think we should enable run test on JDK 21 too, if the code is target to higher JDK version.

We do test with java 21. Check the nightly builds

He-Pin (approved):
Lgtm

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinPoolConstants.scala
pjfanning: done

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala
pjfanning: I added the java-version check because it happens inside the lazy val calculation - see we won't check this over and over.

actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinPoolConstants.scala
He-Pin: Thank you, but Take care, time to sleep 😴 💤 😪

He-Pin:
The code is well done, and the last part should be update the document, that can come up later too.

mdedetrich:
I would be more comfortable putting this in 1.1.0-M2 or maybe 1.2.0 as we are already overloaded with changes and we have just created a regression with JDK

pjfanning:
@He-Pin out of interest, is there a reason why you can't create a custom dispatcher that supports java9+ fork-join-pool? Such an approach would work today, no reason to wait for a Pekko 1.1 release.

He-Pin:
> @He-Pin out of interest, is there a reason why you can't create a custom dispatcher that supports java9+ fork-join-pool? Such an approach would work today, no reason to wait for a Pekko 1.1 release.

1. Custom dispatcher is much change for most crud available Java boies/girls.
2. Out of box solution is better, my systems is full async, but not all.
3. This is a demand from the pekko community, not by me.

I'm find to postpone this to later version eg 1.1.0-m2 or m3 , and I can help to test it.

mdedetrich:
@He-Pin Argument would be a lot more convincing if creating a JDK 9 `ForkJoinPool` would be impossible which is not the case. Although having it default to the better version of JDK9+ would be nice, its also not a deal breaker and we can also document this as a stop gap.

The reason I am hesitant is that the current Pekko build with all of the JDK9 classes, multi-jvm etc etc is a bit of a mess, note that I am _**NOT**_ blaming anyone specifically here, there are legitimate reasons for this mess, i.e. Akka was a really old sbt project that had to implement various workarounds which are no longer needed but those workarounds were still left in the code and then there is the case of working around even current day sbt limitations which is what is causing the sbt-osgi problems.

In short I would like to start devoting actual effort to cleaning up the sbt build which is something that I have already started doing rather than dumping additional features that involve non trivial changes to the build which then end up causing regressions (either at the start or later down the track due to how the build interactions work). sbt-osgi is a good example of this, I took initiate/responsibility to get it working and as you can see we are still getting problems 3-4 months down the track.

We need to be more strategic/intelligent about what features we push and for this feature specifically I would prefer if we actually put it on hold until I figure out and then stabilize all of the sbt-multi-jar/sbt-multi-jvm/sbt-osgi interactions.

He-Pin:
Yes, I want to polish the toolchain too. WDYT, a document still needs to be done to ship this. but the code should be mostly ready.
Let's delay it to 1.2.x. That will make our life easier because we must fix the toolchain first.

mdedetrich:
If 1.1.0-M2 happens and it takes a long time because of other reasons then it can still land for 1.1.0 but to make it clear, we shouldn't delay 1.1.0 just for this feature.

I would also prefer to implement this properly using multi-jar-release rather than reflection.

He-Pin:
That's true, I expect 1.1.0-M1 to be released soon, and I will not submit any feature pr before the 1.1.x branch cut.

pjfanning:
@He-Pin would you support having this in v1.1.0 ?

He-Pin:
> @He-Pin would you support having this in v1.1.0 ?

Yes, I would like to have it ,even in m1.

pjfanning:
@mdedetrich can we merge this? I can rewrite this in another release to use Multi-Release jar support. The impl is private and controlled solely by config so it is easy to just completely rewrite it. We can document that the current support is experimental.

mdedetrich (approved):
lgtm, as long as we can improve on it later without breaking backwards compatibility

Roiocam (approved):
LGTM

PR #487, opened 2023-07-15
Update branch in publish-1.0-docs.yml

pjfanning:
* will need to backport/cherry-pick to `1.0.x` branch
* the advantage of having this checked-in as opposed to having a GitHub CI parameter (that can be entered when running the task) is that changes need peer review


mdedetrich (approved):
lgtm

PR #488, opened 2023-07-15
=ret Change the comment of Akka to Pekko.

He-Pin:
It should be Pekko I think.

### DIFF
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -499,7 +499,7 @@ pekko {
         applied-adapters = []
 
         # The default remote server port clients should connect to.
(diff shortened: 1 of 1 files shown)


pjfanning (commented):
7355 are PEKK (on a mobile phone keypad).

It might be better to just remove the `(AKKA)` and not go into detail about where 7355 comes from.

gmethvin (commented):
Originally the comment meant that 2552 was AKKA on a telephone keypad, but this wasn't exactly obvious. We changed the port to 7355 which is PEKK. So IMO this should say `(PEKK on a telephone keypad)` or just remove the parenthetical entirely since it's not really important.

gmethvin (approved):

pjfanning (commented):
lgtm

Issue #239, opened 2023-03-09
Drop the shaded protobuf?


He-Pin:
[protobuf](https://github.com/apache/incubator-pekko/tree/main/protobuf/src/main/java/org/apache/pekko/protobuf
)

Maybe we can dropped it in next major version?

pjfanning:
I think it's safe to remove this in v1.1

pjfanning:
#489 merged

PR #489, opened 2023-07-15
remove shaded protobuf-java v2 jar (pekko-protobuf)

pjfanning:
* relates to https://github.com/apache/incubator-pekko/issues/239
* we have protobufv3
* protobuf-java(v2) is old and not maintained by Google - and has open CVEs


jsbrucker:
It is possible what I have written below isn't relevant to this discussion as I'm not entirely sure of the impact of this proposal. But I am concerned that this is suggesting removing support for devs/users of Pekko from using proto2 for communication between remote actors (or for integration with other systems).

proto2 doesn't strictly need to be "maintained" (by an org like google) as it is a stable spec, it's like maintaining tcp. proto3 is a competing standard to proto2 not a replacement. proto2 still has significant usage, because among other reasons getting rid of `required`/`optional` and not sending "default" values over the wire don't align well with many use cases.

I'm sure the codebases I have worked on can't be the only ones, but the distinction between `Some("")`, `Some(0)`, `Some(false)`, etc and `None` can/should important semantic meaning, just as the distinction between `required` and `optional` allows users or proto2 to ensure the code leveraging the contracts aren't breaking any invariants (they just have to worry about forward/backward compatibility themselves).

Protobuf isn't just used for archived data at rest and as APIs that require 100% forward/backward compatibility in perpetuity. For use cases where developers have control over the deployment of all uses of a given proto (implementation detail of a collection of services, or actors...) it often makes much more sense to use proto2 and handle backward/forward compatibility directly and be able to rely on the compiler erroring if the contract is violated (particularly useful in a monorepo) instead of having to explicitly write validators for every field.

In my experience it is usually an anti-pattern for a `case class` or API to make every property/parameter an `Option` then write runtime validation to ensure they are set or worse make them all non-Option types but set defaults for all of them. Both situations result in code that is much more difficult to maintain. To my mind protos effectively map to `case class`es or APIs so I struggle to see why the patterns should be different.

pjfanning:
If this gets merged - it will be merged to v1.1 branch. Protobuf2 users can keep using Pekko 1.0.x. Protobuf 2.x has multiple unfixed CVEs. Most of our users would appreciate if we remove dependencies on insecure software libraries.

The Akka team have also recently dropped support for the shaded protobuf2 jar.

I think you should still be able to optionally use unshaded protobuf libs if you so choose.

jsbrucker:
I'm not sure what you mean by:
> protobuf-java (v2)
Doesn't https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java/3.23.4 support code generated from `proto2`?
> protobuf-java (v2) has multiple unfixed CVEs
The version above doesn't appear to have any vulnerabilities listed.

jsbrucker:
I apologize for the confusion, I thought this was discussing removing support for proto2 (syntax) based protobufs, not about removing support for a particular version of the java library.

mdedetrich (approved):
lgtm , would also like a review from @raboof / @jrudolph just to be sure

jrudolph:
The shaded versions are not for user consumptions and as far as I understand the v2 variant was only kept alive for reasons not relevant to pekko at all.

Users should use the official versions.

LICENSE
```diff
@@ -353,12 +353,6 @@ Copyright 2014 - 2016 Real Logic Ltd.
 
 ---------------
 
-pekko-protobuf contains the sources of Google protobuf 2.5.0 runtime support,
```
jrudolph, line 356: Do we already have a notice that we publish the v3 versions? If not, maybe we can repurpose this one.

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
```diff
@@ -484,8 +484,7 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
   // include "com.google.protobuf.GeneratedMessage" = proto in configured serialization-bindings.
   private def checkGoogleProtobuf(className: String): Boolean = checkClass("com.google.protobuf", className)
 
-  // pekko-protobuf is now not a dependency of remote so only load if user has explicitly added it
-  // remove in v1.1
+  // pekko-protobuf-v3
```
jrudolph, line 487: The method itself should probably also go? The v3 package is called `org.apache.pekko.protobufv3`.

jrudolph (approved):
lgtm with a few comments

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
He-Pin: Maybe user still have an explicitly dependence of it?

LICENSE
pjfanning: We don't have the sources of protobuf-java v3 copied into our code base. With protobuf-java v3, sbt-assembly is used to take the official google jar and to produce a new pekko-protobuf-v3 jar with shaded package names. The LICENSE file accounts for source code. We have the `legal` folder where we have specific LICENSE files that we add to our jars. There is one of those for pekko-protobuf-v3 jar that acknowledges that the classes are derived from Google classes.

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
pjfanning: This function is still used inside Serialization.scala.

If it is ok, I can create a separate issue to investigate what happens if we remove this function but I suspect it could cause trouble for the code that uses pekko-protobuf-v3 jar if we remove it.

He-Pin:
@pjfanning I think ProtobufSerializer need some update too

LICENSE
jrudolph: Thanks for clarifying, makes sense.

jrudolph:
> I think ProtobufSerializer need some update too, there are some comments about `org.apache.pekko.protobuf` there

This is done using reflection and configured in the setting `akka.serialization.protobuf.allowed-classes` which contains / contained `org.apache.pekko.protobuf.GeneratedMessage`.

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
pjfanning: just 1 test in org.apache.pekko.remote.artery.HandshakeRetrySpec -- could be an intermittent test issue, I'll run it again

He-Pin (approved):
lgtm

actor/src/main/scala/org/apache/pekko/serialization/Serialization.scala
pjfanning: that was an intermittent failure - I have included the change in this PR now - so the function is now removed

Issue #491, opened 2023-07-17
Undefined behavior of PekkoPduProtobufCodec$


nicolasstucki:
The Scala spec states that names in the source should not contain $. These are reserved for code generation. Using a name with a $ will have undefined behavior.

https://github.com/apache/incubator-pekko/blob/fbac70e9dd57050449b6d00b7b9c6a2669d600d3/remote/src/main/scala/org/apache/pekko/remote/transport/PekkoPduCodec.scala#L131

A change in this undefined behavior was found in https://github.com/lampepfl/dotty/issues/18227

Solution: change the name of this object.

pjfanning:
@mdedetrich any idea why the $ was added? It was added during one of your commits when we were renaming the akka packages. The Akka 2.6 code does not have a $ in the equivalent name.

mdedetrich:
This was likely a mistake, @nicolasstucki does that mean we should make a new release?

pjfanning:
we can start by release noting that this affects users of Scala 3.3.2 (which is only in preview)

mdedetrich:
> * we can start by release noting that this affects users of Scala 3.3.2 (which is only in preview)
> * still works with Scala 3.3.1-RC4 as per @nicolasstucki
> * we should release Pekko 1.0.1 but we may not need to start the RC process yet
> * we may still be better off starting Pekko HTTP 1.0.0 release process first

So just to clarify (and @nicolasstucki can either confirm or deny this), but afaik this issue of using `$` in class names is not Scala 3 specific, its just that because incubator-pekko got added to Scala 3 integration testing (i.e. open community build) where it got revealed but its a general Scala isssue.

The core problem is that Scala (regardless if its Scala 2 or Scala 3) you should never have a `$` in class names that you define because Scala compiler itself will generate classes internally which can contain the `$`.

For this reason I would recommend making a new release irrespective of Scala 3, we still need to go through the release process. If possible Scala 2/3 should also add checks/warnings to prevent this in the future.

pjfanning:
> i think this is not a bug of pekko but a scala3 regression.

Scala 3 is not likely to be changed to allow us not to fix the class name.

I'm happy to change this package private class name since the Scala spec says we shouldn't have $ suffix in class names.

He-Pin:
@mdedetrich So why the compiler did not prevent this but produced an artifact?

mdedetrich:
> @mdedetrich So why the compiler did not prevent this but produced an artifact?

[...]

PR #492, opened 2023-07-17
remove $ from class name

pjfanning:
* relates to #491

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/PekkoPduCodec.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/PekkoPduCodec.scala
@@ -128,7 +128,7 @@ private[remote] trait PekkoPduCodec {
  * INTERNAL API
  */
 @nowarn("msg=deprecated")
-private[remote] object PekkoPduProtobufCodec$ extends PekkoPduCodec {
+private[remote] object PekkoPduProtobufCodec extends PekkoPduCodec {
   import PekkoPduCodec._
 
   private def ackBuilder(ack: Ack): AcknowledgementInfo.Builder = {
--- a/remote/src/main/mima-filters/1.0.x.backwards.excludes/pducodec.backwards.excludes
+++ b/remote/src/main/mima-filters/1.0.x.backwards.excludes/pducodec.backwards.excludes
@@ -1,0 +1,12 @@
+# rename PekkoPduProtobufCodec$ - https://github.com/apache/incubator-pekko/issues/491
+ProblemFilters.exclude[StaticVirtualMemberProblem]("org.apache.pekko.remote.transport.PekkoPduProtobufCodec.decodeMessage")
+ProblemFilters.exclude[StaticVirtualMemberProblem]("org.apache.pekko.remote.transport.PekkoPduProtobufCodec.decodePdu")
+ProblemFilters.exclude[StaticVirtualMemberProblem]("org.apache.pekko.remote.transport.PekkoPduProtobufCodec.constructHeartbeat")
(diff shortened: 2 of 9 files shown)


remote/src/main/scala/org/apache/pekko/remote/transport/PekkoPduCodec.scala
```diff
@@ -128,7 +128,7 @@ private[remote] trait PekkoPduCodec {
  * INTERNAL API
  */
 @nowarn("msg=deprecated")
-private[remote] object PekkoPduProtobufCodec$ extends PekkoPduCodec {
```
He-Pin, line 131: Hum

mdedetrich:
So MIMA is breaking which is kind of expected, otherwise it looks good (will have to set MIMA to 1.0.1)

mdedetrich:
@pjfanning Regarding the MIMA changes, I would actually advocating to disable MIMA and/or ignore the CI rather than adding more filters because if someone breaks this again in the future then MIMA won't catch it.

I think the most practical would be to see if CI works with the newly added MIMA filters and then remove the filters just before merge

mdedetrich (changes requested):
Blocking PR due to https://github.com/apache/incubator-pekko/pull/492#issuecomment-1638142692, will unblock when the MIMA filters get removed.

remote/src/main/scala/org/apache/pekko/remote/transport/PekkoPduCodec.scala
He-Pin: image

He-Pin: Seems IDEA messed this.

pjfanning:
> Blocking PR due to [#492 (comment)](https://github.com/apache/incubator-pekko/pull/492#issuecomment-1638142692), will unblock when the MIMA filters get removed.

Why do we need to remove the mima filters? There is no 1.0.1 release to compare with.

Are you advocating that we should turn off MIMA checking altogether?

pjfanning:
> @pjfanning Do you plan to have this get merged and cherry-pick a pr against the 1.0.x branch?

If you check the mailing list, @mdedetrich wants us to delay using the `1.0.x` branch.

And the mailing list also includes a suggested patching process that involves merging to `main` first and backporting selected PRs to `1.0.x` as opposed to merging to `1.0.x` first.

pjfanning:
> @pjfanning Regarding the MIMA changes, I would actually advocate to disable MIMA and/or ignore the MIMA CI check rather than adding more filters because if someone breaks this again in the future then MIMA won't catch it.
> 
> I think the most practical would be to see if CI works with the newly added MIMA filters and then remove the filters just before merge and then after this PR is merged set MIMA to work on Pekko 1.0.1.

Can't we rebase the mima checks to v1.0.1 after it is released and remove the mima filters in this PR as part of that.

mdedetrich:
> Pekko HTTP does not use pekko-remote (where the issue is). It's not
very tidy but only Pekko HTTP users who also need Pekko Remote and
Clustering features would need to upgrade to Pekko (Core) 1.0.1.

As long as we don't forget, sure. I will review the PR by end of tonight (want to also check other sections of code to make sure we catch everything).

mdedetrich (approved):
lgtm, but lets make sure that we don't merge anything into main unless necessary (which should be a given considering that we are making a new release which will create a code freeze) and also undo the MIMA checks that this PR adds since the filters that are being added are covering a public API.

pjfanning:
The mima filters are for PekkoPduProtobufCodec - a package private object. There is no public API affected by this change or its mima filters.

PR #500, opened 2023-07-26
Remove sonatype commands that shouldn't be run in release process

mdedetrich:
As documented in https://github.com/apache/incubator-pekko-site/wiki/Pekko-Release-Process#deploy-the-jars-to-apache-maven-repository-staging there are commands that shouldn't be executed. This PR removes these commands (which are brought in by sbt-sonatype) so release managers cannot accidentally execute them as part of release process.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,12 @@ ThisBuild / versionScheme := Some(VersionScheme.SemVerSpec)
 sourceDistName := "apache-pekko"
 sourceDistIncubating := true
 
+commands := commands.value.filterNot { command =>
+  command.nameOption.exists { name =>
+    name.contains("sonatypeRelease") || name.contains("sonatypeBundleRelease")
+  }
+}
+
 enablePlugins(
   UnidocRoot,
   UnidocWithPrValidation,


nvollmar (approved):
lgtm

build.sbt
```diff
@@ -18,6 +18,12 @@ ThisBuild / versionScheme := Some(VersionScheme.SemVerSpec)
 sourceDistName := "apache-pekko"
 sourceDistIncubating := true
 
+commands := commands.value.filterNot { command =>
+  command.nameOption.exists { name =>
+    name.contains("sonatypeRelease") || name.contains("sonatypeBundleRelease")
```
He-Pin, line 23: I would like to use `StringUtil.containsAny(...)` :)

He-Pin (approved):
lgtm

jrudolph (approved):
lgtm (I guess there's little risk that someone is going to miss these commands, easy enough to grep the cause)

build.sbt
mdedetrich: I don't think we can use that because this is an sbt build and hence the dependency that brings in `StringUtil` isn't available

pjfanning: seems better to stick with plain Java/Scala

pjfanning (approved):
lgtm

PR #502, opened 2023-07-28
mention pekko and pekko.tcp URL schemes

pjfanning:
migration docs

### DIFF
--- a/docs/src/main/paradox/project/migration-guides.md
+++ b/docs/src/main/paradox/project/migration-guides.md
@@ -16,9 +16,10 @@ These migration notes are designed for users migrating from Akka 2.6 to Pekko 1.
 * Pekko packages start with "org.apache.pekko" instead of "akka" - e.g. `import org.apache.pekko.actor` instead of `import akka.actor`
 * Where class names have "Akka" in the name, the Pekko ones have "Pekko" - e.g. PekkoException instead of AkkaException
 * Config names use "pekko" prefix instead of "akka", e.g. `pekko.actor.provider` instead of `akka.actor.provider`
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/project/migration-guides.md
```diff
@@ -16,9 +16,10 @@ These migration notes are designed for users migrating from Akka 2.6 to Pekko 1.
 * Pekko packages start with "org.apache.pekko" instead of "akka" - e.g. `import org.apache.pekko.actor` instead of `import akka.actor`
 * Where class names have "Akka" in the name, the Pekko ones have "Pekko" - e.g. PekkoException instead of AkkaException
 * Config names use "pekko" prefix instead of "akka", e.g. `pekko.actor.provider` instead of `akka.actor.provider`
+* The Pekko node URLs use `pekko://` and `pekko.tcp//` instead of `akka://` and `akka.tcp//` respectively.
```
He-Pin, line 19: Maybe  a table like would be better :

|before|after|
|-|-|
|akka://|pekko://|
|akka.tcp//|pekko.tcp//|

He-Pin (changes requested):
should be akka.tcp:// and pekko.tcp://

docs/src/main/paradox/project/migration-guides.md
pjfanning: I've tried to create a table but Paradox build is not rendering it properly. Something about trying to put a table inside a bulleted list is causing trouble.

pjfanning: I reformatted the bulleted list so that the 2 URL schemes each get their own indented list entry

He-Pin (approved):
lgtm

PR #504, opened 2023-07-28
Update sbt-reproducible-builds and set check resolver to staging

mdedetrich:
So its easier to verify future builds

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,7 @@
  * This file is part of the Apache Pekko project, which was derived from Akka.
  */
 
+import net.bzzt.reproduciblebuilds.ReproducibleBuildsPlugin.reproducibleBuildsCheckResolver
 import org.apache.pekko._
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -24,6 +25,9 @@ commands := commands.value.filterNot { command =>
   }
 }
 
+ThisBuild / reproducibleBuildsCheckResolver :=
+  "Apache Pekko Staging".at("https://repository.apache.org/content/groups/staging/")
```
pjfanning, line 29: I understand that this useful but users might want to verify against maven central or some other repo. Would it be possible to add a comment about other possible repo values - possibly a link to some centralised doc.

mdedetrich: I was planning to document this in our [release process](https://github.com/apache/incubator-pekko-site/wiki/Pekko-Release-Process). This makes sense as a default because almost all of the time the use case is to verify against jars in staging, if something has already been released and its not in staging then something went wrong.

pjfanning (approved):
lgtm

PR #511, opened 2023-07-30
Remove org.apache.pekko package from sbt build

mdedetrich:
The main goal of this PR is to both remove the `package org.apache.pekko` from sbt build source files (i.e. sources in `project`) and hence also clean up the imports. The reason/s for doing this are

* The `package org.apache.pekko` in sbt build source files are not needed since they are not going to appear in any JVM binary, these files are **only** for loading the sbt build
* The use of the `package org.apache.pekko` was entirely inconsistent, some sbt build source files had it and others didn't
* The source files in `project` didn't follow the idiomatic scala/java build convention (i.e. the source files should be placed in the `org/apache/pekko` dir structure. Removing the package altogether is far less disruptive solution

In addition there were cases where the `package sbt` was used but not really needed because a better solution existed (see comments).

### DIFF
--- a/project/ScalafixSupport.scala
+++ b/project/ScalafixSupport.scala
@@ -11,11 +11,9 @@
  * Copyright (C) 2018-2022 Lightbend Inc. <https://www.lightbend.com>
  */
 
-package sbt
-import Keys.baseDirectory
+import sbt._
 
-import org.apache.pekko.ProjectFileIgnoreSupport
-import sbt.Keys.unmanagedSources
+import Keys._
 
 trait ScalafixSupport {
   private val ignoreConfigFileName: String = ".scalafix.conf"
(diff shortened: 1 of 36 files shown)


project/ScalafixSupport.scala
mdedetrich, line 1: Initially this source file needed the `package sbt` in order to access the `SimpleCommand` datastructure which is package private for `sbt`. Using `SimpleCommand` in isn't intentional, instead you are supposed to use the `nameOption` method, see https://github.com/sbt/sbt/blob/e1ffc2af08f59464e6a4063026337708d390610a/main-command/src/main/scala/sbt/Command.scala#L31-L34

Which matches against `SimpleCommand` and returns you the `name` if its a `SimpleCommand`

jrudolph (approved):
lgtm

PR #522, opened 2023-08-01
Make sbt-header work against all Scala versions

mdedetrich:
To mirror https://github.com/apache/incubator-pekko/blob/main/.github/workflows/headers.yml#L30


He-Pin (approved):
Lgtm

PR #532, opened 2023-08-03
Automatically handle MiMa for all versions

mdedetrich:
The primary goal of this PR is to automatically handle checking for MiMa while in the process removing the hardcoding of `latestPatchOf10` (which is currently incorrect since it should be `1` due to Pekko 1.0.1 being released but the fact that we forgot to increment this strengthens the argument for doing this automatically).

The PR is designed to handle the following cases

* Minor being bumped (i.e. when we just went from `1.0.x` to `1.1.x`) but no version yet released. For this case we compare to the entire previous minor release (i.e. `1.0.0` and `1.0.1` with the example of `1.1.x`)
* Just after the first release of a minor (i.e. `1.0.0`) which implies a snapshot in which case we just compare with `1.0.0`
* The standard case, i.e. with `1.0.1` released then we compare to all of the previous versions from the major series up until this one.

The code is cleanly commented and its also easy to test, i.e. in `build.set` set

```sbt
ThisBuild / version := "<VERSION>"
```

to whatever version you want to test with and then just run `sbt mimaPreviousArtifacts`. This will give you a list of all of computed artifacts to check against.

### DIFF
--- a/project/MiMa.scala
+++ b/project/MiMa.scala
@@ -11,15 +11,13 @@
  * Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
  */
 
-import scala.collection.immutable
 import sbt._
 import sbt.Keys._
+import sbtdynver.DynVerPlugin.autoImport.previousStableVersion
 import com.typesafe.tools.mima.plugin.MimaPlugin
 import com.typesafe.tools.mima.plugin.MimaPlugin.autoImport._
 
 object MiMa extends AutoPlugin {
-
-  private val latestPatchOf10 = 0
 
   override def requires = MimaPlugin
   override def trigger = allRequirements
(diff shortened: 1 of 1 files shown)


pjfanning:
I think it be easier to explicitly name the version that we are diffing against. For me, both 1.0.x and main should diff against 1.0.0 release.

mdedetrich:
> I think it be easier to explicitly name the version that we are diffing against. 

The problem is that this is brittle, if we forget to bump the version (as we just did) then we can introduce bincompat problems. The logic of what we want to do is perfectly easy to automate

> For me, both 1.0.x and main should diff against 1.0.0 release.

Yes and this is retained

mdedetrich:
Okay PR is ready to review.

pjfanning:
build failed because we have no 1.1.0-M0 jars to compare against.

mdedetrich:
> build failed because we have no 1.1.0-M0 jars to compare against.

So this solution would be much more cleanly solved by just publishing a milestone artifact. Tagging something as milestone but not publishing its artifact is all things considered not idiomatic. We can discuss whether it makes sense to bundled a `M0` release together with the last patch version of the previous minor since technically speaking the source will be the same, its just that the versioning is changing since we are creating a branch.

Otherwise the only other way to solve this problem is to hack some kind of workaround and/or not use milestone tags and deal with the fact that until the first release is made for a new minor the snapshots will not line up.

He-Pin:
Let me close and reopen to trigger a rebuild.

He-Pin:
@mdedetrich There is no milestone release

project/MiMa.scala
```diff
@@ -29,35 +27,36 @@ object MiMa extends AutoPlugin {
...
+            Set.empty
+          else
+            Set(organization.value %% moduleName.value % previousStableVersion)
+        case None => Set.empty
+      }
+    },
+    checkMimaFilterDirectories := checkFilterDirectories(baseDirectory.value, version.value))
+
+  def checkFilterDirectories(moduleRoot: File, version: String): Unit = {
+    val strippedPatchRegex = """(\d+)(.*)""".r
```
gmethvin, line 48: This regex doesn't make sense if we're matching against the whole version—if you have a version of `1.1.0` then you're extracting `.1.0` as the patch version which doesn't parse as an int. Did you want something like `\d+[.]\d+[.](\d+).*`?

PR #533, opened 2023-08-04
fix up some docs with version numbers that are Akka versions

pjfanning:
* also removes some Akka migration docs that are not relevant to Pekko users - our migration guide says that we expect users migrating from Akka to first upgrade to Akka 2.6.20 and to follow through on any necessary Akka migrations.


gmethvin (approved):

He-Pin (approved):
Lgtm

PR #486, opened 2023-07-15
 Migrate multi node testkit to Netty 4. 

He-Pin:
refs: https://github.com/apache/incubator-pekko/issues/462

I verified it locally with`cluster/MultiJvm/test`

Issue #462, opened 2023-07-01
Migrate multi-node-testkit to netty 4


He-Pin:
After drop the classical remoting which depends on netty 3, I think we can migrate the multi-node-testkit to netty 4

pjfanning:
I guess that will need to be discussed on the dev mailing list. I have no objection to removing it but we haven't yet really discussed versioning strategy and the priorities for future development.

pjfanning:
depends on https://github.com/apache/incubator-pekko/issues/41

He-Pin:
I have a working local pr, will submit a pr tomorrow.

PR #539, opened 2023-08-05
!test Migrate multi node testkit to Netty 4.

He-Pin:
refs: https://github.com/apache/incubator-pekko/issues/462

I verified it locally with ` sbt cluster/MultiJvm/test`
previous: https://github.com/apache/incubator-pekko/pull/486


project/SbtMultiJvm.scala
```diff
@@ -372,7 +373,13 @@ object MultiJvmPlugin extends AutoPlugin {
         val connectInput = input && index == 0
         log.debug("Starting %s for %s".format(jvmName, testClass))
         log.debug("  with JVM options: %s".format(allJvmOptions.mkString(" ")))
-        (testClass, Jvm.startJvm(javaBin, allJvmOptions, runOptions, jvmLogger, connectInput))
+        val testClass2Process = (testClass, Jvm.startJvm(javaBin, allJvmOptions, runOptions, jvmLogger, connectInput))
+        if (index == 0) {
+          log.debug("%s for %s 's started as `Controller`, waiting 2 seconds before start clients.".format(jvmName,
+            testClass))
+          TimeUnit.SECONDS.sleep(2)
```
He-Pin, line 380: This is needed to get it works on linux and mac.

pjfanning:
I agree that this can be merged to pekko main (when it is approved) and eventually released with pekko 1.1.0. 

I think that anyone who needs this code before the pekko 1.1.0 can easily create their own jar. It would probably be useful for a pekko-multi-node-netty4-testkit project to be created in GitHub for that project to release jars. The project can be archived after the pekko 1.1.0 release.

project/SbtMultiJvm.scala
pjfanning: I don't like time based waits. On a slow machine, they can stop working. Our CI builds on VMs that can be very slow.

Is there any way that this code could be changed to ping the netty cluster and keep doing so until it gets notification that the cluster is ready?

pjfanning (changes requested):

He-Pin:
@pjfanning `ping` and `telnet` is not available on Windows, then I need to implement another `telnet` in java to test the controller is started up.

pjfanning:
> @pjfanning `ping` and `telnet` is not available on Windows, then I need to implement another `telnet` in java to test the controller is started up.

When I say ping, I don't mean the Unix command. Wiktionary has this meaning (among many) of the word 'ping'.

A packet which a remote host is expected to echo, thus indicating its presence.

He-Pin:
I will udpate the pr wit a `java` ping then.

project/SbtMultiJvm.scala
```diff
@@ -372,11 +374,41 @@ object MultiJvmPlugin extends AutoPlugin {
...
+      }
+    }
+
+    val startTime = System.currentTimeMillis()
+    var connectivity = false
+    while (!connectivity && (System.currentTimeMillis() - startTime < timeoutInMillis)) {
+      connectivity = telnet(inetSocketAddress, 1000)
+      TimeUnit.MILLISECONDS.sleep(100)
+    }
+  }
```
He-Pin, line 410: @pjfanning I think it can be running on a slower box now.

pjfanning: seems like a good solution

cluster/src/multi-jvm/resources/logback-test.xml
```diff
@@ -0,0 +1,13 @@
...
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="info">
+        <appender-ref ref="STDOUT" />
+    </root>
+    <logger name="io.netty.util.Recycler" level="ERROR" />
```
mdedetrich, line 11: Is this something that was needed just because of netty4 or is it generally helpful? If it is generally helpful I would do this change in a separate PR so that we can backport it to Pekko 1.0.x

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
```diff
@@ -94,44 +95,88 @@ private[pekko] case object Client extends Role
...
+          .option[java.lang.Boolean](ChannelOption.TCP_NODELAY, true)
+          .option[java.lang.Boolean](ChannelOption.SO_KEEPALIVE, true)
+          .connect(sockaddr)
+        new RemoteConnection {
+          override def channelFuture: ChannelFuture = cf
+
+          override def shutdown(): Unit = {
+            try {
+              channelFuture.channel().close().sync()
+              eventLoopGroup.shutdownGracefully(1L, 1L, TimeUnit.SECONDS)
```
mdedetrich, line 140: Is it worth it to make these shutdown gracefully constants configurable in typesafe config? As I understand these values are the underlying reasons why the implementation initially didn't work on Mac/Linux and if so I think it makes sense for users to be able to configure this without needing a new release incase there are further issues in the future.

cluster/src/multi-jvm/resources/logback-test.xml
mdedetrich, line 11: Is this something that was needed just because of netty4 or is it generally helpful? If it is generally helpful I would do this change in a separate PR so that we can backport it to Pekko 1.0.x

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
```diff
@@ -94,44 +95,88 @@ private[pekko] case object Client extends Role
...
+          .bind(sockaddr)
+          .sync()
+
+        new RemoteConnection {
+          override def channelFuture: ChannelFuture = cf
+
+          override def shutdown(): Unit = {
+            try {
+              channelFuture.channel().close().sync()
+              parentEventLoopGroup.shutdownGracefully(1L, 1L, TimeUnit.SECONDS)
```
mdedetrich, line 168: Same as previous comment about configuration

project/SbtMultiJvm.scala
```diff
@@ -167,10 +168,11 @@ object MultiJvmPlugin extends AutoPlugin {
...
       // just concatenate conflicting text files
       assembly / assemblyMergeStrategy := {
-        case n if n.endsWith(".class") => MergeStrategy.first
-        case n if n.endsWith(".txt")   => MergeStrategy.concat
-        case n if n.endsWith("NOTICE") => MergeStrategy.concat
-        case n                         => (assembly / assemblyMergeStrategy).value.apply(n)
+        case n if n.endsWith(".class")  => MergeStrategy.first
+        case n if n.endsWith(".txt")    => MergeStrategy.concat
+        case n if n.endsWith("NOTICE")  => MergeStrategy.concat
+        case n if n.endsWith("LICENSE") => MergeStrategy.concat
```
mdedetrich, line 174: Although this change was needed for this feature specifically, I think it would be wiser to make this change in a separate PR because in general we should concatenate all `LICENSE` statements.

project/SbtMultiJvm.scala
```diff
@@ -372,11 +374,41 @@ object MultiJvmPlugin extends AutoPlugin {
         val connectInput = input && index == 0
         log.debug("Starting %s for %s".format(jvmName, testClass))
         log.debug("  with JVM options: %s".format(allJvmOptions.mkString(" ")))
-        (testClass, Jvm.startJvm(javaBin, allJvmOptions, runOptions, jvmLogger, connectInput))
+        val testClass2Process = (testClass, Jvm.startJvm(javaBin, allJvmOptions, runOptions, jvmLogger, connectInput))
+        if (index == 0) {
+          log.debug("%s for %s 's started as `Controller`, waiting before can be connected for clients.".format(jvmName,
+            testClass))
+          val controllerHost = hosts.head
+          val serverPort: Int = Integer.getInteger("multinode.server-port", 4711)
```
mdedetrich, line 382: As with previous comments about magic constants should this be made configurable (although in this specific case since we are dealing with test classes in sbt project build it should be done by cli args as is done elsewhere).

project/SbtMultiJvm.scala
```diff
@@ -372,11 +374,41 @@ object MultiJvmPlugin extends AutoPlugin {
...
+      } catch {
+        case _: Exception => false
+      } finally {
+        socket.close()
+      }
+    }
+
+    val startTime = System.currentTimeMillis()
+    var connectivity = false
+    while (!connectivity && (System.currentTimeMillis() - startTime < timeoutInMillis)) {
```
mdedetrich, line 406: Same with previous comment about constants.

mdedetrich (changes requested):
Great work, PR generally looks fine. My main remarks are regarding magic constants and some changes I think be split out and merged on their own so they can be backported.

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
He-Pin: Noop, The reason it not working because of the `Player` not wanting the `Controller` started in the sbtMultiJvm. the old `shutdown` method is deprecated but still can be used.

mdedetrich:
I have just run the cluster tests on the latest version of this PR and I can confirm that they all pass, good work! After the comments are addressed the PR looks to be in good shape to me.

I have attached the test log for those curios

[multi-jvm-test-main-6.txt](https://github.com/apache/incubator-pekko/files/12269096/multi-jvm-test-main-6.txt)

cluster/src/multi-jvm/resources/logback-test.xml
He-Pin: It because there are so much netty log when you run testing.

project/SbtMultiJvm.scala
He-Pin: It's configurable with the `-Dmultinode.server-port`.

cluster/src/multi-jvm/resources/logback-test.xml
mdedetrich: I thought so, do you mind making this change in a separate PR so it can be back-ported separately to Pekko 1.0.x?

I can also do this, and technically speaking you shouldn't even need to update the PR since it would get resolve via a merge.

project/SbtMultiJvm.scala
mdedetrich: I will resolve this for now, I don't want to block the PR because of this and if people have issues we can always update this later (its just tests)

cluster/src/multi-jvm/resources/logback-test.xml
mdedetrich: There was a miscommunication, I realize that `PoolThreadCache` and `Recycler` classes only exist in Netty 4 and not Netty 3 (just checked now), resolving conversation.

mdedetrich (approved):
So I am approving changes, previous comments are resolved. I am not fully aware of this part of the code so it would be good to get some more eyes before merging (we aren't in a massive rush to merge this right now).

project/SbtMultiJvm.scala
mdedetrich: Regarding this point, maybe it makes sense to make this a separate issue to look at after this PR gets merged? The reason I am stating this is because the best way to find out if these time based constants are too low is by seeing how often the tests fail in our nightly github CI action and we can then iterate from there.

Of course most ideal would be to ping the netty cluster until its ready but I don't feel its valid to block the PR over this, for me its critical to get people to test this directly (via snpashots/building locally), hopefully in production to see if there aren't any regressions well before a release happens.

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
pjfanning: I agree with @mdedetrich - this would be better as a setting in reference.conf

project/SbtMultiJvm.scala
He-Pin: I think that's ok, as it's just in tests and no one will touch those again, otherwise we need another：
```
multinode.max-time-waiting-to-connect-in-seconds=20 
```
like thing

project/SbtMultiJvm.scala
He-Pin: Should I seperate this to a dedicated PR? but seems like only Netty 4 contains this thing.

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
He-Pin: I just updated the code without this, but which may waiting max to 15 seconds. the old code is deprecated.

project/SbtMultiJvm.scala
```diff
@@ -372,11 +374,41 @@ object MultiJvmPlugin extends AutoPlugin {
         val connectInput = input && index == 0
         log.debug("Starting %s for %s".format(jvmName, testClass))
         log.debug("  with JVM options: %s".format(allJvmOptions.mkString(" ")))
-        (testClass, Jvm.startJvm(javaBin, allJvmOptions, runOptions, jvmLogger, connectInput))
+        val testClass2Process = (testClass, Jvm.startJvm(javaBin, allJvmOptions, runOptions, jvmLogger, connectInput))
+        if (index == 0) {
+          log.debug("%s for %s 's started as `Controller`, waiting before can be connected for clients.".format(jvmName,
+            testClass))
```
He-Pin, line 380: @mdedetrich This is the real fix which can be a sleep(2000) or like thing. the error was an `CONNECT(2)` from the networking , which I was thing caused by resolver or my code bug.

project/SbtMultiJvm.scala
mdedetrich: Yes please make a separate PR for this. Although it's true that this only effects netty 4 for now, in general of there multiple `LICENSE`'s we do want to concat them (i.e. this isn't just specific to netty 4)

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
He-Pin: @mdedetrich I updated it to `.shutdownGracefully(0L, 0L, TimeUnit.SECONDS)` which alian the old behavior.

pjfanning (approved):
lgtm

He-Pin:
I run several times on gitpod. it works, I will do a final validation before merge this.

He-Pin:
I ran it several times again and it do passed, so it would be good to go.

PR #540, opened 2023-08-05
WIP: Migrate the classic remoting to Netty 4.

He-Pin:
Based on https://github.com/apache/incubator-pekko/pull/539

For a preview, may not be landing in 1.1.0  if community dediced to remove the classic remoting in 1.1.0.

Personally,+1 for removing classic remoting in 1.2.0 

Will only continue after the the vote done.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -48,4 +48,5 @@ SortImports.blocks = [
   "com.sun."
   "org.apache.pekko."
   "org.reactivestreams."
+  "io.netty."
 ]
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -28,7 +28,6 @@ object Dependencies {
   // needs to be inline with the aeron version, check
   // https://github.com/real-logic/aeron/blob/1.x.y/build.gradle
   val agronaVersion = "1.15.1"
-  val nettyVersion = "3.10.6.Final"
   val netty4Version = "4.1.96.Final"
   val protobufJavaVersion = "3.19.6"
   val logbackVersion = "1.2.11"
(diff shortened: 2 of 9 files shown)


pjfanning:
The discussion on pekko mailing list has not reached a conclusion. I agree with @jrudolph that we are better off removing classic remoting. His point is that it is very hard to test remoting and there is no need to have 2 styles.

So -1 from me on this.

pjfanning (changes requested):
-1 from me - many of us think that we should remove classic remoting and only support artery - or failing that do nothing

pjfanning:
The Pekko community conversation is at https://lists.apache.org/thread/grzp3jw398rtpc6oqfybzxnry2mtyh64

I would just prefer to see consensus there before we consider merging a PR. There's no harm in testing this solution but I'd prefer if it wasn't merged until the community decides on what we want.

He-Pin:
> The Pekko community conversation is at https://lists.apache.org/thread/grzp3jw398rtpc6oqfybzxnry2mtyh64
> 
> I would just prefer to see consensus there before we consider merging a PR. There's no harm in testing this solution but I'd prefer if it wasn't merged until the community decides on what we want.

That's  true, I was testing the idea in this draft.

He-Pin:
@pjfanning and @mdedetrich as The poll result, I think I will close this PR.
https://www.reddit.com/r/scala/comments/160pvne/upgrade_pekkos_classical_transport_to_netty_4_or/

mdedetrich:
Your free to close the PR if you want but as was pointed out in https://www.reddit.com/r/scala/comments/160pvne/comment/jxnuvbf/?utm_source=reddit&utm_medium=web2x&context=3 if we decide to strictly follow semver we may not have a choice here unless we want to ship code with CVE's

He-Pin:
I see the value of keep it and upgrade to Netty 4, but it will need more time to polish too, if it's pr is doomed and will not get merged, then maybe better not put more time on it.

Maybe @pjfanning could share some views too.

PR #541, opened 2023-08-05
better license header example

pjfanning:
* the contrib guide links to a POI source file but that header is formatted differently from what we use - the Spark example matches our style
* the existing DRY link is not great - the text display has char encoding issues and the article is truncated because they want you to create an account - the wiki link is fine

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -298,13 +298,13 @@ For a pull request to be considered at all, it has to meet these requirements:
 1. A pull request must be [linked to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) it aims to resolve in the PR's description (or comments). This can be achieved by writing "Fixes #1234" or similar in PR description.
 1. Licensing rules:
    - Existing files with copyright statements must leave those copyright statements intact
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm

CONTRIBUTING.md
```diff
@@ -298,13 +298,13 @@ For a pull request to be considered at all, it has to meet these requirements:
 1. A pull request must be [linked to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) it aims to resolve in the PR's description (or comments). This can be achieved by writing "Fixes #1234" or similar in PR description.
 1. Licensing rules:
    - Existing files with copyright statements must leave those copyright statements intact
-   - New files should have an Apache license header instead. For an example of this, see [this file](https://github.com/apache/poi/blob/trunk/poi/src/main/java/org/apache/poi/POIDocument.java).
+   - New files should have an Apache license header instead. For an example of this, see [this file](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/SparkContext.scala).
```
He-Pin, line 301: maybe Flink, as flink is using pekko.

pjfanning: * it's just a link to a file with an Apache header in the format that we want it in
* it's not an endorsement of Spark or Flink
* It might be nice to have a list of tools that use Pekko but that should be its own page

PR #549, opened 2023-08-08
=sbt Do retry connection in Player.

He-Pin:
refs: https://github.com/apache/incubator-pekko/issues/547
and 
refs: https://github.com/apache/incubator-pekko-http/pull/297

PR #550, opened 2023-08-08
=test Reschedule the reconnect when exception caught.

He-Pin:
Lost it in the last pr https://github.com/apache/incubator-pekko/pull/549.

But as we do a `sync()` when connect, so this will never reached.


pjfanning (approved):
lgtm

PR #561, opened 2023-08-11
mapAsyncPartitioned / mapAsyncPartitionedUnordered

pjfanning:
* code based on https://github.com/jaceksokol/akka-stream-map-async-partition
* relates to #553

There are differences from the new Akka mapAsyncPartitioned function.
* The Akka function has parallelism and perPartition as input params - we do not support perPartition


legal/pekko-stream-jar-license.txt
```diff
@@ -0,0 +1,231 @@
...
+
+pekko-stream contains code derived from jaceksokol/akka-stream-map-async-partition
+<https://github.com/jaceksokol/akka-stream-map-async-partition>
+wich is licensed under an MIT License.
+
+ - stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartition.scala
+
+MIT License
+
+Copyright (c) 2021 Jacek Sokol
```
jaceksokol, line 213: I guess you don't want my name here 😉

jaceksokol:
LGTM

legal/pekko-stream-jar-license.txt
pjfanning: This is a copy of the license in https://github.com/jaceksokol/akka-stream-map-async-partition

gmethvin: @jaceksokol it's up to you whether we include this. If you don't mind, you can choose to contribute this work under Pekko's standard Apache license, without your copyright notice. This would make things a bit simpler for us but it's ultimately your call.

jaceksokol: I added an ordered version of the operator, please check this out: https://github.com/pjfanning/incubator-pekko/pull/1. I'm not sure if we need both versions. When I was developing the operator a couple of years ago, the use case I had did not require the order of the output elements to be the same as in the input.

pjfanning: I think we can support both.

mdedetrich (approved):
lgtm

He-Pin:
@pjfanning @mdedetrich I will need more time to review this, I will try to after work today, overall ,looking very good.

He-Pin:
@pjfanning It would be nice to compare the performance boost with `mapAsync`.

pjfanning:
> @pjfanning It would be nice to compare the performance boost with `mapAsync`.

This is a new set of functions - so I don't think a perf test is necessary. @jaceksokol would you have a sample scenario where the mapAsyncPartitioned was better for you than mapAsync?

He-Pin:
> > @pjfanning It would be nice to compare the performance boost with `mapAsync`.
> 
> This is a new set of functions - so I don't think a perf test is necessary. @jaceksokol would you have a sample scenario where the mapAsyncPartitioned was better for you than mapAsync?

I think when the a head of line problem occurs ？

jaceksokol:
Yes, this operator may help mitigate the head-of-line problem. The main performance optimisation is related to the fact that values for different keys can be usually processed independently (in parallel in this case).

He-Pin:
@pjfanning Do you have more to update on this, I will go through it this weekend.

pjfanning:
@He-Pin I'm not planning anything more on this, My view is it is big enough already. The impl can be tweaked in follow up PRs. If there is anything fundamentally wrong with the code then of course, the PR will not be merged until that issue is addressed,

He-Pin (approved):
lgtm

He-Pin:
@pjfanning I think we can merge this, and tweak can comes later.

PR #564, opened 2023-08-12
jackson 2.16 support

pjfanning:
see https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.16.0/com/fasterxml/jackson/core/StreamReadConstraints.html

Users who want this feature now can use these libs instead:
* https://github.com/pjfanning/pekko-serialization-jackson215
* https://github.com/pjfanning/pekko-serialization-jackson216

For Pekko 1.1, I still want to get this merged centrally.

### DIFF
--- a/serialization-jackson/src/main/resources/reference.conf
+++ b/serialization-jackson/src/main/resources/reference.conf
@@ -46,6 +46,8 @@ pekko.serialization.jackson {
     max-number-length = 1000
     max-string-length = 20000000
     max-name-length = 50000
+    # max-document-length of 0 means unlimited
+    max-document-length = 0
   }
   write {
     # see https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.16.0/com/fasterxml/jackson/core/StreamWriteConstraints.html
--- a/serialization-jackson/src/main/resources/reference.conf
+++ b/serialization-jackson/src/main/resources/reference.conf
@@ -47,7 +47,7 @@ pekko.serialization.jackson {
     max-string-length = 20000000
     max-name-length = 50000
   }
-  read {
+  write {
     # see https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.16.0/com/fasterxml/jackson/core/StreamWriteConstraints.html
     # these defaults are the same as the defaults in `StreamWriteConstraints`
     max-nesting-depth = 1000
--- a/serialization-jackson/src/main/resources/reference.conf
+++ b/serialization-jackson/src/main/resources/reference.conf
@@ -37,6 +37,18 @@ pekko.serialization.jackson {
 
 }
 
+#//#stream-read-constraints
+pekko.serialization.jackson {
+  read {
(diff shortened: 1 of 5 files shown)


serialization-jackson/src/main/resources/reference.conf
```diff
@@ -37,6 +37,18 @@ pekko.serialization.jackson {
...
 }
 
+#//#stream-read-constraints
+pekko.serialization.jackson {
+  read {
+    # see https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.15.2/com/fasterxml/jackson/core/StreamReadConstraints.html
+    # these defaults are the same as the defaults in `StreamReadConstraints`
+    max-nesting-depth = 1000
+    max-number-length = 1000
+    max-string-length = 20000000
```
He-Pin, line 47:
```
constraints {
    max-nesting-depth = 1000
    max-number-length = 1000
    max-string-length = 20000000
}
```
Seems more clear

pjfanning: I honestly don't think adding the word `constraints` to the config names here makes them any clearer. Config names need to balance descriptiveness with conciseness.

pjfanning:
jackson 2.16.0 should be out soon and it has some more constraints - I will repurpose this PR to further update to jackson 2.16.0 and support the extra constraints

serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala
```diff
@@ -81,15 +83,31 @@ object JacksonObjectMapperProvider extends ExtensionId[JacksonObjectMapperProvid
       config: Config,
       baseJsonFactory: Option[JsonFactory]): JsonFactory = {
 
+    val streamReadConstraints = StreamReadConstraints.builder()
+      .maxNestingDepth(config.getInt("read.max-nesting-depth"))
+      .maxNumberLength(config.getInt("read.max-number-length"))
+      .maxStringLength(config.getInt("read.max-string-length"))
+      .maxNameLength(config.getInt("read.max-name-length"))
+      .maxDocumentLength(config.getLong("read.max-document-length"))
```
laglangyue, line 91: Do these configurations have default values, or users must specify the configuration?

pjfanning: We ship reference conf with the defaults. Users can override the values by setting them in their own application.conf or application.json.

pjfanning: This is how Akka/Pekko has always worked.

pjfanning:
@mdedetrich if we are getting close to creating an RC for the M1 of pekko 1.1.0, I would like to get this merged.

mdedetrich:
Okay ill look at it in more detail tomorrow

nvollmar (approved):
lgtm

He-Pin (approved):
Lgtm

PR #572, opened 2023-08-15
Extract AtomicCancellable in Scheduler.

He-Pin:
Motivation:
Reduce code duplication and make the code simpler to understand.

Result:
Clean Code.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
@@ -25,6 +25,7 @@ import scala.util.control.NonFatal
 
 import com.typesafe.config.Config
 import org.apache.pekko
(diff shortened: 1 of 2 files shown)


actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -30,6 +30,8 @@ import pekko.event.LoggingAdapter
 import pekko.util.Helpers
 import pekko.util.Unsafe.{ instance => unsafe }
 
+import org.apache.pekko.actor.Scheduler.AtomicCancellable
```
pjfanning, line 33: could you match the import style above?

pjfanning (approved):
lgtm

He-Pin:
I have some offline pair view, too, from @WayneWang12 and @dengziming, so will be good to go.

PR #575, opened 2023-08-16
add name with space test case for ActorSystemSpec

dongxuwang:
Add a more common bad case for system name


actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
```diff
@@ -135,6 +135,7 @@ class ActorSystemSpec extends PekkoSpec(ActorSystemSpec.config) with ImplicitSen
       for (n <- Seq(
           "-hallowelt",
           "_hallowelt",
+          "hallo welt",
           "hallo*welt",
           "hallo@welt",
           "hallo#welt",
```
He-Pin, line 141: This maybe better with some random generation  properties tests.

dongxuwang: You mean to replace all those strings with a random generation?

He-Pin: Yes, I think that should be better, what do you think.

He-Pin:
```scala
  if (!name.matches("""^[a-zA-Z0-9][a-zA-Z0-9-_]*$"""))
    throw new IllegalArgumentException(
      "invalid ActorSystem name [" + name +
      "], must contain only word characters (i.e. [a-zA-Z0-9] plus non-leading '-' or '_')")
```

dongxuwang: Yes, agree

actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
```diff
@@ -125,23 +123,17 @@ object ActorSystemSpec {
...
-          "hallo@welt",
-          "hallo#welt",
-          "hallo$welt",
-          "hallo%welt",
-          "hallo/welt")) intercept[IllegalArgumentException] {
-        ActorSystem(n)
+      forAll { (name: String) =>
+        whenever(!name.matches("""^[a-zA-Z0-9][a-zA-Z0-9-_]*$""")) {
+          intercept[IllegalArgumentException](ActorSystem(name))
+        }
```
pjfanning, line 136: I prefer the old way - you can actually see what is being tested. With the change, I have to hope that use cases are tested.

pjfanning (changes requested):
-1 from me. With negative testing, there is a lot to be said for having some tests with human readable test values.

Property based tests can be good but I do not see why the old test code should be removed.

actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
He-Pin: Then there will be infinite cases to add

dongxuwang: Makes sense. Actually I felt like adding one more common case "hallo welt".

pjfanning: keep the old test and add an extra new test that uses the property based approach

dongxuwang: updated, PTAL

actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
```diff
@@ -15,15 +15,13 @@ package org.apache.pekko.actor
...
-
 import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
 import scala.language.postfixOps
-
 import scala.annotation.nowarn
 import com.typesafe.config.{ Config, ConfigFactory }
-
 import org.apache.pekko
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
```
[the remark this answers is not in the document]
dongxuwang: updated

PR #582, opened 2023-08-22
=str Add IterableSource.

He-Pin:
was https://github.com/akka/akka/pull/31372

This introduce a dedicated IterableSource, which can be later used for performance optimization. And also obey the supervisionStrategy now.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/SourceSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/SourceSpec.scala
@@ -311,19 +311,67 @@ class SourceSpec extends StreamSpec with DefaultTimeout {
     }
 
     "use decider when iterator throws" in {
+
+      // using stopping decider
+      Source
+        .fromIterator(() => (1 to 5).iterator.map(k => if (k != 3) k else throw TE("a")))
+        .withAttributes(ActorAttributes.supervisionStrategy(Supervision.stoppingDecider))
+        .grouped(10)
+        .runWith(Sink.head)
+        .failed
+        .futureValue shouldBe a[TE]
+
+      // using stopping decider with recover
+      Source
+        .fromIterator(() => (1 to 5).toIterator.map(k => if (k != 3) k else throw TE("a")))
+        .withAttributes(ActorAttributes.supervisionStrategy(Supervision.stoppingDecider))
+        .recoverWithRetries(1, { case _ => Source.empty })
+        .grouped(10)
+        .runWith(Sink.head)
+        .futureValue shouldBe List(1, 2)
+
+      // failing on every elements, using stopping decider
+      Source
+        .fromIterator(() => (1 to 5).toIterator.map(_ => throw TE("b")))
+        .withAttributes(ActorAttributes.supervisionStrategy(Supervision.stoppingDecider))
+        .grouped(10)
+        .runWith(Sink.headOption)
+        .failed
+        .futureValue shouldBe a[TE]
+
+      // failing on every elements, using stopping decider and recover
+      Source
+        .fromIterator(() => (1 to 5).toIterator.map(_ => throw TE("b")))
+        .withAttributes(ActorAttributes.supervisionStrategy(Supervision.stoppingDecider))
+        .recoverWithRetries(1, { case _ => Source.empty })
+        .grouped(10)
+        .runWith(Sink.headOption)
+        .futureValue shouldBe None
+
+      // using resuming decider
+      Source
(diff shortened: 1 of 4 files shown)


stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/SourceSpec.scala
```diff
@@ -311,19 +311,60 @@ class SourceSpec extends StreamSpec with DefaultTimeout {
...
+        .futureValue should ===(List(1, 2, 1, 2, 1, 2, 1, 2, 1, 2))
 
       Source
         .fromIterator(() => (1 to 5).toIterator.map(_ => throw TE("b")))
         .withAttributes(ActorAttributes.supervisionStrategy(Supervision.restartingDecider))
         .grouped(10)
         .runWith(Sink.headOption)
-        .futureValue should ===(None)
+        .failed
+        .futureValue shouldBe an[TE]
```
pjfanning, line 351: `shouldBe a[TE]` would be better

pjfanning (approved):
lgtm - but I'd prefer if some other reviewers also had a look before we merge

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/SourceSpec.scala
```diff
@@ -311,19 +311,60 @@ class SourceSpec extends StreamSpec with DefaultTimeout {
     }
 
     "use decider when iterator throws" in {
+
```
nvollmar, line 314: I'd prefer to have separate tests with descriptive titles if they test different edge cases, or at least a comment per run hinting what each is testing.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/IterableSource.scala
```diff
@@ -0,0 +1,76 @@
...
+                tryPushNextOrComplete()
+            }
+        }
+
+      private def tryPushNextOrComplete(): Unit =
+        if (currentIterator.hasNext) {
+          if (isAvailable(out)) {
+            push(out, currentIterator.next())
+          }
+          if (!currentIterator.hasNext) {
```
nvollmar, line 65: This check only makes sense if we pushed an element -> move it inside the `if (isAvailable(out)) {` block

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/IterableSource.scala
```diff
@@ -0,0 +1,76 @@
...
+  override val shape: SourceShape[T] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with OutHandler {
+      private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+      private var currentIterator: Iterator[T] = _
+
+      override def onPull(): Unit =
+        try {
+          if (currentIterator eq null) {
```
nvollmar, line 45: couldn't we just initialize the variable with `elements.iterator` and avoid the if check/assignment here?

He-Pin: There are some cases where the `iterator` method call invoving some underling resources creation, that's why I defer the creation in the first place.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/IterableSource.scala
```diff
@@ -0,0 +1,76 @@
...
+            decider(ex) match {
+              case Supervision.Stop   => failStage(ex)
+              case Supervision.Resume => tryPushNextOrComplete()
+              case Supervision.Restart =>
+                currentIterator = elements.iterator
+                tryPushNextOrComplete()
+            }
+        }
+
+      private def tryPushNextOrComplete(): Unit =
```
nvollmar, line 60: That would even be simpler:

```
      private def tryPushNextOrComplete(): Unit = {
         if (currentIterator.hasNext && isAvailable(out)) {
           push(out, currentIterator.next())
         }

         if (!currentIterator.hasNext) {
           completeStage()
         }
      }
```

He-Pin: But this will do another `currentIterator.hasNext` checking.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/IterableSource.scala
He-Pin: IIRC the JOOQ does this, and our internal Lindorm Nosql db driver does the same too.

PR #590, opened 2023-08-27
=str Resort the OnNext extractor order.

He-Pin:
Motivotion:
Resort the extractor order for better performance.


mdedetrich (approved):
Safe change that provides low hanging fruit performance benefit, lgtm

jxnu-liguobin (approved):

PR #599, opened 2023-08-29
use sbt paradox release (#597)

pjfanning:
cherry pick 99cc4e794d7457938a4fcafa9d6f12c4265b58e7

### DIFF
--- a/project/Paradox.scala
+++ b/project/Paradox.scala
@@ -76,9 +76,6 @@ object Paradox {
       "fault-tolerance-sample.html"))
 
   val themeSettings = Seq(
-    // TODO: allow access to snapshots for pekko-sbt-paradox, remove when pekko is released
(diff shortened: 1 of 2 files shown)


project/Paradox.scala
```diff
@@ -114,6 +111,5 @@ object Paradox {
     sourceGeneratorSettings ++
     Seq(
       Compile / paradox / name := "Pekko",
-      resolvers += Resolver.jcenterRepo,
```
He-Pin, line 117: because it's now published to maven center?

He-Pin (approved):
lgtm

project/Paradox.scala
pjfanning: jcenter is no longer supported - it was important years ago - but not now

https://blog.gradle.org/jcenter-shutdown

PR #602, opened 2023-08-31
=str Drop the partial function usage in Flow#zipWithIndex.

He-Pin:
Motivation:

The paritial fucntion is not needed, my memory is blurry 

Result:
Clean code.


mdedetrich (approved):
Nice improvement, thanks

Issue #596, opened 2023-08-28
StatefulMap can call `onComplete` twice if it throw


He-Pin:
The current statefulmap only update the `needInvokeOnCompleteCallback` after the `onComplete(state)` been called successfully, if it throw, then the mark is unset and will be called called again in the postStop block.

Workaround:
Do not throw in 'onComplete' callback

PR #603, opened 2023-08-31
Fix statefulMap may call onComplete twice if it throws

GreyPlane:
#596

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
@@ -23,6 +23,7 @@ import pekko.stream.testkit.StreamSpec
 import pekko.stream.testkit.TestSubscriber
 import pekko.stream.testkit.scaladsl.TestSink
 import pekko.stream.testkit.scaladsl.TestSource
+import pekko.stream.testkit.scaladsl.StreamTestKit
 
 import scala.annotation.nowarn
 import scala.concurrent.Await
@@ -34,6 +35,8 @@ import scala.util.control.NoStackTrace
 class FlowStatefulMapSpec extends StreamSpec {
 
   val ex = new Exception("TEST") with NoStackTrace
+
+  val dtorException = new Exception("bad dtor") with NoStackTrace
 
   "A StatefulMap" must {
     "work in the happy case" in {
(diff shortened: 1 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -2273,25 +2273,27 @@ private[pekko] final class StatefulMap[S, In, Out](create: () => S, f: (S, In) =
...
           case None       => failStage(ex)
         }
-        needInvokeOnCompleteCallback = false
       }
 
       override def onPull(): Unit = pull(in)
 
       override def postStop(): Unit = {
         if (needInvokeOnCompleteCallback) {
+          println("changed 2")
```
pjfanning, line 2296: can you remove the printlns?

pjfanning (changes requested):

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
```diff
@@ -35,6 +36,8 @@ class FlowStatefulMapSpec extends StreamSpec {
 
   val ex = new Exception("TEST") with NoStackTrace
 
+  val dtorException = new Exception("bad dtor") with NoStackTrace
```
He-Pin, line 39: dtor means `destructor`?

He-Pin:
LGTM only two comments.

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
```diff
@@ -280,4 +283,64 @@ class FlowStatefulMapSpec extends StreamSpec {
...
+          badDtor()
+          None
+        })
+      .runWith(Sink.ignore)
+
+    StreamTestKit.assertAllStagesStopped(None)
+    counter shouldBe 1
+  }
+
+  "won't call onComplete twice when upstream finished even if onComplete throws" in {
```
He-Pin, line 327: `in assertAllStagesStopped {`

pjfanning:
@GreyPlane could we get you to fill in an Apache CLA for this? https://www.apache.org/licenses/contributor-agreements.html

For a comparatively small change like this, we might not always require the CLA but this issue/PR has become a potential legal issue since @kerr has looked at the Akka solution (and it is not licensed under a license that is compatible with use by us). We don't want any risk that anyone could claim that Category X code has been copied into our project.

https://www.apache.org/legal/resolved.html#category-x

He-Pin:
@pjfanning The fix of my came out independently,  but the tests were what I copied for quick validation, let's make that clear, and then i checked the Akka fix which is not the same as mine.
So it's the tests I took not the fix.

Many ones in Wechat group can prove this.

I was want to make sure the onComplete to be called and then mark the boolean to false, but that will need the onComplete be idempotent and not easy to use.

pjfanning:
@kerr Let us do a clean room fix.

He-Pin:
And you may find someone not in tech to check and validation those fix is not the same

He-Pin:
@GreyPlane Sorry, I think this fix should be closed, even I knew you come up this independently, but as you are involved/leved comments in the akka pr, so this fix should be closed and come out with someone else, thank you very much.

Issue #601, opened 2023-08-31
Deprecate `statefulMapConcat`


He-Pin:
Motivation:

As the `statefulMapConcat` not handing the `upstreamFinished` event, so when the stream is complete, the ongoing state ke
eped in the closure can be lost.

I suggest we annotate that in the scaladoc and deprecate the usage, and infavor the `statefulMap` 

As the tests in https://github.com/Kotlin/kotlinx.coroutines/pull/3868 shows and the origin Akka issue for `statefulMap: https://github.com/akka/akka/issues/31077

pjfanning:
Is this something that can be done in Pekko 1.0.2? We should at least consider updating the Pekko 1.0 docs.

He-Pin:
Can we deprecate things in the 1.0.x ?

pjfanning:
> Can we deprecate things in the 1.0.x ?

yes, we can deprecate code in 1.0.x

something like

`@deprecated("This function has been replace by <otherFunctionName>", "1.0.2")` (the 2nd string is the version number)

PR #606, opened 2023-09-01
=str deprecate `statefulMapConcat` operator.

He-Pin:
refs: https://github.com/apache/incubator-pekko/issues/601

need rebase on https://github.com/apache/incubator-pekko/pull/607


stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
```diff
@@ -773,6 +773,9 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
    * The returned `Iterable` MUST NOT contain `null` values,
    * as they are illegal as stream elements - according to the Reactive Streams specification.
    *
+   * This operator doesn't handle upstream's completion signal, the state kept in the closure can be lost,
```
mdedetrich, line 776: Minor nit, but this sounds better

> This operator doesn't handle upstream's completion signal since the state kept in the closure can be lost. Use [[FlowOps.statefulMap]] instead.

He-Pin:
this pr seems broken, will resubmit.

PR #607, opened 2023-09-01
=str Make use of statefulMap to implement zipWithIndex.

He-Pin:
Signed-off-by: He-Pin <hepin1989@gmail.com>
(cherry picked from commit a855e58bfcd7e48043c5e6d34536c32cd4a97b3f)

### DIFF
--- a/bench-jmh/src/main/scala/org/apache/pekko/stream/ZipWithIndexBenchmark.scala
+++ b/bench-jmh/src/main/scala/org/apache/pekko/stream/ZipWithIndexBenchmark.scala
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


mdedetrich:
I am not sure we should be cherry picking this PR's onto the `1.0.x` branch. While I understand this fix is largely harmless it sends the wrong impression on the goals of the branch, which is that it is meant to be, as much as possible, a largely untouched version of Akka 2.6 just before BSL changes.

Improvements to areas such as the build (sbt), formatting etc etc are fine and welcome since they don't actually effect the actual runtime behaviour of Pekko but unless we are dealing with CVE's, critical bugs or possible migration features to help people move off Akka we shouldn't be backporting such fixes.

@pjfanning wdyt?

He-Pin:
@mdedetrich as we plan to deprecated the statefulmapconcat, if I don't back port this , I need to annotated it with `@nowarn`

mdedetrich:
> @mdedetrich as we plan to deprecated the statefulmapconcat, if I don't back port this , I need to annotated it with `@nowarn`

Then I think using `@nowarn` makes sense, lets see what @pjfanning says

He-Pin:
Think again about this, And I think we should keep this branch as clean as possible.

PR #613, opened 2023-09-02
=act Mark TaskHolder a final class.

He-Pin:
This class is visible from Java and should be final.


pjfanning:
I don't see why a protected class should be final. I don't see the point in this.

He-Pin:
wo, yes.

PR #614, opened 2023-09-02
=str Remove `futureCB` in MapAsyncUnordered.

He-Pin:
Motivation:
This field is unessesary needed.


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -1405,13 +1405,14 @@ private[stream] object Collect {
     new GraphStageLogic(shape) with InHandler with OutHandler {
       override def toString = s"MapAsyncUnordered.Logic(inFlight=$inFlight, buffer=$buffer)"
 
-      val decider =
+      private lazy val decider =
```
pjfanning, line 1408: Could you keep the lazy val change separate? I'm sure lazy val is probably ok here but lazy vals are converted into `synchronized` code when compiled to Java byte code.

He-Pin: Ok, but I think that can be aline with MapAsync too.

pjfanning (approved):
lgtm

PR #615, opened 2023-09-02
=str Switch the type parameter order of UnfoldResourceSource.

He-Pin:
Motivotion:

The type paramaeter `S` is really not great and the order is wrong

NOTE:
The user facing api is untouched, as that will break code.

Result:
Clean code.

Expects an mima error?

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/FusingSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/FusingSpec.scala
@@ -103,7 +103,7 @@ class FusingSpec extends StreamSpec {
     }
 
     // an UnfoldResourceSource equivalent without an async boundary
-    case class UnfoldResourceNoAsyncBoundry[T, S](create: () => S, readData: (S) => Option[T], close: (S) => Unit)
+    case class UnfoldResourceNoAsyncBoundry[R, T](create: () => R, readData: (R) => Option[T], close: (R) => Unit)
         extends GraphStage[SourceShape[T]] {
       val stage_ = new UnfoldResourceSource(create, readData, close)
       override def initialAttributes: Attributes = Attributes.none
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSource.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSource.scala
@@ -20,79 +20,84 @@ import org.apache.pekko
 import pekko.annotation.InternalApi
 import pekko.stream._
 import pekko.stream.ActorAttributes.SupervisionStrategy
+import pekko.stream.Attributes.SourceLocation
 import pekko.stream.impl.Stages.DefaultAttributes
 import pekko.stream.stage._
 
 /**
  * INTERNAL API
  */
-@InternalApi private[pekko] final class UnfoldResourceSource[T, S](
-    create: () => S,
-    readData: (S) => Option[T],
-    close: (S) => Unit)
+@InternalApi private[pekko] final class UnfoldResourceSource[R, T](
+    create: () => R,
+    readData: (R) => Option[T],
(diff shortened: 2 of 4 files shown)


pjfanning:
I'm going to take some convincing that Pekko should go this way. Like it or not, most Pekko users are former or current Akka users. I don't see why making small adjustments to functions like this really helps. I favour keeping APIs including the type params as close as possible to Akka. New features would be absolutely great. Improving the internals of functions in a way that doesn't affect the API is great.

He-Pin:
The userfacing api did not change the type parameter order, and this is safe to compile both in source and binary, only be a problem when you use a reflection api to retrieve the typevariable's value.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
```diff
@@ -825,12 +825,15 @@ object Source {
...
-   */
-  def unfoldResource[T, S](
-      create: function.Creator[S],
-      read: function.Function[S, Optional[T]],
-      close: function.Procedure[S]): javadsl.Source[T, NotUsed] =
-    new Source(scaladsl.Source.unfoldResource[T, S](create.create _, (s: S) => read.apply(s).toScala, close.apply))
+   * @tparam T - the element type
+   * @tparam R - the resource type
+   */
+  def unfoldResource[T, R](
```
He-Pin, line 831: @pjfanning Do you mean keep the typeparameter `S` here?

pjfanning:
> The userfacing api did not change the type parameter order, and this is safe to compile both in source and binary, only be a problem when you use a reflection api to retrieve the typevariable's value.

Even if the code compiles and bin compatibility is unaffected - there is the problem with the documentation and why our type params have different names from the Akka ones. In the end. how is R better than S. They are both unhelpful.

Keeping the type param names as is and adding typeparam documentation would be more useful, in my opinion.

He-Pin:
@pjfanning https://doc.akka.io/api/akka/2.8/akka/stream/scaladsl/Flow.html#mapWithResource[R,T](create:()=>R)(f:(R,Out)=>T,close:R=>Option[T]):FlowOps.this.Repr[T]

I this later, pr , `R` is used, that's why I want a Resource relative api here to be `R` not `S`. and in fs2, `  def bracket[F[_], R](acquire: F[R])(release: R => F[Unit]): Stream[F, R] = ...`

stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSource.scala
```diff
@@ -20,79 +20,84 @@ import org.apache.pekko
...
-              close(blockingStream)
-              failStage(ex)
-            case Supervision.Restart =>
-              restartState()
-              resumingMode = true
-            case Supervision.Resume =>
-              resumingMode = true
-          }
+      override def preStart(): Unit = {
+        resource = create()
```
He-Pin, line 49: It was named `blockingStream`, which leads to some kind of confusing.

He-Pin:
@pjfanning @mdedetrich I think this can be merged now.

stream-tests/src/test/scala/org/apache/pekko/stream/FusingSpec.scala
```diff
@@ -103,7 +103,7 @@ class FusingSpec extends StreamSpec {
     }
 
     // an UnfoldResourceSource equivalent without an async boundary
-    case class UnfoldResourceNoAsyncBoundry[T, S](create: () => S, readData: (S) => Option[T], close: (S) => Unit)
+    case class UnfoldResourceNoAsyncBoundry[R, T](create: () => R, readData: (R) => Option[T], close: (R) => Unit)
```
pjfanning: nit: could you fix the class name? Boundary instead of Boundry.

pjfanning (approved):
lgtm

PR #616, opened 2023-09-02
=str Switch the type parameter order of UnfoldResourceSourceAsync

He-Pin:
Same as https://github.com/apache/incubator-pekko/pull/615

Motvation:
Switch the type parameter order and change the `S` to `R`.

Result:
Clean code.

Modification:
Previous the resource can be null, but now can't, that's resonable.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
@@ -961,11 +961,13 @@ object Source {
    * @param read - function that reads data from opened resource. It is called each time backpressure signal
    *             is received. Stream calls close and completes when `Future` from read function returns None.
    * @param close - function that closes resource
+   * @tparam T the element type
+   * @tparam R the resource type
    */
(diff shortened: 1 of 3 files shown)


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
```diff
@@ -961,11 +961,13 @@ object Source {
...
+   * @tparam T the element type
+   * @tparam R the resource type
    */
-  def unfoldResourceAsync[T, S](
-      create: () => Future[S],
-      read: (S) => Future[Option[T]],
-      close: (S) => Future[Done]): Source[T, NotUsed] =
+  def unfoldResourceAsync[T, R](
+      create: () => Future[R],
+      read: (R) => Future[Option[T]],
```
jxnu-liguobin, line 969:
```suggestion
      read: R => Future[Option[T]],
```

He-Pin:
@pjfanning @mdedetrich I think this can be merged now.

pjfanning (approved):
lgtm

PR #620, opened 2023-09-02
=str Use OptionVal in Recover instead of Option.

He-Pin:
Motivation:
Reduce allocation

Result:
Lesser object allocation.


mdedetrich (approved):
lgtm

PR #621, opened 2023-09-02
=str Make use of applyOrElse in MapError operator.

He-Pin:
Motivation:
The isDefineAt will check twice, avoid use it.

Result:
Better performance


mdedetrich (approved):
lgtm

PR #625, opened 2023-09-03
Having `@deprecated` and `@Deprecated` annotations on same function breaks scala3 build

pjfanning:
* see https://github.com/apache/incubator-pekko/actions/runs/6061206002/job/16446137883
* seems to happen when dotty doc phase kicks in
* we have already had to remove the `@Deprecated` annotations but these were added by https://github.com/apache/incubator-pekko/commit/2ce8a894e9f36338e85313518ea25170e59badc8
* I will forward fit this to `main` branch if this PR is accepted


He-Pin (approved):
lgtm, let me report this to dotty too.

PR #627, opened 2023-09-03
=str Make decider caculation lazy

He-Pin:
Motivation:
We don't expect much exception throws, and it's `lazy val` in `MapAsync`, so better to keep those two the same.

Result:
Consistent with `MapAsync`.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -664,7 +664,7 @@ private[stream] object Collect {
 
   def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) with InHandler with OutHandler {
-      val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+      lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
 
       private var aggregator: Out = zero
       private var aggregating: Future[Out] = Future.successful(aggregator)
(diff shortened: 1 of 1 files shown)


mdedetrich:
So I don't know what to really think of this, I know that in other cases we use `lazy` for the decider however its kind of arbitrary whether its strict or lazy and I am actually leaning towards it being strict because it is slightly faster.

He-Pin:
Take a look at the MapAsync，is that slow？

jxnu-liguobin:
> We don't expect much exception throws, and it's `lazy val` in `MapAsync`, so better to keep those two the same.

Is this just guess? Because it's `Try`

He-Pin:
@jxnu-liguobin Noop, I think it was not lazy just because out of sight in the first place.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -1402,8 +1402,7 @@ private[stream] object Collect {
     new GraphStageLogic(shape) with InHandler with OutHandler {
       override def toString = s"MapAsyncUnordered.Logic(inFlight=$inFlight, buffer=$buffer)"
 
-      private val decider =
-        inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+      private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
```
pjfanning, line 1405: Can we make this a `def`? In most places in this source file, this is a `def`.

The `def` means no synchronize block. `lazy val`s come with the overhead of having `synchronize` in the generated byte code.

He-Pin: 1. `def` will do recaculation, and that should be avoid.
2. `lazy val` will only do synchronization the first time, and the supervision strategy method is expected to be quick for calculation, so even't the undering Thread is a VirtualThread, it will not cause any problem too.

pjfanning (approved):
lgtm

pjfanning:
You don't need to close a PR to rerun the tests. Go to the broken build and there are buttons to rerun failed builds.

He-Pin:
the streamRefspec is flasky

PR #629, opened 2023-09-05
Use fetch tags from latest checkout action

mdedetrich:
The latest version of the checkout github actions now supports a `fetch-tags` flag which does as described, i.e. it will fetch all of our tags which removes the various workarounds that was used before (i.e. manually running `git fetch` command) etc etc, see https://github.com/actions/checkout/pull/579

This will also help in the erroneous errors that we get in CI, i.e. 

```
[error] Failed to derive version from git tags. Maybe run `git fetch --unshallow` or `git fetch upstream` on a fresh git clone from a fork? Derived version: 0.0.0+1-337943bd-SNAPSHOT`
```

### DIFF
--- a/.github/workflows/timing-tests.yml
+++ b/.github/workflows/timing-tests.yml
@@ -16,9 +16,10 @@ jobs:
     steps:
 
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          fetch-tags: true
 
       - name: Setup Java 11
         uses: actions/setup-java@v3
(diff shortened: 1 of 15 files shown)


.github/workflows/timing-tests.yml
```diff
@@ -56,19 +57,3 @@ jobs:
           report_paths: '**/target/test-reports/TEST-*.xml'
           fail_if_no_tests: false
           skip_publishing: true
-
```
mdedetrich, line 59: I have removed this for now because I think sending emails for when timing tests fail isn't the best way to solve this problem (people just end up ignoring the emails) and if we need it back we can always check git history.

pjfanning (approved):
lgtm

PR #630, opened 2023-09-05
GH Workflows: Separate validateCompile step from validatePullRequest:…

mdedetrich:
… (#300)

This will make the source of errors in steps clearer, without adding any extra pipeline time.

steps in github jobs are dependent on one another, so the build files can be passed onto prValidation without the need for PR validation to compile again

(cherry picked from commit d51b7b30def6ab70c718cf2491df01fb207aa4bb)


He-Pin (approved):
lgtm

PR #632, opened 2023-09-06
scala 3.3.1

pjfanning:
* an experiment
* not sure if we need to merge this

### DIFF
--- a/multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/Player.scala
+++ b/multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/Player.scala
@@ -215,8 +215,9 @@ private[pekko] class ClientFSM(name: RoleName, controllerAddr: InetSocketAddress
     case Event(msg: NetworkOp, _) =>
       log.error("received {} instead of Done", msg)
       goto(Failed)
-    case Event(_: ServerOp, _) =>
-      stay().replying(Status.Failure(new IllegalStateException("not connected yet")))
+    // Scala 3.3.1 compiler says this is unreachable
+    // case Event(_: ServerOp, _) =>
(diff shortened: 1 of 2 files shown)


He-Pin:
Result of 3.3.1 con be consumed by 3.3.0

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/Player.scala
```diff
@@ -215,8 +215,9 @@ private[pekko] class ClientFSM(name: RoleName, controllerAddr: InetSocketAddress
     case Event(msg: NetworkOp, _) =>
       log.error("received {} instead of Done", msg)
       goto(Failed)
-    case Event(_: ServerOp, _) =>
-      stay().replying(Status.Failure(new IllegalStateException("not connected yet")))
+    // Scala 3.3.1 compiler says this is unreachable
+    // case Event(_: ServerOp, _) =>
+    //  stay().replying(Status.Failure(new IllegalStateException("not connected yet")))
```
He-Pin, line 220: I checked it and I think the compiler is right, reported it at https://github.com/lampepfl/dotty/issues/18519 too

He-Pin: I think this line can be removed, the compiler is right.

He-Pin: Do a swap
```scala
    case Event(msg: ServerOp, _) =>
      log.error("received {} instead of Done", msg)
      goto(Failed)
    case Event(_: NetworkOp, _) =>
      stay().replying(Status.Failure(new IllegalStateException("not connected yet")))
```

He-Pin:
Why github close this...

He-Pin (approved):
Lgtm

PR #635, opened 2023-09-06
=test Switch pattern matching cases in Player.

He-Pin:
A seprated PR to fix https://github.com/apache/incubator-pekko/pull/632

As the all `ServerOp` extends `NetworkOp`


pjfanning (changes requested):
you should move the code associated with the case

pjfanning (approved):
lgtm

mdedetrich (approved):
lgtm

PR #636, opened 2023-09-06
protobuf-java 3.19.6

pjfanning:
* try to use a newer jar - we use newer versions of protobuf-java already in grpc and connectors projects
* Using the latest protobuf-java jars have caused issues when testing in other repos - so the aim is to try to do smaller upgrades and see how they work out as we test them across the ecosystem


He-Pin (approved):
Lgtm.

I'm using 3.21 at my work

PR #638, opened 2023-09-08
=remote Clean up unused field.

He-Pin:
Those fields are unused.

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
@@ -14,7 +14,7 @@
 package org.apache.pekko.remote.transport.netty
 
 import java.net.{ InetAddress, InetSocketAddress, SocketAddress }
-import java.util.concurrent.{ CancellationException, ConcurrentHashMap, Executors }
+import java.util.concurrent.{ CancellationException, Executors }
 import java.util.concurrent.atomic.AtomicInteger
 
 import scala.annotation.nowarn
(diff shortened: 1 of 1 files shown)


remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -184,8 +182,6 @@ class NettyTransportSettings(config: Config) {
     case value => value.toInt
   }
 
-  val SslSettings: Option[SSLSettings] = if (EnableSsl) Some(new SSLSettings(config.getConfig("security"))) else None
```
pjfanning, line 187: strictly speaking: we can't remove a public field without deprecating it first

He-Pin: let's keep it then.

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -514,11 +507,7 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
       address <- addressToSocketAddress(Address("", "", settings.BindHostname, bindPort))
     } yield {
       try {
-        val newServerChannel = inboundBootstrap match {
-          case b: ServerBootstrap         => b.bind(address)
-          case b: ConnectionlessBootstrap => b.bind(address)
-          case _                          => throw new IllegalStateException() // won't happen, compiler exhaustiveness check pleaser
-        }
```
pjfanning, line 521: This new code changes the logic. How are you so sure that inboundBootstrap can never be a ConnectionlessBootstrap? I prefer the old code, to be frank.

He-Pin: yes, but there will never has a ConnectionlessBootstrap , because udp support is completed removed.

He-Pin: I think it was let there because of out of sight.

pjfanning: I made this change - #639 

Would that be tidier?

He-Pin: Both ok, i did not want to change the return type, this can rebase on yours

pjfanning (approved):
lgtm

mdedetrich (approved):
lgtm

PR #640, opened 2023-09-08
link validation job fails

pjfanning:
* broken by recent upgrade to scala 2.13.12 (#637)
* also an issue with rate limiting on https://dl.acm.org/


He-Pin (approved):
Lgtm

PR #648, opened 2023-09-14
Update to SBT 1.9.6 and use Apache Staging repo constant

mdedetrich:
SBT 1.9.5. added the `Resolver.ApacheMavenStagingRepo` constant

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,8 +24,7 @@ commands := commands.value.filterNot { command =>
   }
 }
 
-ThisBuild / reproducibleBuildsCheckResolver :=
-  "Apache Pekko Staging".at("https://repository.apache.org/content/groups/staging/")
+ThisBuild / reproducibleBuildsCheckResolver := Resolver.ApacheMavenStagingRepo
 
 enablePlugins(
   UnidocRoot,
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.9.4
+sbt.version=1.9.6


nvollmar (approved):
lgtm

He-Pin (approved):
lgtm
```
[info] compiling 189 Scala sources and 3 Java sources to /home/runner/work/incubator-pekko/incubator-pekko/stream/target/scala-2.12/classes ...
[error] /home/runner/work/incubator-pekko/incubator-pekko/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/GzipDecompressor.scala:29:52: Generated class org.apache.pekko.stream.impl.io.compression.createLogic$$anon differs only in case from org.apache.pekko.stream.impl.io.compression.createLogic$$anon (defined in DeflateDecompressor.scala).
[error]   Such classes will overwrite one another on case-insensitive filesystems.
[error]   override def createLogic(attr: Attributes) = new DecompressorParsingLogic {
[error]                                                    ^
[error] /home/runner/work/incubator-pekko/incubator-pekko/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/GzipDecompressor.scala:40:26: Generated class org.apache.pekko.stream.impl.io.compression.createLogic$$anon$inflating$ differs only in case from org.apache.pekko.stream.impl.io.compression.createLogic$$anon$inflating$ (defined in DeflateDecompressor.scala).
[error]   Such classes will overwrite one another on case-insensitive filesystems.
[error]     override case object inflating extends Inflate(false) with Step
[error]                          ^
[error] two errors found
```

mdedetrich:
@He-Pin I just noticed the error, I don't think this PR is causing it did someone forget to update mima filters or something?

pjfanning:
> @He-Pin I just noticed the error, I don't think this PR is causing it did someone forget to update mima filters or something?

The CI job fails before even doing the mima check. There must be at least a chance that sbt 1.9.5 and something that it brings in (eg an upgraded copy of Zinc) are leading to this. It could be a real issue that has been hidden before but it could just be a bug in the compiler.

mdedetrich:
> > @He-Pin I just noticed the error, I don't think this PR is causing it did someone forget to update mima filters or something?
> 
> The CI job fails before even doing the mima check. There must be at least a chance that sbt 1.9.5 and something that it brings in (eg an upgraded copy of Zinc) are leading to this. It could be a real issue that has been hidden before but it could just be a bug in the compiler.

SBT 1.9.5 did update zinc along with a new compiler bridge for Scala 2.13.x. Ill bring this upstream

mdedetrich:
@sethtisue @eed3si9n  @lrytz  I am not sure if this is an issue with zinc, sbt or scala 2 compiler itself (with the update to the new compiler bridge)

Asking because I need to know where to report this.

SethTisue:
The location of the root cause isn't obvious to me. I would say start in https://github.com/sbt/sbt/issues and we'll figure it out from there.

I'll run the sbt 1.9.5 upgrade through the Scala 2 community build and see if anything breaks.

He-Pin:
@mdedetrich Update to 1.9.6~

mdedetrich:
On it

pjfanning (approved):
lgtm

He-Pin (approved):
Lgtm

mdedetrich:
Everythings looking good, merging now

PR #651, opened 2023-09-15
=remote Make ues of Netty's default resolver.

He-Pin:
refs: https://github.com/akka/akka/issues/12960


He-Pin:
Need a tweak around the bind operation, have no power adapter around, will update later.

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -462,23 +462,15 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
...
-    case _ => Future.failed(new IllegalArgumentException(s"Address [$addr] does not contain host or port information."))
-  }
-
   override def listen: Future[(Address, Promise[AssociationEventListener])] = {
     @nowarn("msg=deprecated")
     val bindPort = settings.BindPortSelector
 
     for {
-      address <- addressToSocketAddress(Address("", "", settings.BindHostname, bindPort))
+      _ <- Future.unit
```
jxnu-liguobin, line 470: It no longer needs `for`, you don't even need a unit.

He-Pin: That will invoke a larger range of rewrite.

He-Pin: As you see, the diff is larger.

He-Pin:
@jxnu-liguobin @mdedetrich I have updated the code with 4 more lines changes , but drop the `Future.unit` usage totally.

He-Pin:
@pjfanning @mdedetrich This isn ready

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -462,23 +462,12 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
 
   override def isResponsibleFor(address: Address): Boolean = true // TODO: Add configurable subnet filtering
 
-  // TODO: This should be factored out to an async (or thread-isolated) name lookup service #2960
-  def addressToSocketAddress(addr: Address): Future[InetSocketAddress] = addr match {
```
pjfanning, line 466: This is a public method on a public class - I'm very surprised the mima check didn't fail - and I'm still suspicious that we are not checking mima properly for our PRs. All in all, I'm not sure that this method should be removed - it should probably be deprecated first.

pjfanning (changes requested):
not sure about removing public methods

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -511,28 +500,34 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
...
-    }
+    })
   }
 
   // Need to do like this for binary compatibility reasons
   private[pekko] def boundAddress = boundTo
 
+  private def extractHostAndPort(addr: Address): (String, Int) = addr match {
+    case Address(_, _, Some(host), Some(port)) => (host, port)
+    case _                                     => throw new IllegalArgumentException(s"Address [$addr] does not contain host or port information.")
```
pjfanning, line 520: this message is wrong - should say `host and port`

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -511,28 +500,34 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
...
+
   override def associate(remoteAddress: Address): Future[AssociationHandle] = {
     if (!serverChannel.isActive) Future.failed(new NettyTransportException("Transport is not bound"))
     else {
       val bootstrap: ClientBootstrap = outboundBootstrap(remoteAddress)
 
       (for {
-        socketAddress <- addressToSocketAddress(remoteAddress)
-        readyChannel <- NettyFutureBridge(bootstrap.connect(socketAddress)).map { channel =>
+        (host, port) <- Future.fromTry(Try(extractHostAndPort(remoteAddress)))
```
pjfanning, line 529: this seems bad to wrap the extractHostAndPort in a Try and a Future. Why do this call before the for comprehension so you won't need to wrap the result.

He-Pin: Because it throws

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
He-Pin: Ok, let's keep it :(

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
He-Pin: it should be `or`, because , if the port is not there, then the first case will not match.

pjfanning: I disagree. The 1st case will only match if you have both host and port. In English `host or port` would mean that you only 1 of the 2.

pjfanning (approved):
lgtm

PR #652, opened 2023-09-16
Remove blocking

He-Pin:
Motivation:
Less blocking after migrated to netty 4

based on https://github.com/apache/incubator-pekko/pull/651

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
@@ -68,6 +68,17 @@ object NettyFutureBridge {
     p.future
   }
 
+  private[transport] def apply[T](nettyFuture: io.netty.util.concurrent.Future[T]): Future[T] = {
+    val p = Promise[T]()
+    nettyFuture.addListener((future: io.netty.util.concurrent.Future[T]) =>
+      p.complete(
+        Try(
+          if (future.isSuccess) future.get()
+          else if (future.isCancelled) throw new CancellationException
+          else throw future.cause())))
+    p.future
+  }
+
   def apply(nettyFuture: ChannelGroupFuture): Future[ChannelGroup] = {
     import pekko.util.ccompat.JavaConverters._
     val p = Promise[ChannelGroup]()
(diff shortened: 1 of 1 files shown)


remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -525,21 +530,22 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
   // Need to do like this for binary compatibility reasons
   private[pekko] def boundAddress = boundTo
 
+  private def extractHostAndPort(addr: Address): (String, Int) = addr match {
+    case Address(_, _, Some(host), Some(port)) => (host, port)
+    case _                                     => throw new IllegalArgumentException(s"Address [$addr] does not contain host or port information.")
```
pjfanning, line 535: should be
```
Address [$addr] does not contain host and port information.
```

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
```diff
@@ -462,23 +473,15 @@ class NettyTransport(val settings: NettyTransportSettings, val system: ExtendedA
...
-    case _ => Future.failed(new IllegalArgumentException(s"Address [$addr] does not contain host or port information."))
-  }
-
   override def listen: Future[(Address, Promise[AssociationEventListener])] = {
     @nowarn("msg=deprecated")
     val bindPort = settings.BindPortSelector
 
     for {
-      address <- addressToSocketAddress(Address("", "", settings.BindHostname, bindPort))
+      _ <- Future.unit
```
mdedetrich, line 481: Is there a reason why we are using `Future.unit` like this, from what I can tell you are just lifting a value into `Future` which makes the `for` comprehension a bit redundant (i.e. you would use Future.apply`/`Future.successful`/`Future.failed` instead)?

He-Pin: That will change more code (4 lines more), I have done that before but then reverted.

mdedetrich: I personally have a preference for clearer code versus minimizing the diff, especially for something as trivial as 4 lines.

To me the `Future.unit` in a for comprehension by itself is a case of code smell, i.e. no one would write this code from a clean slate/starting point.

Also if you use something like `Future.fromTry` (which I think is the correct constructor to be used here) it would also be faster since putting inside `for` invokes a call to an executionContext

He-Pin: @mdedetrich see another PR, I will rebase this pr once https://github.com/apache/incubator-pekko/pull/651 get merged.

mdedetrich: Ah I see what you mean now, nvm then.

pjfanning (approved):
lgtm

He-Pin:
Thanks for the review, and then another blocking is removed:)

Issue #661, opened 2023-09-20
ForkJoinPoolStarvationSpec


He-Pin:
```
- must not starve tasks arriving from external dispatchers under high internal traffic *** FAILED *** (6 seconds, 14 milliseconds)
[09-20 10:18:16.027] [info]   java.lang.AssertionError: assertion failed: timeout (6 seconds) during expectMsg while waiting for All fine
[09-20 10:18:16.027] [info]   at scala.Predef$.assert(Predef.scala:223)
[09-20 10:18:16.027] [info]   at org.apache.pekko.testkit.TestKitBase.expectMsg_internal(TestKit.scala:472)
[09-20 10:18:16.028] [info]   at org.apache.pekko.testkit.TestKitBase.expectMsg(TestKit.scala:449)
[09-20 10:18:16.028] [info]   at org.apache.pekko.testkit.TestKitBase.expectMsg$(TestKit.scala:449)
[09-20 10:18:16.028] [info]   at org.apache.pekko.testkit.TestKit.expectMsg(TestKit.scala:982)
[09-20 10:18:16.028] [info]   at org.apache.pekko.dispatch.ForkJoinPoolStarvationSpec.$anonfun$new$6(ForkJoinPoolStarvationSpec.scala:78)
[09-20 10:18:16.028] [info]   at org.apache.pekko.dispatch.ForkJoinPoolStarvationSpec.$anonfun$new$6$adapted(ForkJoinPoolStarvationSpec.scala:75)
[09-20 10:18:16.028] [info]   at scala.collection.immutable.Range.foreach(Range.scala:158)
[09-20 10:18:16.028] [info]   at org.apache.pekko.dispatch.ForkJoinPoolStarvationSpec.$anonfun$new$2(ForkJoinPoolStarvationSpec.scala:75)
[09-20 10:18:16.029] [info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[09-20 10:18:16.029] [info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[09-20 10:18:16.029] [info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[09-20 10:18:16.029] [info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[09-20 10:18:16.029] [info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[...]

PR #666, opened 2023-09-20
POC =act Make use of externalSubmit to yield.

He-Pin:
refs: https://github.com/apache/incubator-pekko/issues/661

Some previous work done by @jrudolph years ago?

The Current VirtualThrad in JDK 21 making use of the `externalSubmit`
```java
if (s == YIELDING) {   // Thread.yield
            setState(RUNNABLE);

            // notify JVMTI that unmount has completed, thread is runnable
            notifyJvmtiUnmount(/*hide*/false);

            // external submit if there are no tasks in the local task queue
            if (currentThread() instanceof CarrierThread ct && ct.getQueuedTaskCount() == 0) {
                externalSubmitRunContinuation(ct.getPool());
            } else {
                submitRunContinuation();
            }
        }
```

I think this change will affect the performance.

I think a better way should be explicit yield after some run, just as the `cede` of CE, which @armanbilge metioned.

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinPoolStarvationSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinPoolStarvationSpec.scala
@@ -18,7 +18,6 @@ import com.typesafe.config.ConfigFactory
 import org.apache.pekko
 import pekko.actor.{ Actor, Props }
 import pekko.testkit.{ ImplicitSender, PekkoSpec }
-import pekko.util.JavaVersion
 
 object ForkJoinPoolStarvationSpec {
   val config = ConfigFactory.parseString("""
@@ -63,8 +62,8 @@ class ForkJoinPoolStarvationSpec extends PekkoSpec(ForkJoinPoolStarvationSpec.co
 
     "not starve tasks arriving from external dispatchers under high internal traffic" in {
       // TODO issue #31117: starvation with JDK 17 FJP
-      if (JavaVersion.majorVersion >= 17)
-        pending
+//      if (JavaVersion.majorVersion >= 17)
+//        pending
 
       // Two busy actors that will occupy the threads of the dispatcher
       // Since they submit to the local task queue via fork, they can starve external submissions
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
@@ -99,8 +99,9 @@ object ActorSystemSpec {
       override protected[pekko] def registerForExecution(
           mbox: Mailbox,
           hasMessageHint: Boolean,
-          hasSystemMessageHint: Boolean): Boolean = {
-        val ret = super.registerForExecution(mbox, hasMessageHint, hasSystemMessageHint)
+          hasSystemMessageHint: Boolean,
+          needYield: Boolean): Boolean = {
+        val ret = super.registerForExecution(mbox, hasMessageHint, hasSystemMessageHint, needYield)
         doneIt.switchOn {
           TestKit.awaitCond(mbox.actor.actor != null, 1.second)
           mbox.actor.actor match {
--- a/actor/src/main/scala/org/apache/pekko/dispatch/AbstractDispatcher.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/AbstractDispatcher.scala
@@ -72,6 +72,14 @@ private[pekko] trait LoadMetrics { self: Executor =>
 /**
  * INTERNAL API
  */
+private[pekko] trait LazyExecuteSupport {
+  self: Executor =>
+  def lazyExecute(runnable: Runnable): Unit = self.execute(runnable)
+}
+
+/**
+ * INTERNAL API
+ */
 private[pekko] object MessageDispatcher {
   val UNSCHEDULED = 0 // WARNING DO NOT CHANGE THE VALUE OF THIS: It relies on the faster init of 0 in AbstractMessageDispatcher
   val SCHEDULED = 1
@@ -160,7 +168,7 @@ abstract class MessageDispatcher(val configurator: MessageDispatcherConfigurator
    */
   final def attach(actor: ActorCell): Unit = {
     register(actor)
-    registerForExecution(actor.mailbox, false, true)
+    registerForExecution(actor.mailbox, false, true, false)
   }
 
   /**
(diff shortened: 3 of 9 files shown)


actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinPoolStarvationSpec.scala
```diff
@@ -63,8 +62,8 @@ class ForkJoinPoolStarvationSpec extends PekkoSpec(ForkJoinPoolStarvationSpec.co
 
     "not starve tasks arriving from external dispatchers under high internal traffic" in {
       // TODO issue #31117: starvation with JDK 17 FJP
-      if (JavaVersion.majorVersion >= 17)
-        pending
+//      if (JavaVersion.majorVersion >= 17)
+//        pending
```
He-Pin, line 66: I tested it locally.

He-Pin:
@jrudolph Yes, I think the current VirtualThread has the same problem, which only yield to the submissionQueue when there is no works to do, and VirutalThread just do `lazySubmit` for most the time.

It very hard to make things right, the code changes from JDK release to release, and the current impl will hurt performance.

refs: https://github.com/typelevel/cats-effect/discussions/3070
refs: https://github.com/typelevel/cats-effect/releases

I asked @armanbilge , he said the WSTP can be used outside of CE too.

pjfanning:
Anyone know if there is anything in the sbt ecosystem to support [Multi-Release jars](https://www.baeldung.com/java-multi-release-jar). That would allow us to have tailored versions of some classes that are used when you use a particular JDK/JRE version at runtime. This would be more efficient than having one version of the code and checking the JDK/JRE version using if stataments.

He-Pin:
@pjfanning I asked @Glavo once and she said the toolchian is not very well

mdedetrich:
> Anyone know if there is anything in the sbt ecosystem to support [Multi-Release jars](https://www.baeldung.com/java-multi-release-jar). That would allow us to have tailored versions of some classes that are used when you use a particular JDK/JRE version at runtime. This would be more efficient than having one version of the code and checking the JDK/JRE version using if stataments.

Its a request feature that is yet to be implemented, see https://github.com/sbt/sbt/discussions/7174#discussioncomment-5288982. I was thinking of working on this at some point but there are already so many OS things that I am working on flying around atm.

I don't think it should be too hard, I think SBT already has functionality to detect multiple JDK's, what needs to be fleshed out is how the matrix will work, i.e. will we have `scala-jdk8` folders for sources only compiled for JDK 8 and will stuff like `scala-2-jdk8`/`scala-2.13+-jdk11` work?

pjfanning:
> I don't think it should be too hard, I think SBT already has functionality to detect multiple JDK's, what needs to be fleshed out is how the matrix will work, i.e. will we have `scala-java8` folders for sources only compiled for JDK 8 and will stuff like `scala-2-java8`/`scala-2.13+-java11` work?

Sounds reasonable but we might also need to support `src/main/java11` too (java code that should be used for Java 11+).

mdedetrich:
> Sounds reasonable but we might also need to support `src/main/java11` too (java code that should be used for Java 11+).

Yes thats implied

jrudolph:
One thing to keep in mind is that Pekko/Akka is already somewhat biased against fairness by choosing relatively big batching values both for the actor "throughput" setting and addition to that for the stream "sync processing limit". Therefore, a lot of the scheduling overhead that other effect systems might see because of pushing every single thunk through the general scheduler (fine-grained parallelism) might not apply to Pekko, especially when using streams. In general, this can lead to higher tail latencies in high load scenarios, though it depends a lot on actual setup of streams etc.

In general, providing fairness to get more predictable latency is a worthwhile goal. My impression is that low-level optimizations will likely not cut it. A high-level reason is that under overload scenarios (the only ones where fairness is relevant), some work is more important than other (work which drives the control flow vs actual CPU intensive workloads). However you structure your work queues as long as you are not able to prioritize more important work against greedy CPU intensive workload tasks (which the one in this example is with a tight loop of actors sending messages back and forth), you will not be able to solve the issues of calculations getting stuck because control flow is starved of CPU resources. (See also, how tight loops can prevent GCs from happening). One approach is to bulk-head CPU heavy but low-priority work out of the general purpose queues that drive control flow (e.g. introduction of internal dispatcher, running CPU workloads out of the streams, creating dispatchers for blocking work, creating extra OS threads to push back fairness issues to the OS scheduler). Unfortunately, these kind of solutions require manual interventions and understanding of the systems.

mdedetrich:
> In general, providing fairness to get more predictable latency is a worthwhile goal. My impression is that low-level optimizations will likely not cut it. A high-level reason is that under overload scenarios (the only ones where fairness is relevant), some work is more important than other (work which drives the control flow vs actual CPU intensive workloads). However you structure your work queues as long as you are not able to prioritize more important work against greedy CPU intensive workload tasks (which the one in this example is with a tight loop of actors sending messages back and forth), you will not be able to solve the issues of calculations getting stuck because control flow is starved of CPU resources. (See also, how tight loops can prevent GCs from happening). One approach is to bulk-head CPU heavy but low-priority work out of the general purpose queues that drive control flow (e.g. introduction of internal dispatcher, running CPU workloads out of the streams, creating dispatchers for blocking work, creating extra OS threads to push back fairness issues to the OS scheduler). Unfortunately, these kind of solutions require manual interventions and understanding of the systems.

If I understand correctly is this basically asking for a higher level API approach where we can assigned certain priorities to flows within a pekko stream (I guess via `Attributes`) so that in cases of CPU starvation the pekko actor/stream runtime has knowledge of what is more critical to continue and what isn't?

jrudolph:
No, I just want to prevent that we do microoptimizations based on findings of different projects with potentially different premises without having real world problems to solve. ForkJoinPoolStarvationSpec  is imo not enough to warrant doing any changes here.

mdedetrich:
> No, I just want to prevent that we do microoptimizations based on findings of different projects with potentially different premises without having real world problems to solve. ForkJoinPoolStarvationSpec is imo not enough to warrant doing any changes here.

No disagreements here, I think I the incorrect impression (I thought you were speaking generally rather than this PR specifically).

Maybe we can make another general discussion thread if its relevant, but I do agree we need to be careful about these cases.

He-Pin (commented):
POC exploring use of `externalSubmit` for yield in ForkJoinPool dispatcher.

1. **Exploratory nature**: This is clearly marked as a POC. The author notes that JDK 21's VirtualThread already uses `externalSubmit` for yield, suggesting a potential performance improvement.

2. **Performance concern**: The author themselves notes 'I think this change will affect the performance' -- meaning it could be negative. Benchmarking would be essential before any production use.

3. **Alternative approach**: The author suggests an explicit yield after some runs (like Cats Effect's `cede`) might be a better approach, referencing @armanbilge's suggestion.

4. **Staleness**: Open since September 2023 with no reviews. As a POC that the author themselves expressed performance concerns about, this should either be completed with benchmarks or closed. Given the age, closing and revisiting with a fresh approach would likely be best.

He-Pin:
Closing this POC. As discussed, there isn't a real-world problem to solve here that warrants changes to the dispatcher scheduling, and @jrudolph noted the risk of negative throughput impact for marginal fairness gains in specific cases. If someone wants to revisit this with concrete benchmarks showing a realistic workload benefit, a fresh PR can be opened.

Issue #669, opened 2023-09-21
backoff supervision lost stash message


Roiocam:
The original issue on: https://stackoverflow.com/questions/75065817/akka-why-are-stashed-messages-with-backoff-supervision-lost

I have been verify it, seem like this is a bug.

PR #670, opened 2023-09-21
fix: keep stashBuffer when exception throw again #669

Roiocam:
References #669

after investigate, the reason why message log is `Supervisor` clear stash immediately(L401) while actor restart, no matter the new actor failure again or not.

It is no issue on classic BackoffOnRestartSupervisor, because they don't have StashBuffer. 

https://github.com/apache/incubator-pekko/blob/e94e7b971b74eeadc05a5b3dc28d1e17c434ad7f/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/Supervision.scala#L397-L411

### DIFF
--- a/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
+++ b/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
@@ -1161,7 +1161,8 @@ class SupervisionSpec extends ScalaTestWithActorTestKit("""
       }
     }
 
-    "ensure unhandled message retention during consecutive restarts" in new FailingRestartBackoffTestSetup(cntValue = 5) {
+    "ensure unhandled message retention during consecutive restarts" in new FailingRestartBackoffTestSetup(
+      cntValue = 5) {
       val exception = new RuntimeException("mockException")
       val msg: Throw = Throw(exception)
       val ref = spawn(behv)
--- a/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
+++ b/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
@@ -66,14 +66,15 @@ object SupervisionSpec {
   def targetBehavior(
       monitor: ActorRef[Event],
       state: State = State(0, Map.empty),
-      slowStop: Option[CountDownLatch] = None): Behavior[Command] =
+      slowStop: Option[CountDownLatch] = None,
+      slowRestart: Option[CountDownLatch] = None): Behavior[Command] =
     receive[Command] { (context, cmd) =>
       cmd match {
         case Ping(n) =>
           monitor ! Pong(n)
           Behaviors.same
         case IncrementState =>
-          targetBehavior(monitor, state.copy(n = state.n + 1), slowStop)
+          targetBehavior(monitor, state.copy(n = state.n + 1), slowStop, None)
         case GetState =>
           val reply = state.copy(children = context.children.map(c => c.path.name -> c.unsafeUpcast[Command]).toMap)
           monitor ! reply
@@ -94,6 +95,8 @@ object SupervisionSpec {
       case (_, sig) =>
         if (sig == PostStop)
           slowStop.foreach(latch => latch.await(10, TimeUnit.SECONDS))
+        else if (sig == PreRestart)
+          slowRestart.foreach(latch => latch.await(10, TimeUnit.SECONDS))
         monitor ! ReceivedSignal(sig)
         Behaviors.same
     }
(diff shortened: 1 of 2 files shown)


actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/Supervision.scala
```diff
@@ -398,8 +404,10 @@ private class RestartSupervisor[T, Thr <: Throwable: ClassTag](initial: Behavior
       val newBehavior = Behavior.validateAsInitial(Behavior.start(initial, ctx.asInstanceOf[TypedActorContext[T]]))
       val nextBehavior = restartingInProgress match {
         case OptionVal.Some((stashBuffer, _)) =>
+          val behavior = stashBuffer.unstashAll(newBehavior.unsafeCast)
+          // reset stash state if not more exception throw
           restartingInProgress = OptionVal.None
-          stashBuffer.unstashAll(newBehavior.unsafeCast)
+          behavior
```
He-Pin, line 410: Only reset if the unstashAll processed success, otherwise the unprocessed messages will be lost right?

actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
```diff
@@ -1133,6 +1161,19 @@ class SupervisionSpec extends ScalaTestWithActorTestKit("""
       }
     }
 
+    "not msg lost on restart multiple times" in new FailingRestartBackoffTestSetup(cntValue = 5) {
```
He-Pin, line 1164: Doesn't lost stashBuffer...

actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
```diff
@@ -1133,6 +1161,19 @@ class SupervisionSpec extends ScalaTestWithActorTestKit("""
...
+      val exception = new RuntimeException("mockException")
+      val msg: Throw = Throw(exception)
+      val ref = spawn(behv)
+      for (_ <- 1 to cntValue) {
+        ref ! msg
+      }
+      latch.await(1000, TimeUnit.MILLISECONDS)
+      for (_ <- 1 to cntValue) {
+        probe.expectMessage(msg)
+      }
```
He-Pin, line 1174: this test doesn't test the behavior

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/Supervision.scala
Roiocam: Yes, because unhandled messages are stored in the StashBuffer within the `restartingInProgress` field.

actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
Roiocam: The "StashBuffer" is utilized only during the restart phase and can be cleared once the restart is successful and all messages have been processed. This PR essentially addresses the following issue:

- Reusing the existing "StashBuffer" object during the preparation of a restart to prevent unhandled message loss.
- Resetting the "restartingInProgress" object only after all unhandled messages have been processed following the completion of the restart, thus avoiding the loss of the "StashBuffer".

He-Pin: Yes, that's what I learnt from the code.

Roiocam: It is that two-issue mean `Restart` and `RestartBackoff` drop an unhandled message,
or "keep `restartingInProgress` before all messages handled" and "refill buffer on preperRestart".

I am struggling with this, and there is another case where, although retaining "restartingInProgress," buffer messages are still lost when encountering an "UnstashException" and the number of retries exceeds`maxRestart`.

He-Pin: When the restart time exceeded, drop the buffer seems fine.

Roiocam (commented):
PTAL, @He-Pin

actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
```diff
@@ -294,6 +297,44 @@ class SupervisionSpec extends ScalaTestWithActorTestKit("""
...
+    class FailingConstructor(context: ActorContext[Command], monitor: ActorRef[Event])
+        extends AbstractBehavior[Command](context) {
+      monitor ! Started
+      context.spawn(targetBehavior(probe.ref, slowStop = Some(slowStop)), nextName())
+      if (failCounter.getAndIncrement() < failCount) {
+        throw TestException("simulated exc from constructor")
+      }
+
+      override def onMessage(message: Command): Behavior[Command] = {
+        monitor ! Pong(0)
```
He-Pin, line 314: maybe better with a `Pong(counter++)`, it  always reply `Pong(0)`.

Roiocam: Makes sense, In this way, we can gain the guarantee of message ordering.

Roiocam:
@He-Pin I am sorry for that closed this PR for no reason, this PR is not only my contribution.

He-Pin (approved):
lgtm

PR #676, opened 2023-09-23
=str Tweak the stream mapAsyncPartitioned operator

He-Pin:
Motivation: 
1. Tweak the mapAsyncPartitioned  & mapAsyncPartitionedUnordered  document
2. Removed some methods which is used only once.
3. Add these method to SubSource and SubFlow.

refs: https://github.com/apache/incubator-pekko/pull/561

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
@@ -23,102 +23,29 @@ import scala.util.control.{ NoStackTrace, NonFatal }
 import scala.util.{ Failure, Success, Try }
 
 import org.apache.pekko
+import pekko.annotation.InternalApi
 import pekko.dispatch.ExecutionContexts
 import pekko.stream.ActorAttributes.SupervisionStrategy
-import pekko.stream.Attributes.{ Name, SourceLocation }
-import pekko.stream.MapAsyncPartitioned._
-import pekko.stream.scaladsl.{ Flow, FlowWithContext, Source, SourceWithContext }
 import pekko.stream.stage._
+import pekko.util.OptionVal
 
+/**
+ * Internal API
+ */
+@InternalApi
 private[stream] object MapAsyncPartitioned {
+  private val NotYetThere = Failure(new Exception with NoStackTrace)
 
-  private def extractPartitionWithCtx[In, Ctx, Partition](extract: In => Partition)(tuple: (In, Ctx)): Partition =
-    extract(tuple._1)
-
-  private def fWithCtx[In, Out, Ctx, Partition](f: (In, Partition) => Future[Out])(tuple: (In, Ctx),
-      partition: Partition): Future[(Out, Ctx)] =
-    f(tuple._1, partition).map(_ -> tuple._2)(ExecutionContexts.parasitic)
-
-  def mapSourceOrdered[In, Out, Partition, Mat](source: Source[In, Mat], parallelism: Int)(
-      extractPartition: In => Partition)(
-      f: (In, Partition) => Future[Out]): Source[Out, Mat] =
-    source.via(new MapAsyncPartitioned[In, Out, Partition](orderedOutput = true, parallelism, extractPartition, f))
-
-  def mapSourceUnordered[In, Out, Partition, Mat](source: Source[In, Mat], parallelism: Int)(
-      extractPartition: In => Partition)(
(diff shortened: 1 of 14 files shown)


stream/src/main/scala/org/apache/pekko/stream/javadsl/FlowWithContext.scala
```diff
@@ -173,39 +173,36 @@ final class FlowWithContext[In, CtxIn, Out, CtxOut, +Mat](
...
+    viaScala(_.mapAsyncPartitioned(parallelism)(partitioner(_))(f(_, _).asScala))
   }
 
   /**
-   * Transforms this stream. Works very similarly to [[#mapAsyncUnordered]] but with an additional
-   * partition step before the transform step. The transform function receives the an individual
-   * stream entry and the calculated partition value for that entry.
+   * Context-preserving variant of [[pekko.stream.javadsl.Flow.mapAsyncPartitionedUnordered]].
    *
-   * @since 1.1.0
```
pjfanning, line 201: Why remove the since markers? I think they are very useful.

He-Pin: Oops, I will add it back.

pjfanning:
Can you update the `.md` files for these 2 operations? If we are calling the partition function `partitioner` then that change should be in the `.md` files too.

stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
```diff
@@ -23,102 +23,29 @@ import scala.util.control.{ NoStackTrace, NonFatal }
...
-
-  private[stream] final class Holder[In, Out](
-      val in: In,
-      var out: Try[Out],
-      callback: AsyncCallback[Holder[In, Out]]) extends (Try[Out] => Unit) {
-
-    // To support both fail-fast when the supervision directive is Stop
-    // and not calling the decider multiple times (#23888) we need to cache the decider result and re-use that
-    private var cachedSupervisionDirective: Option[Supervision.Directive] = None
+  private final class Holder[In, Out](val in: In, var out: Try[Out], val cb: AsyncCallback[Holder[In, Out]]) extends (
```
pjfanning, line 39: could we use the name `callback` instead of `cb`?

jaceksokol:
> @jaceksokol could you have a look? The changes generally seem reasonable to me. I intend to do a more thorough check within the next few days.

Most of the changes are related to code style / structure - LGTM.

stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
He-Pin: - I changed it to `cb` as the same in `Holder` in `MapAsync`.
- the `cb` is ok, Cats-Effect use `cb` in unsafeRunAsync too, maybe that's fine to keep it?

He-Pin:
@pjfanning It is `partitioner` in the `mapAsyncPartitioned.md`

He-Pin:
@jaceksokol Thanks, I also add a description about the behavior, which the max parallelism of per partition is `1`, can you confirm that too:)

stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
```diff
@@ -128,27 +55,32 @@ private[stream] object MapAsyncPartitioned {
...
+ */
+@InternalApi
+private[stream] final class MapAsyncPartitioned[In, Out, Partition](
     parallelism: Int,
-    extractPartition: In => Partition,
+    orderedOutput: Boolean,
+    partitioner: In => Partition,
     f: (In, Partition) => Future[Out]) extends GraphStage[FlowShape[In, Out]] {
+  require(parallelism >= 1, "parallelism must be at least 1")
+  require(partitioner != null, "partitioner function should not be null")
```
pjfanning, line 75: Do we need to do the null checks? They have a small overhead.

He-Pin: This should be fine, because it only occur at dsl build time not run time, otherwise it will throw NPE when the stage is running

mdedetrich: Indeed this is correct, they will only occur once when DSL is building so its fine.

pjfanning (approved):
lgtm

PR #677, opened 2023-09-23
Make use of SAM instead.

He-Pin:
Motivation:
Make use of SAM.

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/ActorContextImpl.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/ActorContextImpl.scala
@@ -284,7 +284,7 @@ import scala.util.Success
   def pipeToSelf[Value](
       future: CompletionStage[Value],
       applyToResult: pekko.japi.function.Function2[Value, Throwable, T]): Unit = {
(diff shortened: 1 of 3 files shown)


actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/ActorContextImpl.scala
```diff
@@ -284,7 +284,7 @@ import scala.util.Success
   def pipeToSelf[Value](
       future: CompletionStage[Value],
       applyToResult: pekko.japi.function.Function2[Value, Throwable, T]): Unit = {
-    future.whenComplete { (value, ex) =>
+    future.handle[Unit] { (value, ex) =>
```
Roiocam, line 287: Is it necessary to make changes? If so, why hasn't PipeToSupport made improvements in this regard?

He-Pin: Yes, `handle` is faster than `whenComplete`, especially in the failed case.

mdedetrich (approved):
lgtm. Nice improvement especially since we don't support JDK versions lower than 8/Scala versions lower than 2.12

PR #680, opened 2023-09-24
feat: Immutable builder for interceptor and supervisor

Roiocam:
Draft because I haven't verified everything and I am not sure if this is the optimal approach.

Example usage could be like this:

```scala
val rawBehavior: Behavior[Command] = Behaviors.empty
val interceptor : BehaviorInterceptor[Command,Command] = ???
val monitorProbe = TestProbe[Command]("probe")

val behavior: Behavior[Command] = Behaviors.interceptorBuilder(rawBehavior)
  .onFailure[Exc1](SupervisorStrategy.restart.withStashCapacity(100))
  .onFailure[Exc2](SupervisorStrategy.restart.withStashCapacity(100))
  .intercept[Command](() => interceptor)
  .onFailure[Exc3](SupervisorStrategy.restart.withStashCapacity(100))
  .monitor(monitorProbe.ref)
  .build()
```

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/scaladsl/InterceptorBuilder.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/scaladsl/InterceptorBuilder.scala
@@ -18,10 +18,12 @@
 package org.apache.pekko.actor.typed.scaladsl
 
 import org.apache.pekko
+import org.apache.pekko.actor.typed.ActorRef
 import org.apache.pekko.actor.typed.Behavior
 import org.apache.pekko.actor.typed.BehaviorInterceptor
 import org.apache.pekko.actor.typed.SupervisorStrategy
 import org.apache.pekko.actor.typed.internal.BehaviorImpl
+import org.apache.pekko.actor.typed.internal.MonitorInterceptor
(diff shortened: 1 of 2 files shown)


actor-typed/src/main/scala/org/apache/pekko/actor/typed/scaladsl/InterceptorBuilder.scala
```diff
@@ -0,0 +1,81 @@
...
+    new InterceptorBuilder(BehaviorImpl.intercept(behaviorInterceptor)(currentBehavior))
+
+  /**
+   * Behavior decorator that copies all received message to the designated
+   * monitor [[pekko.actor.typed.ActorRef]] before invoking the wrapped behavior. The
+   * wrapped behavior can evolve (i.e. return different behavior) without needing to be
+   * wrapped in a `monitor` call again.
+   * The simple way of as [[Behaviors.monitor]]
+   */
+  def monitor(monitor: ActorRef[T]): InterceptorBuilder[T] = {
```
Roiocam, line 69: After incrementing some additional methods, I have discovered that most of the methods can be flattened using this builder. It seems to be more like a `BehaviorBuilder` rather than an `InterceptorBuilder`.

PR #682, opened 2023-09-25
Fix invalid scaladoc links which cannot be found (#353)

ydash:
Refs #353 

I fixed invalid scaladoc links.
* Add enough package name
* Fix invalid syntax of links to a method

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/ActorRef.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/ActorRef.scala
@@ -49,8 +49,8 @@ trait ActorRef[-T] extends RecipientRef[T] with java.lang.Comparable[ActorRef[_]
   /**
    * Unsafe utility method for widening the type accepted by this ActorRef;
    * provided to avoid having to use `asInstanceOf` on the full reference type,
-   * which would unfortunately also work on non-ActorRefs. Use it with caution,it may cause a [[ClassCastException]] when you send a message
-   * to the widened [[ActorRef[U]]].
+   * which would unfortunately also work on non-ActorRefs. Use it with caution,it may cause a [[java.lang.ClassCastException]] when you send a message
+   * to the widened [[ActorRef ActorRef[U]]].
    */
   def unsafeUpcast[U >: T @uncheckedVariance]: ActorRef[U]
 
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/Scheduler.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/Scheduler.scala
@@ -34,7 +34,7 @@ trait Scheduler {
    * Scala API: Schedules a Runnable to be run once with a delay, i.e. a time period that
    * has to pass before the runnable is executed.
    *
-   * @throws IllegalArgumentException if the given delays exceed the maximum
+   * @throws java.lang.IllegalArgumentException if the given delays exceed the maximum
    * reach (calculated as: `delay / tickNanos > Int.MaxValue`).
    *
    * Note: For scheduling within actors `Behaviors.withTimers` or `ActorContext.scheduleOnce` should be preferred.
@@ -45,7 +45,7 @@ trait Scheduler {
    * Java API: Schedules a Runnable to be run once with a delay, i.e. a time period that
    * has to pass before the runnable is executed.
    *
-   * @throws IllegalArgumentException if the given delays exceed the maximum
+   * @throws java.lang.IllegalArgumentException if the given delays exceed the maximum
    * reach (calculated as: `delay / tickNanos > Int.MaxValue`).
    *
    * Note: For scheduling within actors `Behaviors.withTimers` or `ActorContext.scheduleOnce` should be preferred.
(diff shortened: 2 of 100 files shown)


actor-typed/src/main/scala/org/apache/pekko/actor/typed/ActorRef.scala
```diff
@@ -49,8 +49,8 @@ trait ActorRef[-T] extends RecipientRef[T] with java.lang.Comparable[ActorRef[_]
   /**
    * Unsafe utility method for widening the type accepted by this ActorRef;
    * provided to avoid having to use `asInstanceOf` on the full reference type,
-   * which would unfortunately also work on non-ActorRefs. Use it with caution,it may cause a [[ClassCastException]] when you send a message
-   * to the widened [[ActorRef[U]]].
+   * which would unfortunately also work on non-ActorRefs. Use it with caution,it may cause a [[java.lang.ClassCastException]] when you send a message
+   * to the widened [[ActorRef ActorRef[U]]].
```
He-Pin, line 53: this seems invalid?

actor/src/main/java/org/apache/pekko/japi/pf/ReceiveBuilder.java
```diff
@@ -19,7 +19,7 @@
 
 /**
  * Used for building a partial function for {@link
- * org.apache.pekko.actor.AbstractActor#createReceive() AbstractActor.createReceive()}.
+ * org.apache.pekko.actor.AbstractActor#createReceive AbstractActor.createReceive()}.
```
He-Pin, line 22: this seems wrong too.

actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala
```diff
@@ -477,7 +477,7 @@ trait Scheduler {
    * Java API: Schedules a Runnable to be run once with a delay, i.e. a time period that
    * has to pass before the runnable is executed.
    *
-   * @throws IllegalArgumentException if the given delays exceed the maximum
+   * @throws java.lang.IllegalArgumentException if the given delays exceed the maximum
```
He-Pin, line 480: I think we can just import this , then there will be no `java.lang.`

actor-typed/src/main/scala/org/apache/pekko/actor/typed/Scheduler.scala
```diff
@@ -45,7 +45,7 @@ trait Scheduler {
    * Java API: Schedules a Runnable to be run once with a delay, i.e. a time period that
    * has to pass before the runnable is executed.
    *
-   * @throws IllegalArgumentException if the given delays exceed the maximum
+   * @throws java.lang.IllegalArgumentException if the given delays exceed the maximum
```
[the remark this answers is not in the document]
pjfanning: java.lang is imported implicity - same as the top level scala package. If scaladoc is not linking these java.lang classes unless you prefix them like this - then there seems to be a bug that we should report to scala team. I'll have to test out locally if this change actually improves the scaladoc or whether the issue is with links to Java classes generally.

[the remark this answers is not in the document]
He-Pin: Yes, if it can't lik, then its a bug of scaladoc.

actor-typed/src/main/scala/org/apache/pekko/actor/typed/ActorRef.scala
ydash: It's seems invalid.
Due to `[[ActorRef[U]]]`, following warn log is appeared:
```
[warn] /path/actor-typed/src/main/scala/org/apache/pekko/actor/typed/ActorRef.scala:49:3: Could not find any member to link for "ActorRef[U".
```

ydash:
> @ydash thanks, this is epic, but the changeset is large than I expected, will import these classes directly reduce the lines of change?

Probably, it's not possible.
If you import classes which are used in only scaladoc link, following compile error is appeared:
```
[error] /path/actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala:16:18: Unused import
[error] Applicable -Wconf / @nowarn filters for this fatal warning: msg=<part of the message>, cat=unused-imports, site=org.apache.pekko.actor
[error] import java.lang.IllegalArgumentException
[error]                  ^
[error] one error found
[error] (actor / Compile / compileIncremental) Compilation failed
```

He-Pin:
@ydash I think that should be a bug in Scaladoc, because in Java, we can do it. would you like to report it scala/bug ?

actor-typed/src/main/scala/org/apache/pekko/actor/typed/Scheduler.scala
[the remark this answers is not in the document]
ydash: In my environment, linking to classes under the top level `scala` package is also failed without those `scala.` prefix:

https://github.com/apache/incubator-pekko/pull/682/files#diff-d77b49e58aab175ed0d750b8b282f3ad01daab1ba88f38b3a310dc3c2e89e179L126

ydash:
@He-Pin You mean it is a bug that failures of linking to classes under `java.lang` package without the `java.lang.` prefix?

pjfanning:
> @He-Pin You mean it is a bug that failures of linking to classes under `java.lang` package without the `java.lang.` prefix?

Yes @ydash this is what is being requested. Scaladoc should not require us to hack the scaladoc like you are doing in this PR.

pjfanning (approved):
lgtm - I'm willing to merge this. I must admit to not wanting to run this on my laptop first because the build takes so long. If we merge this and it proves not to work well, we can revert it.

He-Pin:
@ydash If you don't have time, I will openup an issue in scala/bug.

He-Pin (approved):
lgtm, checked, we can make if simpler once scala fix this.

refs: https://github.com/scala/bug/issues/12884

pjfanning:
@ydash @He-Pin this change is the cause of #687 - I don't know if reverting some parts of this will be enough or if we would be better off doing a full revert and starting again. 

Scaladoc has so many bugs that it may not be worth our while trying to improve it.

Issue #663, opened 2023-09-20
LineNumberSpec fails with Scala 3


pjfanning:
* org.apache.pekko.util.LineNumberSpec
* https://github.com/apache/incubator-pekko/actions/runs/6242272017/job/16945926518
* tested only in nightly build
* has been failing a while
* I think we can relax the line number matching to make the test pass.

pjfanning:
#683 merged

PR #683, opened 2023-09-25
Update LineNumberSpec because it fails after Scala 3.3.1 upgrade

pjfanning:
see https://github.com/apache/incubator-pekko/issues/663


He-Pin (approved):
Lgtm

PR #688, opened 2023-09-28
revert some scaladoc changes because of scala 2.12 build issues

pjfanning:
I've tested this and this revert allows me to publish the pekko-cluster-sharding-typed jar locally while it fails without this revert.

Reverts part of #682

### DIFF
--- a/cluster-sharding-typed/src/main/scala/org/apache/pekko/cluster/sharding/typed/ReplicatedEntityProvider.scala
+++ b/cluster-sharding-typed/src/main/scala/org/apache/pekko/cluster/sharding/typed/ReplicatedEntityProvider.scala
@@ -34,7 +34,7 @@ object ReplicatedEntityProvider {
   /**
    * Java API:
    *
-   * Provides full control over the [[ReplicatedEntity]] and the [[javadsl.Entity]]
+   * Provides full control over the [[ReplicatedEntity]] and the [[Entity]]
    * Most use cases can use the [[createPerDataCenter]] and [[createPerRole]]
    *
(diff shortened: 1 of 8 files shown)


mdedetrich:
@pjfanning Wouldn't it be safer to use FQCN rather than just removing the prefix?

pjfanning:
> @pjfanning Wouldn't it be safer to use FQCN rather than just removing the prefix?

This is a partial revert of https://github.com/apache/incubator-pekko/pull/682 because our builds are broken.

Scaladoc - especially ScalaDod in Scala 2.12 has bugs. It is a minefield trying to get it right. My main objective is to fix the build. I can't get a build to test the sbt-assembly issue due to the broken build.

mdedetrich:
> This is a partial revert of #682 because our builds are broken.
> 
> Scaladoc - especially ScalaDoc in Scala 2.12 has bugs. It is a minefield trying to get it right. My main objective is to fix the build. I can't get a build to test the sbt-assembly issue due to the broken build.

I understand, but looking at the diff we are not changing from FQCN to a nested package prefix, rather we are changing from a nested packaged prefix to another nested package prefix.

I am quite sure that if you use a complete FQCN i.e. changing `[[javadsl.Entity]]` to `[[org.apache.pekko.cluster.sharding.typed.javadsl.Entity]]` (or w/e it *should * be) would actually remove all brittleness regarding the scaladoc classname lookup and help prevent future problems.

pjfanning:
closing in favour of #688

mdedetrich:
> closing in favour of #688

@pjfanning I think you forgot to close?

PR #689, opened 2023-09-28
Update Netty to 4.1.100.final

Roiocam:
## Motivation

Netty 4.1.98.Final was an unsafe version:

- crash the JVM when using JDK21+
- HTTP/2 Rapid Reset Attack 

## References

- https://mail.openjdk.org/pipermail/hotspot-gc-dev/2023-September/043696.html
- https://netty.io/news/2023/09/21/4-1-99-Final.html
- https://nvd.nist.gov/vuln/detail/CVE-2023-44487
- https://netty.io/news/2023/10/10/4-1-100-Final.html

### DIFF
--- a/docs/src/main/paradox/additional/osgi.md
+++ b/docs/src/main/paradox/additional/osgi.md
@@ -16,7 +16,7 @@ To use Apache Pekko in OSGi, you must add the following dependency in your proje
 ## Background
 
 [OSGi](https://www.osgi.org/resources/where-to-start/) is a mature packaging and deployment standard for component-based systems. It
-has similar capabilities as [Project Jigsaw](https://openjdk.java.net/projects/jigsaw/) (originally scheduled for JDK 1.8), but has far stronger facilities to
+has similar capabilities as [Project Jigsaw](https://openjdk.org/projects/jigsaw/) (originally scheduled for JDK 1.8), but has far stronger facilities to
 support legacy Java code. This is to say that while Jigsaw-ready modules require significant changes to most source files
 and on occasion to the structure of the overall application, OSGi can be used to modularize almost any Java code as far
 back as JDK 1.2, usually with no changes at all to the binaries.
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -28,7 +28,7 @@ object Dependencies {
   // needs to be inline with the aeron version, check
   // https://github.com/real-logic/aeron/blob/1.x.y/build.gradle
   val agronaVersion = "1.19.2"
-  val nettyVersion = "4.1.98.Final"
+  val nettyVersion = "4.1.99.Final"
   val protobufJavaVersion = "3.19.6"
   val logbackVersion = "1.2.11"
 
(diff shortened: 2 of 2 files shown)


pjfanning (changes requested):
-1 from me. This is the umpteenth recent PR for netty upgrades. This is time consuming and of little benefit. We want to set up an automated approach based on Scala Steward. This is a much better way to do this and than random manual PRs.

Roiocam:
> -1 from me. This is the umpteenth recent PR for netty upgrades. This is time consuming and of little benefit. We want to set up an automated approach based on Scala Steward. This is a much better way to do this and than random manual PRs.-1 

The responsibility for that should be with #56, rather than with this issue. Automating such version upgrades is far superior to manual PRs, so why #56 has not made further progress?

Roiocam:
@pjfanning you are one of Pekko PMC, you should take the responsibility to push on #56 and formulate a roadmap, Instead of being overly critical of the contributors' positive responses.

He-Pin:
I think this pr is fine

Even the current pekko is not affected. Which only happens when the native epoll transport is been used.

But it always better to fix critical issues like, and let users feel safe.


We are happy to announce the release of netty 4.1.99.Final. This release fixes a bug which could lead to SIGSERV while running on JDK21+. As JDK21 was recently released and is considered an LTS release we did decided to do this emergency release. Because of the nature of this bug (crashing the JVM) you should upgrade ASAP.


Netty suggests upgrading ASAP.

But yes, the scala steward is better but we can have this in before that.

@Roiocam Would you like to reopen it?

pjfanning:
Netty 4 support for Pekko has never had a release. We don't need to rush netty upgrades. All they affect are our CI runs. netty 4.1.100 will probably be out soon. We can just upgrade occasionally or once in the run up to a release.

He-Pin:
> Netty 4 support for Pekko has never had a release. We don't need to rush netty upgrades. All they affect are our CI runs. netty 4.1.100 will probably be out soon. We can just upgrade occasionally or once in the run up to a release.
> 
> Java 21 release is also a real mess because of a crazy last minute license change - see https://adoptium.net/en-GB/blog/2023/09/temurin21-delay/

Good to know, but yes, not to rush means we have a clean commits line too.

pjfanning:
4.1.100.Final is out but this is a security release so we will want to update

https://nvd.nist.gov/vuln/detail/CVE-2023-44487

Roiocam:
> 4.1.100.Final is out but this is a security release so we will want to update
> 
> https://nvd.nist.gov/vuln/detail/CVE-2023-44487

no doubt right

He-Pin (approved):
lgtm

docs/src/main/paradox/additional/osgi.md
```diff
@@ -16,7 +16,7 @@ To use Apache Pekko in OSGi, you must add the following dependency in your proje
 ## Background
 
 [OSGi](https://www.osgi.org/resources/where-to-start/) is a mature packaging and deployment standard for component-based systems. It
-has similar capabilities as [Project Jigsaw](https://openjdk.java.net/projects/jigsaw/) (originally scheduled for JDK 1.8), but has far stronger facilities to
```
Roiocam, line 19: this link seem likes has been change.

He-Pin: `https://openjdk.org/projects/jigsaw/` I can access it.

Maybe should drop osgi, as I just used it once ten years ago.

PR #704, opened 2023-10-11
chore: Temporarily disable the license report generation to speed up build

Roiocam:
# Motivation

Refs: original from #326 and #332, then #1019 

This PR improves the speed of paradox build  by disabling `sbt-license-report` on specific build.

PTAL, @mdedetrich @He-Pin


.github/workflows/link-validator.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         uses: coursier/setup-action@v1.3.3
 
       - name: Create the Pekko site
-        run: sbt -Dpekko.genjavadoc.enabled=true "Javaunidoc/doc; Compile/unidoc; docs/paradox"
+        run: sbt -Dpekko.genjavadoc.enabled=true -Dpekko.genlicensereport.enabled=false "Javaunidoc/doc; Compile/unidoc; docs/paradox"
```
Roiocam, line 38: skip the `license-report` page generation in link-validator step.

project/ProjectIndexGenerator.scala
```diff
@@ -0,0 +1,75 @@
...
+    )
+    val markdownFilesAfterLicense = Seq(
+      "../additional/faq.md",
+      "../additional/books.md",
+      "examples.md",
+      "links.md"
+    )
+
+     CliOptions.generateLicenseReportEnabled.ifTrue(
+      markdownFilesBeforeLicense ++ "license-report.md"
```
Roiocam, line 55: Use the CLI option to selectively generate the index.

scripts/link-validator.conf
```diff
@@ -77,60 +77,9 @@ site-link-validator {
     "http://www.jaist.ac.jp/~defago/files/pdf/IS_RR_2004_010.pdf"
     # Sigar site has been removed
     "http://www.hyperic.com/products/sigar"
-    "http://www.tom-e-white.com/2007/11/consistent-hashing.html"
+    "http://web.archive.org"
+    "http://www.reactivemanifesto.org/glossary"
     # genereated by @apidoc
     "http://www.scala-lang.org/api/2.13."
-    # following links are generated by sbt-license-report
```
Roiocam, line 83: Does not need those link whitelist cause we can skip the license report now.

Roiocam (commented):
Approach that skip license-report

Roiocam:
Looks like last CI has passed. And since the PR was opened, every `Link Validator` job in each commit has been successfully passed. It seems that this approach can address the issues caused by the license report.

mdedetrich:
I had another way of solving this issue which involved adding an `ignore-files` config to my forked https://github.com/mdedetrich/site-link-validator (the fork is also there so we don't rely on akka). Such a configuration would provide a far simpler solution as you would just do something like

```
ignore-files = "link-validator.html"
```

in `scripts/link-validator.conf`

Which makes this really annoying because a lot of time was spent on this :/

Roiocam:
> I had another way of solving this issue which involved adding an `ignore-files` config to my forked https://github.com/mdedetrich/site-link-validator (the fork is also there so we don't rely on akka). 

I have to say that this approach also crossed my mind initially. However, considering the additional maintenance cost of another repo, I have decided to abandon this solution for this issue and consider a simpler approach.

This PR approach more than ignores license report files and could skip the task of license report generation, it can make PR checks run faster.

In addition, maintaining a new repo could provide ongoing solutions for future issues that may arise. However, the biggest obstacle for me is the sustainability of maintaining an additional repo. Throughout my career, I have encountered situations where repositories were no longer updated due to a lack of maintenance personnel, such as Apache Common DButils.

 Anyway, I am willing to accept whichever solution is ultimately accepted because both of them address the problem.

Best regards.

mdedetrich:
So my biggest apprehension is the complexity of the build which this PR does increase, I have an overarching goal in my head to reduce to try and gradually simplify the sbt build (I haven't really formalized this yet) as I see that sbt's build complexity is creating a barrier of entry to those trying to contribute.

The reduction of the build time I don't see as a critical priority in this specific case, also do note that there will be movement in the future to try and procure our own hardware.

Roiocam:
@mdedetrich Another view is that this undertaking is of modest scale and does not necessitate regular updates. Consequently, the absence of maintenance is not a significant concern to be pondered upon.

Scala is not my primary language, but I still can make the commit like this PR. I acknowledge that the sbt build can be a bit complex, but also offers a lot of possibilities. Considering that sbt builds are not something that needs frequent changes, I believe it won't hinder those who are trying to contribute.

Perhaps we can merge this PR first to avoid those annoying check failure. Once the PR for the other repo is merged and released, we can remove the redundant build task from this PR (This PR also includes a fix part for the broken links). Unless the repo of https://github.com/mdedetrich/site-link-validator was proceeding well.

Roiocam:
@mdedetrich any feedback of this issue will be great.

.github/workflows/link-validator.yml
He-Pin: As this is a ci, can we keep it? or we add a dedicated task for `genlicensereport`

project/ProjectIndexGenerator.scala
```diff
@@ -0,0 +1,75 @@
...
+      "issue-tracking.md",
+      "licenses.md")
+    val markdownFilesAfterLicense = Seq(
+      "../additional/faq.md",
+      "../additional/books.md",
+      "examples.md",
+      "links.md")
+
+    val markdownFiles = if (CliOptions.generateLicenseReportEnabled.get) {
+      markdownFilesBeforeLicense ++ Seq("license-report.md") ++ markdownFilesAfterLicense
```
He-Pin, line 54: I think we can add a `paradoxFast` and `paradoxFastBrowse` task which turn this off?

docs/src/main/paradox/project/project-index.md
```diff
@@ -8,16 +9,16 @@
...
 * [migration-guides](migration-guides.md)
 * [rolling-update](rolling-update.md)
 * [issue-tracking](issue-tracking.md)
 * [licenses](licenses.md)
-* [license-report](license-report.md) 
-* [faq](../additional/faq.md)
-* [books](../additional/books.md)
+* [license-report](license-report.md)
+* [../additional/faq](../additional/faq.md)
+* [../additional/books](../additional/books.md)
```
He-Pin, line 21: seems this can be split to another pr.

mdedetrich:
So I am not sure how this PR relates to https://github.com/apache/incubator-pekko/issues/1019 specifically but as explained [here](https://github.com/apache/incubator-pekko/issues/1019#issuecomment-1903371167) the majority of time is taken up in sbt-license-report's `updateLicenses` task.

I would divert effort into solving that fundamental issue

project/ProjectIndexGenerator.scala
Roiocam: Have tried those approach:

- `(Compile / paradox / javaOptions += "-Dpekko.genlicensereport.enabled=false")`
-  `(javaOptions += "-Dpekko.genlicensereport.enabled=false")`
- `Compile / paradoxMarkdownToHtml / sourceGenerators := {
       (Compile / paradoxMarkdownToHtml / sourceGenerators).value.filterNot(_ == Paradox.licenseReportGeneratorTask.taskValue)
    }`
- `Compile / paradoxMarkdownToHtml / sourceGenerators := {
      Seq.empty
    }`

None of them can be filtered in the license report generated, will keep find a new approach tomorrow.

docs/src/main/paradox/project/project-index.md
Roiocam: I have rolled back these lines, and to avoid any misunderstanding I provide some redundant explanations here:

- The purpose of changing these lines was to allow ProjectIndexGenerator to match the text and skip writing to this file (when license report is enabled). After I complete the rollback commit, this file will always perform write operations.
- ProjectIndexGenerator cannot be completed in a separate PR because when LicenseReport is disabled, the link in the index file will complain lack of a file, resulting in a compilation failure.

pjfanning:
@Roiocam could I ask you to get an iCLA? https://www.apache.org/licenses/contributor-agreements.html

It makes the team's life easier if regular contributors have iCLAs

Roiocam:
> @Roiocam could I ask you to get an iCLA? https://www.apache.org/licenses/contributor-agreements.html
> 
> It makes the team's life easier if regular contributors have iCLAs

I was sent the ICLA submission on 2023/9/23, is any thing I have missed?

mdedetrich:
So I am fine with this being a temporary fix but when https://github.com/sbt/sbt-license-report/issues/87 is resolved I would suggest to revert this commit because there isn't any reason to have this feature disable i.e. its currently a workaround for that bug.

This is because when https://github.com/sbt/sbt-license-report/issues/87 is resolved, the `updateLicenses` task should be near instant because it will actually re-use all of the dependency resolution that was already done when sbt is loaded.

Roiocam:
> So I am fine with this being a temporary fix but when [sbt/sbt-license-report#87](https://github.com/sbt/sbt-license-report/issues/87) is resolved I would suggest to revert this commit because there isn't any reason to have this feature disable i.e. its currently a workaround for that bug.

Yes, that's why we still need another ISSUE #1019  to associate the problem.

> This is because when [sbt/sbt-license-report#87](https://github.com/sbt/sbt-license-report/issues/87) is resolved, the `updateLicenses` task should be near instant because it will actually re-use all of the dependency resolution that was already done when sbt is loaded.

No doubt it, this PR is just sugar for contributors now.

mdedetrich (approved):
Okay I am going to approve it, would be good if @pjfanning and @He-Pin also looked at it. @Roiocam Can you make an standard issue to revert this PR once https://github.com/sbt/sbt-license-report/issues/87 is solved so we don't forget.

Thanks for the work!

Roiocam:
> Okay I am going to approve it, would be good if @pjfanning and @He-Pin also looked at it. @Roiocam Can you make an standard issue to revert this PR once [sbt/sbt-license-report#87](https://github.com/sbt/sbt-license-report/issues/87) is solved so we don't forget.

Yes, I will stay updated on the progress.

He-Pin (approved):
Thanks

He-Pin:
@pjfanning Would you like to take a look at this, I think with this , all our contributors can have a faster local development exprirence.

PR #707, opened 2023-10-13
Improve .scala-steward.conf

mdedetrich:
Let me know if any other additions/removals need to be done

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,7 +1,4 @@
-pullRequests.frequency = "@monthly"
-
 updates.ignore = [
-  { groupId = "com.google.protobuf", artifactId = "protobuf-java" }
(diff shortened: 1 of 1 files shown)


.scala-steward.conf
```diff
@@ -1,7 +1,4 @@
-pullRequests.frequency = "@monthly"
-
 updates.ignore = [
-  { groupId = "com.google.protobuf", artifactId = "protobuf-java" }
```
pjfanning, line 4: I think we should ignore protobuf-java. Newer versions have breaking changes.

mdedetrich: Done, I added it back in but in a better spot

pjfanning (approved):
lgtm

PR #708, opened 2023-10-14
Sidestep reproducibility issue

raboof:
Avoids https://github.com/lampepfl/dotty/issues/18248

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/ExtensionsImpl.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/ExtensionsImpl.scala
@@ -42,7 +42,7 @@ private[pekko] trait ExtensionsImpl extends Extensions { self: ActorSystem[_] wi
     /*
      * @param throwOnLoadFail Throw exception when an extension fails to load (needed for backwards compatibility)
      */
-    def loadExtensions(key: String, throwOnLoadFail: Boolean): Unit = {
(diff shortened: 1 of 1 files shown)


pjfanning:
Tentatively, marked as needing a 1.1.0 release. Renaming public methods is unfortunate but I think this generally ok - the change is useful.

raboof:
> Tentatively, marked as needing a 1.1.0 release. Renaming public methods is unfortunate but I think this generally ok - the change is useful.

I don't think this is a public method though: it's an inner method inside the (confusingly named the same) public method. With the change:

```
$ public static void loadExtensions$(org.apache.pekko.actor.typed.internal.ExtensionsImpl);
  public default void loadExtensions();
  public static java.lang.Object $anonfun$loadExtensions$1(org.apache.pekko.actor.typed.internal.ExtensionsImpl, boolean, java.lang.String);
  private void loadExtensionsFor$1(java.lang.String, boolean);
  public static org.apache.pekko.actor.typed.ExtensionId $anonfun$loadExtensions$3(java.lang.Class);
  public static scala.util.Try $anonfun$loadExtensions$2(java.lang.Class);
```

even from the Java perspective `loadExtensionsFor` is private, only `loadExtensions` (unchanged) is public. Mima seems fine with it as well.

pjfanning (approved):
lgtm - I only looked at the GitHub diff before - now that I see its an inner function, this seems fine

pjfanning:
@raboof would you have time to create a backport PR (branch 1.0.x)?

raboof:
:+1: #739

PR #748, opened 2023-10-18
use slf4j v2

pjfanning:
* more complicated than #733 
* need logback 1.3 because it supports slf4j 2 and jsk 8 (logback 1.4 drops jdk 8 support)


mdedetrich:
Let me think about this, it may be that we can only update slf4j in 2.0.x if pekko-http cannot work with an earlier version pekko (i.e. the 1.0.x version of Pekko working with Pekko Http 1.1.x that has slf4j in 2.0.x)

pjfanning:
> Let me think about this, it may be that we can only update slf4j in 2.0.x if pekko-http cannot work with an earlier version pekko (i.e. the 1.0.x version of Pekko working with Pekko Http 1.1.x that has slf4j in 2.0.x)
> 
> Am I correct in this?

There is no slf4j code in pekko-http. I ran a test with pekko-connectors and it works ok with slf4j v2.

mdedetrich (approved):
lgtm

PR #749, opened 2023-10-18
remove unused log4j dependency

pjfanning:


He-Pin (approved):
lgtm

mdedetrich:
@pjfanning Shall we backport this to `1.0.x`?

He-Pin:
It's never been used, so it's safe.

mdedetrich:
> As @He-Pin says, it isn't used so it doesn't cause any harm to leave it. It is just useful to tidy up the `main` branch so that we don't have it indefinitely.

Oh by isn't used I thought you meant it was part of the dependency tree but not actually needed, but instead its just a standalone variable (was looking at the PR on my phone so didn't really notice).

All good then

PR #751, opened 2023-10-19
[pekko-testkit-typed] Support for JUnit5 in testkit 

thmue:
Due to some issues with the scala setup i created a new PR, which is based on the current main branch. 

This PR replaces https://github.com/apache/incubator-pekko/pull/445

The extension is activated through an annotation:
@ExtendWith(TestKitJUnit5Extension.class)
on the class level.

Additionally, a testKit with a @Junit5TestKit annotation needs to be created.
A builder pattern is used to construct the Testkit instance.

@JUnit5TestKit
public ActorTestKit testKit = new JUnit5TestKitBuilder().build();

The extension itself uses the JUnit5 AfterAllCallback to shutdown the testkit after all tests are executed.


actor-testkit-typed/src/test/java/jdocs/org/apache/pekko/actor/testkit/typed/javadsl/LogCapturingExampleTest.java
```diff
@@ -26,7 +26,7 @@
 
 public class LogCapturingExampleTest {
 
-  @ClassRule public static final TestKitJunitResource testKit = new TestKitJunitResource();
+  @ClassRule public static final TestKitJUnitResource testKit = new TestKitJUnitResource();
```
pjfanning, line 29: Can we leave the legacy Junit 4 classes alone?

I must also admit to preferring that the new classes are `Junit5` to match the pre-existing standard. I am going to -1 the PR due to the renaming of the existing classes.

pjfanning (changes requested):

actor-testkit-typed/src/test/java/jdocs/org/apache/pekko/actor/testkit/typed/javadsl/LogCapturingExampleTest.java
thmue: ok i changed it back :-)

actor-testkit-typed/src/main/java/org/apache/pekko/actor/testkit/typed/annotations/JUnit5TestKit.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package org.apache.pekko.actor.testkit.typed.annotations;
+
+import java.lang.annotation.*;
+
+@Documented
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface JUnit5TestKit {}
```
pjfanning, line 25: * thanks for reverting the changes to the junit4 classes
* with the junit5 classes, can we be consistent - it's either `JUnIt` or it's `Junit` but we in this PR, there is a mix.

thmue: Thanks for spotting this inconsistency, fixed

actor-testkit-typed/src/test/java/jdocs/org/apache/pekko/actor/testkit/typed/javadsl/LogCapturingExampleTest.java
mdedetrich: I disagree, I think the class should be named `JUnit5` and we should have a deprecated alias for `Junit5`.

mdedetrich (changes requested):
While its true that we shouldn't rename the `Junit4` class names, we should actually fix them and have the old class name as an alias that is deprecated (so we don't break binary compatibility).

There is no reason not to improve the current state of things and consistency with current incorrectly named code isn't a compelling argument if you can fix the original issue without breaking users so I would recommend the following

* Rename the classes back to `JUnit4`/`JUnit5` so its actually consistent with how `JUnit` is named.
* Create deprecated `Junit4`/`Junit5` classes that extend `JUnit4`/`JUnit5` so it doesn't break binary compatibility. These can then be dropped in Pekko 2.0.x

actor-testkit-typed/src/main/java/org/apache/pekko/actor/testkit/typed/annotations/JUnit5TestKit.java
mdedetrich: I disagree, I think the class should be named `JUnit5` and we should have a deprecated alias for `Junit5`.

thmue:
ok i changed all Junit5 names/labels to JUnit5. I suggest to rename the other Junit4 labels / class names in a different PR.

mdedetrich:
@thmue There are some compile errors for Scala 3

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJUnit5Extension.scala
```diff
@@ -29,8 +29,8 @@ final class TestKitJUnit5Extension() extends AfterAllCallback with BeforeTestExe
    * Get a reference to the field annotated with @JUnit5Testkit  [[JUnit5TestKit]]
    */
   override def beforeTestExecution(context: ExtensionContext): Unit = {
-
-    context.getTestInstance.ifPresent((instance: AnyRef) => {
+    val testInstance: Option[AnyRef] = Option.when(context.getTestInstance.isPresent)(context.getTestInstance.get())
```
pjfanning, line 32: I think Option.when only appeared in Scala 2.13. We need to support Scala 2.12 still.

actor-typed/src/main/scala-jdk-9/org/apache/pekko/actor/typed/internal/jfr/Events.scala
```diff
@@ -18,7 +18,6 @@ import jdk.jfr.Enabled
 import jdk.jfr.Event
 import jdk.jfr.Label
 import jdk.jfr.StackTrace
-
```
pjfanning, line 21: can you undo this?

actor-testkit-typed/src/test/resources/application.conf
```diff
@@ -2,3 +2,4 @@
 
 # used by ActorTestKitSpec
 test.from-application = yes
+test.value = someValue
```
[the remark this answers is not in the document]
thmue: ok done

actor-typed/src/main/scala-jdk-9/org/apache/pekko/actor/typed/internal/jfr/Events.scala
thmue: should now be the same as in main

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/javadsl/TestKitJUnit5Extension.scala
thmue: i replaced Option.when with an if statement

pjfanning (approved):
lgtm - @mdedetrich wdyt? - it may be worth merging as is and worrying about prospective changes to the junit4 existing code later

pjfanning:
@mdedetrich is it possible to merge this?

mdedetrich:
I would like to add the aliases as I mentioned before, if it's tok much work then I can do it in a follow up PR

pjfanning:
> I would like to add the aliases as I mentioned before, if it's tok much work then I can do it in a follow up PR

I think that would be better in a separate PR

mdedetrich (approved):
lgtm, lets do alias changes in separate PR

aseychell:
I would like to write tests with Junit5 and it seems this PR was merged 8 months ago but this is still not available on any stable release. Are there any plans to make this available please?

raboof:
Good question! I agree we should be working towards an 1.1.0 release. I think the best way to help with that is to look at the issues and PRs linked to the 1.1.0-M2 milestone and see if there's anything in there that you'd feel comfortable picking up.

pjfanning:
> I would like to write tests with Junit5 and it seems this PR was merged 8 months ago but this is still not available on any stable release. Are there any plans to make this available please?

If you don't want to use 1.1.0-M1 (and it is not a full release), you could copy the 4 small classes from this PR into your code base and use them Pekko 1.0.x.

PR #593, opened 2023-08-28
Update javafmt

mdedetrich:
Updates and applies javafmt

PR #756, opened 2023-10-21
turn off static imports format for java formatter.

He-Pin:
refs: https://github.com/apache/incubator-pekko/pull/723 & https://github.com/apache/incubator-pekko/pull/593

Update: When I run the format on Windows, there are `314` files change, so better to leave it in this pr.

### DIFF
--- a/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/AggregatorTest.java
+++ b/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/AggregatorTest.java
@@ -36,10 +36,12 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+// format: off
 import static jdocs.org.apache.pekko.typed.AggregatorTest.IllustrateUsage.HotelCustomer;
 import static jdocs.org.apache.pekko.typed.AggregatorTest.IllustrateUsage.Hotel1;
 import static jdocs.org.apache.pekko.typed.AggregatorTest.IllustrateUsage.Hotel2;
 import static org.junit.Assert.assertEquals;
+// format: on
 
 public class AggregatorTest extends JUnitSuite {
   @ClassRule public static final TestKitJunitResource testKit = new TestKitJunitResource();
--- a/actor-typed-tests/src/test/java/org/apache/pekko/actor/typed/javadsl/ActorContextPipeToSelfTest.java
+++ b/actor-typed-tests/src/test/java/org/apache/pekko/actor/typed/javadsl/ActorContextPipeToSelfTest.java
@@ -30,8 +30,10 @@ import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
+// format: off
 import static org.hamcrest.CoreMatchers.startsWith;
 import static org.junit.Assert.*;
+// format: on
 
 public final class ActorContextPipeToSelfTest extends JUnitSuite {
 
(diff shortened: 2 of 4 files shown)


.sbt-java-formatter.conf
```diff
@@ -8,6 +8,11 @@ ignored-files = [
   "StreamRefMessages.java",
   //in tests
   "ProtobufProtocol.java"
+  //formtat imports cause errors
+  "AggregatorTest.java",
+  "InteractionPatternsTest.java",
+  "ReplicatedShardingTest.java",
+  "ReplicatedAuctionExampleTest.java"
```
He-Pin, line 15: I add thoese files to ignore list.

He-Pin:
Have not tried the https://github.com/google/google-java-format/issues/137#issuecomment-567714532 yet

.sbt-java-formatter.conf
He-Pin: Have not tried the https://github.com/google/google-java-format/issues/137#issuecomment-567714532 yet

Roiocam: Fair enough. 

We correctly using spotless for java formatting, it has configable than raw `google-java-formatter`.
and then we meet the same issue of turn off the snippet formatting. In the end, we using new feature of spotless plugin, it sovled my problem. 

There is some context of this feature and issue: https://github.com/diffplug/spotless/issues/275

sbt has a plugin for spotless, but it has not been updated for a long time.

mdedetrich:
I swear I mentioned this before but I guess I was writing a comment on my mobile and it was lost somewhere but I am not sure what the use of upgrading the sbt-java-formatter is if its changing the meaning of the program by incorrectly formatted imports that is forcing us to add files to the ignore list and/or using `format: off`/`format: on`.

It would be far better to figure out the underlying root cause, i.e. maybe there is an option to keep the import order as it was before?

He-Pin:
@mdedetrich There are no config for this formatter.

I was added these file to `ignored-files` in `.sbt-java-formatter.conf` , it works too.

When I ran the `javafmtAll` there are 314 files affected, but seems less when ran the command from Linux.

mdedetrich:
> @mdedetrich There are no config for this formatter.

Then to be frank I think we should look for another formatter. If a formatter is changing the meaning of a program and at least doesn't allow a configuration to prevent that then this is a terrible design.

He-Pin:
@mdedetrich And seems all those tests are related with `*Behavior` javadsl.

He-Pin:
@mdedetrich but as your comments in anther pr, it's a JDK 8 bug.

mdedetrich:
> @mdedetrich but as your comments in anther pr, it's a JDK 8 bug.

I don't remember saying this, can you provide a link?

He-Pin:
@mdedetrich from https://github.com/apache/incubator-pekko/pull/593
> https://stackoverflow.com/questions/50145575/why-does-rearranging-imports-cause-compilation-to-fail might be related, in which case we should change the formatting style or add these files into ignore

mdedetrich:
Ah right now I remember, still at odds on this because we still support JDK 8 and will likely do so for some time (especially if multi release jar module is made). Lets see what other people say

.sbt-java-formatter.conf
Roiocam: I just verified if this issue can be fixed using `spotless`, the answer is no, there is exception i have meet, i am very sure it will break the `paradox`.

```java
com.google.googlejavaformat.java.FormatterException: error: Imports not contiguous (perhaps a comment separates them?)
```

He-Pin:
@mdedetrich I think this is fine, anther workaround is delete the static imports, by this change we can make some progress anyway

mdedetrich:
> @mdedetrich I think this is fine, anther workaround is delete the static imports, by this change we can make some progress anyway

My current stance is to actually drop it and to add sbt-java-formatter to the ignore list for scala steward until we happen to drop JDK 8.

Its not just about current code with the `format: off`/`format: on` workaround, any new code that is written in Java can be then be autoformatted to reveal this JDK 8 bug and there is no reason for us to upgrade anyways since we are stuck with "JDK 8" java style.

He-Pin:
Okay, let's close it, and pin the sbt-java-formatter at 0.7.0.

PR #765, opened 2023-10-22
Handle mixed akka/pekko protocol names

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko/issues/108

Currently in draft

### DIFF
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -184,6 +184,20 @@ pekko {
     # is 'off'. Set this to 'off' to suppress these.
     warn-unsafe-watch-outside-cluster = on
 
+    # When receiving requests from other remote actors, what are the valid
+    # prefix's to check against. Useful for when dealing with rolling cluster
+    # migrations with compatible systems such as Lightbend's Akka.
+    accept-protocol-names = ["pekko", "akka"]
+
+    # The protocol name to use when sending requests to other remote actors.
+    # Useful when dealing with rolling migration, i.e. temporarily change
+    # the protocol name to match another compatible actor implementation
+    # such as Lightbend's "akka" (whilst making sure accept-protocol-names
+    # contains "akka") so that you can gracefully migrate all nodes to Apache
+    # Pekko and then change the protocol-name back to "pekko" once all
+    # nodes have been are running on Apache Pekko
+    protocol-name = "pekko"
+
     # Settings for the Phi accrual failure detector (http://www.jaist.ac.jp/~defago/files/pdf/IS_RR_2004_010.pdf
     # [Hayashibara et al]) used for remote death watch.
     # The default PhiAccrualFailureDetector will trigger if there are no heartbeats within
--- a/cluster/src/main/scala/org/apache/pekko/cluster/ConfigUtil.scala
+++ b/cluster/src/main/scala/org/apache/pekko/cluster/ConfigUtil.scala
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.cluster
+
+import com.typesafe.config.{ Config, ConfigValue, ConfigValueFactory, ConfigValueType }
+
+import scala.annotation.nowarn
+
+private[cluster] object ConfigUtil {
+
+  @nowarn("msg=deprecated")
+  def addAkkaConfig(cfg: Config, akkaVersion: String): Config = {
+    import scala.collection.JavaConverters._
+    val innerSet = cfg.entrySet().asScala
+      .filter(e => e.getKey.startsWith("pekko.") && e.getValue.valueType() != ConfigValueType.OBJECT)
+      .map { entry =>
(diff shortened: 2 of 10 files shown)


pjfanning:
There is also https://github.com/apache/incubator-pekko/blob/00c6bac1a70d2e0dc0408a4699d136d599855462/remote/src/main/scala/org/apache/pekko/remote/transport/PekkoProtocolTransport.scala#L76

mdedetrich:
> incubator-pekko/remote/src/main/scala/org/apache/pekko/remote/transport/PekkoProtocolTransport.scala

Thanks pushed, lets see if it breaks anything

remote/src/main/resources/reference.conf
```diff
@@ -184,6 +184,20 @@ pekko {
     # is 'off'. Set this to 'off' to suppress these.
     warn-unsafe-watch-outside-cluster = on
 
+    # When receiving requests from other remote actors, what are the valid
+    # prefix's to check against. Useful for when dealing with rolling cluster
+    # migrations with compatible systems such as Lightbend's Akka.
+    accept-protocol-names = ["pekko", "akka"]
```
pjfanning, line 190: could we mention that these 2 properties were only added in v1.0.2?

pjfanning (commented):
is the intention to add some `akka` protocol tests in this PR or maybe in a follow up PR?

remote/src/main/resources/reference.conf
mdedetrich: Do you have an example elsewhere in the docs on the phrasing/formatting of this?

mdedetrich:
> is the intention to add some `akka` protocol tests in this PR or maybe in a follow up PR?

The intention is to add the tests in the same PR, hence why its a draft.

remote/src/main/resources/reference.conf
pjfanning: We have comment text in reference.conf files in this repo for things like deprecations (and the version that the deprecation happened in). I thought it would be uncontroversial to have text that says config was only added in v1.0.2.

pjfanning: I would have thought that something like `# This config was added in v1.0.2.` would be ok.

mdedetrich: It's a matter of style not controversy, all good I'll find an example

mdedetrich: So I just search both `1.` and `2.` in only `*.conf` files and it appears that historically Akka never documents when config values are added, likely because its documented elsewhere (both in actual documentation and migration notes). 

Basically the version that a config is added is implicitly derived from the fact that the documentation will refer to it (and if the config isn't there then its not going to be in the documentation).

pjfanning: I think it is better to doc this in the reference.conf even if we also document it in the paradox docs.

mdedetrich: Ill have a think about it but I am not seeing the value here given that the config literally won't exist in the docs in earlier versions and if was an issue we would have at least a single example of it.

pjfanning:
@mdedetrich what would you think of merging this soon more or less as is but we document it as experimental. Then we can release v1.0.2 with this and see if we can crowdsource the testing. Alternatively, we add it to main only, for now and release v1.1.0-M1 and again see if we can get people to test with it.

mdedetrich:
> @mdedetrich what would you think of merging this soon more or less as is but we document it as experimental. Then we can release v1.0.2 with this and see if we can crowdsource the testing. Alternatively, we add it to main only, for now and release v1.1.0-M1 and again see if we can get people to test with it.

Not really confident merging this as is because @He-Pin said that the issue is more complicated and we also need to do port roll-over to.

pjfanning:
Ports are configurable already. For now, mixed cluster users can use Akka port numbers. They can even use Akka ports after the whole cluster is Pekko. We can add support for mixed port numbers later.

pjfanning:
I think this PR is bettered targeted at main branch. We can backport it when we have something that we are confident in. We can merge experimental or partial impls to main.

mdedetrich:
> I think this PR is bettered targeted at main branch. We can backport it when we have something that we are confident in. We can merge experimental or partial impls to main.

@He-Pin Brought up an interesting point in that this may only be needed in 1.0.x because there is a presumption that once you migrate from an Akka cluster to a Pekko one running 1.0.x, there is no reason that the feature would be needed in 1.1.x (even though we are technically "breaking users").

Given this and the fact that there is a community agreement that all migration (from Akka) related features should land in Pekko 1.0.x I think leaving this to target 1.0.x is the right decision.

pjfanning:
I disagree about having this as a v1.0.x only feature. We've already had users who have tried to switch to Pekko from a version of Akka newer than v2.6. In the real world, we will have Akka users who reach the size where they need to pay license fees. If we add this feature, there is no reason not to keep it for a while. And if we then agree that it is useful to support this for a while (maybe 2 years as an example) then that makes it tidier to develop this on main branch - especially since we can release milestone versions.

mdedetrich:
> I disagree about having this as a v1.0.x only feature. We've already had users who have tried to switch to Pekko from a version of Akka newer than v2.6. In the real world, we will have Akka users who reach the size where they need to pay license fees. If we add this feature, there is no reason not to keep it for a while. And if we then agree that it is useful to support this for a while (maybe 2 years as an example) then that makes it tidier to develop this on main branch - especially since we can release milestone versions.

I'm not disagreeing about it being in both versions, I am just saying that it has to land in 1.0.x as agreed upon and if it's going in 1.1.x as well then it makes sense for the feature to be complete and tested before the 1.1.0 full release otherwise we will have feature gaps

pjfanning:
I want to see this in 1.1.0-M1. It makes it easier for people to try it out. M1 is also a release that we can caveat. A 1.0.3 release would be expected to have well tested, working code. This change still feels experimental but worth experimenting with nonetheless.

pjfanning:
@mdedetrich noone seems to object to a 1.0.3-M1. Can we merge this? If we end up with an urgent 1.0.x bug fix issue, we could revert this again - but the likelihood is that the next 1.0.x release will be 1.0.3-M1. It would be nice to have these changes in a snapshot for some pre 1.0.3-M1 testing.

mdedetrich:
> @mdedetrich noone seems to object to a 1.0.3-M1. Can we merge this? If we end up with an urgent 1.0.x bug fix issue, we could revert this again - but the likelihood is that the next 1.0.x release will be 1.0.3-M1. It would be nice to have these changes in a snapshot for some pre 1.0.3-M1 testing.

Sure, changed it to review status. It just needs a review.

pjfanning (approved):
lgtm

PR #771, opened 2023-11-01
do not render env variables in configs

pjfanning:
see https://github.com/lightbend/config/pull/798

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
@@ -400,13 +400,14 @@ class ActorSystemSpec extends PekkoSpec(ActorSystemSpec.config) with ImplicitSen
           ConfigFactory
             .parseString("""pekko.test.java.property.home = "${user.home}"""")
             .withFallback(PekkoSpec.testConf))
-      val debugText = system.settings.toString
-      val username = System.getProperty("user.name")
-      val userHome = System.getProperty("user.home")
-      (debugText should not).include(username)
-      (debugText should not).include(userHome)
-      debugText should include("<username>")
-      shutdown(system)
+      try {
+        val debugText = system.settings.toString
+        val username = System.getProperty("user.name")
+        val userHome = System.getProperty("user.home")
+        (debugText should not).include(username)
+        (debugText should not).include(userHome)
+        debugText should include("<username>")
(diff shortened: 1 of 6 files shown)


He-Pin (approved):
lgtm, do we need to change the code in spec too?

mdedetrich:
Can we add a simple unit test for this?

mdedetrich:
If it requires too much work we can skip it I guess, its just usually better to have a test for these kinds of things. Isn't `sys.env` mutable which might make things easier?

pjfanning:
> If it requires too much work we can skip it I guess, its just usually better to have a test for these kinds of things. Isn't `sys.env` mutable which might make things easier?

I can play around with `sys.env` and see if that work. I also contacted you in Slack about another related issue that I don't want to make public yet.

pjfanning:
@mdedetrich It looks like sys env is read only - likewise with `System.getenv()`. We might need to rely on the whether config jar is well tested. I have also added my own code to redact username.

mdedetrich:
@pjfanning Okay, just ping me when the PR is ready. If its not possible to test then I would just go with the initial implementation (iirc it was simpler)

pjfanning:
@mdedetrich I've kept the centralised method for rendering with redactions. My code does an additional redaction of the username and may in future do more. The toString of the config is very large and I wouldn't be surprised if we need to do more changes in future.

mdedetrich:
Thanks ill check it out tomorrow

actor-tests/src/test/scala/org/apache/pekko/actor/ActorSystemSpec.scala
```diff
@@ -392,6 +392,22 @@ class ActorSystemSpec extends PekkoSpec(ActorSystemSpec.config) with ImplicitSen
...
+          ConfigFactory
+            .parseString("""pekko.test.java.property.home = "${user.home}"""")
+            .withFallback(PekkoSpec.testConf))
+      val debugText = system.settings.toString
+      val username = System.getProperty("user.name")
+      val userHome = System.getProperty("user.home")
+      (debugText should not).include(username)
+      (debugText should not).include(userHome)
+      debugText should include("<username>")
+      shutdown(system)
```
mdedetrich, line 409: Minor nit, you can put this into a `try`/`finally` block so the actor system is shutdown even if an exception is thrown

mdedetrich (approved):
Minor nit but otherwise lgtm

PR #772, opened 2023-11-02
try to fix link validation

pjfanning:
* dzone article no longer available - at least not for free
* reactivemanifesto link works in browser but fails in CI job due to PKIX cert issue


pjfanning:
The most recent run failed because manning.com has/had an outage but that is probably temporary.

nvollmar (approved):
lgtm

PR #778, opened 2023-11-05
[DRAFT] netty 4 support

pjfanning:
idea is to possibly support netty 4 for pekko 1.0 users

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -28,7 +28,8 @@ object Dependencies {
   // needs to be inline with the aeron version, check
   // https://github.com/real-logic/aeron/blob/1.x.y/build.gradle
   val agronaVersion = "1.15.1"
-  val nettyVersion = "3.10.6.Final"
+  val netty3Version = "3.10.6.Final"
+  val netty4Version = "4.1.100.Final"
   val protobufJavaVersion = "3.16.3"
   val logbackVersion = "1.2.11"
 
@@ -57,7 +58,9 @@ object Dependencies {
     // Compile
 
     val config = "com.typesafe" % "config" % "1.4.3"
-    val netty = "io.netty" % "netty" % nettyVersion
+    val netty3 = "io.netty" % "netty" % netty3Version
+    val `netty4-transport` = "io.netty" % "netty-transport" % netty4Version
+    val `netty4-handler` = "io.netty" % "netty-handler" % netty4Version
 
     val scalaReflect = ScalaVersionDependentModuleID.versioned("org.scala-lang" % "scala-reflect" % _)
 
@@ -272,7 +275,7 @@ object Dependencies {
       Compile.slf4jApi,
       TestDependencies.scalatest.value)
 
-  val remoteDependencies = Seq(netty, aeronDriver, aeronClient)
+  val remoteDependencies = Seq(netty3, `netty4-transport`, `netty4-handler`, aeronDriver, aeronClient)
   val remoteOptionalDependencies = remoteDependencies.map(_ % "optional")
 
   val remote = l ++= Seq(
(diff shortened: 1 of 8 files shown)


mdedetrich:
Am I missing something, isn't the whole ideal that if you want netty4 you should use 1.1.x?

pjfanning:
> Am I missing something, isn't the whole ideal that if you want netty4 you should use 1.1.x?

Not if we intend to delay releasing pekko 1.1 for months and months.

mdedetrich:
> Not if we intend to delay releasing pekko 1.1 for months and months.

Then we should release a M1 of the 1.1.x branch. The whole point of 1.0.x is its behaviour is meant to be as similar as possible to Akka 2.6.20 to help users transition, the change in netty 3 to netty 4 completely changes that proposition.

For these reasons I am against this, unless there is a stronger argument that isn't "users cannot wait for a few more months" (I also haven't heard of anyone complaining about netty4 not being available right now).

pjfanning:
> > Not if we intend to delay releasing pekko 1.1 for months and months.
> 
> Then we should release a M1 of the 1.1.x branch. The whole point of 1.0.x is its behaviour is meant to be as similar as possible to Akka 2.6.20 to help users transition and this is an actual commitment that we collectively made, the change in netty 3 to netty 4 completely changes that proposition.
> 
> For these reasons I am against this, unless there is a stronger argument that isn't "users cannot wait for a few more months" (I also haven't heard of anyone complaining about netty4 not being available right now).
> 
> At the very least this should be discussed on the mailing list.

M1 releases are useful for testing but of no use to production users,

mdedetrich:
> M1 releases are useful for testing but of no use to production users

Thats besides the point, if production users really want netty4 they can test and then use 1.1.0-M1 themselves otherwise they can wait for an actual 1.1.0 release (milestones need to be voted on anyways, just like with any other release).

I really do not see how tainting the entire 1.0.x branch which is explicitly against the entire point of the 1.0.x is worth it for this.

pjfanning:
I would be very disappointed to see users putting a M1 release into production use. Having a 1.1.0 release that is not LTS but where we promise to follow up with an LTS 1.2.0 - I think this would be better. In the absence of any consensus, then this change is low risk - it adds netty 4 support but this is opt-in and the changes do not appear to make it harder for netty 3 users.

mdedetrich:
> I would be very disappointed to see users putting a M1 release into production use.

Why? Its just as tested as any other Pekko release, seeing as it has to go through the entire ASF release process. Furthermore, its actually a good thing to make an 1.1.0-M1 that is tested before an actual release of 1.1.0 because it means we get validated real world.

I mean this is the whole essence behind a milestone.

> In the absence of any consensus, then this change is low risk - it adds netty 4 support but this is opt-in and the changes do not appear to make it harder for netty 3 users.

It doesn't matter if the change is low risk. The entire point of Pekko 1.0.x is that it is essentially Akka 2.6.20 + necessary renames + migration features + any critital fixes/security issues, thats it and nothing more. This was an agreement that was made after many months of discussion by the community and its not fair to upend that, at least without proper consensus (and that does not mean lack of one)

He-Pin:
I think we should only ship it in 1.1.x and we can release M1 in this month? I expect a release of 1.1.0 in before the end of 2023.

mdedetrich:
> I think we should only ship it in 1.1.x and we can release M1 in this month? I expect a release of 1.1.0 in before the end of 2023.

I am expecting full release of 1.1.0 early next year but yes lets see, the inliner changes still haven't been reviewed yet and there may need to be more work there and remember that December is christmas period for us.

M1 should be released after the slf4j 2 changes land in https://github.com/apache/incubator-pekko/pull/784

He-Pin:
@pjfanning as you plan to add a new milestone release for 1.1.x does this still be valid?

pjfanning:
I'll close this.

PR #797, opened 2023-11-17
Update contributing docs to remove some labels

pjfanning:


pjfanning:
@samueleresca what do you think of these changes? I don't want to remove too much in one PR but these bits are not being used by us.

samueleresca:
> @samueleresca what do you think of these changes? I don't want to remove too much in one PR but these bits are not being used by us.

Looks good to me

samueleresca (approved):

PR #804, opened 2023-11-20
Update scala3.md

pjfanning:
This page seems badly out of date.

### DIFF
--- a/docs/src/main/paradox/project/scala3.md
+++ b/docs/src/main/paradox/project/scala3.md
@@ -1,15 +1,9 @@
 # Scala 3 support
 
-Pekko has experimental support for Scala 3.
+Pekko has support for Scala 3. Now that the Scala 3.3 LTS release is out, we can be pretty confident that it is unlikely that we will need to make binary incompatible changes due to Scala 3 changes. See the @ref:[binary compatibility documentation](../common/binary-compatibility-rules.md).
 
 ## Using 2.13 artifacts in Scala 3
 
-You can use [CrossVersion.for3Use2_13](https://scala-lang.org/blog/2021/04/08/scala-3-in-sbt.html#using-scala-213-libraries-in-scala-3)
+This should not be necessary but you can use [CrossVersion.for3Use2_13](https://scala-lang.org/blog/2021/04/08/scala-3-in-sbt.html#using-scala-213-libraries-in-scala-3)
 to use the regular 2.13 Pekko artifacts in a Scala 3 project. This has been
 shown to be successful for Streams, HTTP and gRPC-heavy applications.
-
-## Scala 3 artifacts
-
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/project/scala3.md
```diff
@@ -1,15 +1,9 @@
 # Scala 3 support
 
-Pekko has experimental support for Scala 3.
+Pekko has support for Scala 3. Now that the Scala 3.3 LTS release is out, we can be pretty confident that it is unlikely that we will need to make binary incompatible changes due to Scala 3 changes. See the @ref:[binary compatibility documentation](../common/binary-compatibility-rules.md).
```
He-Pin, line 3: `that it is unlikely that` can be optimized for reading

nvollmar: `Now with Scala 3.3 LTS we can be confident no binary incompatible changes will be necessary due to Scala 3 changes`

nvollmar (approved):
lgtm

docs/src/main/paradox/project/scala3.md
```diff
@@ -1,15 +1,9 @@
...
 
 ## Using 2.13 artifacts in Scala 3
 
-You can use [CrossVersion.for3Use2_13](https://scala-lang.org/blog/2021/04/08/scala-3-in-sbt.html#using-scala-213-libraries-in-scala-3)
+This should not be necessary but you can use [CrossVersion.for3Use2_13](https://scala-lang.org/blog/2021/04/08/scala-3-in-sbt.html#using-scala-213-libraries-in-scala-3)
```
mdedetrich, line 7: We should remove this paragraph entirely, you should only be using the Scala 3 artifacts and considering we are using LTS there shouldn't be an issue here. Also its ideal to not have multiple ways of supporting Scala 3, just use the Scala 3 artifacts.

Also `CrossVersion.for3Use2_13` cannot be used if you are a library that depends on Pekko, it only works with applications.

docs/src/main/paradox/project/scala3.md
mdedetrich, line 3: I think we can also remove the "Now" phrasing in "now that the Scala 3.3 LTS release". Just mention that Pekko is built against Scala 3.3 LTS and because of that we don't forsee any breaking binary compatibility issues in the foreseeable future.

mdedetrich (changes requested):
The bit about `CrossVersion.for3Use2_13` should be entirely removed as we shouldn't be encouraging it, otherwise lgtm.

docs/src/main/paradox/project/scala3.md
pjfanning: reworded the text

samueleresca (approved):
Reviewed after the suggested changes, looks good to me.

mdedetrich (approved):
lgtm

PR #805, opened 2023-11-20
exclude bcpkix-jdk15on

pjfanning:
We import bcpkix-jdk18on explicitly instead - bcpkix-jdk15on is outdated.

It also has CVEs and is causing security alerts on our project. https://github.com/apache/incubator-pekko/security/dependabot/53 and https://github.com/apache/incubator-pekko/security/dependabot/92


nvollmar (approved):
lgtm

mdedetrich (approved):
lgtm, but ideally we should replace this docker client entirely because its no longer maintained https://github.com/spotify/docker-client

PR #796, opened 2023-11-17
increase timeout on flaky test

pjfanning:
see #795

Issue #795, opened 2023-11-17
investigate ClusterActorLoggingSpec


pjfanning:
possibly related to #771

* passes on laptop both for 1.0.x and main branches
* failed in nightly build but passed when rerun
* seems to pass when nightly build is run using the 1.0.x branch - https://github.com/apache/incubator-pekko/actions/runs/6902135347/job/18778244515

He-Pin:
link : https://github.com/apache/incubator-pekko/issues/929

Roiocam:
still flaky in nightly builds: https://github.com/apache/pekko/actions/runs/9993956065/job/27622606159

Roiocam:
Because sometimes the actor don't put value into MDC:

```diff
LoggingEvent(
		INFO,
		org.apache.pekko.cluster.typed.ClusterActorLoggingSpec,
		ClusterActorLoggingSpec-pekko.actor.default-dispatcher-11,
		Starting,
		1721630292397,
		None,	
		None,
+		Map(pekkoAddress -> pekko.tcp://ClusterActorLoggingSpec@10.1.0.31:42905, pekkoSource -> pekko://ClusterActorLoggingSpec/user/$a, sourceActorSystem -> ClusterActorLoggingSpec)
)
LoggingEvent(
		INFO,
		org.apache.pekko.cluster.typed.ClusterActorLoggingSpec,
		ClusterActorLoggingSpec-pekko.actor.default-dispatcher-3,
		Starting,
		1721631006935,
		None,
		None,
-		Map()
)
```

raboof:
> Because sometimes the actor don't put value into MDC

Doesn't that mean the test caught a bug?

Roiocam:
> Doesn't that mean the test caught a bug?

It is flaky depending on how fast SLF4J initialization.

If it loads slower, it will use SubstituteLogger(which uses BasicMDCAdapter) as a fallback.

![截屏2024-07-23 09 32 21](https://github.com/user-attachments/assets/bd646c88-1f44-4752-8f4c-4578fbad1f78)

if we use MDC earlier than SLF4J initialization, the MDCAdapter may no chance to update it.

![image](https://github.com/user-attachments/assets/0bdabb73-c0f3-4dba-bb42-11b3b96d180f)

![截屏2024-07-23 09 24 52](https://github.com/user-attachments/assets/1be1b38d-cf39-4142-ad35-73ae9e034e13)

PR #808, opened 2023-11-21
try to fix ClusterActorLoggingSpec

pjfanning:
relates to #795 - #796 did not help

* it seems like this filter-leeway is the right timeout
* this test never fails for me - so this is a bit painful to get the CI issue fixed


mdedetrich (approved):
lgtm

PR #815, opened 2023-11-25
adjust log-on-startup docs

pjfanning:
we don't want to encourage the use of this in production

### DIFF
--- a/docs/src/main/paradox/logging.md
+++ b/docs/src/main/paradox/logging.md
@@ -119,11 +119,12 @@ pekko {
 }
 ```
 
-This config option is very good if you want to know what config settings are loaded by Pekko:
+This config option is useful if you want to know what config settings are loaded by Pekko:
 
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/logging.md
```diff
@@ -119,11 +119,12 @@ pekko {
 }
 ```
 
-This config option is very good if you want to know what config settings are loaded by Pekko:
+This config option is useful if you want to know what config settings are loaded by Pekko:
 
 ```ruby
 pekko {
   # Log the complete configuration at INFO level when the actor system is started.
+  # This logging can include sensitive values.
```
He-Pin, line 127: Maybe add words like `consider turn this off in production`?

He-Pin (approved):
lgtm

He-Pin (approved):
lgtm

PR #819, opened 2023-11-27
add Akka path parsing tests

pjfanning:
relates to https://github.com/apache/incubator-pekko/pull/765

These tests prove that Pekko doesn't actively check the URI scheme - at least at the initial URI parsing point.


nvollmar (approved):
lgtm

PR #822, opened 2023-12-02
=str Fix maybe throw for `MinimalStage`.

He-Pin:
The `MinimalStage` is also a `CompletableFuture`, but calling `isDone` on it will throw `UnsupportedOperationException`.

This will cause problem when user call `CompletableFuture.completedStage` on Java 9+.

### DIFF
--- a/stream-tests/src/test/java-jdk9-only/org/apache/pekko/stream/javadsl/FlowUnfoldAsyncTest.java
+++ b/stream-tests/src/test/java-jdk9-only/org/apache/pekko/stream/javadsl/FlowUnfoldAsyncTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


stream-tests/src/test/java-jdk9-only/org/apache/pekko/stream/javadsl/FlowUnfoldAsyncTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    public void testFoldAsync() throws Exception {
+        final Integer result = Source.unfoldAsync(
+                0,
+                idx -> {
+                    if (idx >= 10) {
+                        return CompletableFuture.completedStage(Optional.empty());
+                    } else {
+                        return CompletableFuture.completedStage(Optional.of(Pair.create(idx + 1, idx)));
```
He-Pin, line 49: This method returns a `MinimalStage` , still not know why douglee make the MinimalStage extends CompletableFuture.

He-Pin:
One this got merged, I think this should be back port to 1.0.x too.

mdedetrich:
> One this got merged, I think this should be back port to 1.0.x too.

I disagree on this part as this is a behavioural change which we want to keep to a minimum in 1.0.x branch (and I wouldn't classify this as a critical bug).

For 1.1.x its fine

pjfanning (approved):
lgtm

For now, let's target for 1.1.0.

If users hit this issue in 1.0.x - we can tell them how to workaround it. We can think about a backport if it starts affecting lots of users.

Issue #834, opened 2023-12-07
Potentially unnecessary `DnsClient` warning logs


SakulK:
First of all, thank you for all you work with this fork :heart: 
`org.apache.pekko.io.dns.internal.DnsClient` is logging warnings with messages like this:
```
DNS response id -32068 question [Question(<redacted>,RecordType(28,AAAA),RecordClass(1,IN))] question asked [Question(<redacted>,RecordType(28,AAAA),RecordClass(1,IN))]
```
Both questions are exactly the same, which is expected I assume. It seems to be coming from here:
https://github.com/apache/incubator-pekko/blob/main/actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala#L172-L175
Maybe this log should be at the debug level instead?

PR #836, opened 2023-12-08
rework noisy dns log (#835)

pjfanning:
cherry pick f6af100531e037b4f245f564d7be8c70f0093a4f #835 #834

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/internal/DnsClient.scala
@@ -168,11 +168,13 @@ import pekko.pattern.{ BackoffOpts, BackoffSupervisor }
               msg.id,
               msg.questions.mkString(","),
               orig.questions.mkString(","))
-          case Some((_, orig)) =>
-            log.warning("DNS response id {} question [{}] question asked [{}]",
-              msg.id,
-              msg.questions.mkString(","),
-              orig.questions.mkString(","))
+          case Some((_, _)) =>
+            if (log.isDebugEnabled) {
+              log.debug("DNS response id {} has response code {}: question [{}]",
+                msg.id,
+                msg.flags.responseCode,
+                msg.questions.mkString(","))
+            }
             val (recs, additionalRecs) =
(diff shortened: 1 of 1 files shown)


mdedetrich:
Hmm, is this needed for Pekko 1.0.x? Technically speaking it is a behavioural change and its not actually fixing any critical bugs/security fixes.

I would actually argue that such a change is **not** a good thing because if people migrate from Akka to Pekko 1.0.x and they suddenly see a difference in logs their initial assumption (even if incorrect) may be that Pekko might have "broken" something or their migration "didn't work".

pjfanning:
I'll go with the concensus but this log is not an Akka log - it was added by https://github.com/apache/incubator-pekko/commit/c56edca78f775934ca081f7967175960a18cb99c

My opinion is that this log message is incorrect in what it says, is unnecessary and logs excessively.

mdedetrich:
So just to clarify, the change in https://github.com/apache/incubator-pekko/commit/c56edca78f775934ca081f7967175960a18cb99c modified the logging (so its now spitting more logs then it used to) and this cherry pick commit fixes that?

pjfanning:
> So just to clarify, the change in [c56edca](https://github.com/apache/incubator-pekko/commit/c56edca78f775934ca081f7967175960a18cb99c) modified the logging (so its now spitting more logs then it used to) and this cherry pick commit fixes that?

Yes. The change in Pekko may the logs noisier. The log statement that I am modifying seems to be a copy paste with small mods from the previous case - but this case is basically the success path and it issues a warning log when nothing is wrong. 

@SakulK you reported the original issue (#834) - would you be able to comment on this?

mdedetrich (approved):
Thanks for clearing that up, I will tentatively approve this (by tentatively I mean it would be good to get some feedback from others, i.e. @SakulK)

SakulK:
This log wasn't happening with Akka, so during migration to Pekko it got me worried that something is broken (especially since the log level is `WARN`). So I guess this PR would revert an observable difference between Akka and Pekko 1.0.x

mdedetrich:
> This log wasn't happening with Akka, so during migration to Pekko it got me worried that something is broken (especially since the log level is `WARN`). So I guess this PR would revert an observable difference between Akka and Pekko 1.0.x

Thanks for confirming, in this case it's good to merge.

PR #841, opened 2023-12-11
Use `recoverWith` instead of `recover` in `SnapshotSerializer` akka backwards compatibility

SakulK:
@pjfanning after taking another look I think this should make more sense, otherwise `snapshotFromBinary` can either return `T` or `Try[T]` (got bitten by `AnyRef` here..). I'm not sure why the previous version didn't blow up when testing this with real mongo and existing akka persistence snapshots.


pjfanning:
@SakulK have you tried out this change in an end to end use case? I would like to avoid making a large number of PRs here. I'm happy to change this but I would prefer to get it right this time around.

SakulK:
@pjfanning yes, this works for me end-to-end

pjfanning (approved):
lgtm

PR #859, opened 2023-12-23
=str code clean up in Sliding.

He-Pin:
remove unnecessary condition because the opposite check has already been checked - so `else` is enough

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -903,7 +903,7 @@ private[stream] object Collect {
           if (buf.size == n) {
             push(out, buf)
           } else pull(in)
-        } else if (step > n) {
+        } else {
           if (buf.size == step) {
             buf = buf.drop(step)
           }


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -903,7 +903,7 @@ private[stream] object Collect {
           if (buf.size == n) {
             push(out, buf)
           } else pull(in)
-        } else if (step > n) {
```
mdedetrich, line 906: So this condition is pointless?

jxnu-liguobin, line 906: Useless?

He-Pin: it's always true.

pjfanning:
Can we get a description on the PR? What is intends to do, etc.?

Sometimes, PRs are self-evident but with edits to existing runtime code, I think they need to be justified.

pjfanning (changes requested):
needs justification

He-Pin:
@mdedetrich @pjfanning see line at 901,it checks step ＜= n

pjfanning (approved):
lgtm - I updated the PR description

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: `else if (step <= n)` at line 901

mdedetrich (approved):
lgtm

PR #860, opened 2023-12-23
=str Avoid subMaterialization when the provided recover source is empty.

He-Pin:
perf: avoid subMaterialization  when source is empty.


jxnu-liguobin (approved):

mdedetrich (approved):
lgtm

PR #862, opened 2023-12-23
=str Optimize concat for javadsl.Source.empty.

He-Pin:
It was only optimized for scaladsl, now works for javadsl too.


mdedetrich (approved):
lgtm

PR #864, opened 2023-12-23
perf: Drop the usages of SupervisedGraphStageLogic to reduce allocation.

He-Pin:
Motivation:
There is a allocation of `Some` in `SupervisedGraphStageLogic` and for performance reason, I want to remove that.
I can choose the `Option[T]` to `OptionVal[T]`,  but that will cause the problem if the user specified function return `null`.
So I choose to drop the usage of `SupervisedGraphStageLogic`.

Result:
Reduce allocation, but repeated code.

Status: Mima file is not added yet, open for discussion, and I think should remove the `SupervisedGraphStageLogic ` too.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -178,25 +178,26 @@ import pekko.util.ccompat._
   override def initialAttributes: Attributes = DefaultAttributes.dropWhile and SourceLocation.forLambda(p)
 
   def createLogic(inheritedAttributes: Attributes) =
-    new SupervisedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
-      override def onPush(): Unit = {
-        val elem = grab(in)
-        withSupervision(() => p(elem)) match {
-          case Some(flag) =>
-            if (flag) pull(in)
-            else {
-              push(out, elem)
-              setHandler(in, rest)
+    new GraphStageLogic(shape) with InHandler with OutHandler {
+      private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+
+      override def onPush(): Unit =
+        try {
+          val elem = grab(in)
+          if (p(elem)) {
+            pull(in)
+          } else {
+            push(out, elem)
+            setHandler(in, () => push(out, grab(in)))
+          }
+        } catch {
+          case NonFatal(ex) =>
+            decider(ex) match {
+              case Supervision.Stop    => failStage(ex)
+              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
+              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
             }
-          case None => // do nothing
         }
-      }
-
-      def rest = new InHandler {
-        def onPush() = push(out, grab(in))
-      }
-
-      override def onResume(t: Throwable): Unit = if (!hasBeenPulled(in)) pull(in)
 
       override def onPull(): Unit = pull(in)
 
@@ -252,23 +253,25 @@ private[stream] object Collect {
   override def initialAttributes: Attributes = DefaultAttributes.collect and SourceLocation.forLambda(pf)
 
   def createLogic(inheritedAttributes: Attributes) =
-    new SupervisedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
-
+    new GraphStageLogic(shape) with InHandler with OutHandler {
+      private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
       import Collect.NotApplied
 
-      val wrappedPf = () => pf.applyOrElse(grab(in), NotApplied)
-
-      override def onPush(): Unit = withSupervision(wrappedPf) match {
-        case Some(result) =>
-          result match {
+      override def onPush(): Unit =
+        try {
+          pf.applyOrElse(grab(in), NotApplied) match {
             case NotApplied             => pull(in)
             case result: Out @unchecked => push(out, result)
             case _                      => throw new RuntimeException() // won't happen, compiler exhaustiveness check pleaser
           }
-        case None => // do nothing
-      }
-
-      override def onResume(t: Throwable): Unit = if (!hasBeenPulled(in)) pull(in)
+        } catch {
+          case NonFatal(ex) =>
+            decider(ex) match {
+              case Supervision.Stop    => failStage(ex)
+              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
+              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
+            }
+        }
 
       override def onPull(): Unit = pull(in)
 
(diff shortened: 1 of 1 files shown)


laglangyue:
I have reviewed some of your PRs, but due to my lack of familiarity with the project, I did not comment. But I noticed that there is a '=str' in the title all of them, and I don't quite understand what it means.

He-Pin:
> I have reviewed some of your PRs, but due to my lack of familiarity with the project, I did not comment. But I noticed that there is a '=str' in the title all of them, and I don't quite understand what it means.

I was using the old style, maybe chore nowadays ?

He-Pin:
> > I have reviewed some of your PRs, but due to my lack of familiarity with the project, I did not comment. But I noticed that there is a '=str' in the title all of them, and I don't quite understand what it means.
> 
> I was using the old style, maybe chore nowadays ?

=means I added no api change, and str is abbr for stream module.

mdedetrich:
Im with @pjfanning here, I would rather drop this convention (i.e. `=str`) because no one else is doing it so it kind of defeats the point.

For me, simple git commit messages with active tense as well as making sure commits are atomic as possible is whats most important, i.e. general guidelines as from https://reflectoring.io/meaningful-commit-messages/

mdedetrich:
> but that will cause the problem if the user specified function return `null`.

Is this a legitimate issue, I mean we can do behavioural changes if its well justified?

He-Pin:
> > but that will cause the problem if the user specified function return `null`.
> 
> Is this a legitimate issue, I mean we can do behavioural changes if its well justified?

I think we can't,  if I change to make use of  OptionVal, then it will just ignore the value and do nothing, which was a Some(null) and throw an NPE.

and I will update the commit message then.

laglangyue:
> I was using the old style, maybe chore nowadays ?
> =means I added no api change, and str is abbr for stream module. BTW, I see you from China, are you in the wechat group ？

yeah, I am in Hangzhou,Zhejiang, and I am newcommner for pekko/akka，I plan to introduce Pekko into my project in future.
it's my wechatId `laglangyue`, could you invite me to join the wechat group.
I am not picky about style, and just a simple question.

He-Pin:
@laglangyue Are you using Pekko at work too, welcome and you are welcome to review this PR too.

laglangyue:
I don't have a deep understanding of akka/pekko, so I can't comment on this PR.

> @laglangyue Are you using Pekko at work too, welcome and you are welcome to review this PR too.

He-Pin:
@laglangyue it's simple Java/Kotlin like code, try catch thing:)

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -178,26 +178,31 @@ import pekko.util.ccompat._
...
+          } else {
+            push(out, elem)
+            setHandler(in, rest)
+          }
+        } catch {
+          case NonFatal(ex) =>
+            decider(ex) match {
+              case Supervision.Stop    => failStage(ex)
+              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
+              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
```
laglangyue, line 198: case _,or is it has a default case option

Roiocam (approved):
My first thought is that this is a trade-off between abstraction and performance.

Too much duplicate code is not good to maintenance and code readability.

Considering that other Operators (such as Fold) in Ops also use inline instead of `SupervisedGraphStageLogic`, I will not be too picky about the duplicate code brought by this PR.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -178,26 +178,31 @@ import pekko.util.ccompat._
...
             }
-          case None => // do nothing
         }
-      }
 
-      def rest = new InHandler {
-        def onPush() = push(out, grab(in))
+      private def rest: InHandler = new InHandler {
```
Roiocam, line 202: Maybe we should keep the code simple and let the compiler do more work, isn't it?

He-Pin:
Thank you both from China Scala community to review this, I was plan to add onErrorReturn and OnErrorResume and OnErrorComplete operators, which need very handy control around the supervision.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -178,26 +178,31 @@ import pekko.util.ccompat._
   override def initialAttributes: Attributes = DefaultAttributes.dropWhile and SourceLocation.forLambda(p)
 
   def createLogic(inheritedAttributes: Attributes) =
-    new SupervisedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
-      override def onPush(): Unit = {
-        val elem = grab(in)
-        withSupervision(() => p(elem)) match {
```
laglangyue, line 184: remove the method `withSupervision`, and add a `try catch`
 can we concise the implement

He-Pin: I was plan to remove this `SupervisedGraphStageLogic`, this call class will add at least 3 stack frame too.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: image

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: Good, I was want to avoid scala 2.12 bug.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -178,25 +178,26 @@ import pekko.util.ccompat._
...
+      private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
+
+      override def onPush(): Unit =
+        try {
+          val elem = grab(in)
+          if (p(elem)) {
+            pull(in)
+          } else {
+            push(out, elem)
+            setHandler(in, () => push(out, grab(in)))
```
Roiocam, line 191: One last complain, i think this line lambda will generate anonymous class, it is hard to identify on profiler tools.

But i will still approve this change, because i think this method won't be top of hotspot tree.

Roiocam (approved):

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: That's true, was a anonymous class and now a lambda$, I think this should be fine as there are other cases where a lambda is been used,eg:`emit`.

He-Pin:
@mdedetrich @pjfanning ping~

mdedetrich:
> Too much duplicate code is not good to maintenance and code readability.

I wouldn't be surprised if the inliner (i.e. https://github.com/apache/incubator-pekko/pull/857) will automatically do some of these performance improvements, unfortunately it would only work for Scala 2

mdedetrich (approved):
lgtm

PR #869, opened 2023-12-24
chore: Add special `null` value handling doc for mapAsync and mapAsyncUnordered

He-Pin:
Motivation:
User was only knowning this by reading code, including me.

Result:
More clear.


jxnu-liguobin (approved):
LGTM

pjfanning:
this is something that could be backported to 1.0.x branch

PR #870, opened 2023-12-24
chore: Update scalafmt to 3.7.17

He-Pin:


pjfanning:
I'm not sure. I won't block this but we use 3.7.11 everywhere else. `scalafmt-native` is tied to the scalafmt version. For me, it best to use the same scalafmt version everywhere. The upgrade is not even useful because the rule changes don't affect us - no code needs reformatting.

He-Pin:
I saw there are some bug fixed.

pjfanning:
the 'bugs' don't seem to affect us

He-Pin:
Let me close this and update when we needed to.

PR #873, opened 2023-12-24
perf: Fold In&OutHanlder in FlatMapPrefix.

He-Pin:
Motivation:
The in and out handler in flatmapPrefix can be fold into one.

Result:
Less allocation for per materlization.


pjfanning (approved):
lgtm

PR #878, opened 2023-12-25
[DO-NOT-MERGE] chore: bump scala version to 3.4.0-RC1

He-Pin:
Just test how it works with new Scala version

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -38,7 +38,7 @@ object Dependencies {
 
   val scala212Version = "2.12.18"
   val scala213Version = "2.13.12"
-  val scala3Version = "3.3.1"
+  val scala3Version = "3.4.0-RC1"
   val allScalaVersions = Seq(scala213Version, scala212Version, scala3Version)
 
   val reactiveStreamsVersion = "1.0.4"


He-Pin:
Ha, does catch some errors, Do we need to drop 2.12.x if we want the code compiles with 3.4.x too? @SethTisue ping

SethTisue:
It's normal that 3.4 may throw some warnings that aren't fixable while still cross-compiling to 2.12 (and, in a smaller number of cases, even to 2.13). I would suggest simply silencing the warnings in question.

However, it may be that you don't want to upgrade to 3.4 at all, as that would require users to move to 3.4, but 3.4 is not an LTS version.

Note that Akka itself dropped 2.12 support in akka/akka#32125.

pjfanning:
Thanks @SethTisue for the insights. I just checked our Maven Central stats and our Scala 2.12 jars seem to be at least as popular as Scala 2.13, with Scala 3 jars trailing (quite a bit). Scala 2.12 jar downloads dominated last month but the Unique IPs stats show higher numbers for Scala 2.13 than for 2.12. Pretty erratic stats overall.

mdedetrich:
For the reasons that @pjfanning stated, like with JDK 8 we are not going to dropping Scala 2.12 support anytime soon. Aside from the "significant amount of people are still using 2.12" argument, sbt is indirectly making Scala 2.12 the pseudo minimum Scala 2 version.

Its fine if other projects want to drop it, but in our case Scala 2.12 support isn't really causing any significant burdern. Most of it has been abstracted away with our various `org.apache.pekko.util.*` functions and its one of those things that once there you don't worry about it.

Regarding the PR on hand, while I have no problem in people trying Scala 3.4 with Pekko, its kind of ultimately a waste of time because no Pekko module will use Scala 3.4 since due to how Scala 3 works with LTS, picking a Scala version effectively creates a floor that transitively forces everyone to use at least that version.

So the only minor Scala version that we would ever move to would be the next LTS which I assume would be a few years out. Even then its likely that we would still pick to Scala 3.3 because it would take quite a while for the Scala ecosystem to move to the next LTS.

PR #887, opened 2023-12-28
fix: reject zero and negative periodic tasks schedule

Roiocam:
create a zero periodic tasks on scheduler will let scheduler busy and can not scheduler other task.

on jdk `ScheduledThreadPoolExecutor`, the zero period task currently is forrbidden.

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
@@ -770,12 +770,14 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
     @volatile var time: Long = start
     val sched = new LightArrayRevolverScheduler(config.withFallback(system.settings.config), log, tf) {
       override protected def clock(): Long = {
+        // println(s"clock=$time")
         time
       }
 
       override protected def getShutdownTimeout: FiniteDuration = (10 seconds).dilated
 
       override protected def waitNanos(ns: Long): Unit = {
+        // println(s"waiting $ns")
         prb.ref ! ns
         try time += (lbq.get match {
             case q: LinkedBlockingQueue[Long] => q.take()
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
@@ -662,7 +662,7 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
     }
 
     "using scheduleAtFixedRate" must {
-      "reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
+      "fixed rate reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
         val zeroInterval = 0.seconds
         import system.dispatcher
         intercept[IllegalArgumentException] {
@@ -671,7 +671,7 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
         expectNoMessage(1.second)
       }
 
-      "reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
+      "fixed rate reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
         val negativeDelay = -1.seconds
         import system.dispatcher
         intercept[IllegalArgumentException] {
@@ -682,7 +682,7 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
     }
 
     "using scheduleWithFixedDelay" must {
-      "reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
+      "fixed delay reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
         val zeroInterval = 0.seconds
         import system.dispatcher
         intercept[IllegalArgumentException] {
@@ -691,7 +691,7 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
         expectNoMessage(1.second)
       }
 
-      "reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
+      "fixed delay reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
         val negativeDelay = -1.seconds
         import system.dispatcher
         intercept[IllegalArgumentException] {
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
@@ -661,6 +661,45 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
       }
     }
 
+    "using scheduleAtFixedRate" must {
+      "reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
+        val zeroInterval = 0.seconds
+        import system.dispatcher
+        intercept[IllegalArgumentException] {
+          system.scheduler.scheduleAtFixedRate(100.millis, zeroInterval, testActor, "Not Allowed")
+        }
+        expectNoMessage(1.second)
+      }
+
+      "reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
+        val negativeDelay = -1.seconds
+        import system.dispatcher
+        intercept[IllegalArgumentException] {
+          system.scheduler.scheduleAtFixedRate(100.millis, negativeDelay, testActor, "Not Allowed")
+        }
+        expectNoMessage(1.second)
+      }
+    }
+
+    "using scheduleWithFixedDelay" must {
+      "reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
+        val zeroInterval = 0.seconds
+        import system.dispatcher
+        intercept[IllegalArgumentException] {
+          system.scheduler.scheduleWithFixedDelay(100.millis, zeroInterval, testActor, "Not Allowed")
+        }
+        expectNoMessage(1.second)
+      }
+
+      "reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
+        val negativeDelay = -1.seconds
+        import system.dispatcher
+        intercept[IllegalArgumentException] {
+          system.scheduler.scheduleWithFixedDelay(100.millis, negativeDelay, testActor, "Not Allowed")
+        }
+        expectNoMessage(1.second)
+      }
+    }
     // same tests for fixedDelay and fixedRate
     List(new ScheduleWithFixedDelayAdapter, new ScheduleAtFixedRateAdapter).foreach { scheduleAdapter =>
       s"using $scheduleAdapter" must {
@@ -681,6 +720,24 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
           expectMsg("OK")
           intercept[IllegalArgumentException] {
             scheduleAdapter.schedule(100.millis, maxDelay + tickDuration, testActor, "Too long")
+          }
+          expectNoMessage(1.second)
+        }
+
+        "reject periodic tasks scheduled with zero interval" taggedAs TimingTest in {
+          val zeroInterval = 0.seconds
+          import system.dispatcher
+          intercept[IllegalArgumentException] {
+            scheduleAdapter.schedule(100.millis, zeroInterval, testActor, "Not Allowed")
+          }
+          expectNoMessage(1.second)
+        }
+
+        "reject periodic tasks scheduled with negative interval" taggedAs TimingTest in {
+          val negativeInterval = -1.seconds
+          import system.dispatcher
+          intercept[IllegalArgumentException] {
+            scheduleAdapter.schedule(100.millis, negativeInterval, testActor, "Not Allowed")
           }
           expectNoMessage(1.second)
         }
@@ -713,14 +770,12 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
     @volatile var time: Long = start
     val sched = new LightArrayRevolverScheduler(config.withFallback(system.settings.config), log, tf) {
       override protected def clock(): Long = {
-        // println(s"clock=$time")
         time
       }
 
       override protected def getShutdownTimeout: FiniteDuration = (10 seconds).dilated
 
       override protected def waitNanos(ns: Long): Unit = {
-        // println(s"waiting $ns")
         prb.ref ! ns
         try time += (lbq.get match {
             case q: LinkedBlockingQueue[Long] => q.take()
--- a/actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
@@ -196,7 +196,7 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
     }
 
   private def checkPeriod(delay: FiniteDuration): Unit =
-    if (delay.toNanos <= 0)
+    if (delay.length <= 0)
       throw new IllegalArgumentException(
         s"Task scheduled with [${delay.toSeconds}] seconds delay, which means creating an infinite loop. " +
         s"The expected delay must be greater than 0.")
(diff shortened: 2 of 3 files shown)


actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala
```diff
@@ -75,8 +75,8 @@ trait Scheduler {
    * If the `Runnable` throws an exception the repeated scheduling is aborted,
    * i.e. the function will not be invoked any more.
    *
-   * @throws java.lang.IllegalArgumentException if the given delays exceed the maximum
-   * reach (calculated as: `delay / tickNanos > Int.MaxValue`).
+   * @throws java.lang.IllegalArgumentException if the given delays is zero, negative or exceed the maximum
+   * reach (calculated as: `delay / tickNanos > Int.MaxValue`)
```
pjfanning, line 79: can you undo this change?

Roiocam: what's wrong, i thinkg it is good to have a clarity javadoc.

pjfanning: It's the `.` that was removed - can you add that back?

pjfanning:
@Roiocam have you seen the busy loop behaviour in the real world? What about the case where some sets a very short value (eg 1) - isn't that almost as bad?

Roiocam:
> @Roiocam have you seen the busy loop behaviour in the real world? What about the case where some sets a very short value (eg 1) - isn't that almost as bad?

The issue was discovered in our unit test on CI, and the reason was that our heartbeat configuration was not being loaded correctly.

The "short value" might cause a busy loop, as the CPU is a preemptive resource, so we always get a chance to rest. However, in the case of a zero period, this can lead to the actor system never being able to shut down. 

Please refer to the specific unit test on the link for more details: https://gist.github.com/Roiocam/d317683d54bdbf3afe75b1b945c7f115

pjfanning:
tests fail

seems to be down to a duplicate test name

```
8T08:21:22.8083608Z [12-28 08:21:22.808] [0m[[0m[0minfo[0m] [0m[0m[31morg.apache.pekko.actor.LightArrayRevolverSchedulerSpec *** ABORTED *** (1 millisecond)[0m[0m
2023-12-28T08:21:22.8088023Z [12-28 08:21:22.808] [0m[[0m[0minfo[0m] [0m[0m[31m  org.scalatest.exceptions.DuplicateTestNameException was thrown inside "using scheduleWithFixedDelay" must, construction cannot continue: "Duplicate test name: A LightArrayRevolverScheduler when using scheduleWithFixedDelay must reject periodic tasks scheduled with zero interval" (SchedulerSpec.scala:727)[0m[0m
```

He-Pin:
I checked both Netty and Java, which will throw a IllegalArgumentException.

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -105,12 +105,14 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
...
   override def scheduleWithFixedDelay(initialDelay: FiniteDuration, delay: FiniteDuration)(runnable: Runnable)(
       implicit executor: ExecutionContext): Cancellable = {
+    checkZeroPeriod(delay.toNanos)
     checkMaxDelay(roundUp(delay).toNanos)
     super.scheduleWithFixedDelay(initialDelay, delay)(runnable)
   }
 
   override def schedule(initialDelay: FiniteDuration, delay: FiniteDuration, runnable: Runnable)(
       implicit executor: ExecutionContext): Cancellable = {
+    checkZeroPeriod(delay.toNanos)
```
He-Pin, line 115: I think we should add the check to `scheduleAtFixedRate` and `scheduleWithFixedDelay`.

And inline the method,otherwise it will add a stackframe

Roiocam: make sense,  this is not something that wroth to abstract, use inline we could gain less stack frame

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -200,6 +202,11 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
         s"Task scheduled with [${delayNanos.nanos.toSeconds}] seconds delay, " +
         s"which is too far in future, maximum delay is [${(tickNanos * Int.MaxValue).nanos.toSeconds - 1}] seconds")
 
+  private def checkZeroPeriod(delayNanos: Long): Unit =
+    if (delayNanos <= 0)
+      throw new IllegalArgumentException(
+        "Task scheduled with zero or negative period, which is create an an infinite loop")
```
He-Pin, line 208: use:
- `delay :xxx expected > 0` or `delay should > 0`.

user do not know what's `delayNanos`

Roiocam: what about "Task scheduled with [${delay.toSeconds}] seconds delay, which means creating an infinite loop. The expected delay must be greater than 0."?

He-Pin:
I will go to airport, will try to review this when free

actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
```diff
@@ -713,14 +770,12 @@ class LightArrayRevolverSchedulerSpec extends PekkoSpec(SchedulerSpec.testConfRe
...
     val sched = new LightArrayRevolverScheduler(config.withFallback(system.settings.config), log, tf) {
       override protected def clock(): Long = {
-        // println(s"clock=$time")
         time
       }
 
       override protected def getShutdownTimeout: FiniteDuration = (10 seconds).dilated
 
       override protected def waitNanos(ns: Long): Unit = {
-        // println(s"waiting $ns")
```
pjfanning, line 723: remove the printlns

pjfanning (changes requested):
stray println comments

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -105,12 +105,20 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
 
   override def scheduleWithFixedDelay(initialDelay: FiniteDuration, delay: FiniteDuration)(runnable: Runnable)(
       implicit executor: ExecutionContext): Cancellable = {
+    if (delay.toNanos <= 0)
+      throw new IllegalArgumentException(
+        s"Task scheduled with [${delay.toSeconds}] seconds delay, which means creating an infinite loop. " +
+        s"The expected delay must be greater than 0.")
```
pjfanning, line 111: I'm not going to block this over this - but I would I strongly argue that this should be a function - like checkMaxDelay.

I'm not convinced by all this `inline` debate. Modern JIT compilers are good - in fact, very good - at this. I struggle to accept that this coding pratcice is justified without someone proving it to me.

mdedetrich: Since this is an exception it shouldn't be evaluated in the hot path anyways, or is this an exception which is expected to be thrown and caught in the standard business logic flow?

pjfanning: the pre-existing line right after this does a similar check and throws the same type of exception - `checkMaxDelay(roundUp(delay).toNanos)`

I have no problem with validating the values here.

The scaladoc for this method already has `@throws java.lang.IllegalArgumentException`

Roiocam: You are right about the compiler's knowledge, but this change is mostly about reducing the stack frame.

```diff
java.lang.IllegalArgumentException: Task scheduled with [0] seconds delay, which means creating an infinite loop. The expected delay must be greater than 0.
+       at org.apache.pekko.actor.LightArrayRevolverScheduler.checkDelay(LightArrayRevolverScheduler.scala:205)
	at org.apache.pekko.actor.LightArrayRevolverScheduler.scheduleWithFixedDelay(LightArrayRevolverScheduler.scala:108)
	at org.apache.pekko.actor.Scheduler.scheduleWithFixedDelay(Scheduler.scala:157)
	at org.apache.pekko.actor.Scheduler.scheduleWithFixedDelay$(Scheduler.scala:149)
	at org.apache.pekko.actor.LightArrayRevolverScheduler.scheduleWithFixedDelay(LightArrayRevolverScheduler.scala:50)
	at org.apache.pekko.actor.LightArrayRevolverSchedulerSpec.$anonfun$new$52(SchedulerSpec.scala:688)
```

It seems like everything is okay because most of the time we only care about the top frame of the stack.

Roiocam: This is acceptable, even in real-world case, where the dispatcher stack frame is only a few. while other frames are more like noise.

```log
[2024-01-04 10:09:59,876] [ERROR] [ispatcher-55] [akka.actor.typed.Behavior$    ]: Supervisor StopSupervisor saw failure: Task scheduled with [42949672] seconds delay, which is too far in future, maximum delay is [21474835] seconds MDC: {akkaAddress=akka://bank-account@127.0.0.1:2551, akkaSource=akka://bank-account/system/sharding/xxxxxxxxxx, sourceActorSystem=bank-account}
java.lang.IllegalArgumentException: Task scheduled with [42949672] seconds delay, which is too far in future, maximum delay is [21474835] seconds
	at akka.actor.LightArrayRevolverScheduler.checkMaxDelay(LightArrayRevolverScheduler.scala:216)
	at akka.actor.LightArrayRevolverScheduler.scheduleWithFixedDelay(LightArrayRevolverScheduler.scala:98)
	at akka.actor.typed.internal.adapter.SchedulerAdapter.scheduleWithFixedDelay(SchedulerAdapter.scala:44)
	at akka.actor.typed.internal.TimerSchedulerImpl.startTimer(TimerSchedulerImpl.scala:127)
	at akka.actor.typed.internal.TimerSchedulerImpl.startTimerWithFixedDelay(TimerSchedulerImpl.scala:101)
	at akka.actor.typed.internal.TimerSchedulerCrossDslSupport.startTimerWithFixedDelay(TimerSchedulerImpl.scala:63)
	at akka.actor.typed.internal.TimerSchedulerCrossDslSupport.startTimerWithFixedDelay$(TimerSchedulerImpl.scala:62)
	at akka.actor.typed.internal.TimerSchedulerImpl.startTimerWithFixedDelay(TimerSchedulerImpl.scala:83)
	at akka.actor.typed.javadsl.BuiltBehavior.receive(BehaviorBuilder.scala:197)
	at akka.actor.typed.javadsl.BuiltBehavior.receive(BehaviorBuilder.scala:186)
	at akka.actor.typed.Behavior$.interpret(Behavior.scala:274)
	at akka.actor.typed.Behavior$.interpretMessage(Behavior.scala:230)
	at akka.actor.typed.internal.InterceptorImpl$$anon$2.apply(InterceptorImpl.scala:57)
	at akka.actor.typed.internal.InterceptorImpl.receive(InterceptorImpl.scala:87)
	at akka.actor.typed.Behavior$.interpret(Behavior.scala:274)
	at akka.actor.typed.Behavior$.interpretMessage(Behavior.scala:230)
	at akka.actor.typed.internal.InterceptorImpl$$anon$2.apply(InterceptorImpl.scala:57)
	at akka.actor.typed.internal.SimpleSupervisor.aroundReceive(Supervision.scala:131)
	at akka.actor.typed.internal.InterceptorImpl.receive(InterceptorImpl.scala:85)
	at akka.actor.typed.Behavior$.interpret(Behavior.scala:274)
	at akka.actor.typed.Behavior$.interpretMessage(Behavior.scala:230)
	at akka.actor.typed.internal.adapter.ActorAdapter.handleMessage(ActorAdapter.scala:128)
	at akka.actor.typed.internal.adapter.ActorAdapter.aroundReceive(ActorAdapter.scala:107)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:579)
	at akka.actor.ActorCell.invoke(ActorCell.scala:547)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:270)
	at akka.dispatch.Mailbox.run(Mailbox.scala:231)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:243)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:175)
```

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -105,12 +105,14 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
 
   override def scheduleWithFixedDelay(initialDelay: FiniteDuration, delay: FiniteDuration)(runnable: Runnable)(
       implicit executor: ExecutionContext): Cancellable = {
+    checkPeriod(delay)
```
Roiocam, line 108: I think `checkPeriod` is clarify than `checkDelay`, delay may be misunderstand to initialDelay.

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
Roiocam: There is another viewpoint on this: "It is fine because it will always be at the top of the stack frame." I have made an example to verify it, see the  another comment on below.

mdedetrich:
> stray println comments

@pjfanning Do you want to recheck this?

pjfanning:
> > stray println comments
> 
> @pjfanning Do you want to recheck this?

the printlns are commented out still but the lines should be removed altogether

Roiocam:
> > the printlns are commented out still but the lines should be removed altogether
> 
> @Roiocam Do you want to remove this so we can get the PR over the finish line?

Done, let's keep this pull request simple.

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -193,6 +195,12 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
       task
     }
 
+  private def checkPeriod(delay: FiniteDuration): Unit =
```
He-Pin, line 198: @Roiocam @mdedetrich I think when we are not using @inline and iniline def, we may just online this method？but I think the jvm should do the same after some iterations.

He-Pin (approved):
Lgtm

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
He-Pin: How about test with the length field directly？

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
mdedetrich, line 198: I think @He-Pin made a comment here before which now seems to be gone regarding inlining.

I just wanted to say you can inline this method if you want, you just have to make a `scala-2` version using the `@inline` annotation and a `scala-3` version with the `inline` keyword.

The easiest way to do this would be to place the `checkPeriod` function into a `private[pekko]` trait in the `scala-2`/`scala-3` source folders respectively and make `class LightArrayRevolverScheduler` extend that trait.

He-Pin: Yes, we can do that in a separate pr, I have to say, this is a hot path.

pjfanning: Hotspot Compiler is very good at optimising this. There is zero evidence that this code needs excess optimisation.

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
Roiocam: done.

He-Pin (approved):
lgtm

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
mdedetrich: > Hotspot Compiler is very good at optimising this. There is zero evidence that this code needs excess optimisation.

I am also skeptical but it can be proved with benchmarks in a separate PR if need be

mdedetrich:
@pjfanning Do you want to re-review it now?

pjfanning (approved):
lgtm

pjfanning:
@Roiocam could you create a backport PR (to 1.0.x branch)?

He-Pin:
Thank you @Roiocam

PR #894, opened 2023-12-31
Remove Scala 3 inline keyword from Java specific conversion methods

mdedetrich:
The response in https://github.com/lampepfl/dotty/issues/19346#issuecomment-1872478851 confirms that its by design that methods marked `inline` in Scala 3 are not meant to be called within Java, it is intended for such inlined methods  that its meant to **_enforce/guarantee_** the method is inlined which for obvious reasons is not possible if the method is called within Java considering how `inline` in Scala 3 has become a generic construct (this is different from the `@inline` annotation in Scala 2 which you can freely call within Java, it just won't be inlined).

Due to this, this PR removes the Scala 3 `inline` keyword from methods which are designed to be called within Java as well as adding a test to ensure that these designated Java methods can actually be compiled within a Java source. You can add the `inline` keyword to the relevant methods to confirm that the `JavaConverterScala3InlineTest` test properly catches this.

Going forward what this means is that if any of the code within Pekko projects happen to call these conversion methods, if the method is being called within a `*.scala` source then it should use the extension method versions (i.e. the methods defined within `implicit final class`) in order to get the inline benefits when compiled with Scala 3. Thats the best that can be done in this situation and it also explains the rest of the changes in this PR.

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.util;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+
+import scala.Option;
+import scala.concurrent.Future;
+
+/**
+ * These tests are here to ensure that methods from {@link org.apache.pekko.util.FutureConverters},
+ * {@link org.apache.pekko.util.JavaDurationConverters} and {@link
+ * org.apache.pekko.util.OptionConverters} for use within Java can be compiled from with Java
+ * sources. This is because methods marked with the Scala 3 inline keyword cannot be called from
+ * within Java (see https://github.com/lampepfl/dotty/issues/19346)
+ */
+public class JavaConverterScala3InlineTest {
+  public void compileTest() {
+    OptionConverters.toScala(Optional.empty());
+    OptionConverters.toScala(OptionalDouble.of(0));
+    OptionConverters.toScala(OptionalInt.of(0));
+    OptionConverters.toScala(OptionalLong.of(0));
+    OptionConverters.toJava(Option.empty());
+
+    FutureConverters.asJava(Future.successful(""));
+    FutureConverters.asScala(CompletableFuture.completedFuture(""));
+
+    JavaDurationConverters.asFiniteDuration(Duration.ofMillis(0));
+  }
+}
--- a/actor/src/main/scala/org/apache/pekko/actor/ActorSelection.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/ActorSelection.scala
@@ -108,8 +108,10 @@ abstract class ActorSelection extends Serializable {
    * supplied `timeout`.
    */
   @deprecated("Use the overloaded method resolveOne which accepts java.time.Duration instead.", since = "Akka 2.5.20")
-  def resolveOneCS(timeout: FiniteDuration): CompletionStage[ActorRef] =
-    FutureConverters.asJava[ActorRef](resolveOne(timeout))
+  def resolveOneCS(timeout: FiniteDuration): CompletionStage[ActorRef] = {
+    import FutureConverters._
+    resolveOne(timeout).asJava
+  }
 
   /**
    * Java API for [[#resolveOne]]
@@ -133,8 +135,9 @@ abstract class ActorSelection extends Serializable {
    * supplied `timeout`.
    */
   def resolveOne(timeout: java.time.Duration): CompletionStage[ActorRef] = {
+    import FutureConverters._
     import JavaDurationConverters._
-    FutureConverters.asJava[ActorRef](resolveOne(timeout.asScala))
+    resolveOne(timeout.asScala).asJava
   }
 
   override def toString: String = {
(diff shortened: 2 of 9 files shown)


actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
```
laglangyue, line 7: this  is a new File,can we remove the description about akka

actor/src/main/scala/org/apache/pekko/actor/ActorSelection.scala
```diff
@@ -108,8 +108,10 @@ abstract class ActorSelection extends Serializable {
    * supplied `timeout`.
    */
   @deprecated("Use the overloaded method resolveOne which accepts java.time.Duration instead.", since = "Akka 2.5.20")
-  def resolveOneCS(timeout: FiniteDuration): CompletionStage[ActorRef] =
-    FutureConverters.asJava[ActorRef](resolveOne(timeout))
+  def resolveOneCS(timeout: FiniteDuration): CompletionStage[ActorRef] = {
+    import FutureConverters._
+    resolveOne(timeout).asJava
```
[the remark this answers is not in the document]
mdedetrich: > this changs seems not related with this pr. 

It is, specifically this part from the PR description

> Going forward what this means is that if any of the code within Pekko projects happen to call these conversion methods, if the method is being called within a *.scala source then it should use the extension method versions (i.e. the methods defined within implicit final class) in order to get the inline benefits when compiled with Scala 3. Thats the best that can be done in this situation and it also explains the rest of the changes in this PR.

> The API here should be shown to Java programmers, and I think the original way is easier for Java programmers to understand.

This change has zero effect on Java code calling `def resolveOneCS(timeout: FiniteDuration): CompletionStage[ActorRef]`, the inlining is only ocurring in the internal implementation (if thats what you are were implying in your question).

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
mdedetrich: @pjfanning Whats your views on this, original Apache header also makes sense but this header is the one copied from https://github.com/apache/incubator-pekko/blob/b0fdac259bd57fdd481483f3fe9a7aec6e1ff38a/actor-tests/src/test/scala/org/apache/pekko/util/Scala212CompatTest.scala#L1-L8 which is a comparable new test that was written

He-Pin: A copy should keep the lightbend header

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * These tests are here to ensure that methods from {@link org.apache.pekko.util.FutureConverters},
+ * {@link org.apache.pekko.util.JavaDurationConverters} and {@link
+ * org.apache.pekko.util.OptionConverters} for use within Java can be compiled from with Java
+ * sources. This is because methods marked with the Scala 3 inline keyword cannot be called from
+ * within Java (see https://github.com/lampepfl/dotty/issues/19346)
+ */
+public class JavaConverterScala3InlineTest {
+  public void Test() {
```
He-Pin, line 27: @Test

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
mdedetrich: This isn't a copy, it's a new source file

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
```diff
@@ -0,0 +1,39 @@
...
+    OptionConverters.toScala(Optional.empty());
+    OptionConverters.toScala(OptionalDouble.of(0));
+    OptionConverters.toScala(OptionalInt.of(0));
+    OptionConverters.toScala(OptionalLong.of(0));
+    OptionConverters.toJava(Option.empty());
+
+    FutureConverters.asJava(Future.successful(""));
+    FutureConverters.asScala(CompletableFuture.completedFuture(""));
+
+    JavaDurationConverters.asFiniteDuration(Duration.ofMillis(0));
```
He-Pin, line 37: Add Assertions

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
He-Pin: If the content is new ,then only asf header

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
mdedetrich: The point of the rest isn't to validate that the methods are correct but rather that the `*.java` compiles.

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
mdedetrich: Same as https://github.com/apache/incubator-pekko/pull/894#discussion_r1438808197

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
pjfanning: brand new files should use the standard ASF header - unless it has code copied in from an existing file in which case it should use the headers from that file

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
pjfanning: shouldn't method names start with lower case chars? - so `Test()` doesn't seem like a good method name

actor-tests/src/test/java/org/apache/pekko/util/JavaConverterScala3InlineTest.java
pjfanning: If anything Scala212CompatTest seems to have the wrong header - that file looks like it was written as a fresh set of tests by the Pekko team.

mdedetrich: > If anything Scala212CompatTest seems to have the wrong header - that file looks like it was written as a fresh set of tests by the Pekko team.

Thanks, ill fix the header for that file in another PR as well as fixing the headers in this PR

mdedetrich:
I have updated the PR, fixing the header for `JavaConverterScala3InlineTest` and renaming the method to `compileTest` (so that its also clear that the point of the test is to see that the code compiles, not the actual runtime behaviour of the code being executed).

pjfanning (approved):
lgtm

looks like this is needed to fix the broken job for pekko-connectors https://github.com/apache/incubator-pekko-connectors/actions/runs/7373523484

pravega connector compile of a Java class fails with Scala 3

mdedetrich:
> pravega connector compile of a Java class fails with Scala 3

Indeed this appears to be related, merging now

mdedetrich:
@pjfanning I can confirm that this PR did indeed fix the issue with pekko-connectors (see https://github.com/apache/incubator-pekko-connectors/actions/runs/7391882740)

PR #902, opened 2024-01-03
feat: Reimplement MapConcat operator without statefulMapConcat.

He-Pin:
Motivation:
Reduce the usage of `statefulMapConcat`.

Result:
A dedicated operator implementation for `mapConcat` operator.

Note:
Does not implemente the `ContextPropagation`, as I think that was for internal commercial hook?

refs: https://github.com/apache/incubator-pekko-http/issues/395 and chatting on slack.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/MapConcat.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/MapConcat.scala
@@ -50,7 +50,7 @@ private[pekko] final class MapConcat[In, Out](f: In => IterableOnce[Out])
 
       private var currentIterator: Iterator[Out] = _
 
-      @inline private def hasNext = currentIterator != null && currentIterator.hasNext
+      private def hasNext = currentIterator != null && currentIterator.hasNext
 
       override def onPush(): Unit =
         try {
@@ -58,7 +58,7 @@ private[pekko] final class MapConcat[In, Out](f: In => IterableOnce[Out])
           tryPushAndPull()
         } catch handleException
 
-      @inline private def tryPushAndPull(): Unit =
+      private def tryPushAndPull(): Unit =
         try {
           if (hasNext) {
             push(out, currentIterator.next())
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/MapConcat.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/MapConcat.scala
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


mdedetrich (changes requested):
I noticed that you put in `@inline` for  various methods. If you are relying on the inlining, then you need to make a version in `scala-3` that uses the `inline` keyword while moving this implementation to the `scala-2` folder so that the performance is consistent. 

Otherwise remove the `@inline`

He-Pin:
@mdedetrich Ok. let me remove it for now, and let the JVM jit do the thing, as these are private method.

mdedetrich:
> @mdedetrich Ok. let me remove it for now, and let the JVM jit do the thing, as these are private method.

If you can prove that both the scala-2 and scala-3 inliner makes a performance impact then you can make both the `@inline` and `inline` keyword variants. Also do note that the Scala 2 inliner is now enabled for the entire codebase, so if the default heuristics for the Scala 2 inliner think that the method should be inlined then it will.

If you want to verify that the scala 2 inliner is actually inlining those methods (or not) then you can use the `sbt -Dpekko.no.inline=yes` to turn off the inliner (make sure to do `clean` inbetween switching on and off the inliner).

mdedetrich (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
```diff
@@ -1007,8 +1007,7 @@ trait FlowOps[+Out, +Mat] {
    *
    * '''Cancels when''' downstream cancels
    */
-  @nowarn("msg=deprecated")
-  def mapConcat[T](f: Out => IterableOnce[T]): Repr[T] = statefulMapConcat(() => f)
+  def mapConcat[T](f: Out => IterableOnce[T]): Repr[T] = via(new MapConcat(f))
```
laglangyue, line 1010: Maybe not the correct opinion : 
(1) Does javadsl require similar modifications? maybe refer to `org.apache.pekko.stream.javadsl.Flow#statefulMapConcat`
(2) can we use object or case object for `MapConcat`, it is refered to `org.apache.pekko.stream.impl.fusing.MapAsync`.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/MapConcat.scala
```diff
@@ -0,0 +1,90 @@
...
+import pekko.util.ccompat._
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+@ccompatUsedUntil213
+private[pekko] final class MapConcat[In, Out](f: In => IterableOnce[Out])
+    extends GraphStage[FlowShape[In, Out]] {
+  require(f != null, "f function should not be null")
```
laglangyue, line 40: It seems to be defensive, is it nessnary check?
We should specify the parent of the function, like `f function of MapConcat should not be null` ?

He-Pin: Yes，better than a materilixation error.
As the stack trace can show the detail, that's information can be ommited

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
He-Pin: 1. the javadsl's mapconcat is using scaladsl's mapconcat
2. case class will make the jar bigger and we have no pattern matching here.

Issue #900, opened 2024-01-03
ByteStringBuilder can stackoverflow when call addAll with LinearSeq


He-Pin:
```scala
      ByteString.newBuilder.addAll(List[Byte]('a'))
```

results:
```scala
java.lang.StackOverflowError
	at org.apache.pekko.util.ByteStringBuilder.addAll(ByteString.scala:1082)
	at scala.collection.mutable.Growable.$plus$plus$eq(Growable.scala:69)
	at scala.collection.mutable.Growable.$plus$plus$eq$(Growable.scala:69)
	at org.apache.pekko.util.ByteStringBuilder.addAll(ByteString.scala:1193)
	at org.apache.pekko.util.ByteStringBuilder.addAll(ByteString.scala:1082)
	at scala.collection.mutable.Growable.$plus$plus$eq(Growable.scala:69)
	at scala.collection.mutable.Growable.$plus$plus$eq$(Growable.scala:69)
	at org.apache.pekko.util.ByteStringBuilder.addAll(ByteString.scala:1193)
```

seems like some overloading issue.

pjfanning:
@He-Pin @mdedetrich I have a possible fix in https://github.com/apache/incubator-pekko/pull/903

pjfanning:
Until we get a new Pekko 1.0.3 release out, the workaround is to use `++=` instead of `addAll`.

PR #903, opened 2024-01-03
fix ByteStringBuilder.addAll

pjfanning:
see #900 

++= is an alias for addAll but super.++= incorrectly aliases the subclass addAll not the super.addAll.

My test in this PR reproduced the stackoverflow and started working after my code change to ByteStringBuilder

### DIFF
--- a/actor-tests/src/test/scala-2.13+/org/apache/pekko/util/ByteStringBuilderSpec.scala
+++ b/actor-tests/src/test/scala-2.13+/org/apache/pekko/util/ByteStringBuilderSpec.scala
@@ -26,5 +26,9 @@ class ByteStringBuilderSpec extends AnyWordSpec with Matchers {
       val result: ByteString = ByteString.newBuilder.addAll(List[Byte]('a')).result()
       result shouldEqual ByteString("a")
     }
+    "handle ++=" in {
+      val result: ByteString = ByteString.newBuilder.++=(List[Byte]('a')).result()
+      result shouldEqual ByteString("a")
+    }
   }
 }
--- a/actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
@@ -1190,7 +1190,7 @@ final class ByteStringBuilder extends Builder[Byte, ByteString] {
           _length += seq.length
         }
       case _ =>
-        super.++=(xs)
+        super.addAll(xs)
     }
     this
   }
--- a/build.sbt
+++ b/build.sbt
@@ -140,6 +140,14 @@ lazy val actor = pekkoModule("actor")
 lazy val actorTests = pekkoModule("actor-tests")
   .dependsOn(testkit % "compile->compile;test->test", actor)
   .settings(Dependencies.actorTests)
+  .settings(Test / unmanagedSourceDirectories ++= {
+    if (scalaBinaryVersion.value == "2.13" || scalaBinaryVersion.value.startsWith("3")) {
+      Seq(
+        (Test / scalaSource).value.getParentFile / "scala-2.13+")
+    } else {
+      Seq.empty
+    }
+  })
   .enablePlugins(NoPublish)
   .disablePlugins(MimaPlugin)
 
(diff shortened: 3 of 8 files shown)


mdedetrich (approved):
lgtm. Also think this is a candidate for a backport since current behaviour is completely broken.

actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
mdedetrich, line 1: @pjfanning What about the scala-2.12/scala-3 version of `ByteString`?

pjfanning: I'm looking at those now. I had just noticed that the code varied by Scala version.

Scala 2.12 builder does not support addAll. Testing Scala 3.3 now.

pjfanning: I think I've sorted out the scala variants now. As I said above, Scala 2.12 builder does not support addAll - so I haven't added support there.

mdedetrich: Thanks, resolving

mdedetrich (changes requested):
The changes to `build.sbt` for different source directories is not necessary?

build.sbt
```diff
@@ -140,6 +140,14 @@ lazy val actor = pekkoModule("actor")
 lazy val actorTests = pekkoModule("actor-tests")
   .dependsOn(testkit % "compile->compile;test->test", actor)
   .settings(Dependencies.actorTests)
+  .settings(Test / unmanagedSourceDirectories ++= {
```
mdedetrich, line 143: This looks fishy, why are we doing this? Pekko has already been setup to handle all the different types of `scala-*` folders if we wan't to write tests for specific Scala versions and in cases like this I would prefer to KISS

mdedetrich (changes requested):
The changes to `build.sbt` for different source directories is not necessary?

build.sbt
pjfanning: the tests in `scala-2.13+` did not execute without this change

mdedetrich: https://github.com/apache/incubator-pekko/blob/7bee80e05847cd7a1198775b45f831f5ce3227f3/build.sbt#L132-L135 Is probably what you want then

pjfanning: That code does not handle the '+' in `2.13+` - I don't want to duplicate the test that is in 2.13+. If I duplicated the code in `scala-3` and `scala-2.13` - sbt handles those source dirs automatically. It doesn't appear to handle `scala-2.13+` automatically.

mdedetrich: > It doesn't appear to handle `scala-2.13+` automatically.

Indeed sbt doesn't handle this automatically, it would be ideal find the existing code in the build definition that does this so we can reuse it

actor-tests/src/test/scala-2.13+/org/apache/pekko/util/ByteStringBuilderSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.pekko.util
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class ByteStringBuilderSpec extends AnyWordSpec with Matchers {
+  "ByteStringBuilder" should {
+    "handle addAll" in {
```
He-Pin, line 25: Can adds liner sequence of bytes？

build.sbt
mdedetrich: Ah I found it https://github.com/apache/incubator-pekko/blob/main/project/PekkoBuild.scala#L161-L168 . I think if you modify this so it also affects `Test / unmanagedSourceDirectories` then it should fix the same issue in a more principled way.

actor-tests/src/test/scala-2.13+/org/apache/pekko/util/ByteStringBuilderSpec.scala
pjfanning: I updated the test names

build.sbt
pjfanning: I use IDEA and it is ignoring that file. If I search for strings, it ignores that file - won't even let me open the file to view it. I modified it with a text editor and `sbt ++2.13 "show actor-tests/Test/unmanagedSourceDirectories"` has the right results now. I have committed a change.

mdedetrich (approved):
lgtm

He-Pin (approved):
lgtm with good test

Issue #904, opened 2024-01-04
Can not checkout the working tree on Windows


He-Pin:
![image](https://github.com/apache/incubator-pekko/assets/501740/565b933a-57c8-4179-b105-c79e7b5c3580)

On Windows 11
```
C:\Users\hepin\IdeaProjects>git clone git@github.com:He-Pin/incubator-pekko.git
Cloning into 'incubator-pekko'...
remote: Enumerating objects: 463388, done.
remote: Counting objects: 100% (3929/3929), done.
remote: Compressing objects: 100% (446/446), done.
remote: Total 463388 (delta 3563), reused 3538 (delta 3392), pack-reused 459459
Receiving objects: 100% (463388/463388), 254.24 MiB | 3.54 MiB/s, done.
Resolving deltas: 100% (233033/233033), done.
error: invalid path 'actor/src/main/mima-filters/1.1.0.backwards.excludes/option-converters.excludes '
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

mdedetrich:
Found the issue, PR made at https://github.com/apache/incubator-pekko/pull/905

PR #905, opened 2024-01-04
Fix Windows path issue with option-converters.excludes

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko/issues/904

The issue is that there was a trailing space (i.e. ` `) on the filename extension


He-Pin (approved):
lgtm

PR #912, opened 2024-01-07
Add second sbt-osgi in dependency update ignore list

mdedetrich:
(cherry picked from commit 688eac7bf5bcd7ca99d4c926b7a16d57598c79e4)

While it can be argued that this is pointless since we aren't going to enable scala-steward on the `1.0.x` branch, I still think its a good idea to do this as a defensive strategy in case in the future we change our minds for w/e reason.


He-Pin (approved):
Lgtm

Issue #918, opened 2024-01-09
Exposes the current interpreter when init/prestart the stage.


He-Pin:
Motivation:
Currently, we can't access the current interpreter when the stage is preStart. eg using it with `Flow.fromMaterializer`.

Result:
Can access the interpreter when necessary.

PR #919, opened 2024-01-09
feat: Exposes GraphInterpreter when initing the stages.

He-Pin:
Motivation:
Add support for exposing the current interpreter when starting a stage.

Result:
`GraphInterpreter.currentInterpreter` can be used with `Flow.fromMaterializer`

refs: https://github.com/apache/incubator-pekko/issues/918

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FromMaterializationSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FromMaterializationSpec.scala
@@ -14,6 +14,7 @@
 package org.apache.pekko.stream.scaladsl
 
 import org.apache.pekko
+import org.apache.pekko.stream.impl.fusing.GraphInterpreter
 import pekko.NotUsed
 import pekko.stream.Attributes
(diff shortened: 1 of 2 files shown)


He-Pin (commented):
Exposing `GraphInterpreter` via `GraphInterpreter.currentInterpreter` for use with `Flow.fromMaterializer`.

1. **Use case**: The linked issue #918 would provide context on why this is needed. Exposing internal interpreter state is a significant API decision that needs careful consideration of stability guarantees.

2. **API stability**: Making the interpreter accessible to user code creates a new public API surface that would need to be maintained. Consider whether a more limited/controlled API would suffice.

3. **Staleness**: Open since January 2024 with no reviews. This is a relatively small change (2 files) but has architectural implications. @He-Pin can you provide more context on the use case that requires access to the GraphInterpreter?

He-Pin:
Closing due to staleness (open since January 2024 with no reviews) and merge conflicts. The underlying issue (#918 -- NPE when accessing currentInterpreter during preStart) is a real bug that deserves attention. A fresh PR with the fix would be more actionable.

PR #924, opened 2024-01-09
feat: Add onTermination hook to PekkoForkJoinWorkerThread.

He-Pin:
Motivation:
Internal usage, we are using ThreadLocal for RPC tracing, and some time the ThreadLocal can leak. So I would like to add this hook for later aop usage, where I can do a clean up with `EagleEye.cleanUpRpcContext()`.

I would like to implement it with byte-buddy

Result:
A hook for clean up.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/ThreadPoolBuilder.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/ThreadPoolBuilder.scala
@@ -176,6 +176,9 @@ object MonitorableThreadFactory {
   private[pekko] class PekkoForkJoinWorkerThread(_pool: ForkJoinPool)
       extends ForkJoinWorkerThread(_pool)
       with BlockContext {
+
+    override def onTermination(exception: Throwable): Unit = super.onTermination(exception)
+
     override def blockOn[T](thunk: => T)(implicit permission: CanAwait): T = {
       val result = new AtomicReference[Option[T]](None)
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/ThreadPoolBuilder.scala
```diff
@@ -176,6 +176,9 @@ object MonitorableThreadFactory {
   private[pekko] class PekkoForkJoinWorkerThread(_pool: ForkJoinPool)
       extends ForkJoinWorkerThread(_pool)
       with BlockContext {
+
+    override def onTermination(exception: Throwable): Unit = super.onTermination(exception)
```
pjfanning, line 180: Why would we need to override a function just to call its super class implementation?

He-Pin: That's a hook where we can do aop around it, override it in subclass can limit the scope.

He-Pin: I mean I can use `Byte-Buddy` to aop it then,

pjfanning: I don't understand why this is not inherited anyway from the superclass.

He-Pin: just tried, it works too.

Issue #786, opened 2023-11-11
add Source/Flow.mapWithResource


pjfanning:
Based on https://github.com/apache/incubator-pekko/discussions/553

I think we have covered the other functions mentioned there - mapAsyncPartitioned and contramap

@He-Pin no hurry on this but I'm just tracking it a nice-to-have for the v1.1.0 release.

PR #931, opened 2024-01-13
feat: Add mapWithResource stream operator.

He-Pin:
Motivation:
Make every stream can handle with resource.

Like Flux's `using` operator.

refs:
https://github.com/apache/incubator-pekko/issues/786


docs/src/main/java/docs/ddata/protobuf/msg/TwoPhaseSetMessages.java
```diff
@@ -18,81 +18,87 @@
...
   public static void registerAllExtensions(
       org.apache.pekko.protobufv3.internal.ExtensionRegistry registry) {
-    registerAllExtensions(
-        (org.apache.pekko.protobufv3.internal.ExtensionRegistryLite) registry);
+    registerAllExtensions((org.apache.pekko.protobufv3.internal.ExtensionRegistryLite) registry);
   }
-  public interface TwoPhaseSetOrBuilder extends
+
+  public interface TwoPhaseSetOrBuilder
+      extends
```
pjfanning, line 31: Do we need the code format changes?

pjfanning: the PR is big enough without making us have to read through all the unrelated code reformatting

He-Pin: I was run javafmtall，I CAN revert this tomorrow

pjfanning: if you like, you could do a separate PR that does javafmtAll - we merge that PR and then you can rebase this PR. Is that ok?

He-Pin: I have finally remove these files from the PR :(

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -2225,7 +2225,7 @@ private[pekko] final class StatefulMap[S, In, Out](create: () => S, f: (S, In) =
   private val out = Outlet[Out]("StatefulMap.out")
   override val shape: FlowShape[In, Out] = FlowShape(in, out)
 
-  override protected def initialAttributes: Attributes = DefaultAttributes.statefulMap and SourceLocation.forLambda(f)
+  override protected def initialAttributes: Attributes = Attributes(SourceLocation.forLambda(f))
```
He-Pin, line 2228: This is needed to get the tests in MapWithResourceSpec work, which need the name to be `mapWithResource`

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
```diff
@@ -770,6 +770,47 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
...
+   *
+   * @tparam R the type of the resource
+   * @tparam T the type of the output elements
+   * @param create function that creates the resource
+   * @param f function that transforms the upstream element and the resource to output element
+   * @param close function that closes the resource, optionally outputting a last element
+   * @since 1.1.0
+   */
+  def mapWithResource[R, T](
+      create: java.util.function.Supplier[R],
```
He-Pin, line 806: I checked Flux, where a `Callable` is used, which can throw exception.
But I think, we did not expect any exception to be throw ,so a Supplier.

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowMapWithResourceSpec.scala
```diff
@@ -0,0 +1,416 @@
...
+  private val ex = new Exception("TEST") with NoStackTrace
+
+  private val manyLines = {
+    ("a" * 100 + "\n") * 10 +
+    ("b" * 100 + "\n") * 10 +
+    ("c" * 100 + "\n") * 10 +
+    ("d" * 100 + "\n") * 10 +
+    ("e" * 100 + "\n") * 10 +
+    ("f" * 100 + "\n") * 10
+  }
```
laglangyue, line 58: ('a' to 'f')  .map(_.toString)  .map(_*100).map(_+"/n") ....etc

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowMapWithResourceSpec.scala
```diff
@@ -0,0 +1,416 @@
...
+  private val manyLinesArray = manyLines.split("\n")
+
+  private val manyLinesPath = {
+    val file = Files.createFile(fs.getPath("/testMapWithResource.dat"))
+    Files.write(file, manyLines.getBytes(StandardCharsets.UTF_8))
+  }
+  private def newBufferedReader() = Files.newBufferedReader(manyLinesPath, StandardCharsets.UTF_8)
+
+  private def readLines(reader: BufferedReader, maxCount: Int): List[String] = {
```
laglangyue, line 67: why not use `scala.io.Source`? 
 ```
implicit val enc= Codec.UTF8
 val source = Source.fromFile(new File(""))
    source.getLines()...etc
  ```

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowMapWithResourceSpec.scala
He-Pin: Was copied from another test iirc, it's test code , and very explicitly, anyway, thanks

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowMapWithResourceSpec.scala
```diff
@@ -0,0 +1,416 @@
...
+    if (maxCount == 0) {
+      return List.empty
+    }
+
+    @tailrec
+    def loop(builder: ListBuffer[String], n: Int): ListBuffer[String] = {
```
laglangyue, line 73: I try scala's style and it looks good, why not use api of `scala io` and `scalla stream`
```
    implicit val enc= Codec.UTF8
    val source = Source.fromFile(new File(""))
    val reader = source.bufferedReader()
    reader.lines().takeWhile(_ eq null).toArray.mkString("")
```

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
```diff
@@ -1082,7 +1082,47 @@ trait FlowOps[+Out, +Mat] {
...
+   * '''Cancels when''' downstream cancels
+   *
+   * @tparam R the type of the resource
+   * @tparam T the type of the output elements
+   * @param create function that creates the resource
+   * @param f function that transforms the upstream element and the resource to output element
+   * @param close function that closes the resource, optionally outputting a last element
+   * @since 1.1.0
+   */
+  def mapWithResource[R, T](create: () => R)(f: (R, Out) => T, close: R => Option[T]): Repr[T] =
```
laglangyue, line 1119: For close function
`Using for scala3`
`scala.util.control.Exception.ultimately for scala2`
Do we need to consider such usage case for this function?

He-Pin: Would you like elaborate?

laglangyue: Taking scala3 as an example, when I close or use a stream, I will write this code
```
Using (Source. from (File...)) {
  // Todo something
}
```
I won't implement a close function. as also to Java8 `try (resource) {}`
I hope Pekko can do the same. Can we use `Using feature` and remove the params  `close:R=>Option[T]`
I haven't seen the implementation principle of `Using in scala3` yet, so I'm not sure how to define it, maybe further optimization in another PR

He-Pin: I see , that would be nice addition.
1. directly support for Java's AutoCloseable.
2. directly support for Scala's Using.

But I think those two doesn't related to this PR, because the nice feature of this operator is "optional emit with the resource/final signal".

But and what your suggestion can come up with another PR.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
pjfanning: I don't think we should assume `create` call can never fail. Can you use `Callable` instead?

He-Pin: I think a user should handle the maybe exception explicitly, with some kind of `try/retry` otherwise they will wondering , wow, why my stream just failed,ouch, I need somekind of retry.

mdedetrich: Shouldn't we be using one of the interfaces in https://github.com/apache/incubator-pekko/blob/b0fdac259bd57fdd481483f3fe9a7aec6e1ff38a/actor/src/main/scala/org/apache/pekko/japi/function/Function.scala which are deliberately designed to be used within Pekko because they support throwing exceptions in the `apply`?

He-Pin: eg: fs2's `Stream#bracket` where `def bracket[F[_], R](acquire: F[R])(release: R => F[Unit]): Stream[F, R]` the `F` is not limited to `Try` too. so does the `Resource.make`.

And if a use change the checked exception to a runtime exception, the flow will fails too, eg: using `@SneakyThrows` .

He-Pin: When we expecting an exception to be throw then that function should be used, otherwise, I a more java native one should be better.
For this operator,  user can using a `RestartFlow` with this or with restart strategy to handle exception more properly, the unexpected one.

A function can throw implies that we *wanting* the exception.

mdedetrich: I don't think its even a question of want, its rather that if a user ever tried to supply a standard Java lambda that happens to have a `throw` they cannot actually do this because it would fail to compile, i.e. not even a 

```scala
new java.lang.Runnable {
  override def run () {
     ...// do some logic
     if (somethingFatal)
        throws new IllegalArgumentException()
  }
}
```

can compile because `Runnable` doesn't allow exceptions in the `@run` method. This is why those `FunctionalInterface`'s can exist in Pekko.

So unless we can say thats it in **_every single case_** its wrong to throw an exception in the provided lambda, we should use the Pekko provided functional interfaces.

mdedetrich: So I just did a search and found another case where we have a [similar `create` argument ](https://github.com/apache/incubator-pekko/blob/43bab55616f40c0cf5ece07d8cd07110a3027c7d/stream/src/main/scala/org/apache/pekko/stream/javadsl/Sink.scala#L514) and as you can see it uses the pekko `Creator` functional interface (see https://github.com/apache/incubator-pekko/blob/b0fdac259bd57fdd481483f3fe9a7aec6e1ff38a/actor/src/main/scala/org/apache/pekko/japi/function/Function.scala#L90-L97)

I think its genuinely expected that `create` like methods such as this can possible throw exceptions.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
mdedetrich: I think we need a github discussion on this, there are lots of ways we can do automatic resource management i.e. anything from scala.util.Using (which is also available in Scala 2.12) to more extensive methods available in cats-effect/ZIO.

This is something that really needs someone with extensive knowledge to look into it because its actually incredibly difficult to guarantee that resources are cleaned up in all cases.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
mdedetrich: > If we expecting exceptions, we will need change every function, then the question is why scala dsl doesn't using these?

What do you mean? As far as I can tell, every single `create` allocator style parameter in pekko-streams uses `org.apache.pekko.japi.function.Creator`. While it is true that `java.util.function.Supplier` is also used in pekko-streams, its only for allocate methods where you create an initial data structure (i.e. https://github.com/apache/incubator-pekko/blob/cf704782010db237bd07f197e0942c9b02ffe23c/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala#L4360) and for a data structure that makes sense (i.e. creating a `Map` doesn't throw any exceptions).

However as far as I can tell, since we are dealing with `mapWithResource` you can easily supply a function in `create` that can throw an exception even in legitimate scenarios i.e. creating a file so using `java.util.function.Supplier` is actually wrong here

> Ad a daily java developer, I really don't like the japi.function.Functions.

Why? Since it implements `@FunctionalInterface` and its a SAM (single abstract method) if you use lambda syntax its exactly the same or am I missing something here?

He-Pin: 1. for the unexpected exception, simple: 
image
2. if the current usage of `function.Function*` is totally right, why the `aggregateWithBoundary` use `java.util.function.*` and `mergeSorted` use a java `Comparator`?
3. if the we expect api consistency ,why scaladsl doesn't use these fancy handy `@Exception` methods but the one from stdlib?
4. why java's Stream api doesn't use the `X` methods which can throw？
5. why reactor-core, rxjava and smallrye-mutiny don't use these `throw Exception` methods?

I think we should keep the Api more Java **Native**, not define for the end javadsl user, everytime I see those *function.Function*, I feel wired and very hard to promote pekko/akka to java developers.

A user can live with java function, java stream, reactor-core, rxjava and mutiny, they can live with this javaapi too.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
pjfanning: +1 about a discussion - the API should stay consistent with existing APIs. I think Creator API is the one to use and we can add tools to convert other functions to Creator function in cases where the compiler does not do it automatically.

He-Pin: There is no consistency, and consitency for javadsl, we should move away from the `FI` , `JAPI`, `japi.function` as more as possible, and stay close and consistency with other java native reactive library and java stream api. By this way ,we can bring more users on board.

These APIs exist due to their historical necessity and rationality, but they may not necessarily adapt to future development. We need to look at these issues with a developmental perspective.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
He-Pin: @mdedetrich You can add a `@SneakThrow` on your `run` method.

mdedetrich: > if the current usage of `function.Function*` is totally right, why the `aggregateWithBoundary` use `java.util.function.*` and `mergeSorted` use a java `Comparator`?

I already explained why. Some of these functions create a data structure which legitimately should never throw an exception. Other functions in their `create` method could do anything, in which case you use `Creator`. In short if the supplied lambda in 100% of cases would never throw an exception (i.e. creating a collection) then using `Supplier is fine otherwise use `Creator`. The `allocate` parameter in `aggregateWithBoundary` is for creating a collection, not creating **_anything_**.

>  if the we expect api consistency ,why scaladsl doesn't use these fancy handy `@Exception` methods but the one from stdlib?

Because than that would complicate `javadsl` vs `scaladsl`. The `javadsl` are often just wrappers over `scaladsl` and if both use `org.apache.pekko.stream.javadsl.functional` it makes it easier.

> why java's Stream api doesn't use the `X` methods which can throw？

[It does](https://github.com/apache/incubator-pekko/blob/b0fdac259bd57fdd481483f3fe9a7aec6e1ff38a/stream/src/main/scala/org/apache/pekko/stream/javadsl/RestartSink.scala#L195)

>  why reactor-core, rxjava and smallrye-mutiny don't use these `throw Exception` methods?

I don't know and tbh its kind of not that relevant.

> I think we should keep the Api more Java Native, not define for the end javadsl users, everytime I see those function.Function, I feel wired and very hard to promote pekko/akka to java developers.

This is not a discussion to have in this PR. If we are going to make a big shift from `org.apache.pekko.stream.javadsl.functional` interfaces then it should be discussed in the community and not tagged along a feature PR. Any kind of radical shifts like this has to be agreed on collectively.

Until then, we should be consistent what whats currently in Pekko including adding new features. And in this case the interface to use is `Creator` because the lambda supplied to `create` can be anything.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
laglangyue: There is a disagreement. Should we initiate an email vote for reference? 
I used to mainly write Spark apps. From the perspective of Spark API design, they all have related APIs in Java, Scala, Python and R. The API function names are the same and the semantics are basically the same, but the signatures are not completely consistent. User habits are different, or we should provide some function overloading.

mdedetrich: If we are talking about `Using`, since afaict this isn't used anywhere else in Pekko we should honestly discuss this elsewhere.

Regarding the use of `org.apache.pekko.japi.function.*` vs `java.lang.*` (i.e. `Supplier` vs `Creator`) this is actually entirely consistent right now if you **_take into account the various environments** i.e.

* For `javadsl`, `org.apache.pekko.japi.function.*` is used anytime a supplied lambda could legitimately throw an exception. This is because `java.lang.*` `FunctionalInterface`'s cannot throw exceptions (`@SneakyThrow` does not count, it comes from lombok and its a hacky workaround). Otherwise if if a function should never throw an exception (i.e. creating a collection) then you use functional interfaces in `java.lang.*`.
* For `scaladsl` you do not need `org.apache.pekko.japi.function.*` since you instead use the standard Scala functions (i.e. `X => X`) because standard functions in Scala can throw exceptions. One difference here is that `scaladsl` (unlike `javadsl`) does not use `java.lang.*` functional interfaces for lambda's that shouldn't throw an exception however it can be argued that this is just due to the fact that throwing exceptions in Scala are generally frowned upon due to it not being idiomatic in Scala. If someone wants to change this and suggest using `java.lang.*` intefaces for functions that shouldn't throw in `scaladsl` as well, they can also suggest that as a community discussion separate from this PR.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
He-Pin: >  Some of these functions create a data structure which legitimately should never throw an exception. 

A user can throw everywhere, and a labmda can closingovey everything.
```java
<T extends Throwable, R> R sneakyThrow(Throwable t) throws T {
        throw (T) t;
    }
```
Can hide without a `lombok`. a new user of pekko stream will wonder what's the heck of `function.Function`. 

And you still can't explain why there is no `java.util.function.CheckedConsumer` or `java.util.function.CheckedSupplier` in java 21.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
He-Pin: That's funny, why java developers now suddenly need a `org.apache.pekko.japi.function.*` to ship things? And scala developers do not need it.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
mdedetrich: > A user can throw everywhere, and a labmda can closingovey everything.

Yes and then you end up polluting your code in Java with these kinds of wrappers. The better way to solve this problem is to just use a `@FunctionalInterface` that allows throwing of exceptions which is exactly what `org.apache.pekko.japi.function.Creator` is. Java users can just provide a function using standard Java 8 lamda synta without having to use lombok or create any other kind of workarounds, I don't see whats the big issue here.

> And you still not explain why there is no `java.util.function.CheckedConsumer` or `java.util.function.CheckedSupplier` in java 21.

It doesn't matter, maybe they will add it in Java 23 or 25, who knows?

I would strongly suggest if you want more buy in for this PR, just change it to `Creator` and move on. This is not a hill worth dying on, or to put it differently its not worth it to block the PR over this. If you want to start a discussion about using `java.lang.*` FunctionalInterface's generally within Pekko `javadsl` (which is **_NOT_** the current case) then make a separate discussion on it.

There is still time, the Pekko 1.1.x release isn't going to happen for some time.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
mdedetrich: > That's funny, why java developers now suddenly need a `org.apache.pekko.japi.function.*` to ship things? And scala developers do not need it.

Because Scala functions by default work like the `@FunctionalInterfaces` in `org.apache.pekko.japi.function.*`. In scala if you write a function like `def someMethod(lamba: () => ()) = ???` I can easily throw an exception in the supplied argument i.e. `someMethod(() => throw new Exception())`.

Its not that Scala developers don't need it, its that its not necessary because thats how Scala works by default.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
He-Pin: ` just change it to Creator and move on. ` why not `java.util.concurrent.Callable`? every java developer knows about it.

He-Pin: Yes' we can wait a little and see how other people say, people have different tastes, that's common.

mdedetrich: > `just change it to Creator and move on.` why not `java.util.concurrent.Callable`? every java developer knows about it.

We are going around in circles here, because thats not how the current Pekko codebase is written,  its inconsistent with everything else and I have already shown why. If you want to fundamentally change how we treat functions in `javadsl`, create a separate discussion and don't bundle it with a feature request.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
pjfanning: Could we consider using Pekko functions in javadsl for consistency and creating a new javadsl2 package which has standard Java functions instead? This is probably a big job but I don't think we can just change javadsl existing functions - and new functions should conform to the pre-existing style.

He-Pin: Let's wait a little longer, I will change to `creator/callable` api this week, but I have to say, Java developers want `java.util.*`.

laglangyue: Aggre with pjfanning. This PR should be more conservative. Annother PR will discuss the new Java APIs or method-overload to more confortable for java developer.

mdedetrich: > but I have to say, Java developers want `java.util.*`.

That may be the case (although I have no idea what the technical reason why is) but its outside the scope of this PR. Feel free to make a github discussion on it though!

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
pjfanning: Is there a reason why we couldn't add overloaded mapWithResource methods? One version that uses the Pekko Function interfaces (Creator) and a 2nd one that uses Java Function interfaces (Callable or Supplier).

He-Pin: Maybe ,under another package

pjfanning: I added this to my local copy of javadsl.Flow.

```
  def mapWithResource[R, T](
      create: java.util.function.Supplier[R],
      f: java.util.function.BiFunction[R, Out, T],
      close: java.util.function.Function[R, Optional[T]]): javadsl.Flow[In, T, Mat] = ???

  def mapWithResource[R, T](
      create: function.Creator[R],
      f: java.util.function.BiFunction[R, Out, T],
      close: java.util.function.Function[R, Optional[T]]): javadsl.Flow[In, T, Mat] = ???
```

This compiled ok. Maybe there are other issues when implementations are added but so far, it looks feasible to use overloaded methods.

He-Pin: It will not work with lambda

pjfanning (changes requested):
this PR really needs some javadsl tests

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
mdedetrich: Overloading like this will not work. I also realised that `close` needs to use a `org.apache.pekko.stream.javadsl.functional.Procedure` just likes its done [here](https://github.com/apache/incubator-pekko/blob/6872f9c8b78ae2fababb97cb4600d67d77b4bbd2/stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala#L854) in other parts of the code

mdedetrich (changes requested):
As has been discussed https://github.com/apache/incubator-pekko/pull/931#discussion_r1451590755

- The `create` parameter needs to be changed to a `org.apache.pekko.stream.javadsl.functional.Creator` since its possible that a supplied lambda can throw an exception (i.e. `IOException` when creating a file, see [here](https://github.com/apache/incubator-pekko/blob/43bab55616f40c0cf5ece07d8cd07110a3027c7d/stream/src/main/scala/org/apache/pekko/stream/javadsl/Sink.scala#L514) for an existing example)
- The `close` parameter needs to be changed to a `org.apache.pekko.stream.javadsl.functional.Procedure` since its possible that a supplied lambda can throw an exception (i.e. think of calling `close()` on a file handle which can throw an `IOException`). See [here](https://github.com/apache/incubator-pekko/blob/6872f9c8b78ae2fababb97cb4600d67d77b4bbd2/stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala#L854) for an existing example.

Please lets not get into yet another tangent about whether to use `java.util.*` `@FunctionalInterface` functions (even as overloaded parameters). New functionality has to follow the existing style/design of Pekko unless there is a strong exceptional reason why and in this case there isn't

He-Pin:
I've made the changes. Let's leave it at that for now. We can discuss in a separate post how future versions can provide a more Java-friendly API.

mdedetrich (approved):
lgtm

pjfanning (approved):
lgtm

nvollmar (approved):
lgtm

PR #933, opened 2024-01-13
Sort stream imports for pekko-stream modules

He-Pin:
Motivation:
When I do code modification in pekko-stream related module, after a `sortImports` I have to revert all unrelated changes, which is very annoying.

Result:
Sort the imports in `pekko-stream` and `pekko-stream-tests`.

The `sort-imports` will cause problem when encounter `#xxx` and does not obey the `#scala-fix:off/on` too.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -46,7 +46,10 @@ SortImports.blocks = [
   "scala.",
   "*",
   "com.sun."
-  "org.apache.pekko."
+  "re:^(org\\.apache\\.pekko|pekko\\.)\\w*"
   "org.reactivestreams."
   "io.netty."
+  "org.scalatest."
+  "org.slf4j."
+  "com.typesafe."
 ]
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowOnErrorCompleteSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowOnErrorCompleteSpec.scala
@@ -17,11 +17,12 @@
 
 package org.apache.pekko.stream.scaladsl
 
-import org.apache.pekko.stream.testkit.StreamSpec
-import org.apache.pekko.stream.testkit.scaladsl.TestSink
-
 import scala.concurrent.TimeoutException
 import scala.util.control.NoStackTrace
+
+import org.apache.pekko
+import pekko.stream.testkit.StreamSpec
+import pekko.stream.testkit.scaladsl.TestSink
 
 class FlowOnErrorCompleteSpec extends StreamSpec {
   val ex = new RuntimeException("ex") with NoStackTrace
--- a/stream-testkit/src/test/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreterSpecKit.scala
+++ b/stream-testkit/src/test/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreterSpecKit.scala
@@ -13,11 +13,10 @@
 
 package org.apache.pekko.stream.impl.fusing
 
+import scala.annotation.nowarn
 import scala.collection.{ Map => SMap }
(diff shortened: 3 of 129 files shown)


.scalafix.conf
```diff
@@ -46,7 +46,10 @@ SortImports.blocks = [
   "scala.",
   "*",
   "com.sun."
-  "org.apache.pekko."
+  "re:^(org\\.apache\\.pekko|pekko\\.)\\w*"
   "org.reactivestreams."
   "io.netty."
+  "org.scalatest."
+  "org.slf4j."
+  "com.typesafe."
```
He-Pin, line 54: I updated the conf.

mdedetrich:
Are you sorting this using Intellij?

He-Pin:
@mdedetrich I'm using the `sbt sortImports`

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowOnErrorCompleteSpec.scala
```diff
@@ -17,12 +17,12 @@
 
 package org.apache.pekko.stream.scaladsl
 
-import org.apache.pekko.stream.testkit.StreamSpec
-import org.apache.pekko.stream.testkit.scaladsl.TestSink
-
 import scala.concurrent.TimeoutException
 import scala.util.control.NoStackTrace
 
+import org.apache.pekko.stream.testkit.StreamSpec
```
mdedetrich, line 23: Can we change this to

```
import org.apache.pekko
import pekko.stream.testkit.StreamSpec
import pekko.stream.testkit.scaladsl.TestSink
```

And also re-run `sortImports`

mdedetrich:
> @mdedetrich I'm using the `sbt sortImports`

Nice work, this is looking good especially considering its not messing with the `import org.apache.pekko` style and we can add a check to make sure imports are sorted to the CI later. I just added one comment.

mdedetrich:
Not critical but it also would be good to squash the `chore: sort imports for pekko-stream-tests` and `chore: sort imports for pekko-stream` into a single commit.

laglangyue:
When merging to the main branch, all commits will be converted to a single commit, right？
I often submit multiple times, I just want to make sure if I made a mistake. 

> Not critical but it also would be good to squash the `chore: sort imports for pekko-stream-tests` and `chore: sort imports for pekko-stream` into a single commit.

mdedetrich:
> When merging to the main branch, all commits will be converted to a single commit, right？
I often submit multiple times, I just want to make sure if I made a mistake.

So this is the whole "rebase and merge" vs "squash and merge" debate. Some people have different preferences, but this PR specifically is an exception because since its adding a format commit we want to add it `.git-blame-ignore-revs` so it doesn't turn up in `git blame` however we want to make sure that **_ONLY_** the formatting changes are added.

So ideally for this PR you would have one commit that does all of the necessary changes i.e. https://github.com/apache/incubator-pekko/pull/933/commits/84fb5f8142c24517814eb29bd25411ad41ecf864 and then another commit which applies `sortImports` and its this commit which gets added to https://github.com/apache/incubator-pekko/blob/main/.git-blame-ignore-revs

He-Pin:
@laglangyue I will do a rebase later, it was in this style to make review easier.

He-Pin:
@mdedetrich @pjfanning I would like to have this merged first:)

.scalafix.conf
mdedetrich, line 1: @He-Pin I think there was a confusion. This PR should have 2 commits, the changes to this file (i.e. `.scalafix.conf`) should be its own commit and the rest of the changes that are a result of applying `sbt sortImports` should be another commit.

He-Pin: Fixed

mdedetrich (approved):
lgtm

PR #935, opened 2024-01-13
feat: Add support for `for comprehensions`.

He-Pin:
Motivation:
refs: https://github.com/apache/incubator-pekko/discussions/654

Note:
1. These methods are only added for scaladsl 
2. `foreach`  method just work as fs2, where the `Stream` needs compile aka pekko's `run`. but different in ZIO, where `foreach` === `runForeach`.

Result:
`for comprehensions` works as fs2 for both `Source` and `Flow` in scaladsl

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/ForComprehensionsCompileSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/ForComprehensionsCompileSpec.scala
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.scaladsl
+
+import org.apache.pekko
+import pekko.Done
+import pekko.japi.Util
+import pekko.stream.testkit.StreamSpec
+import pekko.stream.testkit.scaladsl.TestSink
+
+import java.util.concurrent.CopyOnWriteArrayList
+import scala.concurrent.Await
+import scala.concurrent.duration.DurationInt
+
+class ForComprehensionsCompileSpec extends StreamSpec {
+  "A Source" must {
+    "be able to be used in a for comprehension which yield" in {
+      val source = Source(1 to 5)
+      val evenSource = for {
+        i <- source if i % 2 == 0
+      } yield i.toString
+      evenSource.runWith(TestSink[String]())
+        .request(5)
+        .expectNextN(List("2", "4"))
+        .expectComplete()
+    }
+
+    "be able to be used in a for comprehension which flatMap" in {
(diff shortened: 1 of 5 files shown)


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
```diff
@@ -2521,6 +2526,11 @@ trait FlowOps[+Out, +Mat] {
    */
   def flatMapConcat[T, M](f: Out => Graph[SourceShape[T], M]): Repr[T] = map(f).via(new FlattenMerge[T, M](1))
 
+  /**
```
pjfanning, line 2529: Can we add `@since` to new methods. This is required on all new methods.

Can we also annotate the new methods as experimental? It gives us some fallback if we find that users using for comprehensions are having trouble - and that we need to make changes to help them.

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/ForComprehensionsCompileSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+        .request(5)
+        .expectNextN(List("2", "4"))
+        .expectComplete()
+    }
+
+    "be able to be used in a for comprehension which yield a runnable graph" in {
+      val source = Source(1 to 5)
+      val list = new CopyOnWriteArrayList[String]()
+      val future = (for (i <- source if i % 2 == 0) {
+        println(i)
```
pjfanning, line 60: can you remove `println`s?

pjfanning (changes requested):
I would still like the new methods marked as experimental. I just think there are chances that we ay need to tweak this support if users try out `for` comprehensions and find some oddities.

He-Pin:
@pjfanning Do you mean add `@ApiMayChange`?

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/ForComprehensionsCompileSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+
+    "be able to be used in a for comprehension which yield a sink" in {
+      val source = Source(1 to 5)
+      val list = new CopyOnWriteArrayList[String]()
+      val sink = for (i <- Flow[Int] if i % 2 == 0) {
+        println(i)
+        list.add(i.toString)
+      }
+      val future = source.runWith(sink)
+      Await.result(future, 3.seconds) shouldBe Done
```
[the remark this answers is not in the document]
He-Pin: It was `3 seconds` in other tests, and I did not see any fails about those.

pjfanning:
> @pjfanning Do you mean add `@ApiMayChange`?

Maybe not that one - the API signatures won't change but we may need to tweak the internals of the functions. 

Even we add something like this to the Scaladoc for the new functions.

```
Support for `for` comprehensions is experimental and it is possible that we might need to change the internal implementation to better support it.
```

He-Pin:
I think the only in question is `foreach`, where `ZIO` alias to `runForEach` but fs2 alias to `foreach` and then needs a `Stream#compile` to run to a `F[_]`, as Pekko stream always needs an *explicitly* `run`, so I think we can follow what fs2 does, and that's what @gmethvin once suggested too.

and for the `withFilter` and `flatmMap` , I think the currently shape should be fine, but I will update the scaladoc.

pjfanning (approved):
lgtm - I would like to see some other reviews though

laglangyue:
LGTM for code.
This is an Amazing feature, we shoudl update the document also,maybe in `docs/src/main/paradox/stream/operators/Flow/ `.

He-Pin:
> LGTM for code.
> This is an Amazing feature, we shoudl update the document also,maybe in `docs/src/main/paradox/stream/operators/Flow/ `.
> 

I ignore that by purpose,  because I don't expect user use these methods without for expression.

jxnu-liguobin (approved):
lgtm

weiwenweiwen (approved):

He-Pin:
Let me merge this, I think this one should be ready to go, any improvement can come up after.

PR #939, opened 2024-01-14
Fix some non-standard comments

laglangyue:
## description
Fix some non-standard comments
use tool of idea 'replaceAll', and check  them in local
## why need
it will cause some method can't generate doc
please view the site `https://pekko.apache.org/japi/pekko/current/org/apache/pekko/stream/scaladsl/Source.html`
image

and some changes maybe not included in doc, I also believe it is necessary to be more standardized for comment
## test
just add `*` for comment, not need test local,it will passed by CI

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/FSM.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/FSM.scala
@@ -440,7 +440,7 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
   type Timeout = Option[FiniteDuration]
   type TransitionHandler = PartialFunction[(S, S), Unit]
 
-  /*
+  /**
    * “import” so that these are visible without an import
    */
   val Event: FSM.Event.type = FSM.Event
@@ -743,7 +743,7 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
     case x    => x.stateData
   }
 
-  /*
+  /**
    * ****************************************************************
    *                PRIVATE IMPLEMENTATION DETAILS
    * ****************************************************************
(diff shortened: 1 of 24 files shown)


actor/src/main/scala/org/apache/pekko/actor/FSM.scala
```diff
@@ -743,29 +743,29 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
     case x    => x.stateData
   }
 
-  /*
+  /**
```
pjfanning, line 746: please remove this change - this comment is not a method or field comment

actor/src/main/scala/org/apache/pekko/actor/FSM.scala
```diff
@@ -743,29 +743,29 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
...
    * ****************************************************************
    *                PRIVATE IMPLEMENTATION DETAILS
    * ****************************************************************
    */
 
   private[pekko] def debugEvent: Boolean = false
 
-  /*
+  /**
```
pjfanning, line 754: actually, I don't see the point in changing the field comments

Can you just change function comments and only function comments?

pjfanning (changes requested):

actor/src/main/scala/org/apache/pekko/actor/FSM.scala
laglangyue: ok, I will check it again

pjfanning (approved):
lgtm

He-Pin:
Thanks , for new contributor from China:)

PR #947, opened 2024-01-14
feat: Add `create` method to PFBuilder.

He-Pin:
Motivation:

`new PFBuilder<>()` is not that fluent.

Result:
Add a `create` method on `PFBuilder`, alias with others.

### DIFF
--- a/actor/src/main/java/org/apache/pekko/japi/pf/PFBuilder.java
+++ b/actor/src/main/java/org/apache/pekko/japi/pf/PFBuilder.java
@@ -21,6 +21,15 @@ package org.apache.pekko.japi.pf;
  */
 public final class PFBuilder<I, R> extends AbstractPFBuilder<I, R> {
 
+  /**
+   * Create a new {@link PFBuilder}.
+   *
+   * @since 1.1.0
+   */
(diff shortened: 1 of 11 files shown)


actor/src/main/java/org/apache/pekko/japi/pf/PFBuilder.java
```diff
@@ -21,6 +21,11 @@
  */
 public final class PFBuilder<I, R> extends AbstractPFBuilder<I, R> {
 
+  /** Create a new {@link PFBuilder}. */
+  public static <I, R> PFBuilder<I, R> create() {
+    return new PFBuilder<>();
+  }
```
He-Pin, line 27: Just a small helper method for javadsl.

laglangyue:
lgtm

actor/src/main/java/org/apache/pekko/japi/pf/PFBuilder.java
nvollmar: Maybe add `@since 1.1.0`

nvollmar (approved):
lgtm

PR #949, opened 2024-01-14
chore: Deprecate JAPI.

He-Pin:
Motivation:
Depcate the `JAPI`.java

refs:
https://github.com/apache/incubator-pekko/issues/951

### DIFF
--- a/actor/src/main/java/org/apache/pekko/japi/JAPI.java
+++ b/actor/src/main/java/org/apache/pekko/japi/JAPI.java
@@ -15,8 +15,16 @@ package org.apache.pekko.japi;
 
 import scala.collection.Seq;
 
+@Deprecated
 public class JAPI {
 
+  /**
+   * Create a Scala seq from a Java array.
+   *
+   * @deprecated since 1.1.0. Use <code>
+   *     org.apache.pekko.japi.Util.immutableSeq(java.util.Arrays.asList(...));</code> instead.
+   */
+  @Deprecated
   @SafeVarargs
   public static <T> Seq<T> seq(T... ts) {
     return Util.immutableSeq(ts);


actor/src/main/scala/org/apache/pekko/japi/JavaAPI.scala
```diff
@@ -255,13 +254,16 @@ object Util {
...
+  def immutableSeq(arr: Array[Class[_]]): immutable.Seq[Class[_]] =
+    if ((arr ne null) && arr.length > 0) arr.toIndexedSeq else Nil
 
   /**
    * Turns an array into an immutable Scala sequence (by copying it).
    */
-  def immutableSeq[T](arr: Array[T]): immutable.Seq[T] =
-    if ((arr ne null) && arr.length > 0) arr.toIndexedSeq else Nil
+  @SafeVarargs
+  @varargs
```
pjfanning, line 264: this is not a deprecation - please do not add a big change like this into a PR with a name like `chore: Deprecate JAPI`

pjfanning (changes requested):

actor/src/main/java/org/apache/pekko/japi/JAPI.java
```diff
@@ -15,8 +15,11 @@
 
 import scala.collection.Seq;
 
+@Deprecated
```
pjfanning, line 18: could you you use a format like this?

```
    /**
     * @deprecated This method is no longer acceptable to compute time between versions. Use {@link Utils#calculatePeriod(Machine)} instead.
     */
    @Deprecated(since = "1.1.0")
```

The text is grabbed from another example but the idea is that we need to specify what the recommended alternative is.

pjfanning (changes requested):

actor/src/main/java/org/apache/pekko/japi/JAPI.java
He-Pin: Since will require java 9，but I can update it

actor/src/main/java/org/apache/pekko/japi/JAPI.java
```diff
@@ -15,8 +15,14 @@
...
 public class JAPI {
 
+  /**
+   * Use <code>Util.immutableSeq(java.util.Arrays.asList(...))</code> instead. Deprecated since
```
pjfanning, line 22:
```suggestion
   * @deprecated Use <code>Util.immutableSeq(java.util.Arrays.asList(...))</code> instead. Deprecated since
```

pjfanning: remove the `@Deprecated` annotation if the compiler won't let you use both

What is `Util` - can you fully reference it by including the package name?

actor/src/main/java/org/apache/pekko/japi/JAPI.java
```diff
@@ -15,8 +15,14 @@
 
 import scala.collection.Seq;
 
+@Deprecated
 public class JAPI {
 
+  /**
+   * @deprecated Use <code>Util.immutableSeq(java.util.Arrays.asList(...))</code> instead. Deprecated since
```
pjfanning, line 22: I still don't know where this is Util class is. It could mean anything. It is too generic a name not to be fully qualified.

pjfanning (changes requested):

docs/src/test/java-jdk9-only/jdocs/stream/operators/source/AsSubscriber.java
```diff
@@ -42,13 +39,13 @@ static <T> org.apache.pekko.stream.javadsl.Source<T, Subscriber<T>> asSubscriber
 
     static class Row {
         public String getField(String fieldName) {
-            throw new NotImplementedException();
+            throw new UnsupportedOperationException("Not implemented in sample");
```
nvollmar, line 42: Intentionally in this PR?

He-Pin: no, I will need a rebase.

nvollmar (approved):
lgtm

pjfanning (approved):
lgtm

PR #961, opened 2024-01-15
fix up javadoc on Creator trait

pjfanning:
There is no class called java.util.function.Creator - I think java.util.function.Supplier is the closest equivalent to the Pekko Creator interface.


He-Pin (approved):
lgtm

PR #962, opened 2024-01-15
feat: Add expectNextN to StreamTestKit for javadsl.

He-Pin:
Motivation:
Add an `expectN` method to StreamTestKit which supports java List

### DIFF
--- a/stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
+++ b/stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
@@ -23,15 +23,16 @@ import scala.concurrent.duration._
 import scala.reflect.ClassTag
 
 import org.apache.pekko
-import pekko.actor.ClassicActorSystemProvider
-import org.reactivestreams.{ Publisher, Subscriber, Subscription }
 import pekko.actor.{ ActorRef, ActorSystem, DeadLetterSuppression, NoSerializationVerificationNeeded }
+import pekko.actor.ClassicActorSystemProvider
 import pekko.stream._
 import pekko.stream.impl._
 import pekko.testkit.{ TestActor, TestProbe }
 import pekko.testkit.TestActor.AutoPilot
 import pekko.util.JavaDurationConverters
 import pekko.util.ccompat._
+
+import org.reactivestreams.{ Publisher, Subscriber, Subscription }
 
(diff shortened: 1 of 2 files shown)


stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
```diff
@@ -462,6 +463,15 @@ object TestSubscriber {
       self
     }
 
+    /**
+     * Fluent DSL
+     * Expect the given elements to be signalled in order.
+     */
+    def expectNextN(elems: java.util.List[I]): Self = {
+      elems.forEach(e => probe.expectMsg(OnNext(e)))
+      self
```
He-Pin, line 472: Again I will need a `@since 1.1.0`. this is for javadsl

stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
```diff
@@ -462,6 +463,16 @@ object TestSubscriber {
       self
     }
 
+    /**
+     * Fluent DSL
+     * Expect the given elements to be signalled in order.
+     * @since 1.1.0
+     */
+    def expectNextN(elems: java.util.List[I]): Self = {
```
pjfanning, line 471: any chance this could be `java.util.Iterable`?

He-Pin: I think the Iterable not mean sequence/keep the order?

He-Pin: https://openjdk.org/jeps/431 was only added in JDK21

nvollmar: `java.util.Set` implements `Iterable` but would not keep the order. In this context a `List` would probably make most sense to guide the user.

pjfanning: fair points - stick with List

pjfanning:
is there a test that could be added here or elsewhere in the repo that uses this method?

nvollmar (approved):
lgtm

PR #963, opened 2024-01-15
chore: Add collect operator test for stream javadsl.

He-Pin:
Motivation:
The test for javadsl's `collect` is missing.

Result:
Added a test for `Source.collect` javadsl.


nvollmar (approved):
lgtm

Issue #959, opened 2024-01-15
CI should run paradox to check for documentation compile issues


mdedetrich:
See https://github.com/apache/incubator-pekko/pull/958#issuecomment-1892088786

Roiocam:
Got a good news, this issue was cause by #1039 

I have verified in locally.

```
[info] compiling 80 Scala sources and 1 Java source to /Users/roiocam/IdeaProject/incubator-pekko/bench-jmh/target/scala-2.13/classes ...
[error] (actor-typed / CompileJdk9 / compileIncremental) Compilation failed
```

He-Pin:
👍

PR #965, opened 2024-01-15
chore: Check paradox for pr validation.

He-Pin:
Motivation:
Check the paradox too.

refs: https://github.com/apache/incubator-pekko/issues/959

### DIFF
--- a/.github/workflows/generate-paradox-check.yml
+++ b/.github/workflows/generate-paradox-check.yml
@@ -1,0 +1,48 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Generate paradox check
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  generate-doc-check:
+    name: Generate paradox check
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-doc-check.yml
```diff
@@ -50,3 +50,6 @@ jobs:
 
       - name: Compile docs for all Scala versions
         run: sbt +compile:doc
+
+      - name: Compile paradox docs
+        run: sbt paradox
```
pjfanning, line 55: do we need `sbt +compile:doc` if we do `sbt paradox`?

He-Pin: Seems like it would be better to chain the commands.

pjfanning: it's not a necessarily good idea to run separate commands. You will often find that the work is done from scratch, on each call.

.github/workflows/generate-doc-check.yml
```diff
@@ -49,4 +49,4 @@ jobs:
           sudo apt-get install graphviz
 
       - name: Compile docs for all Scala versions
-        run: sbt +compile:doc
+        run: sbt ";+compile:doc; paradox"
```
mdedetrich, line 52: Does this need to be `+paradox`

pjfanning: I think we only need to build this once - with java 11. Paradox doesn't work with Java 17 (you need to add JPMS flags to the java command at least).

mdedetrich (approved):
lgtm, lets double check this before merging by creating a deliberate compilation error in one of the paradox code sources to see if it fails

He-Pin:
I tested it locally , it works
image

mdedetrich:
Perfect, feel free to merge it once you verify with the `test` commit and remove it.

Also cherry pick it to the `1.0.x` branch once merged

He-Pin:
@mdedetrich No sure why ,but even with dedicated tests, it can not detecting anything.

mdedetrich:
I'll try and look into it today

He-Pin:
The `docs/Test/compile` can't catch the compiling error either, wired.

Roiocam:
> @mdedetrich No sure why ,but even with dedicated tests, it can not detecting anything.

Pekko already has build script for those` java-jdk9-only` directory, https://github.com/apache/incubator-pekko/blob/main/project/Jdk9.scala

But you have to run test/compile on JDK11 or higher.

Roiocam:
This could be closed because the problem has been found, see #959 and related issues for more details.

He-Pin:
@Roiocam Thanks for update.

PR #971, opened 2024-01-17
chore: Switch to new new wildcard [?].

He-Pin:
Motivation:
Test and migrate code to use `?`.

known issue: https://github.com/VirtusLab/scala-cli/issues/2684

Note: this pr is for testing

background: https://github.com/apache/incubator-pekko/pull/878
```
[error] -- Error: /home/runner/work/incubator-pekko/incubator-pekko/actor/src/main/scala/org/apache/pekko/actor/TypedActor.scala:277:38 
[error] 277 |      interfaces: immutable.Seq[Class[_]])
[error]     |                                      ^
[error]     |`_` is deprecated for wildcard arguments of types: use `?` instead
[error]     |This construct can be rewritten automatically under -rewrite -source 3.4-migration.
```

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,5 +1,5 @@
 version                                  = 3.7.14
-runner.dialect                           = scala213
+runner.dialect                           = scala213source3
 project.git                              = true
 style                                    = defaultWithAlign
 docstrings.style                         = Asterisk
@@ -75,3 +75,14 @@ project.excludeFilters = [
   "scripts/authors.scala"
 ]
 project.layout         = StandardConvention
+
+# Have to turn if off for now, scala-cli fmt not working for rewrite.
+# The rewriting works when do rewriting with `sbt scalafmtAll`.
+//rewrite.scala3.convertToNewSyntax = true
+//runner {
+//  dialectOverride {
+//    allowAsForImportRename = false
+//    allowStarWildcardImport = false
+//    allowPostfixStarVarargSplices = false
+//  }
+//}
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/Effect.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/Effect.scala
@@ -45,7 +45,7 @@ object Effect {
       with Serializable {
 
     override def equals(other: Any) = other match {
-      case o: Spawned[_] =>
+      case o: Spawned[?] =>
         this.behavior == o.behavior &&
         this.childName == o.childName &&
         this.props == o.props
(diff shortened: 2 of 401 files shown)


.scalafmt.conf
```diff
@@ -1,5 +1,5 @@
-version                                  = 3.7.14
-runner.dialect                           = scala213
+version                                  = 3.7.17
```
mdedetrich, line 1: Can we upgrade the version in a separate PR?

.scalafmt.conf
```diff
@@ -75,3 +75,19 @@ project.excludeFilters = [
...
+}
+
+//rewrite.scala3.convertToNewSyntax = false
+//runner {
+//  dialectOverride {
+//    allowAsForImportRename = false
+//    allowStarWildcardImport = false
+//    allowPostfixStarVarargSplices = false
+//  }
+//}
```
mdedetrich, line 93: nit: newline

He-Pin: How do you think we migrate to the `?`.

mdedetrich: I meant that the file is missing a new line at the end, thats all

He-Pin:
Will need another ignore blame, if this can get merged.

mdedetrich:
So I am a bit on the fence of this, mainly because its such a massive rewrite that its going to cause quite a bit of chaos especially with the amount of open PR's we have. I would be more amicable to it if we only rewrote scala-3 sources rather than the entire project.

Another disadvantage is that it will make backports to `1.0.x` slightly more complicated (I guess we can also apply this migration to the 1.0.x branch as its only syntax that doesn't have any other impact).

I am not against it, but there might be better timing to merge such a PR (i.e. when there are very few open PR's). Willing to hear what others think.

He-Pin:
If I compile it with Scala 3.4, there will be many warning, I want our code prepare for the next LTS early, at least seems not that harm, and in Java we use `<?>` everyday.

mdedetrich:
> If I compile it with Scala 3.4, there will be many warning, I want our code prepare for the next LTS early, at least seems not that harm, and in Java we use `<?>` everyday.

Thanks for providing the context, then I would initially suggest lets wait until Pekko 1.1.0 is out and some more PR's are merged, Scala 3.5.x LTS is still a long way off.

He-Pin:
https://github.com/lampepfl/dotty/issues/19479

@mdedetrich another case why I  want pekko code to be more  scala 3.4 friendly

mdedetrich:
> @mdedetrich another case why I  want pekko code to be more  scala 3.4 friendly 

This appears to be entirely unrelated and type inferencer/compiler bug, not syntax.

He-Pin (commented):
Migrating from `_` to `?` wildcard syntax for Scala 3.4+ compatibility.

1. **Forward compatibility**: This migration is necessary for Scala 3.4+ compatibility where `_` for wildcards is deprecated. Good proactive maintenance.

2. **Scope**: The changes are widespread across many files but are mechanical find-and-replace changes. Low risk.

3. **Known issue**: The PR mentions a scala-cli issue (#2684) -- is this still a blocker or has it been resolved upstream?

4. **Staleness**: Open since January 2024 with no recent activity. The author noted this was 'for testing'. If the migration is still needed (it almost certainly is for Scala 3.4+), this should be rebased and completed. The mechanical nature of changes makes this straightforward.

Issue #980, opened 2024-01-19
Feature request: Resume with Error logging for Stream


He-Pin:
Motivation:
I saw in https://github.com/akka/akka/issues/30267 @nvollmar added this feature for actor, but I think it would be nice to have this in pekko-stream, where user can logging the error with a restart/resume.

I found the related case from @tg44 in https://discuss.lightbend.com/t/recover-from-sink-exceptions/10554

I think this should be very useful for real production usecase.

Result:
More handy control over the exceptions

mdedetrich:
You can do this already due to this change https://github.com/apache/incubator-pekko/pull/252 which has already landed.

You can just make your own supervision decider, i.e. 

```scala
val decider: Supervision.Decider = { e =>
  logger.error("Unhandled exception in stream", e)
  Supervision.Stop
}
```

or am I missing something?

He-Pin:
Yes, I was expecting some factory method, eg `Decider.resume(level:LoggingLevel)`, let's wait for some feedbacks form @tg44 and @nvollmar

tg44:
So, my initial problem was that I have a substream which has a fileIO in it, the fileIO was only a safety measure, so if we loose the db connection we don't loose (too much) data. Then one day the fileIO failed (we couldn't recreate it bcs it opened a file which failed due to hosting provider issues), and it brought down the whole stream, and we lost hours of data when we finally noticed that the stream is down...

My idea was to lazily create a sink, if it fails I can recreate it, if it fails multiple times I can simply change it to a `Sink.ignore` if I want to.

He-Pin:
@mdedetrich I think what's @tg44 needs is :
1. a RestartSink, with is currently missing, maybe that can be done with the RestartFlow.to(IgnoreSink)
2. a fallback, but for the sink, the main sink was a `fileIO` sink ,but after some max retry, switch to a `Sink.ignore`.

```scala

Data --> Process A ----> Process FileIO Sink (main, can retry with backoff restart)
                                      |
                                      |  (only switch after the main sink die and restart up to the max limit)
                                      |
                                      |----> fallback Ignore Sink(logging) maybe.
           
````

And I think this is what he really want.

mdedetrich:
@He-Pin But he is talking about `SubFlow`, so from what he seems to be saying is that he just wants the `SubFlow` to have the ability to restart (and this means by implication if the `SubFlow` is just a sink then the Sink will restart).

Then you can just do 

```scala
[...]

PR #981, opened 2024-01-19
Support Supervision.restart for SubFlow's

mdedetrich:
Possible solution for https://github.com/apache/incubator-pekko/issues/980


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -614,14 +607,31 @@ import pekko.util.ccompat.JavaConverters._
...
+              // Start draining
+              if (!hasBeenPulled(in)) pull(in)
+            }
+          case Supervision.Restart =>
+            if (isClosed(in)) completeStage()
+            else {
+              restartState()
+              // Start draining
+              if (!hasBeenPulled(in)) pull(in)
+            }
```
He-Pin, line 625: I'm not sure about this, it was restart the `sink` not the upstream `subFlow`.

mdedetrich: @He-Pin Is it fine to resolve this conversation? I think the issue you point out was specific to https://github.com/apache/incubator-pekko/issues/980 and not about this feature in general?

mdedetrich: @He-Pin Can you look at the above comment, I want to start working on this today

He-Pin: I will review it after work, but I think you can added a `Sink.orElse`, alias with `Flow.orElse` too. Ping me after You think this ready.

And I think your current implementation is wrong.

mdedetrich: > And I think your current implementation is wrong.

If you can let me know whats wrong about it after work and then I can start writing tests/finishing it off tomorrow

He-Pin: I mean I can't see how this will solve the origin problem. when restarting the flow, the connected sink is not restarted.

mdedetrich: > I mean I can't see how this will solve the origin problem. when restarting the flow, the connected sink is not restarted.

Thats actually the point I am making which is that this PR's is no longer about solving the original problem (i.e. https://github.com/apache/incubator-pekko/issues/980) but is now just generally about supporting `Supervision.restart` for `SubFlow` in way that makes sense. https://github.com/apache/incubator-pekko/issues/980 is going to be solved another way that makes more sense (which we already discussed).

So with that being said, if we had a design goal of adding support for `Supervision.restart` generally in `SubFlow`, does this current implementation make sense or am I missing something?

He-Pin: OK, Now I see your point, I will review your PR against this key result.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -614,14 +607,31 @@ import pekko.util.ccompat.JavaConverters._
 
       override def onDownstreamFinish(cause: Throwable): Unit = {
         substreamCancelled = true
-        if (isClosed(in) || propagateSubstreamCancel(cause)) {
-          cancelStage(cause)
-        } else {
-          // Start draining
-          if (!hasBeenPulled(in)) pull(in)
+        decider(cause) match {
```
He-Pin, line 610: the cause can be `NoMoreElementsNeeded | StageWasCompleted` should test that first.

mdedetrich: Isn't this being tested in a different way later down with `isClosed`/`hasBeenPulled`? Also I had a look at parts of the Pekko codebase that implements `Supervision.restart` and I don't see `NoMoreElementsNeeded | StageWasCompleted` being tested in this way.

This doesn't mean its wrong, but if thats the case then testing `NoMoreElementsNeeded | StageWasCompleted` would be unique to `SubFlow` and I am not sure why?

He-Pin: NoMoreElementsNeeded | StageWasCompleted indicate the downstream cancel without exception, then that should not go through the supervision decider ? @raboof @jrudolph may have better understanding than me , ping~

mdedetrich:
@He-Pin @jxnu-liguobin @Roiocam @raboof

So I just noticed that for `SubFlow`'s in general the supervision strategy doesn't properly propagate (see comment at https://github.com/apache/incubator-pekko/blob/c44c0b7cbdab11d85176cfe062288fdcba16c56a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala#L213) which mean that this PR isn't going to work until we solve the fundamental underlying issue.

Due to this I will set the PR to blocked and make another issue on this point

pjfanning:
I'm planning to cut an RC for 1.1.0-M1. This will likely not be in it - but can be included in 1.1.0-M2.

mdedetrich:
> I'm planning to cut an RC for 1.1.0-M1. This will likely not be in it - but can be included in 1.1.0-M2.

There is an actual bug with the functionality introduced in https://github.com/apache/incubator-pekko/pull/252, i.e. `Supervision.resumeStrategy` doesn't actually resume when an exception is thrown.

This needs to be solved for 1.1.0, ideally in `-M1` but I don't know when you wanted to make the actual release (I guess after incubation is fully passed?)

mdedetrich:
Bug made at https://github.com/apache/incubator-pekko/issues/1205

pjfanning:
is there a chance that we could revert #252 and try a full change for 1.1.0-M2?

mdedetrich:
> is there a chance that we could revert #252 and try a full change for 1.1.0-M2?

We are actually using this feature in production so I would try and avoid that. If fixing the bug ends up taking too long I would propose we fix it in 1.1.0-M2 since no one right now is actually relying on `Supervision.resumeStrategy` catching and resuming exceptions

mdedetrich:
> I'm only to delaying the 1.1.0-M1-RC1 until next week if this a reasonable chance that this can be addressed soon.

Agreed

mdedetrich:
So I had a look at this after it was unblocked and it does seem to be harder than normal, this is because a new `SubstreamHandler` is created from the original input port which makes things complicated if we have already handed over to the new `SubStreamHandler` because we have crossed the boundary of the split decision. I think to solve this properly we need to persist already retrieved elements.

Furthermore its not just `Split` we have to support but also `Group` (the other operator that creates `SubFlow`). For this reason and in the interest of also reducing the number of milestones ill drop this unless I happen to be missing something.

@pjfanning @He-Pin @Roiocam wdyt?

pjfanning:
Is this related to #1205 ?

mdedetrich:
No this is something else, its a new feature change

He-Pin (commented):
Adding supervision restart support for SubFlows is a useful feature that brings SubFlows in line with the supervision capabilities of regular stream operators.

1. **Implementation**: The changes to `StreamOfStreams.scala` are focused (31 additions, 15 deletions). The implementation needs careful review for correctness around stream lifecycle and state management during restart.

2. **Testing**: I notice this PR modifies only one source file and does not appear to include new tests for the restart behavior. Tests covering restart scenarios would be essential before merging.

3. **Staleness**: Open since January 2024 with extensive discussion but no resolution. The conversation appears to have gone through several iterations. @mdedetrich is this still something you want to pursue? If so, adding tests and rebasing would be the next steps.

PR #982, opened 2024-01-19
chore: Clean up some duplicated code.

He-Pin:
Motivation：
I found these code is a little duplicated.

Result:
Simpler code.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -193,9 +193,8 @@ import pekko.util.ccompat._
         } catch {
           case NonFatal(ex) =>
             decider(ex) match {
-              case Supervision.Stop    => failStage(ex)
-              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
-              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
+              case Supervision.Stop => failStage(ex)
+              case _                => pull(in)
             }
         }
 
@@ -267,9 +266,8 @@ private[stream] object Collect {
         } catch {
           case NonFatal(ex) =>
             decider(ex) match {
-              case Supervision.Stop    => failStage(ex)
-              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
-              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
+              case Supervision.Stop => failStage(ex)
+              case _                => pull(in)
             }
         }
 
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -267,7 +267,11 @@ private[stream] object Collect {
           case NonFatal(ex) =>
             decider(ex) match {
               case Supervision.Stop => failStage(ex)
-              case _                => pull(in)
+              case _                =>
+                // The !hasBeenPulled(in) check is not required here since it
+                // isn't possible to do an additional pull(in) due to the nature
+                // of how collect works
+                pull(in)
             }
         }
 


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -267,9 +266,8 @@ private[stream] object Collect {
         } catch {
           case NonFatal(ex) =>
             decider(ex) match {
-              case Supervision.Stop    => failStage(ex)
-              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
-              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
```
pjfanning, line 272: can you explain the change? - does `pull(in)` automatically check `!hasBeenPulled(in)`?

pjfanning (changes requested):
change needs better explanation - you should not expect reviewers to read all the other code for themselves

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: There is no need to check the `(!hasBeenPulled(in))` , because there is no additional `pull(in)`.  these method was copied from the old code, where, an exception maybe throw during the `preStart` but the upstream has been pulled. but here, as the `collect` operator never do some pulling or any other things which can throws after a `pull(in)`, so I think it would be better to just remove those duplicated check.

That's why I remove it.

pjfanning (approved):
lgtm - but it would be good if someone else reviews this too

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -267,9 +266,8 @@ private[stream] object Collect {
         } catch {
           case NonFatal(ex) =>
             decider(ex) match {
-              case Supervision.Stop    => failStage(ex)
-              case Supervision.Resume  => if (!hasBeenPulled(in)) pull(in)
```
mdedetrich, line 271: Can we add a comment here explaining why `!hasBeenPulled(in)` is not necessary.

mdedetrich (changes requested):
While I trust that the `!hasBeenPulled(in)` is not necessary it would be good to add a comment explaining why for future people reading the code.

Would also be good to add a unit test confirming this if its quick to do so.

Ill approve once comment is added

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: Why? the Map and Filter don't have this either.

I would like to add that for in `MapConcat` operator which explain why a `!hasBeenPulled(in)` test is needed.

I think the origin reason about why this `!hasBeenPulled(in)` test is because that method is a abstract class and so that exception can happen when onPush/onPull, so it have to be guard for extending, and reduce the sub class' burden. that's not the case here.

He-Pin:
@mdedetrich @pjfanning I can not push anycode now, GFW blocked my network.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
mdedetrich: > Why? the Map and Filter don't have this either.

We should add the comments there as well then. Its more that when other people that are not so familiar with Akka/Pekko read the code they get a better understanding of why these exceptions are there.

Akka's problem historically is that a lot of the contributions were done within Lightbend and so there is a lot of inhouse knowledge that wasn't visible to outside contributors.

He-Pin: I can't push any code because gfw blocked me, Maybe I can use a web ide?
How about we do that in a separate pr or you take care about it?

He-Pin: https://pekko.apache.org/docs/pekko/current/stream/stream-customize.html
It's detailly described here.

He-Pin: @mdedetrich Please unlock this, I'm not pay for everything, thanks, I just a volunteer, and this kind of information is well covered in the doc, this is code, not a book.

mdedetrich: @He-Pin I was away from laptop, ill just add a commit with comment onto your branch and then merge it.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -267,9 +266,12 @@ private[stream] object Collect {
...
-              case Supervision.Restart => if (!hasBeenPulled(in)) pull(in)
+              case Supervision.Stop => failStage(ex)
+              case _                =>
+                // The !hasBeenPulled(in) check is not required here since it
```
mdedetrich, line 271: @He-Pin Is this comment correct?

He-Pin: Is true, but where this comments come from?

mdedetrich: As a maintainer of Pekko, I have permission to add commits onto branches of other people's PR. This is my commit that I added now.

I will approve PR and squash + merge it, thanks for work!

mdedetrich (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
He-Pin: I see, hah, The chinese networking to github is very unstable, that's why I hate changing and changing, sometime I can't pulling pull too, both via ssh and https

Issue #977, opened 2024-01-18
Feature request: Add Flow#collectFirst operator


He-Pin:
Motivation:
I think this operator is common, even on X, there are server tweets about it. Both ZIO and FS2 has this method too.

I asked the question on discord, and I think this should be nice.

Result: 
Rich operator set.

PR #984, opened 2024-01-19
feat: Add `collectFirst` stream operator.

He-Pin:
Motivation:
Add the `collectFirst` operator, which would be nice to have it , otherwise you will need a `collect(...).take(...))`

refs: https://github.com/apache/incubator-pekko/issues/977

The doc is OK.
image


jxnu-liguobin (approved):
lgtm

PR #991, opened 2024-01-20
chore: Remove unused attributes in Stages.

He-Pin:
Motivation:
These fields are internal and not used in anywhere.

Result:
Clean code.


mdedetrich (approved):
lgtm but you need to add MiMa filters (which is fine since this is `InternalApi`)

He-Pin:
I think they added these because they plan to implement these:)

nvollmar (approved):
lgtm, if I remember correctly those stages were removed with Akka 2.6 or so

PR #992, opened 2024-01-20
chore: Fix incorrectly prefixed config names

He-Pin:
Motivation:
This should be pekko prefix.

Result:
correct prefix.

No need to backport to 1.0.x


pjfanning (approved):
lgtm

PR #997, opened 2024-01-20
improve performance in PEM reader

pjfanning:
* We don't need the ByteArrayInputStream, we can read from the file directly

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/artery/tcp/ssl/PemManagersProvider.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/artery/tcp/ssl/PemManagersProvider.scala
@@ -13,24 +13,16 @@
 
 package org.apache.pekko.remote.artery.tcp.ssl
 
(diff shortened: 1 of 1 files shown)


laglangyue:
LGTM，BTW，I took a look at the method above and it seems that it can also be optimized,
`org.apache.pekko.remote.artery.tcp.ssl.PemManagersProvider#loadPrivateKey`
can use `java.util.Base64.Decoder#decode(byte[]))`，not need to new `String()` and decode, ...I'm not sure if it's possible

pjfanning:
> LGTM，BTW，I took a look at the method above and it seems that it can also be optimized, `org.apache.pekko.remote.artery.tcp.ssl.PemManagersProvider#loadPrivateKey` can use `java.util.Base64.Decoder#decode(byte[]))`，not need to new `String()` and decode, ...I'm not sure if it's possible

That method uses PEMDecoder not Base64.Decoder.

samueleresca (approved):
looks good to me

PR #1005, opened 2024-01-21
Simplify and cleanup MultiJvm, MultiNode, Copyright in sbt build

mdedetrich:
Originally this PR was only meant to remote the `.configs(MultiJvm)` since its not necessary due to https://github.com/apache/incubator-pekko/blob/main/project/SbtMultiJvm.scala#L83. Since then I went into a bit of a rabbit hole due to hitting the really weird way that some parts of the sbt build have been written in Pekko.

Basically the sbt build of the core pekko project is kind of a mess. A lot of the internal sbt plugins that are written are not done properly, i.e. instead of using `object MySbtPlugin extends AutoPlugin` ++ `override def requires` (as you should be doing) currently a lot of plugins use trait composition/super overrides which is not how sbt plugins are supposed to be structured when it comes to composition/inheritance.

This PR fixes all of this, i.e. the current sbt plugins have been changed so its written the proper sbt way which makes the build easier to understand.

Due to this PR changing how `MultiJvm` interacts with `CopyrightPlugin`/`headerCheck` its recommended to verify that `headerCheckAll` works on sources contained within `multi-jvm`, i.e.

```
remote-tests/src/multi-jvm
cluster/src/multi-jvm
cluster-metrics/src/multi-jvm
cluster-typed/src/multi-jvm
cluster-tools/src/multi-jvm
cluster-sharding-typed/src/multi-jvm
distributed-data/src/multi-jvm
cluster-sharding/src/multi-jvm
```

You can test this by manually removing a header from a source file and then doing `remote-tests/headerCheckAll` (using `remote-tests` as an example). I did this myself but best to double check.


He-Pin (approved):
Lgtm, and will run the multi node sites locally later.

mdedetrich:
> Lgtm, and will run the multi node sites locally later.

When fixing this I came across some other problems, seems like the current sbt build was a bit brittle and the headerCheck is now failing.

I will look into this a bit later

project/CopyrightHeader.scala
```diff
@@ -133,5 +133,3 @@ trait CopyrightHeader extends AutoPlugin {
 }
 
 object CopyrightHeader extends CopyrightHeader
-
-object CopyrightHeaderInPr extends CopyrightHeader
```
mdedetrich, line 137: This should have been deleted along with the change in https://github.com/apache/incubator-pekko/commit/bcec7bae0768cf9b3c9715fce3f908437752524b#diff-068e87d37c866101b33e15bb754968e8e77fd43f90f2d3ad50344742b9604113L110-L111 , its now no longer doing anything

project/ValidatePullRequest.scala
```diff
@@ -85,7 +85,6 @@ object PekkoValidatePullRequest extends AutoPlugin {
  */
 object MultiNodeWithPrValidation extends AutoPlugin {
   import PekkoValidatePullRequest._
-  import MultiJvmPlugin.MultiJvmKeys.MultiJvm
```
mdedetrich, line 88: This import is useless

He-Pin:
I'm verify it on gitpod, will report back.

build.sbt
```diff
@@ -35,8 +35,6 @@ enablePlugins(
   UnidocRoot,
   UnidocWithPrValidation,
   NoPublish,
-  CopyrightHeader,
-  CopyrightHeaderInPr,
   ScalafixIgnoreFilePlugin,
```
He-Pin, line 38: So these two are enabled one by one now?

mdedetrich: `CopyrightHeaderInPr` is removed in this PR since it was pointless and `CopyrightHeader` is now an autoPlugin that is by default enabled.

project/CopyrightHeader.scala
```diff
@@ -11,22 +11,18 @@
...
+  override def trigger = allRequirements
 
-  override def requires: Plugins = HeaderPlugin
-
-  override def trigger: PluginTrigger = allRequirements
-
-  protected def headerMappingSettings: Seq[Def.Setting[_]] =
-    Seq(Compile, Test, MultiJvm).flatMap { config =>
+  override def projectSettings = Def.settings(
+    Seq(Compile, Test).flatMap { config =>
```
He-Pin, line 25: Will it now update the header in multi-node folder too?

mdedetrich: Yes, please see https://github.com/apache/incubator-pekko/pull/1005#issue-2092289217 specifically

> You can test this by manually removing a header from a source file and then doing remote-tests/headerCheckAll (using remote-tests as an example). I did this myself but best to double check.

He-Pin (approved):
verifyed, great work,but the cluster and multi-node header checking working .
image

He-Pin:
@mdedetrich I have verified it, I think this one should be good to go.

mdedetrich:
I just rebased but it was because of a spelling error in the commit message

Issue #979, opened 2024-01-19
Doc: We can metion the Java 21's new switch statement.


He-Pin:
Motivation:
When handle an actor's messages, in Scala ,we can use the pattern matching, but now with JDK 21, we can use the new fancy :
```java
switch(msg） {
  case XEvent event -> handle(event);
  default _ -> unhandled(msg);
```
style, and which will leverage the `typeSwitch` and maybe better performance vs the `if else`.

duobei:
When I execute ``, it raised a error:
```
[error] java.lang.RuntimeException: Error creating extended parser class: Could not determine whether class 'org.pegdown.ParserWithDirectives$$parboiled' has already been loaded
[error] 	at org.parboiled.Parboiled.createParser(Parboiled.java:58)
[error] 	at com.lightbend.paradox.markdown.Reader.<init>(Reader.scala:40)
[error] 	at com.lightbend.paradox.sbt.ParadoxPlugin$.$anonfun$baseParadoxSettings$2(ParadoxPlugin.scala:83)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:63)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:69)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:283)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error] 	at sbt.Execute.work(Execute.scala:292)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:283)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:65)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
[...]

PR #1011, opened 2024-01-21
doc: metion for the Java 21's new switch statement in document

duobei:
#979

### DIFF
--- a/docs/src/main/paradox/actors.md
+++ b/docs/src/main/paradox/actors.md
@@ -840,6 +840,41 @@ untyped version. When extending `UntypedAbstractActor` each message is received 
 
 @@snip [ActorDocTest.java](/docs/src/test/java/jdocs/actor/ActorDocTest.java) { #optimized }
 
+In addition, Java 21 introduces [powerful pattern matching for switch](https://openjdk.org/jeps/441) supporting any reference type. We can use `switch` instead of `if ... else ...` for conditional branches:
+
+```java
+static class PatternMatchedActor extends UntypedAbstractActor {
+
+    public static class Msg1 {}
+
+    public static class Msg2 {}
+
+    public static class Msg3 {}
+
+    @Override
+    public void onReceive(Object msg) throws Exception {
+      switch(msg) {
+        case Msg1 msg -> receiveMsg1((Msg1) msg);
+        case Msg2 msg -> receiveMsg2((Msg2) msg);
+        case Msg3 msg -> receiveMsg3((Msg3) msg);
+        default _ -> unhandled(msg);
+      }
+    }
+
+    private void receiveMsg1(Msg1 msg) {
+      // actual work
+    }
+
+    private void receiveMsg2(Msg2 msg) {
+      // actual work
+    }
+
+    private void receiveMsg3(Msg3 msg) {
+      // actual work
+    }
+  }
+```
+
 @@@
 
 <a id="actor-reply"></a>


docs/src/test/java/jdocs/actor/ActorDocTest.java
```diff
@@ -170,6 +170,40 @@ private void receiveMsg3(Msg3 msg) {
...
+
+    public static class Msg3 {}
+
+    @Override
+    public void onReceive(Object msg) throws Exception {
+      switch(msg) {
+        case Msg1 msg -> receiveMsg1((Msg1) msg);
+        case Msg2 msg -> receiveMsg2((Msg2) msg);
+        case Msg3 msg -> receiveMsg3((Msg3) msg);
+        default _ -> unhandled(msg);
```
He-Pin, line 189: Very cool and very clean, thank you.

He-Pin (approved):
LGTM, thanks.

docs/src/test/java/jdocs/actor/ActorDocTest.java
```diff
@@ -170,6 +170,40 @@ private void receiveMsg3(Msg3 msg) {
   }
   // #optimized
 
+  public
+  // #pattern-match-java21
+  static class PatternMatchedActor extends UntypedAbstractActor {
```
He-Pin, line 175: Seems like we need a `jdk-21` support, eg `java-jdk-21`. So seems we need to update the Jdk plugin then.

docs/src/test/java/jdocs/actor/ActorDocTest.java
He-Pin: I think you should move this fild to `java-jdk-9`, thanks.

He-Pin (commented):
Maybe we just attach the code directly in the doc, then there will be no error.

duobei:
> Maybe we just attach the code directly in the doc, then there will be no error.

I'll try later.

He-Pin (approved):
Cool, thanks

pjfanning:
I'm not sure about the positioning of this doc. The Classic Actors page is basically regarded as documented an effectively deprecated usage pattern. This change also goes against the norm in this .md file which allow Scala users to just see Scala code snippets. I may be wrong about how this displays and will check the docs again after the nightly build runs.

I think we would be better off having a dedicated Java 21 page and having links to that from the Classic and the Typed Actors pages. Maybe also useful to link to this from the Stream pages.

wdyt @mdedetrich

He-Pin:
I think this is fine. And the classical actor will not go anywhere, which is very high performance.

For the doc , we can always improve it later, once we get the doc published with jdk 21, we can even let it compile checked.

I can do the split tomorrow, but as this is his first contribution, we should be more nice and welcome.

He-Pin:
And even now, the scala user can refs to javadsl packages , that should not be a big problem, paradox support tagging, I cab improve that.

He-Pin:
@pjfanning @mdedetrich I just chatted with the origin contributor, I will take care it now, I just opened up my laptop and it's 1.49 AM

pjfanning:
@He-Pin there is no hurry to remedy this. In fact, I think we need to slow things down. I think PRs should be open for a few days before they are merged. I have sent an email to dev list with this proposal.

He-Pin:
@pjfanning I have to sleep now, I think it's fine now, new contributors may not know everything about the toolchain, I think that's ok.

PR #1016, opened 2024-01-22
Use lazy val's in sbt build

mdedetrich:
This PR updates the sbt build so it uses `lazy val`'s in the areas where its idiomatic/expected to do so with the intention that the code is both easier to understand and in some cases faster (i.e. avoiding unnecessary re-evaluation or not evaluating values if they aren't needed) specifically

* zero arg `def`'s. There is no reason to use a `def`'s here because computed values will not change in consecutive evaluations so the behaviour of `lazy val` caching the value after initial evaluation is what we want here.
* `val`'s: Any `val`'s which use the sbt dsl/syntax have been changed to `lazy val` because sbt dsl/syntax is expensive to compute for the first time so this means that we only actually compute the values if they are needed. `val`'s that only initiate constants (i.e. string constants) that don't use the sbt dsl/syntax have been untouched because evaluating those is likely as fast (if not faster) than the overhead that `lazy val` has
* For `project/Depedencies.scala` there was a lot of cases where `Def.setting` was used where it didn't need to be (as a quick refresher `Def.setting` should only be used when you need to reference an sbt `SettingKey`/`TaskKey` i.e. `scalaVersion.value`). `Def.setting` was needed in these cases previously because we had different dependency versions based on things like scala versions and the like (such as Scala 2 vs Scala 3) however this is no longer the case since in Pekko 1.1.x we have updated to the latest version's of dependencies which have the same version releases between the Scala versions that we support. In summary the `Def.settings` were accidentally left over when we update the dependencies.
  * I have also removed the redundant `// <LICENSE>` comments as this is no longer needed ([sbt-license-report](https://github.com/sbt/sbt-license-report) solves this need automatically which Pekko already uses)

### DIFF
--- a/project/ParadoxBrowse.scala
+++ b/project/ParadoxBrowse.scala
@@ -19,12 +19,12 @@ import sbt._
 object ParadoxBrowse extends AutoPlugin {
 
   object autoImport {
-    val paradoxBrowse = taskKey[Unit]("Open the docs in the default browser")
+    lazy val paradoxBrowse = taskKey[Unit]("Open the docs in the default browser")
   }
   import autoImport._
 
-  override def trigger = allRequirements
-  override def requires = ParadoxPlugin
+  override lazy val trigger = allRequirements
+  override lazy val requires = ParadoxPlugin
 
   override lazy val projectSettings = Seq(paradoxBrowse := {
     import java.awt.Desktop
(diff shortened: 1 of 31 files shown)


project/ParadoxBrowse.scala
```diff
@@ -19,12 +19,12 @@ import sbt._
 object ParadoxBrowse extends AutoPlugin {
 
   object autoImport {
-    val paradoxBrowse = taskKey[Unit]("Open the docs in the default browser")
+    lazy val paradoxBrowse = taskKey[Unit]("Open the docs in the default browser")
```
He-Pin, line 22: I think this kind should not be a `lazy val`.

He-Pin (approved):
overall lgtm, but first let' me test it locally.

I tested a simple `sbt clean`, seems the same.

He-Pin:
@mdedetrich Would you merge this soon, then we next build will be faster.

nvollmar (approved):
lgtm

PR #1020, opened 2024-01-22
Only use dependabot for alerts

mdedetrich:
We are using scala steward instead of dependabot to make PR's for dependency updates


pjfanning (approved):
lgtm

gmethvin (approved):

PR #1023, opened 2024-01-22
chore: Move java 21 code to a dedicated folder and including code sni…

He-Pin:
Motivation:
Group code in Java and Java21 tab, and as the origin is under in `@@@ div { .group-java }`, so It will not be visible to Scala user too.

image

Continue of https://github.com/apache/incubator-pekko/pull/1011


docs/src/main/java-jdk-21/docs/actors/classical/OptimizedActorWithJava21.java
```diff
@@ -0,0 +1,33 @@
...
+
+  private void receiveMsg2(Msg2 msg) {
+      // actual work
+  }
+
+  private void receiveMsg3(Msg3 msg) {
+      // actual work
+  }
+}
+// #pattern-matching
```
pjfanning, line 33: line ending should be added at EOF

He-Pin: Please kindly change it directly please, I'm behind a firewall:)

duobei:
That's great. So that's how it's done.

nvollmar (approved):
lgtm

docs/src/main/paradox/actors.md
```diff
@@ -838,42 +838,14 @@ that the JVM can have problems optimizing and the resulting code might not be as
...
-    }
+Java
+: @@snip [ActorDocTest.java](/docs/src/test/java/jdocs/actor/ActorDocTest.java) { #optimized }
 
-    private void receiveMsg3(Msg3 msg) {
-      // actual work
-    }
-  }
-```
+Java 21
```
Roiocam, line 847: Looks great, this is kind of like [Tabs](https://docusaurus.io/docs/markdown-features/tabs) feature in docusaurus.

I often use Tabs when writing release notes at work.

I think this is a good example when we add the newest JDk-related features in the future.

docs/src/main/java-jdk-21/docs/actors/classical/OptimizedActorWithJava21.java
```diff
@@ -0,0 +1,33 @@
+
+// #pattern-matching
+
+static class OptimizedActorWithJava21 extends UntypedAbstractActor {
```
Roiocam, line 4: should also add there source directory on the build script, as JDK9.scala does.

Roiocam (commented):
LGTM

PR #983, opened 2024-01-19
chore: Micro optimization for `collect*` operator.

He-Pin:
Motivation:
As `collect` and `collectWhile` are hot, I think it's worth to do this kind of micro optimization, and these code  unlikely to be changed too.
Same trick can be find https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/IterableOnce.scala#L1178
Result:
`IF_ACMPEQ` will be a little faster, and reduce an `INVOKEVIRTUAL java/lang/Object.equals`

> Jmh/run -i 10 -wi 10 -f1 -t1 .*CollectBenchmark.*

first run:
```
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                          Mode  Cnt         Score        Error  Units
[info] CollectBenchmark.benchNewCollect  thrpt   10  14550829.734 ± 110526.538  ops/s
[info] CollectBenchmark.benchOldCollect  thrpt   10  14268840.296 ± 332840.984  ops/s
[success] Total time: 783 s (13:03), completed 2024年1月20日 上午2:04:44
sbt:pekko-bench-jmh> Jmh/run -i 10 -wi 10 -f1 -t1 .*CollectBenchmark.*

info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                          Mode  Cnt         Score        Error  Units
[info] CollectBenchmark.benchNewCollect  thrpt   10  15099034.601 ± 331220.483  ops/s
[info] CollectBenchmark.benchOldCollect  thrpt   10  14517524.914 ± 281804.895  ops/s
[success] Total time: 469 s (07:49), completed 2024年1月20日 上午3:25:36

[...]

PR #1026, opened 2024-01-23
chore: Make use of pattern matching on singleton type instead.

He-Pin:
More info: @som-snytt point out in https://scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html#type-patterns see singleton.

Thanks @som-snytt for this.

continue : https://github.com/apache/incubator-pekko/pull/983

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectWhile.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectWhile.scala
@@ -17,6 +17,7 @@
 
 package org.apache.pekko.stream.impl.fusing
 
+import scala.annotation.nowarn
 import scala.util.control.NonFatal
 
 import org.apache.pekko
@@ -43,6 +44,7 @@ private[pekko] final class CollectWhile[In, Out](pf: PartialFunction[In, Out]) e
       private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
       import Collect.NotApplied
 
+      @nowarn("msg=Any")
       override final def onPush(): Unit =
         try {
           // 1. `applyOrElse` is faster than (`pf.isDefinedAt` and then `pf.apply`)
@@ -50,11 +52,9 @@ private[pekko] final class CollectWhile[In, Out](pf: PartialFunction[In, Out]) e
           //   eg: just a simple `IF_ACMPNE`, and you can find the same trick in `Collect` operator.
           //   If you interest, you can check the associated PR for this change and the
           //   current implementation of `scala.collection.IterableOnceOps.collectFirst`.
-          val result = pf.applyOrElse(grab(in), NotApplied)
-          if (result.asInstanceOf[AnyRef] eq NotApplied) {
-            completeStage()
-          } else {
-            push(out, result.asInstanceOf[Out])
+          pf.applyOrElse(grab(in), NotApplied) match {
+            case _: NotApplied.type   => completeStage()
+            case elem: Out @unchecked => push(out, elem)
           }
         } catch {
           case NonFatal(ex) =>
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -14,8 +14,8 @@
 package org.apache.pekko.stream.impl.fusing
 
 import java.util.concurrent.TimeUnit.NANOSECONDS
-import scala.annotation.nowarn
-import scala.annotation.tailrec
+
+import scala.annotation.{ nowarn, tailrec }
 import scala.collection.immutable
 import scala.collection.immutable.VectorBuilder
 import scala.concurrent.Future
@@ -236,7 +236,9 @@ private[stream] object Collect {
   // Cached function that can be used with PartialFunction.applyOrElse to ensure that A) the guard is only applied once,
   // and the caller can check the returned value with Collect.notApplied to query whether the PF was applied or not.
   // Prior art: https://github.com/scala/scala/blob/v2.11.4/src/library/scala/collection/immutable/List.scala#L458
-  final val NotApplied: Any => Any = _ => Collect.NotApplied
+  object NotApplied extends (Any => Any) {
+    final override def apply(v1: Any): Any = this
+  }
 }
 
 /**
@@ -255,18 +257,17 @@ private[stream] object Collect {
       private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
       import Collect.NotApplied
 
+      @nowarn("msg=Any")
       override def onPush(): Unit =
         try {
-          val result = pf.applyOrElse(grab(in), NotApplied)
           // 1. `applyOrElse` is faster than (`pf.isDefinedAt` and then `pf.apply`)
           // 2. using reference comparing here instead of pattern matching can generate less and quicker bytecode,
           //   eg: just a simple `IF_ACMPNE`, and you can find the same trick in `CollectWhile` operator.
           //   If you interest, you can check the associated PR for this change and the
           //   current implementation of `scala.collection.IterableOnceOps.collectFirst`.
-          if (result.asInstanceOf[AnyRef] eq Collect.NotApplied) {
-            pull(in)
-          } else {
-            push(out, result.asInstanceOf[Out])
+          pf.applyOrElse(grab(in), NotApplied) match {
+            case _: NotApplied.type   => pull(in)
+            case elem: Out @unchecked => push(out, elem)
           }
         } catch {
           case NonFatal(ex) =>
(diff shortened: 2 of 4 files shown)


mdedetrich:
Build is failing

He-Pin:
@mdedetrich should we remove `infer-any`

som-snytt:
I just tried it. I forgot about https://github.com/scala/bug/issues/12569

which I think it's just necessary to `@nowarn`.  Alternatively, `-Wconf` something something.

That is some kind of pattern matching bug, but it's "just a warning".

som-snytt:
Aliasing it and hiding the type seems to work without nowarn.
```
val fallback: AnyRef = NotApplied.asInstanceOf[AnyRef]
          pf.applyOrElse(grab(in), NotApplied) match {
            case _: fallback.type => completeStage()
```

mdedetrich:
Hmm, with all of the workarounds not sure if this is that much better than whats currently there

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -348,10 +350,11 @@ private[stream] object Collect {
...
       import Collect.NotApplied
 
+      @nowarn("msg=Any")
       override def onUpstreamFailure(ex: Throwable): Unit = f.applyOrElse(ex, NotApplied) match {
-        case NotApplied   => super.onUpstreamFailure(ex)
-        case t: Throwable => super.onUpstreamFailure(t)
-        case _            => throw new IllegalStateException() // won't happen, compiler exhaustiveness check pleaser
+        case _: NotApplied.type => super.onUpstreamFailure(ex)
+        case t: Throwable       => super.onUpstreamFailure(t)
+        case _                  => throw new IllegalStateException() // won't happen, compiler exhaustiveness check pleaser
```
He-Pin, line 357: It's better which cover the error case. @mdedetrich

He-Pin:
> val fallback: AnyRef = NotApplied.asInstanceOf[AnyRef]
>           pf.applyOrElse(grab(in), NotApplied) match {
>             case _: fallback.type => completeStage()

@som-snytt what about 
```
          pf.applyOrElse(grab(in), NotApplied: AnyRef) match {
            case _: NotApplied.type => completeStage()
```

He-Pin:
I just run the bench
image

@mdedetrich @som-snytt cc, performance is the same as `eq`.

BTW, the sbt-jmh issue on Windows been verified fixed, thanks.

He-Pin:
```
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                          Mode  Cnt         Score        Error  Units
[info] CollectBenchmark.benchNewCollect  thrpt   10  15246702.956 ± 127702.345  ops/s
[info] CollectBenchmark.benchOldCollect  thrpt   10  14828734.816 ± 112590.257  ops/s
[success] Total time: 773 s (12:53), completed 2024年1月23日 下午10:41:06
```

He-Pin:
@mdedetrich @pjfanning ping, what's your option on this, I tested seems the performance is the same as my previous commit.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectWhile.scala
```diff
@@ -43,18 +44,17 @@ private[pekko] final class CollectWhile[In, Out](pf: PartialFunction[In, Out]) e
...
       import Collect.NotApplied
 
+      @nowarn("msg=Any")
       override final def onPush(): Unit =
         try {
           // 1. `applyOrElse` is faster than (`pf.isDefinedAt` and then `pf.apply`)
           // 2. using reference comparing here instead of pattern matching can generate less and quicker bytecode,
           //   eg: just a simple `IF_ACMPNE`, and you can find the same trick in `Collect` operator.
           //   If you interest, you can check the associated PR for this change and the
           //   current implementation of `scala.collection.IterableOnceOps.collectFirst`.
```
Roiocam, line 54: Will these comments be outdated after this PR?

Roiocam (commented):
```
[info] Benchmark                            Mode  Cnt        Score        Error  Units
[info] CollectBenchmark.benchNewCollect    thrpt    3  9147229.404 ± 255711.579  ops/s
[info] CollectBenchmark.benchOldCollect    thrpt    3  8889505.583 ± 985315.585  ops/s
[info] CollectBenchmark.oldIFACMPNE        thrpt    3  8899066.666 ± 271187.167  ops/s
```
Just a quick bench, looks great.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectWhile.scala
He-Pin: I think the it's still apply, maybe drop the :
>//   If you interest, you can check the associated PR for this change and the current implementation of `scala.collection.IterableOnceOps.collectFirst`.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectWhile.scala
```diff
@@ -43,18 +44,17 @@ private[pekko] final class CollectWhile[In, Out](pf: PartialFunction[In, Out]) e
...
-          val result = pf.applyOrElse(grab(in), NotApplied)
-          if (result.asInstanceOf[AnyRef] eq NotApplied) {
-            completeStage()
-          } else {
-            push(out, result.asInstanceOf[Out])
+          pf.applyOrElse(grab(in), NotApplied) match {
+            case _: NotApplied.type   => completeStage()
```
jrudolph, line 56: Why not match on the singleton instance instead? In any case, it seems uncommon to match on the singleton type.

It's not even easy to say if matching on the type or the reference is supposed to be faster in general, and maybe the Scala compiler should make the right choice. (On one hand, reference comparison is as fast as it gets, on the other hand, getting the singleton instance reference is at least a static field lookup which might not be faster than comparing types, would depend on the JIT compiler).

mdedetrich: > Why not match on the singleton instance instead? In any case, it seems uncommon to match on the singleton type.

I am under the impression it only works if you match on the type and not on the instance, @som-snytt can you confirm?

By only works I mean producing the optimal bytecode which was the point of the change in the first place.

He-Pin:
```scala
  def collect[A](a: A, pf: PartialFunction[A, A]): A = {
    pf.applyOrElse(a, NotApplied) match {
      case NotApplied => a
      case _          => a
    }
  }

  def collect2[A](a: A, pf: PartialFunction[A, A]): A = {
    pf.applyOrElse(a, NotApplied) match {
      case _: NotApplied.type => a
      case _                  => a
    }
  }
```

ByteCode:
image

He-Pin: @mdedetrich @jrudolph I think just as @som-snytt said , the current version should generate less bytecode.

He-Pin:
final results:
Jmh/run -i 10 -wi 10 -f1 -t1 .CollectBenchmark.
```
[info] Benchmark                                      Mode  Cnt                  Score                 Error  Units
[info] CollectBenchmark.benchCollectOnSingleton      thrpt   10  17608004768421276.000 ± 192500018073910.500  ops/s
[info] CollectBenchmark.benchCollectOnSingletonType  thrpt   10  17912827936113308.000 ± 239976332956371.340  ops/s
[info] CollectBenchmark.benchNewCollect              thrpt   10           15283510.041 ±          460993.700  ops/s
[info] CollectBenchmark.benchOldCollect              thrpt   10           14845265.443 ±          178628.150  ops/s
```

He-Pin:
After carefully bench and diff the bytecode, I think this one should be ready to go.

mdedetrich (approved):
lgtm from my side, would also wait for another review from someone else.

pjfanning (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectWhile.scala
som-snytt: The clunky pattern is just to avoid a clunky `eq`.

You might use a local val to avoid "cost of obtaining the value".

This is like pattern match on List with Nil case, where match on `::` first, then just `case nil`. It's nice if you have a type to check.

An example where just "obtaining the value" is expensive is `NoSymbol` in the compiler, because of the "cake". There are edits trying to choose between `eq` and `==` but that is swamped by the fetch. It is outweighed by style concerns.

If you can measure a difference reliably, then of course you are equipped to choose.

PR #1027, opened 2024-01-23
chore: Add PekkoManagedBlocker to reduce memory.

He-Pin:
Motivation:
There is an additional `new AtomicReference[Option[T]](None)` usage, by introduce a `PekkoManagedBlocker` , this reduced it.

Result:
Clean code an less memory usage.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/ThreadPoolBuilder.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/ThreadPoolBuilder.scala
@@ -173,19 +173,23 @@ object MonitorableThreadFactory {
   val doNothing: Thread.UncaughtExceptionHandler =
     new Thread.UncaughtExceptionHandler() { def uncaughtException(thread: Thread, cause: Throwable) = () }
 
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/ThreadPoolBuilder.scala
```diff
@@ -173,19 +173,23 @@ object MonitorableThreadFactory {
   val doNothing: Thread.UncaughtExceptionHandler =
     new Thread.UncaughtExceptionHandler() { def uncaughtException(thread: Thread, cause: Throwable) = () }
 
+  private[pekko] class PekkoManagedBlocker[T](thunk: => T)
```
pjfanning, line 176: why not just private?

He-Pin: `private[pekko] class PekkoForkJoinWorkerThread` is private[pekko] too, maybe its useful somewhere

pjfanning (approved):
lgtm

PR #1034, opened 2024-01-23
[DRAFT] set Apache name in Pekko poms

pjfanning:
for https://github.com/apache/incubator-pekko/issues/1006

I'll generate more draft PRs but the build has been pretty brittle when I've tried changes.

changes were suggested by @laglangyue 

Due to the build brittleness, I want to try getting things working in 1 or 2 repos first before centralising a stable solution in sbt-pekko-build.

PR #1035, opened 2024-01-23
[DRAFT] try to centralise pom name change logic

pjfanning:
This has been causing compile issues for me locally - relates to #1034

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,7 @@
  */
 
 import net.bzzt.reproduciblebuilds.ReproducibleBuildsPlugin.reproducibleBuildsCheckResolver
+import org.apache.commons.text.WordUtils
 
 ThisBuild / scalafixScalaBinaryVersion := scalaBinaryVersion.value
 
@@ -595,6 +596,13 @@ lazy val billOfMaterials = Project("bill-of-materials", file("bill-of-materials"
   .settings(PekkoBuild.defaultSettings)
   .settings(
     name := "pekko-bom",
+    publishM2Configuration := publishM2Configuration.value
+      .withOverwrite(true)
+      .withPublishMavenStyle(true),
+    // rename module
+    makePomConfiguration := makePomConfiguration.value.withModuleInfo(
+      makePomConfiguration.value.moduleInfo.get.withNameFormal(s"Apache Pekko Bill of Materials")
+    ),
     bomIncludeProjects := userProjects,
     description := s"${description.value} (depending on Scala ${CrossVersion.binaryScalaVersion(scalaVersion.value)})")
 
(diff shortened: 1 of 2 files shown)


mdedetrich:
@pjfanning This can be fixed in a much nicer way, likely with a single line at https://github.com/mdedetrich/sbt-apache-sonatype

mdedetrich:
@pjfanning So I went down the rabbit hole of the whole `pom.xml` name field issue and I created a PR for this with tests at https://github.com/mdedetrich/sbt-apache-sonatype/pull/31 with proper tests and everything. The advantage of this implementation is that it leaves everything else regarding the sbt build in tact, only the final pom.xml is being modified (which is what gets published).

What is still unclear is what format should this name field be, i.e. should it be a completely human readable `Apache Pekko Streams` or `apache-pekko-streams` which https://mvnrepository.com/ then pretty prints to `Apache Pekko Streams`. I can’t find any ASF rule on this actually.

I also sent a message to the mailing list on this. If we do actually need a human readable representation in the name field then I need to add a `apacheSonatypeArtifactNamePrettifier` `SettingKey` to that PR which will be a function that will pretty the `apache-pekko-streams` to `Apache Pekko Streams`.

gmethvin:
I think if we're going to bother having a display name that's different from the artifact ID, we might as well have a fully pretty-printed version of the name. If we use something that looks like an artifact ID it might confuse people at first glance.

mdedetrich:
> I think if we're going to bother having a display name that's different from the artifact ID, we might as well have a fully pretty-printed version of the name. If we use something that looks like an artifact ID it might confuse people at first glance.

Agreed, the situation itself is pretty confusing. Would like to get clarity on what the actual ASF rule on this is, human readable name would make it more clear but interestingly it appears that sbt and other build tools were designed around `name` being the machine representation of the project name (which is why in sbt, the `name` setting key is what is displayed in the sbt shell for the project name).

gmethvin:
> interestingly it appears that sbt and other build tools were designed around name being the machine representation of the project name (which is why for example, in sbt, the name setting key is what is displayed in the sbt shell for the project name).

Yeah that's interesting, sbt actually has a `normalizedName` that it generates based on the `name`, so if you did set your `name` to `Pekko Streams` then it would generate `pekko-streams` to be used as the artifact ID. So in that sense `name` is _intended_ to be a human-readable name, but in my experience this "feature" isn't really used because it creates more confusion.

I don't know what the common conventions are for `<name>` in Maven POM files, but it seems like other Apache projects tend to use a fully pretty-printed version of the name.

mdedetrich:
> Yeah that's interesting, sbt actually has a `normalizedName` that it generates based on the `name`, so if you did set your `name` to `Pekko Streams` then it would generate `pekko-streams` to be used as the artifact ID. So in that sense `name` is _intended_ to be a human-readable name, but in my experience this "feature" isn't really used because it creates more confusion.
> 
> I don't know what the common conventions are for `<name>` in Maven POM files, but it seems like other Apache projects tend to use a fully pretty-printed version of the name.

Thanks for the context, it does seem that `name` field should be human readable but due to various reasons people don't do that anymore, in terms of confusion it also doesn't help that sites like https://mvnrepository.com/ take the machine readable values and pretty prints. Lets wait for confirmation from ASF if its indeed meant to be human readable and then I can update the PR as necessary.

pjfanning:
to be clear, this PR is not a solution - I raised it to highlight how it causes compile issues for the Pekko runtime code

PR #1041, opened 2024-01-25
perf: Add dedicated collectFirst implementation.

He-Pin:
Motivation:
Use dedicated collectFirst implementation

Result:
Better performance.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectFirst.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/CollectFirst.scala
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


pjfanning (changes requested):
This new class looks a lot like existing Akka based code. Is it truly new? Otherwise, use the Lightbend and Apache header from the original file.

He-Pin:
Yes it's totally new, I changed it from CollectWhile not the origin Collect, why should I? I think every `collectFirst` will implemented this way. And I'm not looking into the code of `Collect` when implement this and CollectWhile.

pjfanning (approved):
lgtm

mdedetrich (approved):
lgtm

@pjfanning I can also confirm that if you wanted to write this from scratch it would also look like this. While it may look similar to existing akka code, there isn't much you can do about that because generally all of this code has the same style so if you want to write new functionality and keep the existing style it has a very high chance of looking this way.

He-Pin:
Yes, we can ask anyone to write this code, and anyone's code will result in this style.

PR #1049, opened 2024-01-26
Add missing java api for StreamTestKit

### LINKED ISSUE #970: Feature request:Add missing java api for `StreamTestKit`
Motivation:
Currently there are some methods are missing in the StreamTestkit for javadsl. And which is very annoying when you implement  a `GraphStage` in Java and test it in pure Java.

I think this task is easy if anyone else in the community want to take a try. otherwise I would like to prepare this this weekend, because I coding in java everyday.

Result:
More friendly for Java end users.

naosense:
To resolve #970

### DIFF
--- a/stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
+++ b/stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
@@ -241,7 +241,7 @@ object TestPublisher {
      * configuration entry "pekko.test.timefactor", while the min Duration is not.
      *
      * {{{
-     * val ret = within(50 millis) {
+     * val ret = within(Duration.ofMillis(50)) {
      *   test ! "ping"
      *   expectMsgClass(classOf[String])
      * }
@@ -257,7 +257,7 @@ object TestPublisher {
     def within[T](max: FiniteDuration)(f: => T): T = executeAfterSubscription { probe.within(max)(f) }
 
     /**
-     * Same as calling `within(0 seconds, max)(f)`.
+     * Same as calling `within(Duration.ofSeconds(0), max)(f)`.
      */
     def within[T](max: java.time.Duration)(f: => T): T = executeAfterSubscription { probe.within(max.asScala)(f) }
   }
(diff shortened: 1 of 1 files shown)


stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
```diff
@@ -818,11 +886,35 @@ object TestSubscriber {
...
+     * Execute code block while bounding its execution time between `min` and
+     * `max`. `within` blocks may be nested. All methods in this trait which
+     * take maximum wait times are available in a version which implicitly uses
+     * the remaining time governed by the innermost enclosing `within` block.
+     *
+     * Note that the timeout is scaled using Duration.dilated, which uses the
+     * configuration entry "pekko.test.timefactor", while the min Duration is not.
+     *
+     * {{{
+     * val ret = within(50 millis) {
```
He-Pin, line 899: delete or rewrite this code as java:)

stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
```diff
@@ -818,11 +886,35 @@ object TestSubscriber {
...
+     *   test ! "ping"
+     *   expectMsgClass(classOf[String])
+     * }
+     * }}}
+     */
+    def within[T](min: java.time.Duration, max: java.time.Duration)(f: => T): T =
```
He-Pin, line 905: `(f: => T)` should be `f( function.Function<T>)`

stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/StreamTestKit.scala
```diff
@@ -818,11 +886,35 @@ object TestSubscriber {
...
+
     /**
      * Same as calling `within(0 seconds, max)(f)`.
      */
     def within[T](max: FiniteDuration)(f: => T): T = probe.within(max)(f)
 
+    /**
+     * Same as calling `within(0 seconds, max)(f)`.
+     */
+    def within[T](max: java.time.Duration)(f: => T): T = probe.within(max.asScala)(f)
```
He-Pin, line 916: f => T should be `function.Function[Unit, T]`

He-Pin:
Thanks: here is some of my feedback:
1. no curried method in Java dsl
2. use the functions in `japi.function.Function` instead.
3. If you can write some spec to tests this api, then what would be nice too.

pjfanning (changes requested):
looks like a good start

we need the function type changes that @kerr suggested

but we should also really have some basic test coverage of the new functions too

and all new public methods need `@since 1.1.0` in their scaladoc

naosense:
> 3. If you can write some spec to tests this api, then what would be nice too.

@He-Pin there is no java testcase?  
image

He-Pin:
> > 3. If you can write some spec to tests this api, then what would be nice too.
> 
> @He-Pin there is no java testcase?  
> image
> 

If you not sure how to write the test cases, I can do a following pr for your pr with tests.

naosense:
> If you not sure how to write the test cases, I can do a following pr for your pr with tests.

yes please, master pin @He-Pin

pjfanning:
@naosense @kerr any interest in finishing this? - we might still fit this into 1.1.0-M1 if it can be finished soon

He-Pin:
> @naosense @kerr any interest in finishing this? - we might still fit this into 1.1.0-M1 if it can be finished soon

I think I can took this on weekend.

PR #1054, opened 2024-01-27
feat: Add multi java lts support.

He-Pin:
Motivation：

Add support to test JDK 11, JDK 17, and JDK 21 related code and just add the code when the runtime version matches.
This will help us to detect LTS version-related issues at the nightly build.
With JDK 21:
image
With JDK 11:
image

@Roiocam ping for a review too

### DIFF
--- a/docs/src/main/java-jdk-21/docs/actors/classical/OptimizedActorWithJava21.java
+++ b/docs/src/main/java-jdk-21/docs/actors/classical/OptimizedActorWithJava21.java
@@ -1,33 +1,38 @@
-
+package docs.actors.classical;
 // #pattern-matching
 
-static class OptimizedActorWithJava21 extends UntypedAbstractActor {
-  public static class Msg1 {}
+import org.apache.pekko.actor.UntypedAbstractActor;
 
-  public static class Msg2 {}
-
-  public static class Msg3 {}
-
-  @Override
-  public void onReceive(Object msg) throws Exception {
-    switch(msg) {
-      case Msg1 msg -> receiveMsg1((Msg1) msg);
-      case Msg2 msg -> receiveMsg2((Msg2) msg);
-      case Msg3 msg -> receiveMsg3((Msg3) msg);
-      default _ -> unhandled(msg);
+public class OptimizedActorWithJava21 extends UntypedAbstractActor {
+    public static class Msg1 {
     }
-  }
 
-  private void receiveMsg1(Msg1 msg) {
-      // actual work
-  }
+    public static class Msg2 {
+    }
 
-  private void receiveMsg2(Msg2 msg) {
-      // actual work
-  }
+    public static class Msg3 {
+    }
 
-  private void receiveMsg3(Msg3 msg) {
-      // actual work
-  }
+    @Override
+    public void onReceive(Object msg) throws Exception {
+        switch (msg) {
+            case Msg1 msg1 -> receiveMsg1(msg1);
+            case Msg2 msg2 -> receiveMsg2(msg2);
+            case Msg3 msg3 -> receiveMsg3(msg3);
+            default -> unhandled(msg);
+        }
+    }
+
+    private void receiveMsg1(Msg1 msg) {
+        // actual work
+    }
+
+    private void receiveMsg2(Msg2 msg) {
+        // actual work
+    }
+
+    private void receiveMsg3(Msg3 msg) {
+        // actual work
+    }
 }
 // #pattern-matching
--- a/project/CopyrightHeaderForJdk9.scala
+++ b/project/CopyrightHeaderForJdk9.scala
@@ -12,7 +12,6 @@
  */
 
 import de.heikoseeberger.sbtheader.HeaderPlugin.autoImport.headerSources
-import sbt.Keys.sourceDirectory
 import sbt.{ Compile, Def, Test, _ }
 
 object CopyrightHeaderForJdk9 extends AutoPlugin {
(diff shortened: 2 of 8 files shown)


docs/src/main/java-jdk21/docs/actors/classical/OptimizedActorWithJava21.java
```diff
@@ -0,0 +1,38 @@
+package docs.actors.classical;
```
pjfanning, line 1: -1 on the dir rename - we should keep the existing name style

https://github.com/apache/incubator-pekko/blob/a2b5b5df22ec0820691bd84c358fb4352b97cbd0/project/Jdk9.scala#L26

He-Pin:
Updating~~

docs/src/main/java-jdk21/docs/actors/classical/OptimizedActorWithJava21.java
He-Pin: -1 for what? How can you compile a Java 21 code with Java 8/11/17?
The `"java-jdk9` file is still there.

BTW, I can't upgrade because of GFW, you can check it now.

He-Pin: I see, I just upgraded, and the networking issue makes it very ...

He-Pin:
image
@pjfanning Actually we currently how two style :(

pjfanning (changes requested):
the file should be left in java-jdk-21 to match the existing dir names, like java-jdk-9

He-Pin:
@pjfanning I think it should be fine now

docs/src/main/java-jdk-21/docs/actors/classical/OptimizedActorWithJava21.java
```diff
@@ -1,33 +1,38 @@
...
-      case Msg2 msg -> receiveMsg2((Msg2) msg);
-      case Msg3 msg -> receiveMsg3((Msg3) msg);
-      default _ -> unhandled(msg);
+    @Override
+    public void onReceive(Object msg) throws Exception {
+        switch (msg) {
+            case Msg1 msg1 -> receiveMsg1(msg1);
+            case Msg2 msg2 -> receiveMsg2(msg2);
+            case Msg3 msg3 -> receiveMsg3(msg3);
+            default -> unhandled(msg);
```
He-Pin, line 22: Now catch.

project/Jdk9.scala
```diff
@@ -11,36 +11,87 @@
  * Copyright (C) 2017-2022 Lightbend Inc. <https://www.lightbend.com>
  */
 
-import sbt._
 import sbt.Keys._
+import sbt._
 
 object Jdk9 extends AutoPlugin {
```
pjfanning, line 17: shouldn't this be renamed - it has nothing to do with Jdk9 now

He-Pin: I think we can just leave it for now, I changed it to `MultiJavaLTS` support, and then there are many place need to update.

He-Pin: If you like, I think I can create a pending PR to do this, but as I say, many places need to be updated.

project/Jdk9.scala
```diff
@@ -11,36 +11,87 @@
...
   lazy val TestJdk9 = config("TestJdk9").extend(Test).extend(CompileJdk9)
 
-  val SCALA_SOURCE_DIRECTORY = "scala-jdk-9"
-  val SCALA_TEST_SOURCE_DIRECTORY = "scala-jdk9-only"
-  val JAVA_SOURCE_DIRECTORY = "java-jdk-9"
-  val JAVA_TEST_SOURCE_DIRECTORY = "java-jdk9-only"
+  lazy val SCALA_SOURCE_DIRECTORY = getAdditionalSourceDirectoryNames("scala", isTest = false)
+  lazy val SCALA_TEST_SOURCE_DIRECTORY = getAdditionalSourceDirectoryNames("scala", isTest = true)
+
+  lazy val JAVA_SOURCE_DIRECTORY = getAdditionalSourceDirectoryNames("java", isTest = false)
```
pjfanning, line 33: the name need to change to JAVA_SOURCE_DIRECTORIES or better still, use Scala naming convention of JavaSourceDirectories. JAVA_TEST_SOURCE_DIRECTORY should be renamed too.

He-Pin: Change it directly, I just back to my home, very bad network.

pjfanning: This is not my PR. It is also not even urgent. It is not normal practice to request other people to edit your PR for you. You can change this PR when you have free time. If that free time is in a week's time, then no harm.

He-Pin: I just done it locally, will try to update...GFW.

project/Jdk9.scala
```diff
@@ -11,36 +11,87 @@
...
+  lazy val additionalTestSourceDirectories = Def.setting {
+    additionalScalaTestSourceDirectories.value ++ additionalJavaTestSourceDirectories.value
+  }
+
+  private def getAdditionalSourceDirectoryNames(language: String, isTest: Boolean): Seq[String] = {
+    for {
+      version <- supportedJavaLTSVersions if version.toInt <= JdkOptions.JavaVersion.majorVersion
+    } yield {
+      if (isTest) {
+        s"$language-jdk${version}-only"
```
Roiocam, line 73: I notice you are using string interpolation, why not use `suffix` argument rather than a boolean, we could gain beautiful code.

project/CopyrightHeaderForJdk9.scala
```diff
@@ -12,24 +12,25 @@
...
-      Compile / headerSources ++=
-        (((Compile / sourceDirectory).value / SCALA_SOURCE_DIRECTORY) ** "*.scala").get,
-      Test / headerSources ++=
-        (((Test / sourceDirectory).value / SCALA_TEST_SOURCE_DIRECTORY) ** "*.scala").get,
-      Compile / headerSources ++=
-        (((Compile / sourceDirectory).value / JAVA_SOURCE_DIRECTORY) ** "*.java").get,
-      Test / headerSources ++=
-        (((Test / sourceDirectory).value / JAVA_TEST_SOURCE_DIRECTORY) ** "*.java").get)
+    for {
+      dir <- additionalSourceDirectories.value ++ additionalTestSourceDirectories.value
```
Roiocam, line 25: brilliant.

project/Jdk9.scala
```diff
@@ -11,36 +11,87 @@
...
-  val JAVA_SOURCE_DIRECTORY = "java-jdk-9"
-  val JAVA_TEST_SOURCE_DIRECTORY = "java-jdk9-only"
+  lazy val ScalaSourceDirectories: Seq[String] = getAdditionalSourceDirectoryNames("scala", isTest = false)
+  lazy val ScalaTestSourceDirectories: Seq[String] = getAdditionalSourceDirectoryNames("scala", isTest = true)
+
+  lazy val JavaSourceDirectories: Seq[String] = getAdditionalSourceDirectoryNames("java", isTest = false)
+  lazy val JavaTestSourceDirectories: Seq[String] = getAdditionalSourceDirectoryNames("java", isTest = true)
+
+  private lazy val additionalScalaSourceDirectories = Def.setting {
+    getAdditionalSourceDirectories(
```
Roiocam, line 37: I think these methods could be simpler.

project/JdkOptions.scala
```diff
@@ -27,6 +27,19 @@ object JdkOptions extends AutoPlugin {
 
   lazy val specificationVersion: String = sys.props("java.specification.version")
 
+  object JavaVersion {
+    val majorVersion: Int = {
```
Roiocam, line 31: I am fine with that, but it will be great if we have a better solution for this.

Roiocam (commented):
Brilliant works, thanks. Leave some picky suggestion.

He-Pin:
@pjfanning , @Roiocam ping with additional simplification from @Roiocam :)

Roiocam (approved):
Thanks @He-Pin , It's great and useful, we could support more JDK feature now.

He-Pin:
@pjfanning @mdedetrich Would you give this another look？

He-Pin:
@pjfanning ping for another round of review, I think this is done.

pjfanning (approved):
lgtm

Issue #1040, opened 2024-01-25
add CI check for the Java 9+ classes


pjfanning:
see #1039 

I think we could have a check that uses a `unzip -l` call and that searches the output for expected class files.

There are other ways to check too.

He-Pin:
How about using scala-cli
https://scala-cli.virtuslab.org/docs/reference/directives to verfify it.
https://scala-cli.virtuslab.org/docs/cookbooks/introduction/gh-action/
seems like we can use it in github action and use the local repo

pjfanning:
> How about using scala-cli https://scala-cli.virtuslab.org/docs/reference/directives to verfify it. https://scala-cli.virtuslab.org/docs/cookbooks/introduction/gh-action/ seems like we can use it in github action and use the local repo

If you want to try using this, that's fine. The links provided do not appear to mention how to check a jar for the presence of a file. If implement this, I will use the easier solution of using the unzip command.

Roiocam:
Too complex for me, if let me do this, I will be trying there easy way:

- create a wrong test on `java-jdk-9` path, and then run it on verify CI. If the unit test failed, that means jdk9 path was compiled!
- verify `unmanagedSourceDirectories` before build

He-Pin:
seems using `jar xf` and then test if `find` will works too, and simpler. I was thinking using sbt to generate an scala-cli.sc and test the local published jars.

He-Pin:
I think check the unmanagedSourceDirectores will not works:
```scala
//sbt with on Java 21
sbt:pekko-stream> show sourceDirectories
[info] * C:\Users\hepin\IdeaProjects\incubator-pekko\stream\src\main\scala
[info] * C:\Users\hepin\IdeaProjects\incubator-pekko\stream\src\main\scala-2.13
[info] * C:\Users\hepin\IdeaProjects\incubator-pekko\stream\src\main\scala-2
[info] * C:\Users\hepin\IdeaProjects\incubator-pekko\stream\src\main\java
[info] * C:\Users\hepin\IdeaProjects\incubator-pekko\stream\src\main\scala-2.13+
[info] * C:\Users\hepin\IdeaProjects\incubator-pekko\stream\target\scala-2.13\src_managed\main
```

pjfanning:
> I think checking the unmanagedSourceDirectores will not work:
> 
> ```scala
> //sbt with on Java 21
> sbt:pekko-stream> show sourceDirectories
[...]

PR #1064, opened 2024-01-28
Verify JDK 9 classes.

He-Pin:
Motivation:
Use Scala-cli to verify the resulting jar can be used in JDK 9, I think this can be a more complete solution than` unzip ...find.`
refs: https://github.com/apache/incubator-pekko/issues/1040

Result:
Can be tested with scala-cli at nightly job, which was verified with https://github.com/apache/incubator-pekko/actions/runs/7684565710/job/20941211960?pr=1064

ping @Roiocam too

### DIFF
--- a/project/VerifyJDK9Classes.scala
+++ b/project/VerifyJDK9Classes.scala
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+object VerifyJDK9Classes {
+  import sbt._
+  import sbt.Keys._
+
+  lazy val settings: Seq[Setting[_]] = inConfig(Compile) {
+    Seq {
+      sourceGenerators += {
+        generateAndWriteScalaCLIScript(
+          target,
+          _ / "scala-cli" / "VerifyJDK9Classes.sc")
+      }
+    }
+  }
+
+  def generateAndWriteScalaCLIScript(dir: SettingKey[File], locate: File => File): Def.Initialize[Task[Seq[sbt.File]]] =
+    Def.task[Seq[File]] {
+      val script = generateScalaCLIScript(version.value)
+      val file = locate(dir.value)
+      val content = script.stripMargin.format(version.value)
+      if (!file.exists || IO.read(file) != content) IO.write(file, content)
+      // the generated file is not used.
+      Nil
+    }
+
+  private def generateScalaCLIScript(version: String): String =
+    s"""
+      |/*
+      | * Licensed to the Apache Software Foundation (ASF) under one or more
+      | * contributor license agreements. See the NOTICE file distributed with
+      | * this work for additional information regarding copyright ownership.
+      | * The ASF licenses this file to You under the Apache License, Version 2.0
+      | * (the "License"); you may not use this file except in compliance with
+      | * the License. You may obtain a copy of the License at
+      | *
+      | *    http://www.apache.org/licenses/LICENSE-2.0
+      | *
+      | * Unless required by applicable law or agreed to in writing, software
+      | * distributed under the License is distributed on an "AS IS" BASIS,
+      | * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+      | * See the License for the specific language governing permissions and
+      | * limitations under the License.
+      | */
+      |//> using scala 2.13
+      |//> using dep "org.apache.pekko::pekko-stream:${version}"
+      |////> using jvm 11
+      |object VerifyJDK9Classes {
+      |  def main(args: Array[String]): Unit = {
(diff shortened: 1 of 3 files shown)


mdedetrich:
As an alternate suggestion, wouldn't it be simpler to just check the header for some specific `.class` file that we know should be compiled with JDK9?

He-Pin:
@mdedetrich Do you mean `unzip | find`?

.github/workflows/verify-jdk9-classes.yml
```diff
@@ -0,0 +1,65 @@
...
+          sudo apt update
+          sudo apt install scala-cli
+
+      - name: Use Scala-CLI to check
+        run: |-
+          scala-cli --version
+          echo "Starting verification with Scala-CLI"
+          scala-cli stream/target/scala-cli/VerifyJDK9Classes.sc && echo "Verification successful" || (
+          echo "Error when VerifyJDK9Classes"
+          exit 1
```
He-Pin, line 63: I will move this to the nightly build once checked.

He-Pin:
image

He-Pin:
@pjfanning I only added this to the nightly build, which I think should be enough, if you like, I think we can add a dedicated job to test this.

project/VerifyJDK9Classes.scala
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+object VerifyJDK9Classes {
+  import sbt.*
```
pjfanning, line 19: use `._` in imports

He-Pin: Oh, Yes, but It will work now AFAIK.

project/VerifyJDK9Classes.scala
```diff
@@ -0,0 +1,85 @@
...
+      val content = script.stripMargin.format(version.value)
+      if (!file.exists || IO.read(file) != content) IO.write(file, content)
+      // the generated file is not used.
+      Nil
+    }
+
+  private def generateScalaCLIScript(version: String): String =
+    s"""
+      |/*
+      | * Licensed to the Apache Software Foundation (ASF) under one or more
```
pjfanning, line 45: why not just check in this script instead of generating it?

He-Pin: That will not test the result jar, at least I don't know how to now.

mdedetrich:
> @mdedetrich Do you mean `unzip | find`?

Yup, even an awk ontop just to parse out the bytecode header in the `.class` file to see if it matches the JDK 9 bytecode version

He-Pin:
@mdedetrich 
The difference is that this approach can test whether the final product really works properly, rather than just checking if xxx.class exists using grep. I think this is a relatively advantageous point. In addition, it can expand to test multiple platforms' products, while if you use grep xxx.class in the end, can only determine whether xxx.class exists, right? At least I saw that Dotty uses this approach as well.

He-Pin:
Because I'm going back to my hometown the day after tomorrow, I want to help resolve some blocker issues for 1.1.0-M1. At least I think this solution can solve the issues associated with it, and it's not such a strange solution because there is already a VersionGenerator now.

mdedetrich:
> @mdedetrich The difference is that this approach can test whether the final product really works properly, rather than just checking if xxx.class exists using grep. I think this is a relatively advantageous point. In addition, it can expand to test multiple platforms' products, while if you use grep xxx.class in the end, can only determine whether xxx.class exists, right? At least I saw that Dotty uses this approach as well.

I understand and I will accept either solution. My main reason for suggesting the simpler approach is that our issue is quite black and white, either the JDK 9 classes don't get generated at all or they get generated with the correct JDK version and if they do its completely correct.

Hence I think that checking whether it works is a bit excessive because at that point we are basically checking the scalac compiler is working where as what our historic issue has been is that the misconfiguration of sbt-osgi/multijvm/etc etc means that certain classes don't get compiled/included

He-Pin:
I understand you too. But in another post, you said we still need to support Java 8 for a while. If we eventually migrate to only supporting Java 11, then this can naturally be deleted, as the project is still in incubation. I think we can bear this cost to ensure our products don't become a laughing stock and to ensure supply chain security, right?

mdedetrich:
> I understand you too. But in another post, you said we still need to support Java 8 for a while. If we eventually migrate to only supporting Java 11, then this can naturally be deleted, as the project is still in incubation. I think we can bear this cost to ensure our products don't become a laughing stock and to ensure supply chain security, right?

Sure, as I said I will accept either solution.

I think there may be a more elegant/simpler one we can talk about/do later.

He-Pin:
@mdedetrich Of course, I'm looking forward to your improvements, at least our community has polished our build a lot.

project/VerifyJDK9Classes.scala
```diff
@@ -0,0 +1,85 @@
...
+      |  def main(args: Array[String]): Unit = {
+      |    import org.apache.pekko.actor.ActorSystem
+      |    import org.apache.pekko.stream.scaladsl.{ JavaFlowSupport, Source }
+      |
+      |    import java.lang.System.exit
+      |    import scala.concurrent.Await
+      |    import scala.concurrent.duration.DurationInt
+      |    implicit val system: ActorSystem = ActorSystem.create("test")
+      |    val future = Source(1 to 3).runWith(
+      |      JavaFlowSupport.Sink.asPublisher[Int](fanout = false).mapMaterializedValue { p =>
```
Roiocam, line 73: I have reviewed this for a while, i am kind of understand why this PR does.

A simple `grep` pipeline connected to unzip results only checks some of the classes, once these class names change or delete, we have to upgrade the CI too.

Another way that parse out the bytecode header, I think it was too complex.

Roiocam (approved):
LGTM, Thanks. I am glad we have a double insurance for these JDK9 check.

He-Pin:
@mdedetrich @pjfanning I have rebased.

nvollmar (approved):
lgtm

PR #1068, opened 2024-01-28
chore: remove the inline modifier from pekko util.

He-Pin:
Motivation: 

These method is used in other pekko projects, and can not be used in Java.
Remove the `inline def` in scala 3

Or we have to find another way 
refs: https://github.com/apache/incubator-pekko-connectors/pull/412
refs: https://github.com/apache/incubator-pekko-connectors/actions/runs/7686045276/job/20944445485?pr=412#step:5:506

Or we just update the Java code. @mdedetrich Ping with a scala forwarder in these projects.
refs: https://github.com/lampepfl/dotty/issues/19346

Or we need release another 1.0.x version of pekko-http which compiles with the `inline def` version of pekko `1.0.x`, otherwise will encounter this issue in pekko-http anyway.

Asked on: https://contributors.scala-lang.org/t/can-we-still-generate-bytecode-when-inline-def-method-been-called-from-java/6500

This will cause the `pekko-http` 1.0.x can not been used with `pekko` 1.1.x on Scala 3.

### DIFF
--- a/actor/src/main/scala-3/org/apache/pekko/util/FutureConverters.scala
+++ b/actor/src/main/scala-3/org/apache/pekko/util/FutureConverters.scala
@@ -28,7 +28,7 @@ private[pekko] object FutureConverters {
   def asJava[T](f: Future[T]): CompletionStage[T] = javaapi.FutureConverters.asJava(f)
 
   implicit final class FutureOps[T](private val f: Future[T]) extends AnyVal {
-    inline def asJava: CompletionStage[T] = javaapi.FutureConverters.asJava(f)
+    def asJava: CompletionStage[T] = javaapi.FutureConverters.asJava(f)
   }
 
   // Ideally this should have the Scala 3 inline keyword but then Java sources are
@@ -36,6 +36,6 @@ private[pekko] object FutureConverters {
   def asScala[T](cs: CompletionStage[T]): Future[T] = javaapi.FutureConverters.asScala(cs)
 
   implicit final class CompletionStageOps[T](private val cs: CompletionStage[T]) extends AnyVal {
-    inline def asScala: Future[T] = javaapi.FutureConverters.asScala(cs)
+    def asScala: Future[T] = javaapi.FutureConverters.asScala(cs)
   }
 }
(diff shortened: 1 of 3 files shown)


mdedetrich:
So FYI whats the weird thing here is that its known that `inline def` does not work if its called within Java `*.java` source code however its definitely meant to work when called within `*.scala` source code and the whole point of Scala compiler is that its meant to, well, inline that method so with

```
[error] Test docs.javadsl.SqsSourceTest failed: java.lang.NoSuchMethodError: org.apache.pekko.util.FutureConverters$FutureOps$.asJava$extension(Lscala/concurrent/Future;)Ljava/util/concurrent/CompletionStage;, took 0.0 sec
```

you shouldn't even see `org.apache.pekko.util.FutureConverters$FutureOps$.asJava$extension(Lscala/concurrent/Future;)Ljava/util/concurrent/CompletionStage` in the stack trace, it should be just `CompletionStage` at this point.

mdedetrich:
Ah I think I know what the problem is, Pekko Http 1.1.x is not being compiled against Pekko 1.1.x but instead Pekko 1.0.x and because Pekko 1.0.x doesn't have the `inline` keyword it didn't get inlined within Pekko Http 1.1.x so Pekko Http 1.1.x is still expecting the inline methods to exist in bytecode but they don't because they are being stripped.

Ill have to think about this, worst case scenario is that we have to remove all the Scala 3 inline keywords

He-Pin:
That's true, I think @pjfanning will not support we backport this inline support back to 1.0.x.

An in source helper method will not help the Http 1.0.x without inkine backport.

mdedetrich:
> That's true, I think @pjfanning will not support we backport this inline support back to 1.0.x.
> 
> An in source helper method will not help the Http 1.0.x without inkine backport.

I found a possible solution using https://docs.scala-lang.org/scala3/reference/other-new-features/targetName.html, making a PR now

He-Pin:
But not sure whether if that just invalid the inline support, somekind of kotlin 's JvmName annotation thing.

mdedetrich:
> But not sure whether if that just invalid the inline support, somekind of kotlin 's JvmName annotation thing.

Will explain in PR

mdedetrich:
PR created at https://github.com/apache/incubator-pekko/pull/1070

mdedetrich:
@He-Pin Shall we close this since https://github.com/apache/incubator-pekko/pull/1070 is merged? We can re-open it later if that solution doesn't work.

He-Pin:
Ok，

Issue #1066, opened 2024-01-28
Pekko 1.1 snapshots break Pekko Connectors tests


pjfanning:
https://github.com/apache/incubator-pekko-connectors/pull/412

Needs investigation before we can release 1.1.0-M1

We are getting NoSuchMethodErrors on some of the pekko utils classes like OptionConverter and FutureConverter.

It could be that we need to rewrite our pekko-connector test builds - to compile with pekko 1.0 but then to test with pekko 1.1.

```
[error] Test docs.javadsl.SnsPublisherTest failed: java.lang.NoSuchMethodError: org.apache.pekko.util.FutureConverters$FutureOps$.asJava$extension(Lscala/concurrent/Future;)Ljava/util/concurrent/CompletionStage;, took 0.0 sec
[error]     at org.apache.pekko.http.javadsl.Http.shutdownAllConnectionPools(Http.scala:798)
[error]     at docs.javadsl.SnsPublisherTest.tearDownAfterClass(SnsPublisherTest.java:71)
```

```
[error] Test docs.javadsl.OpensearchV1Test failed: java.lang.NoSuchMethodError: 'java.util.concurrent.CompletionStage org.apache.pekko.util.FutureConverters$FutureOps$.asJava$extension(scala.concurrent.Future)', took 0.0 sec
[error]     at org.apache.pekko.http.javadsl.Http.singleRequest(Http.scala:643)
```

Issue seems to happen with Scala 3.3 testing but not with Scala 2.12 or 2.13 testing.

This is a command that fails in incubator-pekko-connectors (don't forget to use docker-compose to start all the necessary containers for this connector). 
```
sbt -Dpekko.build.pekko.version=main "++3.3.1 elasticsearch/test"
```

Seems like pekko-http jars built for Scala 3 with Pekko 1.0 jars do not work Pekko 1.1 jars at runtime.

He-Pin:
May because of the inline？@mdedetrich

pjfanning:
The issue is still happening. Please do not close this issue until we have proof that is fixed.

https://github.com/apache/incubator-pekko-connectors/actions/runs/7686045276/job/20961266768

mdedetrich:
> The issue is still happening. Please do not close this issue until we have proof that is fixed.

I didn't intend to close the issue, it auto closed when PR that referenced the issue was merged.

I guess GitHub added "references" as ons of the keywords that causes linked issues to close when PR does, will have to find another word.

[...]

PR #1070, opened 2024-01-28
Add targetName overloaded methods for Scala 3 inline methods

mdedetrich:
We are currently having the issue where the Scala 3 `inline` keyword actually strips any mention of the method inside the generated JVM bytecode which is causing issues . Thankfully Scala 3 also introduced [targetName annotation](https://docs.scala-lang.org/scala3/reference/other-new-features/targetName.html) which lets you set the target designated symbol name in the bytecode of any method, i.e. directly quoting

> A [`@targetName`](https://scala-lang.org/api/3.x/scala/annotation/targetName.html) annotation on a definition defines an alternate name for the implementation of that definition. Example:

```scala
import scala.annotation.targetName

object VecOps:
  extension [T](xs: Vec[T])
    @targetName("append")
    def ++= [T] (ys: Vec[T]): Vec[T] = ...
```

>Here, the `++=` operation is implemented (in Byte code or native code) under the name `append`. The implementation name affects the code that is generated, and is the name under which code from other languages can call the method. For instance, `++=` could be invoked from Java like this:

```java
VecOps.append(vec1, vec2)
```

>The [`@targetName`](https://scala-lang.org/api/3.x/scala/annotation/targetName.html) annotation has no bearing on Scala usages. Any application of that method in Scala has to use `++=`, not `append`.

The last part is what is critical, `@targetName` methods cannot be called in Scala but they still appear in bytecode with the designated name so they can be called both within Java and they can be linked against at runtime.

We can use this trick to resolve https://github.com/apache/incubator-pekko/issues/1066 by just creating an alias method for every `inline def` and using `targetName` for the original method name and there won't be any conflict in the bytecode since `inline def` methods never generate any JVM bytecode (thats what the original problem is)

### DIFF
--- a/actor/src/main/scala-3/org/apache/pekko/compat/PartialFunction.scala
+++ b/actor/src/main/scala-3/org/apache/pekko/compat/PartialFunction.scala
@@ -15,6 +15,8 @@ package org.apache.pekko.compat
 
 import org.apache.pekko.annotation.InternalApi
 
+import scala.annotation.targetName
+
 /**
  * INTERNAL API
  *
@@ -22,10 +24,15 @@ import org.apache.pekko.annotation.InternalApi
  * against Scala 2.12, 2.13, 3.0
  *
  * Remove these classes as soon as support for Scala 2.12 is dropped!
+ * Remove the @targetName bytecode forwarded methods for Pekko 2.0.x since we only care about source compatibility
  */
 @InternalApi private[pekko] object PartialFunction {
 
   inline def fromFunction[A, B](f: A => B): scala.PartialFunction[A, B] =
+    scala.PartialFunction.fromFunction(f)
+
+  @targetName("fromFunction")
+  def _pekko10FromFunction[A, B](f: A => B): scala.PartialFunction[A, B] =
     scala.PartialFunction.fromFunction(f)
 
 }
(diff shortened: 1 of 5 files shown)


pjfanning:
Conceptually seems ok but would you be able to test it locally to see if it helps locally? Ideally, we wouldn't merge this and then use snapshots and CI builds using those snapshots to try it out for the first time.

mdedetrich:
> Conceptually seems ok but would you be able to test it locally to see if it helps locally? Ideally, we wouldn't merge this and then use snapshots and CI builds using those snapshots to try it out for the first time.

Don't have an issue with this albeit it would probably take more time (in terms of labour) of testing this then merging it and waiting overnight.

If this doesn't work then we would just revert this commit and the one that adds the inline keywords, we will be forced to use the Scala 3 `inline` keyword for Pekko 2.0.x

mdedetrich:
I have just updated the pr to add `_` prefix to these methods to give a signal that they are not meant to be directly called in Scala.

pjfanning (approved):
lgtm

mdedetrich:
Going to merge this so we catch the midnight snapshot deploys. Will take 1-2 days for changes to propagate until connectors.

He-Pin:
@mdedetrich @pjfanning I think we can set the connectors nightly build to 1 AM.

PR #1079, opened 2024-01-29
Remove scala 3 inline keyword since its erased in bytecode

mdedetrich:
Turns out that there is no way to deal with the issue of the Scala 3 inline keyword erasing the method from JVM bytecode so lets remove it, final solution for https://github.com/apache/incubator-pekko/issues/1066


pjfanning (approved):
Lgtm - pending CI

Issue #995, opened 2024-01-20
add toInputStream to ByteString


pjfanning:
See https://github.com/apache/incubator-pekko-http/blob/da6a228372efac2da05aae68c3acb982f7a7e58d/http-core/src/main/scala/org/apache/pekko/http/impl/engine/http2/hpack/ByteStringInputStream.scala#L30

wdyt @mdedetrich

He-Pin:
Zio has `toInputStream` on `ZStream`.

pjfanning:
One problem with adding toInputStream is that we won't be able to use it in pekko-http because we'll want pekko-http to still be able to use pekko-core 1.0.x - and presuambly, we wouldn't backport this to 1.0.x

Looking a pekko-http, I think we could improve performance by calling ByteString.toArrayUnsafe instead of toArray. Pekko-http only uses the byte[] to create a ByteArrayInputStream and this won't mutate the byte[] (but it doesn't clone it either).

That change would save us from cloning the array like ByteString.toArray does. See the hpack code link in the issue description.

He-Pin:
We can source in that code in that code in pekko http, and when we bump the required pekko version, we change the package and recompile.

mdedetrich:
> One problem with adding toInputStream is that we won't be able to use it in pekko-http because we'll want pekko-http to still be able to use pekko-core 1.0.x - and presuambly, we wouldn't backport this to 1.0.x
> 
> Looking a pekko-http, I think we could improve performance by calling ByteString.toArrayUnsafe instead of toArray. Pekko-http only uses the byte[] to create a ByteArrayInputStream and this won't mutate the byte[] (but it doesn't clone it either).
> 
> That change would save us from cloning the array like ByteString.toArray does. See the hpack code link in the issue description.
> 
> So what I think I'm saying is do this ByteString change but also do a separate pekko-http change that doesn't use this change yet.

There are many ways we can save this problem

- Force Pekko-Http to use Pekko 1.0.x (thats what the `minVersion` is meant to specify)
- Use reflection to detect if the method doesn't exist, and if it doesn't then pekko-http can use its own inner sourced variant

I would say even backporting this to Pekko 1.0.x wouldn't entirely solve the issue and hence gives a false sense of semver considering that it wouldn't work for the earlier versions of Pekko 1.0.x

pjfanning:
#1085 merged

PR #1085, opened 2024-01-30
add asInputStream to ByteString

pjfanning:
See https://github.com/apache/incubator-pekko/issues/995

This needs tests but raising it for discussion.

This is probably not that useful for Pekko usage but general ByteString users might appreciate it.

This method gets an InputStream without cloning the array but is not unsafe like toArrayUnsafe which exposes the underlying array data in a way that it can be unsafely changed. The InputStream wraps this unsafe array in a way that doesn't allow it to be modified.

I've done some experimentation and ByteArrayInputStream works just as well as alternative implementations in commons-io and fastutil.

https://github.com/apache/incubator-pekko-http/pull/424 is a good enough solution for Pekko HTTP.


actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -827,6 +827,14 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
    */
   def toArrayUnsafe(): Array[Byte] = toArray
 
+  /**
+   * Return the bytes in this ByteString as an InputStream.
+   *
+   * @return the bytes in this ByteString accessible as an InputStream
+   * @since 1.1.0
+   */
+  def getInputStream(): InputStream = new ByteArrayInputStream(toArrayUnsafe())
```
He-Pin, line 836: For composiable Bytestring, I expect this call lazy.

He-Pin: How about name it `asInputString` or `toInputStream`?

pjfanning: renamed as `asInputStream` because it is similar to `asByteBuffer`

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -827,6 +827,15 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
...
   def toArrayUnsafe(): Array[Byte] = toArray
 
+  /**
+   * Return the bytes in this ByteString as an InputStream.
+   *
+   * @return the bytes in this ByteString accessible as an InputStream
+   * @see [[asByteBuffer]]
+   * @since 1.1.0
+   */
+  final def asInputStream: InputStream = new ByteArrayInputStream(toArrayUnsafe())
```
jtjeferreira, line 837: The `toArrayUnsafe` documentation states:

> If the ByteString is backed by a single array it is returned without any copy. If it is backed by a rope
of multiple ByteString instances a new array will be allocated and the contents will be copied
into it before returning it

So we could avoid that allocation either by:
* defining `def asInputStream: InputStream` as an abstract method, and add different implementations that avoid allocations 
  * using `new ByteArrayInputStream(bytes)` for `ByteString1C`, 
  * using `new ByteArrayInputStream(bytes, start, length)` for `ByteString1`, 
  * using `new SequenceInputStream(bytestrings.map(_.asInputStream))` for `ByteStrings`
* Implement in the super class using `SequenceInputStream(asByteBuffers.map(bb => new ByteBufferBackedInputStream(bb))`

WDYT?

mdedetrich: Was about to write the same, this should be the implementation

pjfanning: good point - I have committed a change based on @jtjeferreira's suggestions

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -827,6 +836,16 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
...
 
+  /**
+   * Return the bytes in this ByteString as an InputStream.
+   *
+   * @return the bytes in this ByteString accessible as an InputStream
+   * @see [[asByteBuffer]]
+   * @since 1.1.0
+   */
+  def asInputStream: InputStream =
+    new SequenceInputStream(asByteBuffers.map(bb => new ByteBufferBackedInputStream(bb)).iterator.asJavaEnumeration)
```
jtjeferreira, line 847: this should be an abstract method, right?
```suggestion
  def asInputStream: InputStream
```

mdedetrich: Agreed as it forces you to implement it for all subtypes which avoids leaving in a default sub-optimal (in terms of performance) implementation if a subtype happens to miss an implementation

pjfanning: removed the default impl

jtjeferreira (approved):
LGTM, but 2 notes:
* mima is complaining
* should some benchmarks be written?

mdedetrich:
> * mima is complaining

```
[error] pekko-actor: Failed binary compatibility check against org.apache.pekko:pekko-actor_2.12:1.0.0! Found 1 potential problems
[error]  * abstract method asInputStream()java.io.InputStream in class org.apache.pekko.util.ByteString is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.util.ByteString.asInputStream")
```

This is a forward compatibility error and is expected, i.e. MiMa complains when you add a new method to a public interface

> * should some benchmarks be written?

Yes they should

pjfanning:
this is still WIP
* more test coverage needed
* basic benchmarks added

mdedetrich:
If it's not too pushy it would be good to get this out for M1, I suggested on mailing list we should do M1 around mid next week

pjfanning:
I have a basic benchmark that shows that ByteString.asInputStream is faster than new ByteArrayInputStream(ByteString.toArray). The latter is the current safe way to create an InputStream for a ByteString.

I have also tested ByteArrayInputStream(ByteString.toArrayUnsafe) which is about as fast as ByteString.asInputStream for simple ByteStrings (single arrays) but is slow like ByteArrayInputStream(ByteString.toArray) when you have byte strings composed of many smaller byte strings.

He-Pin:
I just checked in Netty, there is a `public class ByteBufInputStream extends InputStream implements DataInput`, Do you think we can have a `class ByteStringInputString extends InputStream implements DataInput`?

pjfanning:
> I just checked in Netty, there is a `public class ByteBufInputStream extends InputStream implements DataInput`, Do you think we can have a `class ByteStringInputString extends InputStream implements DataInput`?

I checked that already. Unfortunately, Scala conversion to JavaEnumeration only seem to work on Scala Iterators and not on Scala Collections.

With this change:
```
[error] /Users/pj.fanning/code/incubator-pekko/actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala:586:64: value asJavaEnumeration is not a member of scala.collection.immutable.Vector[java.io.InputStream]
[error]       new SequenceInputStream(bytestrings.map(_.asInputStream).asJavaEnumeration)
```

He-Pin:
> > I just checked in Netty, there is a `public class ByteBufInputStream extends InputStream implements DataInput`, Do you think we can have a `class ByteStringInputString extends InputStream implements DataInput`?
> 
> I checked that already. Unfortunately, Scala conversion to JavaEnumeration only seem to work on Scala Iterators and not on Scala Collections.
> 
> With this change:
> ```
> [error] /Users/pj.fanning/code/incubator-pekko/actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala:586:64: value asJavaEnumeration is not a member of scala.collection.immutable.Vector[java.io.InputStream]
> [error]       new SequenceInputStream(bytestrings.map(_.asInputStream).asJavaEnumeration)
> ```

'bytestrings.iterator.map(_.asInputStream).asJavaEnumeration'

pjfanning:
> I just checked in Netty, there is a `public class ByteBufInputStream extends InputStream implements DataInput`, Do you think we can have a `class ByteStringInputString extends InputStream implements DataInput`?

We are dealing with wrapping byte arrays - the perfect InputStream for wrapping a byte array in ByteArrayInputStream.

He-Pin:
It will defer the mapping if you don't read.

He-Pin (approved):
Lgtm

pjfanning:
> It will defer the mapping if you don't read.

it is being read - there is nothing gained here by thinking there is lazy eval anywhere here. SequenceInputStream takes an evaluated JavaEnumeration. For me, if we adjust this, we could write a custom function that maps a Vector to a JavaEnumeration without requiring the intermediate iterator.

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -566,6 +572,9 @@ object ByteString {
 
     def asByteBuffers: scala.collection.immutable.Iterable[ByteBuffer] = bytestrings.map { _.asByteBuffer }
 
+    override def asInputStream: InputStream =
+      new SequenceInputStream(bytestrings.map(_.asInputStream).iterator.asJavaEnumeration)
```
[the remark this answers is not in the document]
He-Pin: @@pjfanning I will send a suggestion tomorrow about this line, currently on my phone.

He-Pin, line 576:
```suggestion
      new SequenceInputStream(bytestrings.iterator.map(_.asInputStream).asJavaEnumeration)
```

actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
```diff
@@ -578,6 +582,9 @@ object ByteString {
 
     def asByteBuffers: scala.collection.immutable.Iterable[ByteBuffer] = bytestrings.map { _.asByteBuffer }
 
+    override def asInputStream: InputStream =
+      new SequenceInputStream(bytestrings.map(_.asInputStream).iterator.asJavaEnumeration)
```
He-Pin, line 586:
```suggestion
      new SequenceInputStream(bytestrings.iterator.map(_.asInputStream).asJavaEnumeration)
```

actor/src/main/scala-3/org/apache/pekko/util/ByteString.scala
```diff
@@ -579,6 +583,9 @@ object ByteString {
 
     def asByteBuffers: scala.collection.immutable.Iterable[ByteBuffer] = bytestrings.map { _.asByteBuffer }
 
+    override def asInputStream: InputStream =
+      new SequenceInputStream(bytestrings.map(_.asInputStream).iterator.asJavaEnumeration)
```
He-Pin, line 587:
```suggestion
      new SequenceInputStream(bytestrings.iterator.map(_.asInputStream).asJavaEnumeration)
```

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
He-Pin: Feel free, I have back hometown

He-Pin: But still, it's worth optimizing even income is insignificant！

He-Pin (approved):
Lgtm

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
pjfanning: turns out that @He-Pin 's suggestion does seem to run a little faster - so I've made the change

PR #1086, opened 2024-01-30
chore: Remove useless contractor parameters in doc tests.

He-Pin:
Motivation:
Fix the nightly build on Scala 3.3

https://github.com/apache/incubator-pekko/actions/runs/7714027653/job/21025328188

I tested it locally, the tests passed.

I Have no idea why it works on Scala 2.13, as there is an additional argument, and should fail as Scala 3.

### DIFF
--- a/docs/src/test/scala/docs/actor/ActorDocSpec.scala
+++ b/docs/src/test/scala/docs/actor/ActorDocSpec.scala
@@ -345,7 +345,7 @@ class ActorDocSpec extends PekkoSpec("""
       }
       // #import-context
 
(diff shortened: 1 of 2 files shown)


docs/src/test/scala/docs/actor/ActorDocSpec.scala
```diff
@@ -345,7 +345,7 @@ class ActorDocSpec extends PekkoSpec("""
       }
```
He-Pin, line 345: There is no argument needed for `FirstActor`

He-Pin:
I think this is a dotty issue

PR #1088, opened 2024-01-31
Get avoid snapshot interval calculate int overflow

Roiocam:
## Motivation

When the user uses `Integer.MAX_VALUE` as keepNSnapshots, it will touch integer overflow, and persistence will never delete those events.

This change will not change anything besides avoid overflow, because the persistence actor version already designed as a long type.

This screenshot was  happened in the real world:

image

### DIFF
--- a/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/RetentionCriteriaImpl.scala
+++ b/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/RetentionCriteriaImpl.scala
@@ -36,14 +36,14 @@ import pekko.persistence.typed.scaladsl
 
   def deleteUpperSequenceNr(lastSequenceNr: Long): Long = {
     // Delete old events, retain the latest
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
Lgtm

pjfanning (approved):
Lgtm. I think this is a bug fix and should ideally be backported to 1.0.x branch too.

He-Pin:
@Roiocam  just saw this in production @pjfanning

He-Pin:
@Roiocam Would you like to cherry pick this to 1.0.x ? Thanks.

Roiocam:
> @Roiocam Probably hard to test but could you consider if it's possible?

Yes, the reproduction can be implemented on `persistence-plugin`, I will pick it up and check it out if any possible to implement on pekko-persistence

He-Pin:
There is no need to write a plugin for just test this...

Roiocam:
> There is no need to write a plugin for just test this...

I mean verified in such as  pekko-persistence-jdbc

Issue #1089, opened 2024-01-31
Virtual Thread Support: replace synchronized with ReentrantGuard


He-Pin:
Motivation:
As there is onging work to make VT works well with monitor，not sure when , maybe in JDK 25, but we can do better to remove some `synchronized` block.

According to https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9:

>A current limitation of the implementation of virtual threads is that performing a blocking operation while inside a synchronized block or method causes the JDK's virtual thread scheduler to block a precious OS thread, whereas it wouldn't if the blocking operation were done outside of a synchronized block or method. We call that situation "pinning".

Pekko already has org.apache.pekko.util.ReentrantGuard (extends Java's ReentrantLock) which has a withGuard function.
```
   val lock = new ReentrantGuard()

   def fun(): T = {
     lock.withGuard {
       // ... function body returning an instance of T
     }
   }
```

ReentrantGuard.withGuard calls ReentrantLock lock() and waits until the lock is available and when it completes (successfully or not), it calls ReentrantLock unlock().

Running the JVM with `-Djdk.tracePinnedThread=full` is a useful way to be notified about pinned threads.

pjfanning:
I think this should be a discussion. I think we have 1 open already. An issue needs a reasonably concrete set of steps. Lazy Val's used synchronization under the hood. So finding a non synchronized alternative is a good place to start.

He-Pin:
refs: https://github.com/vavr-io/vavr/issues/2760 , yes I just found this.

pjfanning:
@He-Pin can we rework this issue to focus on replacing the use of `this.synchronized` with ReentrantLocks? That is a more achievable issue than 'Virtual thread friendly'. Someone would be more inclined to contribute the ReentrantLock solution.

He-Pin:
@pjfanning YES, PLEASE.

KrnSaurabh:
Hi @He-Pin , I would be happy to contribute. Please let me know this issue can be assigned to me if not picked up by someone already?

pjfanning:
@KrnSaurabh I don't think we want this change at the moment. ReentrantGuards are slower than synchronized blocks.

He-Pin:
@KrnSaurabh In JDK24, this is been fixed, so I think we can just drop this.

PR #1092, opened 2024-02-01
use ReentrantLock in LockUtil (Switch)

pjfanning:
an example change for #1089

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/LockUtil.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/LockUtil.scala
@@ -30,8 +30,9 @@ final class ReentrantGuard extends ReentrantLock {
  */
 class Switch(startAsOn: Boolean = false) {
   private val switch = new AtomicBoolean(startAsOn) // FIXME switch to AQS
+  private val lock = new ReentrantGuard()
 
-  protected def transcend(from: Boolean, action: => Unit): Boolean = synchronized {
+  protected def transcend(from: Boolean, action: => Unit): Boolean = lock.withGuard {
     if (switch.compareAndSet(from, !from)) {
       try action
       catch {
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm

laglangyue:
lgtm

pjfanning:
I won't merge this yet but I'd appreciate some feedback. With Loom in mind, there doesn't appear to be a big reason not to start changing synchronized blocks to ReentrantLocks. There is a fair amount of documentation online encouraging this change.

He-Pin:
I think we should only do this where io is involved.

The only issue of the current implementation is it introduce some allocation with the by name parameter.

pjfanning:
I open to abandoning this. But if we abandon this, I think that #1089 should be be changed to not be a 'good first issue'. It isn't a good first issue if it needs serious expertise to decide on which 'synchronized' blocks need to be changed.

To flip this on it's head, maybe we need to get people to use Pekko with virtual threads and to report to us the pinned thread issues that they hit.

He-Pin:
I will back my hometown tomorrow, I will try to work on the virtual thread support.

mdedetrich:
> I'm open to abandoning this. But if we abandon this, I think that #1089 should be changed to not be a 'good first issue'. It isn't a good first issue if it needs serious expertise to decide on which 'synchronized' blocks need to be changed.
> 
> To flip this on it's head, maybe we need to get people to use Pekko with virtual threads and to report to us the pinned thread issues that they hit.

Completely agree with this, `VirtualThread` support is not a good first issue not only due to it deceptively easy but also because since its a JDK 21 feature as is well known we are having issues on whats the best way to integrate this into Pekko.

I am still heavily in the "lets just document how JDK 21 users can create their own VirtualThreadPool" camp rather than pushing these kind of changes in.

pjfanning:
@mdedetrich this change does not use any new classes - it uses existing classes. `synchronized` has not been improved in Java 21 but locks like ReentrantLock have been. ReentrantLock does not block the carrier thread and other virtual threads can use the carrier thread while the lock is waiting. `synchronized` does block the carrier thread and other virtual threads will not get a chance to use the carrier thread during the synchronized blocking is waiting.

See https://softwaremill.com/what-is-blocking-in-loom/#retrofitting-blocking

mdedetrich:
> @mdedetrich this change does not use any new classes - it uses existing classes. `synchronized` has not been improved in Java 21 but locks like ReentrantLock have been. ReentrantLock does not block the carrier thread and other virtual threads can use the carrier thread while the lock is waiting. `synchronized` does block the carrier thread and other virtual threads will not get a chance to use the carrier thread during the synchronized blocking is waiting (known as pinning).
> 
> See https://softwaremill.com/what-is-blocking-in-loom/#retrofitting-blocking

I wasn't talking about this PR but the other one with VirtualThreadPool.

He-Pin:
The class is gone

pjfanning:
there are fixes going into Java that makes synchronized work better with virtual threads than it did when virtual threads first came out

PR #1095, opened 2024-02-01
don't hardcode exact coursier version so that we can pick up patches

pjfanning:
want to upgrade node version - see https://github.com/coursier/cache-action/releases/tag/v6.4.5


mdedetrich (approved):
lgtm

PR #1096, opened 2024-02-03
additional mailbox selector for typed props

Roiocam:
## Motivation

I have experience in migrating Classic to Typed. In the process, I found that Typed can't customize Mailbox through Props. We should match the same Classic method on Typed.

截屏2024-02-03 17 03 39

## TODO list
- [x] Implementation: Reuse existing classes
- [x] Mention in documentation
- [x] Verify via unit test

PR #1100, opened 2024-02-05
Configuration typo fixes

Roiocam:
Separate PR and cherry-pick from #1096, and solve all module typos.

### DIFF
--- a/cluster-sharding/src/main/resources/reference.conf
+++ b/cluster-sharding/src/main/resources/reference.conf
@@ -22,7 +22,7 @@ pekko.cluster.sharding {
 
   # When this is set to 'on' the active entity actors will automatically be restarted
   # upon Shard restart. i.e. if the Shard is started on a different ShardRegion
-  # due to rebalanced or crash.
+  # due to rebalance or crash.
   remember-entities = off
 
   # When 'remember-entities' is enabled and the state store mode is ddata this controls
--- a/cluster-sharding/src/main/resources/reference.conf
+++ b/cluster-sharding/src/main/resources/reference.conf
@@ -22,7 +22,7 @@ pekko.cluster.sharding {
 
   # When this is set to 'on' the active entity actors will automatically be restarted
   # upon Shard restart. i.e. if the Shard is started on a different ShardRegion
-  # due to rebalance or crash.
+  # due to rebalanced or crash.
   remember-entities = off
 
   # When 'remember-entities' is enabled and the state store mode is ddata this controls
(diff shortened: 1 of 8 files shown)


cluster-sharding/src/main/resources/reference.conf
```diff
@@ -22,7 +22,7 @@ pekko.cluster.sharding {
 
   # When this is set to 'on' the active entity actors will automatically be restarted
   # upon Shard restart. i.e. if the Shard is started on a different ShardRegion
-  # due to rebalance or crash.
+  # due to rebalanced or crash.
```
pjfanning, line 25: better to say `due to a rebalance event or a crash`

cluster/src/main/resources/reference.conf
```diff
@@ -233,7 +233,7 @@ pekko {
     multi-data-center {
 
       # Defines which data center this node belongs to. It is typically used to make islands of the
-      # cluster that are colocated. This can be used to make the cluster aware that it is running
+      # cluster that are collocated. This can be used to make the cluster aware that it is running
```
pjfanning, line 236: the change is not wrong - should be colocated or co-located

remote/src/main/resources/reference.conf
```diff
@@ -492,7 +492,7 @@ pekko {
         applied-adapters = []
 
         # The default remote server port clients should connect to.
-        # Default is 7355 (PEKK on a telephone keypad), use 0 if you want a random available port
```
pjfanning, line 495: 7355 is PEKK - 4 digits, 4 letters - so can you undo this change?

remote/src/main/resources/reference.conf
```diff
@@ -852,7 +852,7 @@ pekko {
         # collected, which is not as efficient as reusing buffers in the pool.
         large-buffer-pool-size = 32
 
-        # For enabling testing features, such as blackhole in pekko-remote-testkit.
+        # For enabling testing features, such as black-hole in pekko-remote-testkit.
```
pjfanning, line 855: blackhole is correct, don't change this

remote/src/main/resources/reference.conf
Roiocam: Thanks, I forgot the previous discussion of pekko default listener port.

remote/src/main/resources/reference.conf
Roiocam: Yes, but I think the `black-hole` may clear that `blackhole`, Never mind, let we keep `blackhole`.

pjfanning (approved):
lgtm

PR #1107, opened 2024-02-09
run prValidation on Java 11

raboof:
As that's what we're using to release as well


raboof:
Moving this to 'draft' since this change would increase the chance that we inadvertently accept a change that would break developing Pekko on Java 8. Discussion around this is ongoing at https://lists.apache.org/thread/ro3rz6s8mvj7j9dp39oszf7rnr6hn19d

He-Pin:
Java 17 now.

PR #1112, opened 2024-02-10
add default for pekko.remote.akka.version

pjfanning:
This property is only needed if you have pekko.remote.accept-protocol-names set up to include "akka". This is the case in the test config for pekko-remote and we get a lot of logging about pekko.cluster.akka.version as a result.

Relates to #765 and can only be merged to 1.0.x because #765 is only merged to that branch.

spotted in https://github.com/apache/incubator-pekko/actions/runs/7854077720 - we don't run the nightly tests on 1.0.x branch (and we should probably fix that too)

As part of the change, I have renamed the config to pekko.remote.akka.version so that the config is with the other new configs.

### DIFF
--- a/cluster/src/main/scala/org/apache/pekko/cluster/SeedNodeProcess.scala
+++ b/cluster/src/main/scala/org/apache/pekko/cluster/SeedNodeProcess.scala
@@ -57,10 +57,8 @@ private[cluster] abstract class SeedNodeProcess(joinConfigCompatChecker: JoinCon
     val cfg = context.system.settings.config
     if (cfg.hasPath("akka.version")) {
       cfg.getString("akka.version")
-    } else if (cfg.hasPath("pekko.cluster.akka.version")) {
-      cfg.getString("pekko.cluster.akka.version")
     } else {
-      "2.6.21"
+      cfg.getString("pekko.remote.akka.version")
     }
   }
 
--- a/cluster/src/main/scala/org/apache/pekko/cluster/SeedNodeProcess.scala
+++ b/cluster/src/main/scala/org/apache/pekko/cluster/SeedNodeProcess.scala
@@ -57,8 +57,10 @@ private[cluster] abstract class SeedNodeProcess(joinConfigCompatChecker: JoinCon
     val cfg = context.system.settings.config
     if (cfg.hasPath("akka.version")) {
       cfg.getString("akka.version")
-    } else {
+    } else if (cfg.hasPath("pekko.cluster.akka.version")) {
       cfg.getString("pekko.cluster.akka.version")
+    } else {
+      "2.6.21"
     }
   }
 
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -185,9 +185,12 @@ pekko {
     warn-unsafe-watch-outside-cluster = on
 
     # When receiving requests from other remote actors, what are the valid
-    # prefix's to check against. Useful for when dealing with rolling cluster
+    # prefixes to check against. Useful for when dealing with rolling cluster
     # migrations with compatible systems such as Lightbend's Akka.
-    accept-protocol-names = ["pekko", "akka"]
+    # By default, we only support "pekko" protocol.
+    # If you want to also support Akka, change this config to:
+    # pekko.remote.accept-protocol-names = ["pekko", "akka"]
+    accept-protocol-names = ["pekko"]
 
     # The protocol name to use when sending requests to other remote actors.
     # Useful when dealing with rolling migration, i.e. temporarily change
@@ -197,6 +200,12 @@ pekko {
     # Pekko and then change the protocol-name back to "pekko" once all
     # nodes have been are running on Apache Pekko
     protocol-name = "pekko"
+
+    # When pekko.remote.accept-protocol-names contains "akka", then we
+    # need to know the Akka version. If you include the Akka jars on the classpath,
+    # we can use the akka.version from their configuration. This configuration
+    # setting is only used if we cabn't find an akka.version setting.
+    akka.version = "2.6.21"
 
     # Settings for the Phi accrual failure detector (http://www.jaist.ac.jp/~defago/files/pdf/IS_RR_2004_010.pdf
     # [Hayashibara et al]) used for remote death watch.
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -204,7 +204,7 @@ pekko {
     # When pekko.remote.accept-protocol-names contains "akka", then we
     # need to know the Akka version. If you include the Akka jars on the classpath,
     # we can use the akka.version from their configuration. This configuration
-    # setting is only used if we cabn't find an akka.version setting.
+    # setting is only used if we can't find an akka.version setting.
     akka.version = "2.6.21"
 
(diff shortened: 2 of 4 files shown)


cluster/src/main/scala/org/apache/pekko/cluster/SeedNodeProcess.scala
```diff
@@ -57,8 +57,10 @@ private[cluster] abstract class SeedNodeProcess(joinConfigCompatChecker: JoinCon
     val cfg = context.system.settings.config
     if (cfg.hasPath("akka.version")) {
       cfg.getString("akka.version")
-    } else {
+    } else if (cfg.hasPath("pekko.cluster.akka.version")) {
       cfg.getString("pekko.cluster.akka.version")
+    } else {
+      "2.6.21"
```
mdedetrich, line 63: Should this be configurable as well or is that taking things too far?

pjfanning: I could set it in pekko-remote test conf too. Should I just then throw an exception if it isn't set. This code path only happens when pekko.remote.accept-protocol-names is set up to include "akka".

pjfanning: I definitely don't think another config setting improves this.

pjfanning: actually - I think this could be worse than I thought

This is in our main pekko-remote reference.conf. I don't think this should be in the reference.conf - fine for tests but not set up as our actual default. I think users should enable it in their confs.
```
    # When receiving requests from other remote actors, what are the valid
    # prefix's to check against. Useful for when dealing with rolling cluster
    # migrations with compatible systems such as Lightbend's Akka.
    accept-protocol-names = ["pekko", "akka"]
```

mdedetrich: I guess fundamentally to me it just feels weird that we have a hardcoded akka constant version in the source code considering that the general premise behind all of these "pekko spoofing itself so it looks like akka" values are configurable via typesafe config.

mdedetrich: > This is in our main pekko-remote reference.conf. I don't think this should be in the reference.conf - fine for tests but not set up as our actual default. I think users should enable it in their confs.
> 
> ```
>     # When receiving requests from other remote actors, what are the valid
>     # prefix's to check against. Useful for when dealing with rolling cluster
>     # migrations with compatible systems such as Lightbend's Akka.
>     accept-protocol-names = ["pekko", "akka"]
> ```

Whats the problem with having this as a default, seems sensible to me or am I missing something wild?

pjfanning: @mdedetrich I might be able to remove this PR if we can agree to change pekko-remote main reference.conf to

```
    # When receiving requests from other remote actors, what are the valid
    # prefixes to check against. Useful for when dealing with rolling cluster
    # migrations with compatible systems such as Lightbend's Akka.
    # By default, we only support "pekko" protocol. If you want to also support Akka, change this config to:
    # pekko.remote.accept-protocol-names = ["pekko", "akka"]
    accept-protocol-names = ["pekko"]
```

pjfanning: > > This is in our main pekko-remote reference.conf. I don't think this should be in the reference.conf - fine for tests but not set up as our actual default. I think users should enable it in their confs.
> > ```
> >     # When receiving requests from other remote actors, what are the valid
> >     # prefix's to check against. Useful for when dealing with rolling cluster
> >     # migrations with compatible systems such as Lightbend's Akka.
> >     accept-protocol-names = ["pekko", "akka"]
> > ```
> 
> Whats the problem with having this as a default, seems sensible to me or am I missing something wild? Or are you suggesting just having `accept-protocol-names = ["pekko"]` (which is something I can get behind) but it still doesn't solve the akka version string issue which is a bit of a gotcha (at minimum it should be documented).

The problem is performance for one. When you allow Akka, you get all the new logic that I enabled to change the compat messages to include `akka` data as well as `pekko` data. This is not free.

mdedetrich: > @mdedetrich I might be able to remove this PR if we can agree to change pekko-remote main reference.conf to
> 
> ```
>     # When receiving requests from other remote actors, what are the valid
>     # prefixes to check against. Useful for when dealing with rolling cluster
>     # migrations with compatible systems such as Lightbend's Akka.
>     # By default, we only support "pekko" protocol.
>     # If you want to also support Akka, change this config to:
>     # pekko.remote.accept-protocol-names = ["pekko", "akka"]
>     accept-protocol-names = ["pekko"]
> ```

Thats fine with me, but if we document that you need to change it to `accept-protocol-names = ["akka", "pekko"]` for the migration scenario we then hit the akka version gotcha, so what should we do there? Should we still add a default for `pekko.cluster.akka.version` even if its only going to be used in the migration scenario?

pjfanning: Sure - I can add `pekko.cluster.akka.version` to the pekko-remote reference.conf, instead of hardcoding the default in the code.

mdedetrich:
Looks much better now, let me know when PR is out of draft and ill review it.

remote/src/main/resources/reference.conf
```diff
@@ -198,6 +201,12 @@ pekko {
     # nodes have been are running on Apache Pekko
     protocol-name = "pekko"
 
+    # When pekko.remote.accept-protocol-names contains "akka", then we
+    # need to know the Akka version. If you include the Akka jars on the classpath,
+    # we can use the akka.version from their configuration. This configuration
+    # setting is only used if we can't find an akka.version setting.
+    akka.version = "2.6.21"
```
mdedetrich, line 208: This might be over-engineering way too much but considering you could put any string values into `accept-protocol-names` it might make sense for this configuration to be an object i.e.

```
version {
  akka = "2.6.21"
}
```
where the key (in this case `akka`) is pointing to the the same value inside of ` accept-protocol-names`.

Otherwise we should do some sanity validation on `accept-protocol-names` and refuse any values that aren't `akka` or `pekko`

pjfanning: I think this is overengineering. The new code I added is very Akka specific (#765). It only kicks if you include "akka" in `accept-protocol-names` (again hardcoded). In the real world, I do not know of anyone else who would need to add a non-Pekko node to a Pekko cluster. If that story ever comes to pass, I will volunteer to help those users to get the cluster compatibility checks to work with their nodes.

mdedetrich: Can we then just add a validation check to the config parsing so that only `akka` and `pekko` can be put in as values (can also be separate PR)?

mdedetrich (approved):
lgtm, might be best to get another reviewer just in case

samueleresca (approved):
Is there already an issue for the missing nightly build schedule in `1.0.x`? Otherwise, looks good to me

pjfanning:
> Is there already an issue for the missing nightly build schedule in `1.0.x`? Otherwise, looks good to me

I logged https://github.com/apache/incubator-pekko/issues/1119

Issue #1119, opened 2024-02-12
set up scheduled nightly build for 1.0.x branch


pjfanning:
* we could add a yml file in main branch that checks out `ref: 1.0.x` - we do this for a few jobs already
* see top answer on https://stackoverflow.com/questions/58798886/github-actions-schedule-operation-on-branch as to why we can't use the existing https://github.com/apache/incubator-pekko/blob/main/.github/workflows/nightly-builds.yml - its schedule applies only to the main branch
* we could possibly use https://github.com/apache/incubator-pekko/blob/main/.github/workflows/nightly-builds.yml but add a matrix element called `checkoutBranch` set to `[main, 1.0.x]` and change the checkout steps to apply `ref: ${{ matrix.checkoutBranch }}`
* in mean time, you can still run the tests manually at https://github.com/apache/incubator-pekko/actions/workflows/nightly-builds.yml and choose the `1.0.x` branch

pjfanning:
this is done

PR #1120, opened 2024-02-12
Create nightly-1.0-builds.yml

pjfanning:
see #1119 

I prefer keeping this yml separate so that we can tailor what we want to run

### DIFF
--- a/.github/workflows/nightly-1.0-builds.yml
+++ b/.github/workflows/nightly-1.0-builds.yml
@@ -1,0 +1,142 @@
+name: Nightly Builds (1.0)
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  pekko-cluster-metrics-sigar:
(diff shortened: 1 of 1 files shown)


raboof (approved):

.github/workflows/nightly-1.0-builds.yml
```diff
@@ -0,0 +1,142 @@
+name: Nightly Builds (1.0)
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+permissions: {}
```
raboof, line 8: I'm curious what this does?

pjfanning: it's an attempt to limit the permissions for the GitHub token - this setting basically says we don't need it at all

raboof: right, so the default permissions are wider than `{}` (https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) :+1: great!

PR #1121, opened 2024-02-12
remove unsupported feature from 1.0 build

pjfanning:
The build added in #1120 includes items from the main branch nightly build. The `sbt dependWalkerCheck` check is only something that was added on main branch so it causes the 1.0 build to break.


mdedetrich (approved):
lgtm

PR #1128, opened 2024-02-15
in TlsActor flush data to user only after handshake has finished

pjfanning:
relates to https://github.com/apache/incubator-pekko-http/issues/485


mdedetrich (approved):
lgtm @pjfanning I assume we don't need to handle the `case Failure(ex) =>` case because `fail(ex, closeTransport = true)` already does the flushing?

raboof (approved):

raboof:
> Another question is should we backport this to 1.0.x?

It's a bugfix for a somewhat rarely-used feature, so it seems fine to expect people to upgrade to get it

PR #1130, opened 2024-02-17
fix numbering in statefulMap.md

pjfanning:


docs/src/main/paradox/stream/operators/Source-or-Flow/statefulMap.md
```diff
@@ -45,15 +45,15 @@ Scala
 Java
 :   @@snip [StatefulMap.java](/docs/src/test/java/jdocs/stream/operators/flow/StatefulMap.java) { #bufferUntilChanged }
 
-In the forth example, repeated incoming elements are only emitted once and then dropped.
+In the third example, repeated incoming elements are only emitted once and then dropped.
```
pjfanning, line 48: if there ever was a third example before, it has been removed

samueleresca (approved):

PR #1134, opened 2024-02-18
Remove unnecessary scalafmt ignore

mdedetrich:
So looking at the original commit at https://github.com/apache/incubator-pekko/commit/949ee7ea21934459d6240f60277bf0854eeff158, I don't think whatever the original reason is applicable anymore (unlike Akka, Pekko uses a modern scalafmt which scala-steward should have issues with).

As they say, what happens in Athens stays in Athens

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -5,8 +5,6 @@ updates.pin = [
 ]
 
 updates.ignore = [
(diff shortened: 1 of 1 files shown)


jxnu-liguobin (approved):

raboof (approved):

Roiocam (approved):
LGTM

PR #1135, opened 2024-02-21
Use sbt-web 1.5.x despite staying on paradox 0.9.2

mkurz:
I am pretty sure this is safe since I am basically the on who maintains sbt-web and v1.5 is pretty much compatible with v1.4 API wise.
I use the same workaround in https://github.com/sbt/sbt-paradox-material-theme/blob/v0.7.0/project/plugins.sbt#L1-L3 actually

With this change I can `+Test/compile` pekko and I am not hitting https://repo.scala-sbt.org/ anymore (since I have blocked it on my machine).

@mdedetrich I am pretty sure this is ok to merge.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -37,6 +37,8 @@ addSbtPlugin(("org.apache.pekko" % "pekko-sbt-paradox" % "1.0.0").excludeAll(
   "com.lightbend.paradox", "sbt-paradox",
   "com.lightbend.paradox" % "sbt-paradox-apidoc",
   "com.lightbend.paradox" % "sbt-paradox-project-info"))
-addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox" % "0.9.2").force())
-addSbtPlugin(("com.lightbend.paradox" % "sbt-paradox-apidoc" % "0.10.1").force())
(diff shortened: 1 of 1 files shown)


mdedetrich:
Running CI just to make sure but I am not expecting any surprises here

mkurz:
Oops sorry fixed formatting

raboof (approved):

pjfanning:
https://github.com/apache/incubator-pekko-http/pull/486 has an exclude on sbt-paradox 0.9.2 but this doesn't.

My preference is to try out the changes in https://github.com/apache/incubator-pekko-http and then when we are happy in that repo, we can copy the full set of changes to our other repos.

mdedetrich:
> [apache/incubator-pekko-http#486](https://github.com/apache/incubator-pekko-http/pull/486) has an exclude on sbt-paradox 0.9.2 but this doesn't.

This exclude is no longer necessary on the latest version of pekko-sbt-paradox because of this change https://github.com/apache/incubator-pekko-sbt-paradox/pull/97. The exception here is if some other library brings in a later version of sbt-paradox/sbt-paradox-theme that overrides the 0.9.2 version which is JDK compatible.

PR #1137, opened 2024-02-21
Rework PhiAccrualFailureDetector to enable monitoring of interval

### LINKED ISSUE #1136: interceptable heartbeat interval to record metric
`PhiAccrualFailureDetector` is not suitable to extend or use intercept to get access to interval for monitoring hearbeat

nvollmar:
resolves #1136


danischroeter (approved):
lgtm

pjfanning:
Would it be possible to write a fuller description to the issue? This is a non-trivial change to our code and we need to be able to release note it.

nvollmar:
@pjfanning I can add more to the issue. But I’d consider this a rather trivial change. 

There is no functional difference, the implementation is now open for Kanela to intercept the call to the new method to report that metric to Kamon. Or to extend this class and overwrite the method to get the interval.

Also the new trait makes it possible to implement a custom failure detector with the same functionality.

mdedetrich (approved):
lgtm, as far as I can tell this is just shuffling code around so certain parts of it can be overridden (i.e. enabling features like monitoring interval).

remote/src/main/scala/org/apache/pekko/remote/PhiAccrualFailureDetector.scala
```diff
@@ -153,16 +152,10 @@ class PhiAccrualFailureDetector(
...
-            eventStream.get.publish(
-              Warning(
-                this.toString,
-                getClass,
-                s"heartbeat interval is growing too large for address $address: $interval millis",
-                Logging.emptyMDC,
-                RemoteLogMarker.failureDetectorGrowing(address)))
-          oldState.history :+ interval
-        } else oldState.history
+          recordInterval(interval)
```
mdedetrich, line 155: nit: Can we remove the unnecessary `{` and `}` in the `if`/`else` statement so its closer to the older style.

nvollmar: will change it back

nvollmar:
Looks like link validator is broken because the url changed from `opensource.org/licenses/BSD-3-Clause` to `opensource.org/blog/license/bsd-3-clause`

mdedetrich:
> Looks like link validator is broken because the url changed from `opensource.org/licenses/BSD-3-Clause` to `opensource.org/blog/license/bsd-3-clause`

This is unrelated to this PR so you can file an issue or fix it in another PR

remote/src/main/scala/org/apache/pekko/remote/FailureDetector.scala
```diff
@@ -40,6 +40,16 @@ trait FailureDetector {
 
 }
 
+trait FailureDetectorWithAddress {
+
+  /**
+   * Address of observed host will be set after detector creation.
+   */
+  def address_=(addr: String): Unit
```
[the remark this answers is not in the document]
jxnu-liguobin: Indeed, but it's OK and convenient   in scala.

He-Pin (commented):
If the usercase is for aop,then we should better have javadsl in mind too.

nvollmar:
The use case would be to intercept or override the call to `recordInterval`

remote/src/main/scala/org/apache/pekko/remote/FailureDetector.scala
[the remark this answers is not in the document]
nvollmar: updated it to make it more java friendly

He-Pin (approved):
Lgtm

nvollmar:
This would be ready to merge unless there are some objections/further remarks

Roiocam (approved):
LGTM

remote/src/main/scala/org/apache/pekko/remote/PhiAccrualFailureDetector.scala
```diff
@@ -119,7 +117,8 @@ class PhiAccrualFailureDetector(
   private val acceptableHeartbeatPauseMillis = acceptableHeartbeatPause.toMillis
 
   // address below was introduced as a var because of binary compatibility constraints
```
mdedetrich, line 119: Can we prefix this comment with `NOTE` so that its easily searchable. Should also make a pekko issue to make this a non var targeting Pekko 2.0.x

mdedetrich, line 119: Can we prefix this comment with `NOTE` so that its easily searchable. Should also make a pekko issue to make this a non var targeting Pekko 2.0.x

remote/src/main/scala/org/apache/pekko/remote/PhiAccrualFailureDetector.scala
```diff
@@ -153,15 +151,7 @@ class PhiAccrualFailureDetector(
         val interval = timestamp - latestTimestamp
         // don't use the first heartbeat after failure for the history, since a long pause will skew the stats
         if (isAvailable(timestamp)) {
```
mdedetrich, line 153: nit: Still a redundant single bracket here, i.e. should be

```scala
if (isAvailable(timestamp))
  recordInterval(interval)
else
  oldState.history
```

You can even put it on a single line

mdedetrich (changes requested):
Just a minor thing, want to add a `NOTE` prefix to comment so that it appears in IDE"s then I will approve

remote/src/main/scala/org/apache/pekko/remote/PhiAccrualFailureDetector.scala
nvollmar: I removed the redundant bracket, since it isn't an assignment I'd leave it on different lines for clarity

remote/src/main/scala/org/apache/pekko/remote/PhiAccrualFailureDetector.scala
nvollmar: I'll create an issue for Pekko 2.0

nvollmar: #1148

mdedetrich (approved):
lgtm, just added a commit to capitalize the note

PR #1149, opened 2024-02-26
Backport heartbeat monitoring

nvollmar:
backporting #1137


pjfanning:
I'm not sure this can be backported like this. It has a MIMA issue so that's something that we should consider carefully before merging. If we do choose to merge this, the MIMA folder name would need to change (it work need to be a 1.0 exclude).

The only thing really blocking a 1.1 milestone now is getting the release notes together.

PR #1150, opened 2024-02-26
in TlsActor flush data to user only after handshake has finished (#1128)

pjfanning:
cherry pick 1e41829bf7abeec268b9a409f35051ed7f4e0090 #1128

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/io/TLSActor.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/io/TLSActor.scala
@@ -421,7 +421,6 @@ import pekko.util.ByteString
       case OK =>
         result.getHandshakeStatus match {
           case NEED_WRAP =>
-            flushToUser()
             // https://github.com/apache/incubator-pekko/issues/442
             // A second workaround for an infinite loop we have not been able to reproduce/isolate,
             // if you see this, and can reproduce consistently, please report back to the Apache Pekko team
@@ -481,6 +480,7 @@ import pekko.util.ByteString
       case Success(()) =>
         currentSession = session
         corkUser = false
+        flushToUser()
       case Failure(ex) =>
         fail(ex, closeTransport = true)
     }


raboof (approved):
I'm not sure this needs backporting, it seems pretty obscure and not have security impact, but I'm not opposed either.

mdedetrich:
Im also on the fence on this one, people may be relying on this behaviour even though its "wrong" and I am not sure I would categorize this as critical

raboof:
> Im also on the fence on this one, people may be relying on this behaviour even though its "wrong"

(that risk seems relatively low to me)

> and I am not sure I would categorize this as critical

agree

mdedetrich:
> (that risk seems relatively low to me)

Oh I agree its quite low, but the whole point of the 1.0.x branch is to keep the behaviour as close as possible to original Akka 2.6.x to minimize surprises when people migrate (hence why no changes unless its security or critical)

@pjfanning wdyt ?

mdedetrich:
We can always decide to backport this later if someone brings it up but its always been the intention that the 1.0.x is just a step up to the 1.1.x branch which contains all the improvements/fixes etc etc

TjarkoG:
Hi
Will 1.1.0 be released soon? otherwise i would ask if this could be reopened and merged, since it allowes the 
fix for https://github.com/apache/pekko/issues/442 to actually work. Without it the flushToUser function resets the workaround counter on every loop before we check if we should break it.
this should be the reason we're seeing the 100% cpu bug again even with the fixing commit being merged:
https://github.com/apache/pekko/issues/842

pjfanning:
@TjarkoG Have you tried v1.1.0-M1? I would hope that v1.1.0 would be released in the next month or 2.
I, personally, wouldn't be against backporting the `case NEED_WRAP` part of this PR's changes to 1.0.x but I would need to get that approved by other Pekko team members.

mdedetrich:
Indeed 1.1.0-M1 has been released so you should be able to use that.

I am on the fence about backporting this, @raboof have your thoughts changed on this?

TjarkoG:
is the 1.1.0-M1 production ready?
i thought that the Milestone-releases should not be used in production.
we have thus not yet used the 1.1.0-M1 release.

raboof:
> I am on the fence about backporting this, @raboof have your thoughts changed on this?

My thoughts haven't changed: I'm not opposed to backporting, but ideally I'd rather see us put effort into releasing v1.1.0 than into backporting.

> is the 1.1.0-M1 production ready? i thought that the Milestone-releases should not be used in production. we have thus not yet used the 1.1.0-M1 release.

Indeed we're not yet recommending 1.1.0-M1 for general purpose production use (https://github.com/apache/pekko/blob/main/docs/src/main/paradox/release-notes/releases-1.1.md#110-m1 - it seems we haven't published that yet?). However, if have any capacity, if you could help by running 1.1.0-M1 and confirming whether it fixes the issue for you (and generally appears to work well), that might speed up how quickly we feel confident to release 1.1.0 .

TjarkoG:
i'm assuming that we stick with semantic versioning and 1.1.0 does not have breaking changes compared to 1.0.x?
we'll test it in some of our less critical applications and i'll give feedback if everything works as expected.
might take a little since the bug is not reproduceable and occurs randomly.

raboof:
> i'm assuming that we stick with semantic versioning and 1.1.0 does not have breaking changes compared to 1.0.x?

We've documented our compatibility rules at https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html - indeed if you weren't using experimental or deprecated APIs on 1.0.x I don't think you should expect anything to break on 1.1.x. Of course testing before going in production is still advised :)

> we'll test it in some of our less critical applications and i'll give feedback if everything works as expected. might take a little since the bug is not reproduceable and occurs randomly.

That'd be much appreciated!

TjarkoG:
sadly we're currently blocked by an error within kamon instrumentation when updating to 1.1.0-M1
https://github.com/kamon-io/Kamon/issues/1352
once this is fixed we'll get back

PR #1156, opened 2024-02-27
fix: proper path when promise actor terminated quickly

Roiocam:
solve #1157 
self-explanatory

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/pattern/AskSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/pattern/AskSpec.scala
@@ -266,6 +266,29 @@ class AskSpec extends PekkoSpec("""
       val (promiseActorRefForSelection, "ask") = p.expectMsgType[(ActorRef, String)]: @unchecked
       promiseActorRefForSelection.path.name should startWith("_user_myName")
     }
+
+    "proper path when promise actor terminated" in {
+      implicit val timeout: Timeout = Timeout(300 millis)
+      val p = TestProbe()
+
+      val act = system.actorOf(Props(new Actor {
+          def receive = {
+            case _ =>
+              val senderRef: ActorRef = sender()
+              senderRef ! "complete"
+              p.ref ! senderRef
+          }
+        }), "pathPrefix")
+
+      val f = (act ? "ask").mapTo[String]
+      val promiseActorRef = p.expectMsgType[ActorRef]: @unchecked
+
+      // verify ask complete
+      val completed = f.futureValue
+      completed should ===("complete")
+      // verify path was proper
+      promiseActorRef.path.toString should include("pathPrefix")
+    }
   }
 
 }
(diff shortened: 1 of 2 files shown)


laglangyue:
LGTM. 
I only use Pekko streams. I have read the  `path` method, some code of the class and  the related test cases. 
I am not an expert in Pekko core, so hope more LGTM.

He-Pin:
Better with a bug report here too.

raboof (approved):
Looks reasonable to me!

jxnu-liguobin (approved):

actor-tests/src/test/scala/org/apache/pekko/pattern/AskSpec.scala
```diff
@@ -266,6 +266,29 @@ class AskSpec extends PekkoSpec("""
...
+      val p = TestProbe()
+
+      val act = system.actorOf(Props(new Actor {
+          def receive = {
+            case _ =>
+              val senderRef: ActorRef = sender()
+              senderRef ! "complete"
+              p.ref ! senderRef
+          }
+        }), "pathPrefix")
```
He-Pin, line 281: For this tests, I think we would better test it several times with some generated pathPrefixs.

laglangyue: random str is ok?

Roiocam: I don't think it's necessary to verify the `pathPrefixs` several times, because this is not a flaky behavior. 

After investigation, I think it is a clear bug, and the unit test explicitly verifies this. Let me explain:

The reason why the path is correct after adding STDOUT is because `toString` is called `path`

https://github.com/apache/incubator-pekko/blob/d884540c92d0d38797c32f217107d97e15cdafbe/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/ActorRefImpl.scala#L52C5-L52C62

Which initialized the path with refPathPrefix(correctly path):

https://github.com/apache/incubator-pekko/blob/d884540c92d0d38797c32f217107d97e15cdafbe/actor/src/main/scala/org/apache/pekko/pattern/AskSupport.scala#L593-L597

But when toString is not called in the `messageFactory` function apply, all this will not happen. When the Actor terminated, the method call to the path will generate a path without refPathPrefix (that is the place of the bug, and the place where the PR is fixed)

Roiocam: This test was enough for this bug. We completed the ask future(which made PromiseActor terminated) on the Actor message handler, and then sent PromiseActorRef to the probe for saving reference.

laglangyue: for me, it's enough for the bugfix.

He-Pin (approved):
lgtm, but it would be better with some generated data other than `"pathPrefix"` in test

He-Pin:
Once this gets merged, should backport to 1.0.x too.

PR #1159, opened 2024-02-28
Update scala-steward.yml with 1.0.x branches

He-Pin:
Motivation:
Add 1.0.x to scala steward for auto-update deps.


pjfanning:
I don't think we need this. I think we should only upgrade dependencies in 1.0.x when there are serious issues. For instance, I would have voted against the Scala upgrades merged today.

pjfanning (changes requested):
One of the major points of the 1.0.0 release cycle was not to upgrade dependencies in 1.0.0 unless they were important. I think we will need a Pekko team vote on this and I will vote -1.

He-Pin:
We can pin some dependencies on 1.0.x too.

pjfanning:
> We can pin some dependencies on 1.0.x too.

As I said above, this needs a vote.

PR #1181, opened 2024-03-06
upgrade some docker-client transitive dependencies

pjfanning:
causing some dependency warnings for us - see https://github.com/apache/incubator-pekko/security/dependabot

https://github.com/docker-java/docker-java/pulls shows that they don't have much interest in dependency upgrades


He-Pin (approved):
Lgtm

raboof (approved):

Issue #1182, opened 2024-03-07
TcpDnsClient cannot recover if registration on TcpConnection times out


nvollmar:
I uncovered this investigating cluster issues on our nightly deployment test. Since we started to use a low power cpu governor during the night we started seeing issues of a Pekko cluster forming during the nightly deployment.

I've tracked it down to the `TcpDnsClient` / `TcpConnection` initialization timing out, leaving it in a state it cannot recover from and never responding to any requests.

The `TcpOutgoingConnection` is connecting and responds with a `Tcp.Connected` message to the `TcpDnsClient`, which in turn registers itself on the connection again:
https://github.com/apache/incubator-pekko/blob/46e60a61fbabce5e3f36a408bfa3d1fb249eef44/actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala#L48-L53

If that message arrives late, the `TcpOutgoingConnection` will stop itself and `TcpDnsClient` has no detection or handling for this case:

https://github.com/apache/incubator-pekko/blob/46e60a61fbabce5e3f36a408bfa3d1fb249eef44/actor/src/main/scala/org/apache/pekko/io/TcpConnection.scala#L104-L108

This is a very unusual case, but it happens almost every deployment for one or two pods when the system is in low power mode.

Proposed fix: `TcpDnsClient` must watch the connection and fail on termination to re-initialize (it is already handled by a backoff supervisor)

nvollmar:
@pjfanning Since we ran into that a couple of time now, I'd like to backport to 1.0

pjfanning:
@nvollmar sure - could you create a cherry pick PR that targets the 1.0.x branch and add that new PR to the 1.0.3 milestone?

nvollmar:
Sure, will do

PR #1183, opened 2024-03-07
Add handling for tcp register timeout leaving connection dead

nvollmar:
Refs #1182

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala
@@ -14,10 +14,9 @@
 package org.apache.pekko.io.dns.internal
 
 import java.net.InetSocketAddress
-
 import org.apache.pekko
 import pekko.PekkoException
-import pekko.actor.{ Actor, ActorLogging, ActorRef, Stash }
+import pekko.actor.{ Actor, ActorLogging, ActorRef, Stash, Terminated }
 import pekko.annotation.InternalApi
 import pekko.io.Tcp
 import pekko.io.dns.internal.DnsClient.Answer
@@ -49,6 +48,7 @@ import pekko.util.ByteString
       log.debug("Connected to TCP address [{}]", ns)
       val connection = sender()
       context.become(ready(connection))
+      context.watch(connection)
       connection ! Tcp.Register(self)
       unstashAll()
     case _: Message =>
@@ -80,7 +80,10 @@ import pekko.util.ByteString
         }
       }
     case Tcp.PeerClosed =>
+      context.unwatch(connection)
       context.become(idle)
+    case Terminated(`connection`) =>
+      throwFailure("TCP connection terminated without closing (register timeout?)", None)
   }
 
   private def parseResponse(data: ByteString) = {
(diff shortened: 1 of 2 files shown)


mdedetrich (approved):
lgtm, @pjfanning @Roiocam @He-Pin @jxnu-liguobin @raboof Should also take a look

pjfanning (approved):
lgtm

pjfanning:
maybe in some further iteration, we could look at retries - but this is an improvement as is

laglangyue:
LGTM

nvollmar:
I was thinking going the other direction and questioning the need for the `Register` message.
As far as I could see from a quick search, the `Register` is always sent by the same actor that opens the connections.

So do we need to have that capability to have a different handler? Or could we inline everything into the connection open and avoid having that timeout entirely?

actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala
```diff
@@ -80,7 +80,10 @@ import pekko.util.ByteString
         }
       }
     case Tcp.PeerClosed =>
+      context.unwatch(connection)
       context.become(idle)
+    case Terminated(`connection`) =>
+      throwFailure("TCP connection terminated without closing (register timeout?)", None)
```
Roiocam, line 86: Did you notice `TcpOutgoingConnection` doing the same thing here?

https://github.com/apache/incubator-pekko/blob/46e60a61fbabce5e3f36a408bfa3d1fb249eef44/actor/src/main/scala/org/apache/pekko/io/TcpOutgoingConnection.scala#L58-L59

It will send `CommandFailed` to `commander(TcpDnsClient)`

Roiocam (commented):
I think we may need to investigate deeper.

He-Pin:
I will TAL at weekend, a little busy at $Work

actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala
nvollmar: Not in the case of the register timeout, there it will just do `context.stop`.

actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala
```diff
@@ -49,6 +48,7 @@ import pekko.util.ByteString
       log.debug("Connected to TCP address [{}]", ns)
       val connection = sender()
       context.become(ready(connection))
+      context.watch(connection)
```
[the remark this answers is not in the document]
nvollmar: Other usages of that connection also watch it to fail on termination:
https://github.com/apache/incubator-pekko/blob/46e60a61fbabce5e3f36a408bfa3d1fb249eef44/stream/src/main/scala/org/apache/pekko/stream/impl/io/TcpStages.scala#L295-L296

That's why I did the same here for a low risk fix of the issue.
In general it would make sense to rethink the construct as a whole as the code is more or less unchanged since at least 2013.

actor/src/main/scala/org/apache/pekko/io/dns/internal/TcpDnsClient.scala
Roiocam: Ahh, I see. The connection is ready, and there are only the last two Acks left. Things from TcpOutgoingConnection to TcpConnection. 

Thanks for your explanation.

Roiocam (approved):

He-Pin (approved):
lgtm, as it was supervised by `org.apache.pekko.pattern.BackoffSupervisor`.

mdedetrich:
Got enough approvals, merging so it can get released with M1

Issue #1180, opened 2024-03-05
Improve Migration Doc layout (main branch)


pjfanning:
See https://pekko.apache.org/docs/pekko/snapshot/project/migration-guides.html

The 1.0 to 1.1 migration doc appears as a strange bullet point with a link at the end the existing 1.0 migration doc.

I would suggest that we should create a new top level migration page with a table of contents that includes the 1.0 migration doc and the 1.0 to 1.1 migration doc.

pjfanning:
Marking as high priority because it would be good to have this as part of the 1.1.0-M1 release.

naosense:
If I understand correctly,  you want to create a new top level  page,  then points to [1.0 Migration Documentation](https://pekko.apache.org/docs/pekko/snapshot/project/migration-guides.html) and [1.0 to 1.1 Migration Documentation](https://pekko.apache.org/docs/pekko/snapshot/project/migration-guide-1.0.x-1.1.x.html)?

pjfanning:
> If I understand correctly, you want to create a new top level page, then points to [1.0 Migration Documentation](https://pekko.apache.org/docs/pekko/snapshot/project/migration-guides.html) and [1.0 to 1.1 Migration Documentation](https://pekko.apache.org/docs/pekko/snapshot/project/migration-guide-1.0.x-1.1.x.html)?

Yes - I think that layout would be better than what we have in https://pekko.apache.org/docs/pekko/snapshot/project/migration-guides.html today.

pjfanning:
#1193 merged

PR #1193, opened 2024-03-15
Improve migration doc layout

naosense:
resove #1180 

hi @pjfanning Could you take a look, if this is what you want?


pjfanning:
@naosense this looks like the right approach

pjfanning (approved):
lgtm

PR #1198, opened 2024-03-15
remove unused scripts

pjfanning:
These scripts are not used.

They also cause trouble at release time because they have no license headers but they were written by the Akka team making it complicated as to whether we can put headers on the files.


He-Pin (approved):
lgtm

Issue #1205, opened 2024-03-18
SubFlow's don't correctly propagate `Supervision.resumeStrategy`


mdedetrich:
In https://github.com/apache/incubator-pekko/pull/252 I made it so that we can set `SupervisionStrategy` rather than using `SubStreamCancelStrategy`. It turns out there is an existing test which was written as pending (so its currently disabled) to test this exact `SupervisionStrategy` functionality (specifically `SupervisionStrategy.resumingDecider` when an exception is thrown) when its adding in the future (which is now) i.e.

https://github.com/apache/incubator-pekko/blob/c44c0b7cbdab11d85176cfe062288fdcba16c56a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala#L213-L262

When I try to enable the test by removing `pending` so the test runs it actually does not pass which means that the `SupervisionStrategy.resumingDecider` is not properly catching and suppressing the exception as its meant to do so.

mdedetrich:
So I set the 1.1.0-M1 milestone for this bug, will do my best to solve it by the time the release is actually cut however if its not done by then can do it in a following `-M2` or full release.

It does however need to be solved by full 1.1.0 release

gwak:
Hi @mdedetrich,

@pjfanning and I just tested the 1.1 milestone release with **[playframework](https://github.com/playframework/playframework/pull/12662)**. We followed the migration guide and explicitly defined the supervision strategy when using the `splitWhen` function like so:

```scala
.splitWhen(_.isLeft)
.withAttributes(ActorAttributes.supervisionStrategy(Supervision.resumingDecider))
.prefixAndTail(1)
```

But when using the deprecated `SubstreamCancelStrategy` it works fine:

```scala
.splitWhen(SubstreamCancelStrategy.drain)(_.isLeft)
```

Is there something we're missing ?

mdedetrich:
@gwak So I still need to work on this, in the meantime I would recommend using the deprecated method as you described.

I just skimmed through the issue to recover some context, iirc the core issue is that in general `ActorAttributes` are not propagating where as in Pekko 1.0 and/or current pekko 1.1 deprecation functionality propagates the strategy in a different/more explicit way

mdedetrich:
Re-opening this as it needs to be solved

mdedetrich:
[...]

PR #1207, opened 2024-03-18
Fix uncaught decider exception in Split with Supervision.resumingDecider

mdedetrich:
Resolves: https://github.com/apache/incubator-pekko/issues/1205

Turns out the error was quite simple, we weren't using a decider to caught the exception in `onPush`

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
@@ -211,8 +211,6 @@ class FlowSplitAfterSpec extends StreamSpec("""
     }
 
     "resume stream when splitAfter function throws" in {
-      info("Supervision is not supported fully by GraphStages yet")
-      pending
       val publisherProbeProbe = TestPublisher.manualProbe[Int]()
       val exc = TE("test")
       val publisher = Source
@@ -250,10 +248,10 @@ class FlowSplitAfterSpec extends StreamSpec("""
       upstreamSubscription.sendNext(6)
       substreamPuppet1.expectNext(6)
       substreamPuppet1.expectComplete()
+      upstreamSubscription.sendNext(7)
       val substream2 = subscriber.expectNext()
       val substreamPuppet2 = StreamPuppet(substream2.runWith(Sink.asPublisher(false)))
       substreamPuppet2.request(10)
-      upstreamSubscription.sendNext(7)
       substreamPuppet2.expectNext(7)
 
       upstreamSubscription.sendComplete()
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
@@ -636,7 +636,12 @@ import pekko.util.ccompat.JavaConverters._
               else substreamSource.push(elem)
             }
           } catch {
-            case NonFatal(ex) => onUpstreamFailure(ex)
+            case NonFatal(ex) =>
+              decider(ex) match {
+                case Supervision.Resume  => pull(in)
+                case Supervision.Stop    => onUpstreamFailure(ex)
+                case Supervision.Restart => onUpstreamFailure(ex) // TODO implement restart?
+              }
           }
         }
 
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitWhenSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitWhenSpec.scala
@@ -327,9 +327,6 @@ class FlowSplitWhenSpec extends StreamSpec("""
     }
 
     "resume stream when splitWhen function throws" in {
-      info("Supervision is not supported fully by GraphStages yet")
-      pending
-
       val publisherProbeProbe = TestPublisher.manualProbe[Int]()
       val exc = TE("test")
       val publisher = Source


stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
```diff
@@ -250,10 +248,10 @@ class FlowSplitAfterSpec extends StreamSpec("""
       upstreamSubscription.sendNext(6)
       substreamPuppet1.expectNext(6)
       substreamPuppet1.expectComplete()
+      upstreamSubscription.sendNext(7)
```
mdedetrich, line 251: It seems like this is a genuine mistake when the test was originally written (remember that this test never passed, it was written for the future when supervision strategy propagation would be implemented which occurred due to https://github.com/apache/incubator-pekko/pull/252).

Logically when you read the test, its expected that when you do `expectNext` that a previously an element should have been sent with `sendNext` and its also the same in  the equivalent [`groupBy` test](https://github.com/apache/incubator-pekko/blob/c44c0b7cbdab11d85176cfe062288fdcba16c56a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowGroupBySpec.scala#L306-L308)

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -640,7 +640,12 @@ import pekko.util.ccompat.JavaConverters._
             else substreamSource.push(elem)
           }
         } catch {
-          case NonFatal(ex) => onUpstreamFailure(ex)
+          case NonFatal(ex) =>
+            decider(ex) match {
+              case Supervision.Resume  => pull(in)
+              case Supervision.Stop    => onUpstreamFailure(ex)
+              case Supervision.Restart => onUpstreamFailure(ex) // TODO implement restart?
```
mdedetrich, line 647: This will be implemented as a result of https://github.com/apache/incubator-pekko/pull/981

mdedetrich:
@He-Pin @Roiocam @jxnu-liguobin @pjfanning @raboof Please take a note of https://github.com/apache/incubator-pekko/pull/1207#pullrequestreview-1942808650 , I might be missing something here

pjfanning (dismissed):
Lgtm

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
mdedetrich: I thought about this again, I may be wrong and the test is indeed valid as its written right now because its `splitAfter` so its its meant to be skipping the next element due to exception being thrown, in which case we have to drop one element and then do `pull(in)`?

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
He-Pin: After I took a look at the code, I think we should handle `Supervision.Restart` as the `Supervision.Resume`,be cause it happens in `SubstreamHandler`

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
He-Pin: Yes, the element which case the exception is dropped now.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
mdedetrich: > After I took a look at the code, I think we should handle `Supervision.Restart` as the `Supervision.Resume`,be cause it happens in `SubstreamHandler`

Does it matter? The code will be updated in https://github.com/apache/incubator-pekko/pull/981? Currently it's a place holder.

Or do you mean  that in https://github.com/apache/incubator-pekko/pull/981 it should behave as you say?

mdedetrich: > We should add test for  at least ` Supervision.Resume ` and ` Supervision.Stop` strategy.

These already exist, only the case of throwing exception with `Supervision.restart` being missed

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
mdedetrich: > Yes, the element which case the exception is dropped now.

So to confirm, the current PRs implementation is correct and the test as originally written was wrong?

He-Pin: As the `downstreamSubscription.request(100)` , I think the orign test should be right too, but its a little strange now ,that we need send the `next(7)` to make the next sub source be ready.

I think we should call 
```scala
      private def pushSubstreamSource(): Unit = {
        push(out, Source.fromGraph(substreamSource.source))
        scheduleOnce(SubscriptionTimer, timeout)
        substreamWaitingToBePushed = false
      }
```

now inside the `SubstreamHandler#onPush` when the decision is `SplitAfter` too to keep the behavior the same, as it just a dummy Source.

mdedetrich: So if I change the logic to

```scala
          case NonFatal(ex) =>
            decider(ex) match {
              case Supervision.Resume  =>
                if (decision == SplitAfter)
                  pushSubstreamSource()
                else
                  pull(in)
              case Supervision.Stop    => onUpstreamFailure(ex)
              case Supervision.Restart => onUpstreamFailure(ex) // TODO implement restart?
            }
```

This test now fails earlier on in the happy path, specifically at https://github.com/apache/incubator-pekko/blob/df3b9fccde6b7fbe7829703996379a055bc27b42/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala#L243

Maybe the test really was just originally written incorrectly?

samueleresca: I was reviewing this test. Isn't the `val substream2 = subscriber.expectNext()` expecting the new stream element (substream2) triggered by the `sendNext(6)` instruction? why do we need to move the `sendNext(7)` above?

mdedetrich: > I was reviewing this test. Isn't the val substream2 = subscriber.expectNext() expecting the new stream element (substream2) triggered by the sendNext(6) instruction? why do we need to move the sendNext(7) above?

Well the test doesn't pass otherwise and if you see the equivalent tests both in [`FlowSplitWhenSpec`](https://github.com/apache/incubator-pekko/blob/df3b9fccde6b7fbe7829703996379a055bc27b42/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitWhenSpec.scala#L246-L278) and [`FlowGroupBySpec`](https://github.com/apache/incubator-pekko/blob/c44c0b7cbdab11d85176cfe062288fdcba16c56a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowGroupBySpec.scala#L281-L323), its written in the same way that this PR changes it to.

There also doesn't appear to be an obvious solution to make the test as it was originally written to pass, hence why considering everything it could just be that the test was mistakingly written incorrectly (which is quite probably considering the test was written just as an example, it was never executed/passing)

He-Pin: @mdedetrich in the test, it requests 10, so the SubSource is expected to be there after 6.

mdedetrich: > @mdedetrich in the test, it requests 10, so the SubSource is expected to be there after 6.

But its also sending a completion because of this condition `if (elem == 3) throw exc else elem % 3 == 0` (i.e. `6 % 3` == 0 so the split happens) which is why the test is failing with

```
Expected OnNext(_), yet no element signaled during 3 seconds
java.lang.AssertionError: Expected OnNext(_), yet no element signaled during 3 seconds
	at org.apache.pekko.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:398)
	at org.apache.pekko.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:389)
	at org.apache.pekko.stream.scaladsl.FlowSplitAfterSpec.$anonfun$new$11(FlowSplitAfterSpec.scala:251)
```

i.e. at `val substream2 = subscriber.expectNext()` because there isn't going to be a `expectNext` due to the completion?

He-Pin: **1. `val substream2 = subscriber.expectNext()`, the `subscriber` is not expecting a `complete`,  otherwise, the call `sendNext(7)` will cause issue because the origin source is already completed.

The problem here is when the `stream2` be generated, I expected the behavior the same, just after the `6 % 3 == 0`, wdyt @samueleresca @mdedetrich**

mdedetrich: > **1. `val substream2 = subscriber.expectNext()`, the `subscriber` is not expecting a `complete`, otherwise, the call `sendNext(7)` will cause issue because the origin source is already completed.
> 
> The problem here is when the `stream2` be generated, I expected the behavior the same, just after the `6 % 3 == 0`, wdyt @samueleresca @mdedetrich**

I am not saying that this is wrong, just that its a completely separate issue from what this PR is changing/solving. Or to put it differently, this behaviour is the same as current Pekko since at this point i.e.

```scala
upstreamSubscription.sendNext(6)
substreamPuppet1.expectNext(6)
substreamPuppet1.expectComplete()
upstreamSubscription.sendNext(7)
val substream2 = subscriber.expectNext()
val substreamPuppet2 = StreamPuppet(substream2.runWith(Sink.asPublisher(false)))
substreamPuppet2.request(10)
substreamPuppet2.expectNext(7)

upstreamSubscription.sendComplete()
subscriber.expectComplete()
substreamPuppet2.expectComplete()
```

we are past recovering from an exception. The critical part of the test specifically dealing with recovering from exception and resuming is

```scala
upstreamSubscription.sendNext(3)
upstreamSubscription.sendNext(4)
substreamPuppet1.expectNext(4) // note that 3 was dropped
```

And this part is completely unchanged from how the test was originally written.

Given that, I think it makes sense to file a separate issue if this behaviour about completing after a split needs to change and hence to tackle it separately

He-Pin (approved):
lgtm, we will see what will happen with nightly build.

mdedetrich:
Thanks, ill go ahead and merge this. With it being included in `-M1` it will get plenty of testing/usage before the final release incase there is any hypothetical issue

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowSplitAfterSpec.scala
samueleresca: Thanks folks, I appreciate the detailed explanation. And yes, I was diverting away (not on purpose) from the original goal of the test 

> Note that at this point, we are dealing with the normal logic of SplitWhen. The new functionality of skipping 
> exceptions with SupervisionDecider.resume occurs earlier when elem == 3 so we are past that and just testing happy > path logic of SplitAfter which is not changed at all (and it shouldn't be either!).

looks good to me

PR #1209, opened 2024-03-18
build with protoc 3.25.3

pjfanning:
matches the protobuf-java version


project/Protobuf.scala
```diff
@@ -40,7 +40,7 @@ object Protobuf {
     // this keeps intellij happy for files that use the shaded protobuf
     Compile / unmanagedJars += (LocalProject("protobuf-v3") / Compile / packageBin).value,
     protoc := "protoc",
-    protocVersion := "3.20.3",
+    protocVersion := "3.25.3",
```
He-Pin, line 43: How about move the version to Dependencies.scala?

He-Pin (approved):
lgtm

project/Protobuf.scala
pjfanning: done

PR #1210, opened 2024-03-19
chore: Pull instead of throw exception in groupBy operator.

He-Pin:
Motivation:
Fix up a groupby operator bug(issue), and let it just behavior the same as the scaladoc.
>If elements arrive for a stream that is in the process of closing these elements might get lost.

Without this ,the stream is completed with an error.

Result:
The elements associated with the closed sub source is just dropped, and the stream continue, and this will help to keep long live stream in production alive.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowGroupBySpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowGroupBySpec.scala
@@ -694,6 +694,19 @@ class FlowGroupBySpec extends StreamSpec("""
       queue.complete()
     }
 
+    "should not fail when one sub stream completed" in {
+      Source(1 to 10)
+        .groupBy(2, _ % 2, allowClosedSubstreamRecreation = false)
+        .take(2)
+        .mergeSubstreams
+        .fold(Set.empty[Int])((set, elem) => set + elem)
+        .runWith(TestSink())
+        .ensureSubscription()
+        .request(10)
+        .expectNext(Set(1, 2, 3, 4))
+        .expectComplete()
+    }
+
   }
 
 }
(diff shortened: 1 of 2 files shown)


stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowGroupBySpec.scala
```diff
@@ -694,6 +692,19 @@ class FlowGroupBySpec extends StreamSpec("""
       queue.complete()
     }
 
+    "should not fail when one sub stream completed" in {
+      Source(1 to 10)
+        .groupBy(2, _ % 2, allowClosedSubstreamRecreation = false)
+        .take(2)
```
He-Pin, line 698: The sub stream is limited to 2, after that, the elements is skiped and keep pulling.

laglangyue:
lgtm

pjfanning:
Can't be merged until M1 is released. Unless we agree to bring it into M1.

mdedetrich:
After having the recent conversation at https://lists.apache.org/thread/rjygk60ho1mojk4rqsmvd2y4244d55lr I think its best to put as many changes as we can in `-M1` to avoid an `-M2` if possible (i.e. there may be some bug fixes after `-M1`  but nothing drastic like new features).

That is my view but I am not going to push hard on it.

mdedetrich (approved):
lgtm, not sure if the change is big enough to warrant updating https://github.com/apache/incubator-pekko/blob/1bdd6ed80e765c9e3e87a6182adf9f60c8355993/docs/src/main/paradox/migration/migration-guide-1.0.x-1.1.x.md

He-Pin:
That is why I do this asap，+1 for have it in m1 too

mdedetrich:
@pjfanning wdyt?

pjfanning (approved):
lgtm

He-Pin:
Thanks for detail review.

PR #1220, opened 2024-03-23
chore: Fix scala constructor finding tests for scala 3.

He-Pin:
Motivation:
Fix https://github.com/apache/pekko/issues/1082

### DIFF
--- a/docs/src/test/scala/docs/actor/ActorDocSpec.scala
+++ b/docs/src/test/scala/docs/actor/ActorDocSpec.scala
@@ -13,7 +13,7 @@
 
 package docs.actor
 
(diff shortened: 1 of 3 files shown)


He-Pin:
Scala 3 encoding different vs Scala 2, updating :(

pjfanning:
unfortunately our PR build does not test this code - so let's get it merged to main when it is ready but then we run the nightly build

We can backport when we know that the main branch nightly tests pass.

He-Pin:
@pjfanning The constructor encoding is different in Scala 3 and Scala 2, so I just move these classes to an object , and let them then closing over nothing.

pjfanning (approved):
lgtm

PR #1240, opened 2024-03-29
chore: Refactor UnfoldResourceSourceAsync.

He-Pin:
Motivation:
Refactory UnfoldResourceSourceAsync without change the order of type parameter instead of https://github.com/apache/pekko/pull/616

Result:
Clean code and less allocation.

I think we should this when reading the code in https://github.com/apache/pekko/pull/1240


Roiocam (approved):
LGTM

He-Pin:
Thanks for the review

PR #1244, opened 2024-03-31
chore: Add Source.iterate operator.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1189

Result:
Source.iterate operator added.

Note:
Because there is an overloading issue in Scala 2, so missing an overloaded method in Scala dsl.

### DIFF
--- a/docs/src/test/java/jdocs/stream/operators/source/Iterate.java
+++ b/docs/src/test/java/jdocs/stream/operators/source/Iterate.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package jdocs.stream.operators.source;
+
+import org.apache.pekko.NotUsed;
+import org.apache.pekko.stream.javadsl.Source;
+
+interface Iterate {
+
+  // #countTo
+  static Source<Long, NotUsed> countTo(long n) {
+    return Source.iterate(1L, i -> true, i -> i + 1).take(n);
+  }
+  // #countTo
+}
--- a/docs/src/test/scala/docs/stream/operators/source/Iterate.scala
+++ b/docs/src/test/scala/docs/stream/operators/source/Iterate.scala
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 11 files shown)


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
```diff
@@ -17,15 +17,15 @@ import java.util.concurrent.CompletionStage
...
 import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration.FiniteDuration
 
 import org.apache.pekko
 import pekko.{ Done, NotUsed }
 import pekko.actor.{ ActorRef, Cancellable }
 import pekko.annotation.InternalApi
-import pekko.stream.{ Outlet, SourceShape, _ }
+import pekko.stream._
 import pekko.stream.impl.{ PublisherSource, _ }
```
Roiocam, line 29:
```suggestion
import pekko.stream.impl._
```

stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
```diff
@@ -275,6 +275,25 @@ object Source {
...
+    new Source(scaladsl.Source.iterate(seed)(elem => p.test(elem), elem => f(elem)))
+
+  /**
+   * Creates an infinite sequential `Source` by iterating with the given function,
+   * starting with the given `seed` value.
+   *
+   * @see [[unfold]]
+   */
+  def iterate[T](seed: T, f: function.Function[T, T]): Source[T, NotUsed] =
+    new Source(scaladsl.Source.iterate(seed)(ConstantFun.anyToTrue, elem => f(elem)))
```
Roiocam, line 295: Should we mention in the document that Java has an API that does not require a predicate?

docs/src/test/java/jdocs/stream/operators/source/Iterate.java
```diff
@@ -0,0 +1,13 @@
+package jdocs.stream.operators.source;
+
+import org.apache.pekko.NotUsed;
+import org.apache.pekko.stream.javadsl.Source;
+
+interface Iterate {
+
+  // #countTo
+  static Source<Long, NotUsed> countTo(long n) {
+    return Source.iterate(1L, i -> i + 1).take(n);
```
Roiocam, line 10: For comparison, it may be more friendly to use the same API as ScalaDsl.

```suggestion
    return Source.iterate(1L, i -> true, i -> i + 1).take(n);
```

Roiocam (approved):
LGTM

stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
He-Pin: I think that's is not needed, we can only add that method when we only compiles with Scala 3, the overloading issue will not be fixed in Scala 2.12/2.13 in anyway.

docs/src/test/scala/docs/stream/operators/source/Iterate.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import org.apache.pekko
+import pekko.NotUsed
+import pekko.stream.scaladsl.Source
+
+object Iterate {
+
+  // #countTo
+  def countTo(n: Long): Source[Long, NotUsed] = Source
+    .iterate(1L)(_ => true, _ + 1)
```
laglangyue, line 28: how about `(_ < n, _ + 1)`?
This show the usage of the predication function
and don't need the take

laglangyue:
LGTM

docs/src/test/scala/docs/stream/operators/source/Iterate.scala
He-Pin: Let's keep it as it for now, which seems simpler.

Roiocam (approved):

PR #1246, opened 2024-04-01
feat: Java API for ExternalShardAllocationStrategy

Roiocam:
# Motivation

`ExternalShardAllocationStrategy` doesn't have a Java API, users have to use `pekko.util.TimeUnit ` rather than `java.time.Duration`

And the documentation also has an issue that doesn't specific AllocateStrategy, see: https://pekko.apache.org/docs/pekko/current/typed/cluster-sharding.html#external-shard-allocation

截屏2024-04-01 15 49 39

### DIFF
--- a/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
+++ b/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
@@ -119,7 +119,7 @@ object ExternalShardAllocationStrategy {
 
 class ExternalShardAllocationStrategy(systemProvider: ClassicActorSystemProvider, typeName: String)(
     // local only ask
-    implicit val timeout: Timeout)
+    implicit val timeout: Timeout = Timeout(5.seconds))
     extends ShardCoordinator.StartableAllocationStrategy {
 
   private val system = systemProvider.classicSystem
--- a/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
+++ b/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
@@ -13,7 +13,6 @@
 
 package org.apache.pekko.cluster.sharding.external
 
-import scala.annotation.varargs
 import scala.collection.immutable
 import scala.concurrent.Future
 import scala.concurrent.duration._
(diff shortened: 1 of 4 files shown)


cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
```diff
@@ -44,6 +45,21 @@ object ExternalShardAllocationStrategy {
 
   type ShardRegion = ActorRef
 
+  /**
+   * Scala API: Create an [[ExternalShardAllocationStrategy]]
+   */
+  def apply(systemProvider: ClassicActorSystemProvider, typeName: String)(
+      implicit timeout: Timeout = 5.seconds): ExternalShardAllocationStrategy =
+    new ExternalShardAllocationStrategy(systemProvider, typeName)(timeout)
```
He-Pin, line 53: 1. remove the `Scala API` which is very clear this is a `Scala API`.
2. remove the default parameter ` 5.seconds, which is not great for binary compabinity .

cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
```diff
@@ -44,6 +45,21 @@ object ExternalShardAllocationStrategy {
...
+
+  /**
+   * Java API: Create an [[ExternalShardAllocationStrategy]]
+   */
+  @varargs
+  def create(systemProvider: ClassicActorSystemProvider, typeName: String, duration: java.time.Duration)
+      : ExternalShardAllocationStrategy =
+    this.apply(systemProvider, typeName)(Timeout.create(duration))
```
He-Pin, line 61: 1. remove `@varargs` ,which is not needed.
2. the `Timeout.create(duration)` can use `duration.asScala` I think.

He-Pin (approved):
Cool, lgtm

cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
```diff
@@ -39,11 +39,35 @@ import pekko.cluster.sharding.ShardRegion.ShardId
...
+   * Create an [[ExternalShardAllocationStrategy]]
+   */
+  def apply(systemProvider: ClassicActorSystemProvider, typeName: String): ExternalShardAllocationStrategy = {
+    implicit val timeout: Timeout = 5.seconds
+    new ExternalShardAllocationStrategy(systemProvider, typeName)
+  }
+
+  /**
+   * Scala API: Create an [[ExternalShardAllocationStrategy]]
+   */
```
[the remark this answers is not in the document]
pjfanning: I don't see why this should be removed. There are plenty of other docs that includes this.

pjfanning (approved):
lgtm

cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/external/ExternalShardAllocationStrategy.scala
Roiocam: Let's keep the `Scala API`, otherwise it will be as confused as ActorSystem for Java newcomers.

PR #1247, opened 2024-04-03
Avoid boxing in Framing

JD557:
Adds a specialized `indexOf(Byte, Int)` to all `ByteString` implementations, similar to what is done on the `ByteIterator`, in order to avoid boxing/unboxing when searching for a byte.

This speeds up framing quite a bit.

Benchmark results (Java 11, Scala 2.13, Apple M3 Max):

<details><summary>Old</summary>
<pre>
Benchmark                 (framePerSeq)  (messageSize)   Mode  Cnt         Score         Error  Units
FramingBenchmark.framing              1             32  thrpt    3  10113843.601 ±  347860.528  ops/s
FramingBenchmark.framing              1             64  thrpt    3   9649511.788 ±  275242.133  ops/s
FramingBenchmark.framing              1            128  thrpt    3   8121220.174 ± 1320460.364  ops/s
FramingBenchmark.framing              1            256  thrpt    3   5466330.854 ± 3260665.704  ops/s
FramingBenchmark.framing              1            512  thrpt    3   2859562.317 ±  489750.599  ops/s
FramingBenchmark.framing              1           1024  thrpt    3   1594647.037 ±   52601.259  ops/s
FramingBenchmark.framing              8             32  thrpt    3   1772199.560 ±   36471.572  ops/s
FramingBenchmark.framing              8             64  thrpt    3   1595775.427 ±   94933.072  ops/s
FramingBenchmark.framing              8            128  thrpt    3   1074911.259 ±  198661.935  ops/s
FramingBenchmark.framing              8            256  thrpt    3    715528.264 ±  117832.304  ops/s
FramingBenchmark.framing              8            512  thrpt    3    388890.655 ±   23628.881  ops/s
FramingBenchmark.framing              8           1024  thrpt    3    218333.423 ±    3502.994  ops/s
FramingBenchmark.framing             16             32  thrpt    3   1073203.353 ±   56959.738  ops/s
FramingBenchmark.framing             16             64  thrpt    3    835697.061 ±   61190.351  ops/s
FramingBenchmark.framing             16            128  thrpt    3    620562.797 ±    4529.555  ops/s
FramingBenchmark.framing             16            256  thrpt    3    380776.167 ±    3069.815  ops/s
FramingBenchmark.framing             16            512  thrpt    3    201483.021 ±   27772.140  ops/s
FramingBenchmark.framing             16           1024  thrpt    3    110883.293 ±    2468.349  ops/s
FramingBenchmark.framing             32             32  thrpt    3    589338.870 ±   48664.767  ops/s
FramingBenchmark.framing             32             64  thrpt    3    438839.627 ±   98301.907  ops/s
FramingBenchmark.framing             32            128  thrpt    3    318200.080 ±    3745.204  ops/s
FramingBenchmark.framing             32            256  thrpt    3    196484.642 ±    1439.287  ops/s
FramingBenchmark.framing             32            512  thrpt    3    102559.681 ±    1066.612  ops/s
FramingBenchmark.framing             32           1024  thrpt    3     55713.288 ±    1021.259  ops/s
FramingBenchmark.framing             64             32  thrpt    3    260941.068 ±   11029.238  ops/s
FramingBenchmark.framing             64             64  thrpt    3    213252.824 ±    2763.842  ops/s
FramingBenchmark.framing             64            128  thrpt    3    152106.933 ±    3544.177  ops/s
FramingBenchmark.framing             64            256  thrpt    3     97187.752 ±   16306.733  ops/s
FramingBenchmark.framing             64            512  thrpt    3     51460.170 ±     851.556  ops/s
FramingBenchmark.framing             64           1024  thrpt    3     27853.852 ±    1002.975  ops/s
FramingBenchmark.framing            128             32  thrpt    3    143485.697 ±    2294.380  ops/s
FramingBenchmark.framing            128             64  thrpt    3    108585.695 ±    1580.502  ops/s
FramingBenchmark.framing            128            128  thrpt    3     74709.074 ±    1104.879  ops/s
FramingBenchmark.framing            128            256  thrpt    3     49446.714 ±    3098.863  ops/s
FramingBenchmark.framing            128            512  thrpt    3     25852.970 ±    2643.999  ops/s
FramingBenchmark.framing            128           1024  thrpt    3     13979.600 ±     221.422  ops/s
</pre>
</details>

<details><summary>New</summary>
<pre>
Benchmark                 (framePerSeq)  (messageSize)   Mode  Cnt         Score        Error  Units
FramingBenchmark.framing              1             32  thrpt    3  12488386.857 ± 550524.197  ops/s
FramingBenchmark.framing              1             64  thrpt    3  10825562.059 ± 619829.640  ops/s
FramingBenchmark.framing              1            128  thrpt    3  10298423.557 ± 379279.669  ops/s
FramingBenchmark.framing              1            256  thrpt    3   7195103.283 ± 489354.505  ops/s
FramingBenchmark.framing              1            512  thrpt    3   5318261.366 ± 271701.067  ops/s
FramingBenchmark.framing              1           1024  thrpt    3   3380494.654 ± 293146.013  ops/s
FramingBenchmark.framing              8             32  thrpt    3   1876127.541 ± 228271.371  ops/s
FramingBenchmark.framing              8             64  thrpt    3   1777053.701 ± 146526.586  ops/s
FramingBenchmark.framing              8            128  thrpt    3   1430280.685 ±  97618.551  ops/s
FramingBenchmark.framing              8            256  thrpt    3   1130490.909 ± 134372.173  ops/s
FramingBenchmark.framing              8            512  thrpt    3    784673.367 ±  42187.315  ops/s
FramingBenchmark.framing              8           1024  thrpt    3    496979.878 ±  65898.234  ops/s
FramingBenchmark.framing             16             32  thrpt    3   1085230.634 ±  80359.239  ops/s
FramingBenchmark.framing             16             64  thrpt    3    972758.829 ±  47127.839  ops/s
FramingBenchmark.framing             16            128  thrpt    3    794532.378 ±  35834.092  ops/s
FramingBenchmark.framing             16            256  thrpt    3    601945.868 ± 120485.938  ops/s
FramingBenchmark.framing             16            512  thrpt    3    414966.392 ± 187654.169  ops/s
FramingBenchmark.framing             16           1024  thrpt    3    259000.528 ±  58471.624  ops/s
FramingBenchmark.framing             32             32  thrpt    3    593169.099 ±   4840.018  ops/s
FramingBenchmark.framing             32             64  thrpt    3    528762.400 ±  23527.622  ops/s
FramingBenchmark.framing             32            128  thrpt    3    429482.997 ±  13070.825  ops/s
FramingBenchmark.framing             32            256  thrpt    3    321971.065 ±  35385.581  ops/s
FramingBenchmark.framing             32            512  thrpt    3    214747.165 ±  34767.609  ops/s
FramingBenchmark.framing             32           1024  thrpt    3    130871.717 ±  30464.051  ops/s
FramingBenchmark.framing             64             32  thrpt    3    318087.922 ±   7819.818  ops/s
FramingBenchmark.framing             64             64  thrpt    3    259738.166 ±   9262.097  ops/s
FramingBenchmark.framing             64            128  thrpt    3    208251.362 ±   9641.613  ops/s
FramingBenchmark.framing             64            256  thrpt    3    154602.853 ±  17896.718  ops/s
FramingBenchmark.framing             64            512  thrpt    3    101678.754 ±  20105.737  ops/s
FramingBenchmark.framing             64           1024  thrpt    3     62230.188 ±  19234.156  ops/s
FramingBenchmark.framing            128             32  thrpt    3    150634.966 ±  15251.295  ops/s
FramingBenchmark.framing            128             64  thrpt    3    127661.514 ±   2498.963  ops/s
FramingBenchmark.framing            128            128  thrpt    3    105351.600 ±   6728.938  ops/s
FramingBenchmark.framing            128            256  thrpt    3     80119.663 ±  11005.101  ops/s
FramingBenchmark.framing            128            512  thrpt    3     52383.251 ±  11319.119  ops/s
FramingBenchmark.framing            128           1024  thrpt    3     28313.662 ±  11664.417  ops/s
</pre>
</details>

<details><summary>Both (Interleaved)</summary>
<pre>
Benchmark                     (framePerSeq)  (messageSize)   Mode  Cnt         Score        Error  Units
FramingBenchmark.framing_old              1             32  thrpt    3  10113843.601 ±  347860.528  ops/s
FramingBenchmark.framing_new              1             32  thrpt    3  12488386.857 ± 550524.197  ops/s
FramingBenchmark.framing_old              1             64  thrpt    3   9649511.788 ±  275242.133  ops/s
FramingBenchmark.framing_new              1             64  thrpt    3  10825562.059 ± 619829.640  ops/s
FramingBenchmark.framing_old              1            128  thrpt    3   8121220.174 ± 1320460.364  ops/s
FramingBenchmark.framing_new              1            128  thrpt    3  10298423.557 ± 379279.669  ops/s
FramingBenchmark.framing_old              1            256  thrpt    3   5466330.854 ± 3260665.704  ops/s
FramingBenchmark.framing_new              1            256  thrpt    3   7195103.283 ± 489354.505  ops/s
FramingBenchmark.framing_old              1            512  thrpt    3   2859562.317 ±  489750.599  ops/s
FramingBenchmark.framing_new              1            512  thrpt    3   5318261.366 ± 271701.067  ops/s
FramingBenchmark.framing_old              1           1024  thrpt    3   1594647.037 ±   52601.259  ops/s
FramingBenchmark.framing_new              1           1024  thrpt    3   3380494.654 ± 293146.013  ops/s
FramingBenchmark.framing_old              8             32  thrpt    3   1772199.560 ±   36471.572  ops/s
FramingBenchmark.framing_new              8             32  thrpt    3   1876127.541 ± 228271.371  ops/s
FramingBenchmark.framing_old              8             64  thrpt    3   1595775.427 ±   94933.072  ops/s
FramingBenchmark.framing_new              8             64  thrpt    3   1777053.701 ± 146526.586  ops/s
FramingBenchmark.framing_old              8            128  thrpt    3   1074911.259 ±  198661.935  ops/s
FramingBenchmark.framing_new              8            128  thrpt    3   1430280.685 ±  97618.551  ops/s
FramingBenchmark.framing_old              8            256  thrpt    3    715528.264 ±  117832.304  ops/s
FramingBenchmark.framing_new              8            256  thrpt    3   1130490.909 ± 134372.173  ops/s
FramingBenchmark.framing_old              8            512  thrpt    3    388890.655 ±   23628.881  ops/s
FramingBenchmark.framing_new              8            512  thrpt    3    784673.367 ±  42187.315  ops/s
FramingBenchmark.framing_old              8           1024  thrpt    3    218333.423 ±    3502.994  ops/s
FramingBenchmark.framing_new              8           1024  thrpt    3    496979.878 ±  65898.234  ops/s
FramingBenchmark.framing_old             16             32  thrpt    3   1073203.353 ±   56959.738  ops/s
FramingBenchmark.framing_new             16             32  thrpt    3   1085230.634 ±  80359.239  ops/s
FramingBenchmark.framing_old             16             64  thrpt    3    835697.061 ±   61190.351  ops/s
FramingBenchmark.framing_new             16             64  thrpt    3    972758.829 ±  47127.839  ops/s
FramingBenchmark.framing_old             16            128  thrpt    3    620562.797 ±    4529.555  ops/s
FramingBenchmark.framing_new             16            128  thrpt    3    794532.378 ±  35834.092  ops/s
FramingBenchmark.framing_old             16            256  thrpt    3    380776.167 ±    3069.815  ops/s
FramingBenchmark.framing_new             16            256  thrpt    3    601945.868 ± 120485.938  ops/s
FramingBenchmark.framing_old             16            512  thrpt    3    201483.021 ±   27772.140  ops/s
FramingBenchmark.framing_new             16            512  thrpt    3    414966.392 ± 187654.169  ops/s
FramingBenchmark.framing_old             16           1024  thrpt    3    110883.293 ±    2468.349  ops/s
FramingBenchmark.framing_new             16           1024  thrpt    3    259000.528 ±  58471.624  ops/s
FramingBenchmark.framing_old             32             32  thrpt    3    589338.870 ±   48664.767  ops/s
FramingBenchmark.framing_new             32             32  thrpt    3    593169.099 ±   4840.018  ops/s
FramingBenchmark.framing_old             32             64  thrpt    3    438839.627 ±   98301.907  ops/s
FramingBenchmark.framing_new             32             64  thrpt    3    528762.400 ±  23527.622  ops/s
FramingBenchmark.framing_old             32            128  thrpt    3    318200.080 ±    3745.204  ops/s
FramingBenchmark.framing_new             32            128  thrpt    3    429482.997 ±  13070.825  ops/s
FramingBenchmark.framing_old             32            256  thrpt    3    196484.642 ±    1439.287  ops/s
FramingBenchmark.framing_new             32            256  thrpt    3    321971.065 ±  35385.581  ops/s
FramingBenchmark.framing_old             32            512  thrpt    3    102559.681 ±    1066.612  ops/s
FramingBenchmark.framing_new             32            512  thrpt    3    214747.165 ±  34767.609  ops/s
FramingBenchmark.framing_old             32           1024  thrpt    3     55713.288 ±    1021.259  ops/s
FramingBenchmark.framing_new             32           1024  thrpt    3    130871.717 ±  30464.051  ops/s
FramingBenchmark.framing_old             64             32  thrpt    3    260941.068 ±   11029.238  ops/s
FramingBenchmark.framing_new             64             32  thrpt    3    318087.922 ±   7819.818  ops/s
FramingBenchmark.framing_old             64             64  thrpt    3    213252.824 ±    2763.842  ops/s
FramingBenchmark.framing_new             64             64  thrpt    3    259738.166 ±   9262.097  ops/s
FramingBenchmark.framing_old             64            128  thrpt    3    152106.933 ±    3544.177  ops/s
FramingBenchmark.framing_new             64            128  thrpt    3    208251.362 ±   9641.613  ops/s
FramingBenchmark.framing_old             64            256  thrpt    3     97187.752 ±   16306.733  ops/s
FramingBenchmark.framing_new             64            256  thrpt    3    154602.853 ±  17896.718  ops/s
FramingBenchmark.framing_old             64            512  thrpt    3     51460.170 ±     851.556  ops/s
FramingBenchmark.framing_new             64            512  thrpt    3    101678.754 ±  20105.737  ops/s
FramingBenchmark.framing_old             64           1024  thrpt    3     27853.852 ±    1002.975  ops/s
FramingBenchmark.framing_new             64           1024  thrpt    3     62230.188 ±  19234.156  ops/s
FramingBenchmark.framing_old            128             32  thrpt    3    143485.697 ±    2294.380  ops/s
FramingBenchmark.framing_new            128             32  thrpt    3    150634.966 ±  15251.295  ops/s
FramingBenchmark.framing_old            128             64  thrpt    3    108585.695 ±    1580.502  ops/s
FramingBenchmark.framing_new            128             64  thrpt    3    127661.514 ±   2498.963  ops/s
FramingBenchmark.framing_old            128            128  thrpt    3     74709.074 ±    1104.879  ops/s
FramingBenchmark.framing_new            128            128  thrpt    3    105351.600 ±   6728.938  ops/s
FramingBenchmark.framing_old            128            256  thrpt    3     49446.714 ±    3098.863  ops/s
FramingBenchmark.framing_new            128            256  thrpt    3     80119.663 ±  11005.101  ops/s
FramingBenchmark.framing_old            128            512  thrpt    3     25852.970 ±    2643.999  ops/s
FramingBenchmark.framing_new            128            512  thrpt    3     52383.251 ±  11319.119  ops/s
FramingBenchmark.framing_old            128           1024  thrpt    3     13979.600 ±     221.422  ops/s
FramingBenchmark.framing_new            128           1024  thrpt    3     28313.662 ±  11664.417  ops/s
</pre>
</details>


actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
```diff
@@ -717,6 +717,82 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       compact.indexOf('g', 5) should ===(5)
       compact.indexOf('g', 6) should ===(-1)
     }
+    "indexOf (specialized)" in {
```
He-Pin, line 720: add a blank line

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -790,6 +843,30 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
...
+   * Finds index of first occurrence of some byte in this ByteString after or at some start index.
+   *
+   * Similar to indexOf, but it avoids boxing if the value is already a byte.
+   *
+   *  @param   elem   the element value to search for.
+   *  @param   from   the start index
+   *  @return  the index `>= from` of the first element of this ByteString that is equal (as determined by `==`)
+   *           to `elem`, or `-1`, if none exists.
+   */
+  def indexOf(elem: Byte, from: Int): Int = indexOf(elem, from)
```
He-Pin, line 857: add `@since 1.1.0`

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -790,6 +843,30 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
...
+  /**
+   * Finds index of first occurrence of some byte in this ByteString.
+   *
+   * Similar to indexOf, but it avoids boxing if the value is already a byte.
+   *
+   *  @param   elem   the element value to search for.
+   *  @return  the index `>= from` of the first element of this ByteString that is equal (as determined by `==`)
+   *           to `elem`, or `-1`, if none exists.
+   */
+  def indexOf(elem: Byte): Int = indexOf(elem, 0)
```
He-Pin, line 868: add `@since 1.1.0`

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
He-Pin: How about add another `indexOf(elem:Byte,fromIndex:Int, toIndex:Int`?
or `def indexOf(fromIndex:Int, toIndex:Int, elem:Byte)`

He-Pin: This mehtod is actually firstIndexOf or byteBefore

He-Pin:
The byte searching can do with SIMD too, but I lack of time to do this:(

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
JD557: Done in 44cf4bc

actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
JD557: I'm a bit confused. Where should I add the blank line? After? Before?

This was copied from the tests above, which don't seem to have any blank lines 😕

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
JD557: Added the `@since` annotation in 44cf4bccf71cfaed828d81625c603ed1e66e3fab

Regarding adding more methods, I think that is a bit out of scope for this PR.

I was thinking a bit about that, and I think it would be nice if, when introducing the new methods, they were in sync with the stdlib.
For example, it would be pretty bad to introduce a `indexOf(elem:Byte,fromIndex:Int, toIndex:Int)` and then the stdlib introduce a `indexOf(elem:Byte, fromIndex:Int, lenght:Int)`.

So, I think one possible way to avoid that would be:
1. Port this PR to https://github.com/mdedetrich/bytestring
2. Try to move the `ByteString` to the stdlib - I think that was one of @mdedetrich main reasons for extracting the code (as discussed in https://contributors.scala-lang.org/t/adding-akkas-bytestring-as-a-scala-module-and-or-stdlib/5967), and I think this should now be possible under [SIP-51](https://github.com/scala/improvement-proposals/pull/54).
3. Propose the new methods to the stdlib, implementing the specialized version where it makes sense.

WDYT?

actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
pjfanning: I'm ok with you keeping the existing style. Blanks lines are not needed if the existing style is like that.

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
pjfanning: Don't worry about the @mdedetrich project? This repo is the source of truth for ByteString. He will copy any changes from this repo if he thinks that they are useful.

The Pekko ByteString is not going away soon. The idea that it will end up in the Scala lib is not gaining traction and even it does it may not end up in every Scala version that Pekko needs it in and it may end up working differently to the way the Pekko one does.

mdedetrich: > The Pekko ByteString is not going away soon. The idea that it will end up in the Scala lib is not gaining traction and even it does it may not end up in every Scala version that Pekko needs it in and it may end up working differently to the way the Pekko one does.

It dis a really gain traction, it's just like qith most things you need to drive it yourself (others won't do it for you) and I didn't have the capacity with all of the other Pekko stuff going on

There is still a lot of merit in Pekko using an eventually stdlib `ByteString` but this will happen in a major breaking change in Pekko and it shouldn't dictate the development of Pekko.

actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
```diff
@@ -702,6 +728,33 @@ object ByteString {
...
+          if (bsIdx >= byteStringsSize) -1
+          else {
+            val bs = bytestrings(bsIdx)
+
+            if (bs.length <= relativeIndex) {
+              find(bsIdx + 1, relativeIndex - bs.length, bytesPassed + bs.length)
+            } else {
+              val subIndexOf = bs.indexOf(elem, relativeIndex)
+              if (subIndexOf < 0) {
+                val nextString = bsIdx + 1
```
He-Pin, line 747: nextBsIdx?

actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
```diff
@@ -702,6 +728,33 @@ object ByteString {
...
+                find(nextString, relativeIndex - bs.length, bytesPassed + bs.length)
```
He-Pin, line 748: or we just inline it to the arguments.

He-Pin (approved):
lgtm, thanks

actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
JD557: Went with this solution in e923cba. This is also what happens in the branch above, so the code looks more symmetrical now. 👍

He-Pin:
@Roiocam @jxnu-liguobin @jrudolph Would you like to get a look into this?

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -433,6 +447,20 @@ object ByteString {
       }
     }
 
+    override def indexOf(elem: Byte): Int = indexOf(elem, 0)
+    override def indexOf(elem: Byte, from: Int): Int = {
+      if (from >= length) -1
+      else {
+        var found = -1
+        var i = math.max(from, 0)
+        while (i < length && found == -1) {
```
He-Pin, line 456: I think if we swap with `found == -1 && i < length)` which can reduce a bit when we just found:)

actor/src/main/scala-2.12/org/apache/pekko/util/ByteString.scala
```diff
@@ -433,6 +447,20 @@ object ByteString {
...
+          if (bytes(startIndex + i) == elem) found = i
+          i += 1
+        }
+        found
```
He-Pin, line 460: how about we extract the `startIndex + 1` to the start of the loop and returning a ` found - startIndex`？

He-Pin:
I think we can delegate the `indexOf[B >: Byte](elem: B, from: Int): Int` to `indexOf(elem: Byte, from: Int): Int` with `Byte.unbox`, wdyt, the reduce the duplicated code. @JD557

JD557:
Unfortunately, I don't think that would work, as the signature requires `[B >: Byte]`. `B` could be `Any` and in that case we need the slow implementation that falls back to `Object.equals` 😕 

Code example:
```scala
case class MyByteSeq(data: List[Byte]) {
  def fastIndexOf(byte: Byte): Int = data.indexOf(byte)
  //def indexOf1[B >: Byte](x: B): Int = fastIndexOf(Byte.unbox(x)) // This won't compile
  def indexOf2[B >: Byte](x: B): Int = fastIndexOf(Byte.unbox(x.asInstanceOf[Object])) // This can fail at runtime
  def indexOf3[B >: Byte](x: B): Int = x match {
    case b: Byte => fastIndexOf(b) // I think the pattern match already unboxes it
    case _ => ??? // I can't call fastIndexOf here, so I need a duplicated version of the code anyway
  }
}
```

He-Pin:
THANKS, But I checked the bytecode, it was Java.lang.Objrct, not sure why that would   fail at Runtime.

He-Pin:
@JD557 do you have any further improvement on this pr , I would like to merge this and improvement can come up later.

JD557:
> @JD557 do you have any further improvement on this pr ?I would like to merge this and improvement can come up later.

Not really, feel free to merge this and improve later

jrudolph (commented):
That still feels like a footgun, since it is quite unclear in which situations the new overload will be selected? Should we document in which case the static overload resolution will choose the new method?

Is it somehow possible to avoid the massive code duplication here (probably hard since the whole existing scheme is built around the function of `Any.==`)? If not it should be documented.

Ultimately, the main problem of `ByteString` has always been the attempt to make it fit seamlessly into the rest of the Scala collections by making it part of the collections type hierarchy (instead of making it fast and useful in the first place and then consider useful conversions/views to the Scala collection types).

actor/src/main/scala-3/org/apache/pekko/util/ByteString.scala
```diff
@@ -825,6 +876,32 @@ sealed abstract class ByteString
...
+   *
+   * Similar to indexOf, but it avoids boxing if the value is already a byte.
+   *
+   *  @param   elem   the element value to search for.
+   *  @param   from   the start index
+   *  @return  the index `>= from` of the first element of this ByteString that is equal (as determined by `==`)
+   *           to `elem`, or `-1`, if none exists.
+   *  @since 1.1.0
+   */
+  def indexOf(elem: Byte, from: Int): Int = indexOf(elem, from)
```
jrudolph, line 891: Infinite loop here (probably impossible to be called)?

jrudolph (changes requested):

actor/src/main/scala-3/org/apache/pekko/util/ByteString.scala
He-Pin: Replaced with a `???` should be better I think.

jrudolph: Can just be left blank, no need for a default implementation afaics.

JD557:
> That still feels like a footgun, since it is quite unclear in which situations the new overload will be selected? Should we document in which case the static overload resolution will choose the new method?

I don't have any strong feelings about the naming. I used this scheme because it's the same that's used by ByteIterator (See: https://github.com/apache/pekko/blob/3cd08016b57e95f9244f0990282dd956c173b82e/actor/src/main/scala-3/org/apache/pekko/util/ByteIterator.scala#L498-L502).

If we change the method here (e.g. to `indexOfByte`, like in 18c5db1), then I think it should also be changed in `ByteIterator`.

However, on that note:

> Is it somehow possible to avoid the massive code duplication here (probably hard since the whole existing scheme is built around the function of Any.==)? If not it should be documented.

Looks like `ByteIterator` avoids the duplication by delegating the comparison to `indexWhere`. Not sure if the lambda could have a performance impact, though. I would need to benchmark this again.

He-Pin:
@JD557 I think we can delegate the current version to the specialized version, even it can happen in type theory but will not happen in real use case if user don't try to do strange thing.

actor/src/main/scala-2.12/org/apache/pekko/util/ByteIterator.scala
```diff
@@ -470,7 +470,7 @@ abstract class ByteIterator extends BufferedIterator[Byte] {
   override def indexWhere(p: Byte => Boolean): Int = indexWhere(p, 0)
   override def indexWhere(p: Byte => Boolean, from: Int): Int = {
     var index = 0
-    while (index < from) {
+    while (index < from && hasNext) {
```
pjfanning, line 473: if there is a bug in ByteIterator, could the bug fix be in a separate PR? It needs to have its own release note and may be backported to 1.0.x,

pjfanning (changes requested):

actor/src/main/scala-2.12/org/apache/pekko/util/ByteIterator.scala
JD557: Extracted in https://github.com/apache/pekko/pull/1282

I'll cleanup the history here in a bit, to avoid potential merge conflicts.

actor/src/main/scala-3/org/apache/pekko/util/ByteIterator.scala
```diff
@@ -482,7 +482,7 @@ abstract class ByteIterator extends BufferedIterator[Byte] {
 
   override def indexWhere(p: Byte => Boolean, from: Int = 0): Int = {
     var index = 0
-    while (index < from) {
+    while (index < from && hasNext) {
```
He-Pin, line 485: the changes in byte Iterator is not needed.

IN Netty, these method is named as firstIndexOf and lastIndexOf, that's an option too.

He-Pin (changes requested):

actor/src/main/scala-3/org/apache/pekko/util/ByteIterator.scala
JD557: Extracted to https://github.com/apache/pekko/pull/1282

actor/src/main/scala-3/org/apache/pekko/util/ByteString.scala
JD557: Removed default implementation in 2cce3ea

I think this will fallback to `indexWhere`

JD557:
So, I was doing some more tests with a specialized version of `indexWhere`, but that's noticeably slower (even though it's faster than `ByteIterator#indexWhere`). So I don't see many ways to reduce the duplication.

As such, I'm not sure how to proceed with this PR:
- Should I introduce a specialized `indexWhere` anyway, leading to code multiplication?
- Do I keep the `indexOf` name (like the `ByteIterator` does) or do I name it something else?

He-Pin (approved):
Lgtm

He-Pin:
@JD557 I think if you want to change the method name, you can using the name of firstIndexOf.

And for the api, because it's using the scala collection's indexOf，so the input B where is actually a Byte, but now can be a any.

As for selection, I think the compiler will choose this new method when it know the elem for testing is a type Byte but will not when It doesn't. 

As for the old index Of, which will always return -1 us the elem is not a Number/can boxed to Byte, so a quick check and delegate to the specified one will not harm too much. Who will using a ByteString to indexOf a any?

He-Pin:
In BoxedRuntime.java
```java
    public static byte unboxToByte(Object b) {
        return b == null ? 0 : ((java.lang.Byte)b).byteValue();
    }

    public static java.lang.Byte boxToByte(byte b) {
        return java.lang.Byte.valueOf(b);
    }
```

In Byte.java
```java
    public boolean equals(Object obj) {
        if (obj instanceof Byte) {
            return value == ((Byte)obj).byteValue();
        }
        return false;
    }
```

JD557:
I don't think that `BoxesRunTime` trick is enough, unfortunately.

Say one asks `ByteString.indexOf(500)`. This should obviously return `-1`, as it is impossible for a byte to have that value.
However `BoxesRunTime.toByte(500)` returns `-12: Byte`.

Ideally we would check if the number is between `Byte.MinValue` and `Byte.MaxValue` before boxing, but even that is not enough, due to floating point values (BoxesRunTime.toByte(127.9) == 127.toByte == 127.0).

Although I imagine most problems would come from `Int` and `Char`.

He-Pin:
Another thing is , when we using code like `indexOf(`a`, 0)` the new method will not be selected too :(

jrudolph:
> As for selection, I think the compiler will choose this new method when it know the elem for testing is a type Byte but will not when It doesn't.

Probably, yes. If the general overload would not exist, the new one would also work for literals like `12` or `'a'`. So, maybe it's really the best we can do right now? With this solution it will at least choose the faster version whenever the user is explicitly looking for a byte.

He-Pin:
@JD557 need another update to make mima happy

JD557:
I can try to fix it, but out of curiosity, isn't it OK to have MiMa issues, since this only targets 1.1.x?

I was ignoring the issue because I thought 1.1.0 was going to break bincompat with 1.0.x anyway

mdedetrich:
> I can try to fix it, but out of curiosity, isn't it OK to have MiMa issues, since this only targets 1.1.x?

Pekko core follows SemVer, so the only acceptable MiMa issues are for internal code (i.e. `@InternalApi`) or with false positives

> I was ignoring the issue because I thought 1.1.0 was going to break bincompat with 1.0.x anyway

No it doesn't.

actor/src/main/scala-3/org/apache/pekko/util/ByteString.scala
```diff
@@ -823,7 +874,33 @@ sealed abstract class ByteString
   override def indexWhere(p: Byte => Boolean, from: Int): Int = iterator.indexWhere(p, from)
 
   // optimized in subclasses
-  override def indexOf[B >: Byte](elem: B, from: Int): Int = indexOf(elem, from)
+  override def indexOf[B >: Byte](elem: B, from: Int): Int = super.indexOf(elem, from)
```
JD557, line 877: I am aware that this is override is a bit weird, but:
- I think the old code was an infinite loop
- Removing this override breaks the MiMa checks

This should never be called anyway, but I think having it like this is a bit safer.

JD557:
OK, I think the MiMA issues should be fixed now.

pjfanning (approved):
lgtm

pjfanning:
@jrudolph do you still think this PR change needs more work?

He-Pin:
@sirthias Would you like to give some input about this too？

sirthias:
The only comment I have is that there would be a potentially _much_ more efficient implementation of `indexOf(elem: Byte, from: Int)` if we somehow had low-level (unsafe) access to the byte array and could read 8 bytes at once into a `long`.
Then we could use SWAR (SIMD within a register) and reduce the loop count by 8 (asymptotically) over the current implementation.
Doesn't pekko already have an `Unsafe` access construct somewhere?
IIRC Akka did...

sirthias:
Yes, there is a (much more complicated) SWAR loop implemented in [borer's JSON parser](https://github.com/sirthias/borer/blob/4d639de35ade1bc2513a50a9da9587451b322476/core/src/main/scala/io/bullet/borer/json/JsonParser.scala#L430).
Here we only have to look for a single known byte rather than a whole set of different characters and we also don't have to copy segments and do UTF8 decoding at the same time.

But the whole thing only makes sense if we can really get down to raw byte access via `Unsafe` or more modern means. And on ScalaJS a SWAR approach will just create overhead and be a lot slower than the simple loop.

pjfanning:
@He-Pin @mdedetrich @Roiocam @samueleresca @raboof @jxnu-liguobin should we get this sorted out before 1.1.0-M1 RC or should we look at this again after the 1.1.0-M1 release?

He-Pin:
Sorry for now response, holidays here. I think we can including this in m1 and try with SWAR after that.

pjfanning:
Merged - thanks @JD557

PR #1249, opened 2024-04-04
chore: Make use of splitIterator directly

He-Pin:
Motivation:
Use splitIterator directly

Result:
Less conversion.the current iterator is created from the splitIterator, which is another adapter, and we need double invocation hasNext and Next, both involving the adapter.

So with this directly using the split iterator, we can reduce the allocation and indirect invocation.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/JavaStreamSource.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/JavaStreamSource.scala
@@ -18,6 +18,9 @@ import pekko.annotation.InternalApi
 import pekko.stream._
 import pekko.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
 
+import java.util.Spliterator
+import java.util.function.Consumer
+
 /** INTERNAL API */
 @InternalApi private[stream] final class JavaStreamSource[T, S <: java.util.stream.BaseStream[T, S]](
     open: () => java.util.stream.BaseStream[T, S])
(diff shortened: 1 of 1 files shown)


pjfanning:
I would really prefer if there was a detailed description of why this is better. Some PRs are pretty obvious. This one isn't.

He-Pin:
Response from my phone, the current iterator is created from the splitIterator, which is another adapter, and we need double invocation hasNext and Next, both involving the adapter.

So with this directly using the split iterator, we can reduce the allocation and indirect invocation.

pjfanning:
> Response from my phone, the current iterator is created from the splitIterator, which is another adapter, and we need double invocation hasNext and Next, both involving the adapter.
> 
> So with this directly using the split iterator, we can reduce the allocation and indirect invocation.

Can you provide some stacktraces or code links to back this up? I checked the Javadoc for Stream iterator and spliterator and the Javadoc defines no behaviour where the iterator must call the spliterator.

He-Pin:
@pjfanning When you call `stream.iterator`, the returns `iterator` is created from a `SplitIterator`. where it's backing by an `Adapter`
see https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/Spliterators.java#L668

stream/src/main/scala/org/apache/pekko/stream/impl/JavaStreamSource.scala
```diff
@@ -27,28 +30,32 @@ import pekko.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
...
+    new GraphStageLogic(shape) with OutHandler with Consumer[T] {
+      private var stream: java.util.stream.BaseStream[T, S] = _
+      private var splitIterator: Spliterator[T] = _
 
-      override def preStart(): Unit = {
+      final override def preStart(): Unit = {
         stream = open()
-        iter = stream.iterator()
+        splitIterator = stream.spliterator()
+        val isEmpty = splitIterator.hasCharacteristics(Spliterator.SIZED) && splitIterator.estimateSize() == 0
```
laglangyue, line 40: How about
`splitIterator.estimateSize() == 0 && splitIterator.hasCharacteristics(Spliterator.SIZED)`

He-Pin: Bit operation is fast

laglangyue:
LGTM

pjfanning:
> @pjfanning When calling `stream.iterator`, the returns `iterator` is created from a `SplitIterator`. where it's backing by an `Adapter` see https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/Spliterators.java#L668

Could you provide the code where stream.iterator creates a spliterator? We may need this optimisation but we might want to only do for certain stream types. We may want to also report to OpenJDK that certain streams need to be improved so that they create iterators directly when users want iterators.

He-Pin:
image
```java
public class Main {
    public static void main(String[] args) {
        System.out.println("Hello World!");
        List<String> list = new ArrayList<String>();
        Iterator<String> itr = list.stream().iterator();
        System.out.println(itr.getClass());
    }
}
```

@pjfanning Just updated

pjfanning:
@He-Pin so Java Collections class has this.

```
    default Stream<E> stream() {
        return StreamSupport.stream(spliterator(), false);
    }
```

So this explians where the spliterator comes from.

My question would still be - are we sure all the streams that we need to handle come from Collections?

He-Pin:
We can't, but it will not be worse than the old one.

stream/src/main/scala/org/apache/pekko/stream/impl/JavaStreamSource.scala
```diff
@@ -27,28 +30,32 @@ import pekko.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
...
+        if (isEmpty) {
```
pjfanning, line 41: do we need the isEmpty val? couldn't we just have that as the if condition?

He-Pin: ok, let me inline it, but it's simpler for read.

pjfanning (approved):
lgtm

PR #1250, opened 2024-04-05
chore: Avoid copy the whole iterable for mapconcat oprerator in Javadsl.

He-Pin:
Motivation:
Currently, it's using the `Util.immutableSeq` which will copy the whole value in the Java iterable and convert it to a vector.
Because in the MapConcat/StatefulMapConcat ops, only the underlying `iterator` is being used,  so I think we can use the JIteratorWrapper instead.

Result:
Avoid the copy


pjfanning (approved):
lgtm - pending CI

Was it like this in Akka 2.6? If not, we might want to backport this.

PR #1266, opened 2024-04-06
we are not ignoring enough jackson modules (scala-steward)

pjfanning:
Jackson 2.17.0 has some issues.

the current setup tries to focus on jackson-core and ignores updates on other jackson modules - there is one jacksonVersion param and ScalaSteward does not recognise that it creates lots of PRs to update the same value (one for each Jackson module)


He-Pin (approved):
lgtm

PR #1288, opened 2024-04-24
Capture build scans on ge.apache.org to benefit from deep build insights

c00ler:
This PR adds a new auto-plugin that configures publishing of a build scan for every CI build on GitHub Actions. Build scans from local builds can be published by authenticated Apache committer only by explicitly opting in. The build will not fail if publishing fails.

The build scans of the Apache Pekko projects are published to the Develocity instance at [ge.apache.org](https://ge.apache.org/), hosted by the Apache Software Foundation and run in partnership between the ASF and Gradle through a [Platinum Targeted Sponsorship](https://news.apache.org/foundation/entry/the-apache-software-foundation-announces-gradle-as-a-platinum-targeted-sponsor). This Develocity instance has all features and extensions enabled and is freely available for use by the Apache Pekko projects and all other Apache projects.

On this Develocity instance, Apache Pekko will have access not only to all of the published build scans but other aggregate data features such as:

* Dashboards to view all historical build scans, along with performance trends over time
* Build failure analytics for enhanced investigation and diagnosis of build failures
* Test failure analytics to better understand trends and causes around slow, failing, and flaky tests
 
Develocity is already used by other Apache projects such as [Beam](https://ge.apache.org/scans?search.rootProjectNames=beam&search.timeZoneId=America%2FChicago), [Kafka](https://ge.apache.org/scans?search.rootProjectNames=kafka&search.timeZoneId=America%2FChicago), [IoTDB](https://ge.apache.org/scans?search.rootProjectNames=Apache%20IoTDB%20Project%20Parent%20POM&search.timeZoneId=America%2FChicago), [Pulsar](https://ge.apache.org/scans?search.rootProjectNames=pulsar&search.timeZoneId=America%2FChicago), and more. Develocity is also used by a number of other [OSS projects](https://gradle.com/enterprise-customers/oss-projects/) for which we sponsor instances of Develocity.

Please let me know if there are any questions about the value of Develocity or the changes in this pull request and I’d be happy to address them.

### DIFF
--- a/project/PekkoDevelocityPlugin.scala
+++ b/project/PekkoDevelocityPlugin.scala
@@ -36,7 +36,7 @@ object PekkoDevelocityPlugin extends AutoPlugin {
 
   override lazy val buildSettings: Seq[Setting[_]] = Def.settings(
     develocityConfiguration := {
-      val isInsideCi = insideCI.value
+      val isInsideCI = insideCI.value
 
       val original = develocityConfiguration.value
       val apacheDevelocityConfiguration =
@@ -49,11 +49,11 @@ object PekkoDevelocityPlugin extends AutoPlugin {
           .withBuildScan(
             original.buildScan
               .withPublishing(Publishing.onlyIf(_.authenticated))
-              .withBackgroundUpload(!isInsideCi)
+              .withBackgroundUpload(!isInsideCI)
               .withObfuscation(
                 original.buildScan.obfuscation
                   .withIpAddresses(_.map(_ => ObfuscatedIPv4Address))))
-      if (isInsideCi) {
+      if (isInsideCI) {
         apacheDevelocityConfiguration
           .withTestRetryConfiguration(
             original.testRetryConfiguration
--- a/project/PekkoDevelocityPlugin.scala
+++ b/project/PekkoDevelocityPlugin.scala
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.gradle.develocity.agent.sbt.DevelocityPlugin
+import com.gradle.develocity.agent.sbt.DevelocityPlugin.autoImport.{
+  develocityConfiguration,
+  FlakyTestPolicy,
+  ProjectId,
+  Publishing
+}
+import sbt.{ url, AutoPlugin, Def, PluginTrigger, Plugins, Setting }
+import sbt.Keys.insideCI
+
+object PekkoDevelocityPlugin extends AutoPlugin {
+
+  private val ApacheDevelocityUrl = url("https://ge.apache.org")
+  private val PekkoProjectId = ProjectId("pekko")
+  private val ObfuscatedIPv4Address = "0.0.0.0"
+
+  override lazy val trigger: PluginTrigger = allRequirements
+  override lazy val requires: Plugins = DevelocityPlugin
+
+  override lazy val buildSettings: Seq[Setting[_]] = Def.settings(
+    develocityConfiguration := {
+      val isCI = insideCI.value
+
+      val original = develocityConfiguration.value
+      val apacheDevelocityConfiguration =
+        original
+          .withProjectId(PekkoProjectId)
+          .withServer(
+            original.server
+              .withUrl(Some(ApacheDevelocityUrl))
+              .withAllowUntrusted(false))
+          .withBuildScan(
+            original.buildScan
+              .withPublishing(Publishing.onlyIf(_.authenticated))
+              .withBackgroundUpload(!isCI)
+              .withObfuscation(
+                original.buildScan.obfuscation
+                  .withIpAddresses(_.map(_ => ObfuscatedIPv4Address))))
+      if (isCI) {
+        apacheDevelocityConfiguration
+          .withTestRetryConfiguration(
+            original.testRetryConfiguration
+              .withMaxRetries(1)
+              .withFlakyTestPolicy(FlakyTestPolicy.Fail) // preserve the original build outcome in case of flaky tests
+          )
+      } else apacheDevelocityConfiguration
+    })
+}
--- a/project/PekkoDevelocityPlugin.scala
+++ b/project/PekkoDevelocityPlugin.scala
@@ -36,7 +36,7 @@ object PekkoDevelocityPlugin extends AutoPlugin {
 
   override lazy val buildSettings: Seq[Setting[_]] = Def.settings(
     develocityConfiguration := {
-      val isCI = insideCI.value
+      val isInsideCi = insideCI.value
 
       val original = develocityConfiguration.value
       val apacheDevelocityConfiguration =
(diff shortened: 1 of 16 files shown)


c00ler:
@pjfanning Thank you for the fast feedback! 

> This PR is causing memory issues in the build. Leading to outright failures.

This is something unexpected. We've done many tests with pekko and other projects and never seen OOM. 
I've noticed one interesting detail. In logs there are GC pressure warnings before the OOM, see [example](https://github.com/apache/pekko/actions/runs/8819286268/job/24210600731?pr=1288#step:6:855): 

```
[warn] In the last 10 seconds, 12.015 (127.8%) were spent in GC. [Heap: 0.01GB free of 3.00GB, max 3.00GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.
```  

The warning says that max heap size is 3GB. 3GB is the default value defined in [.sbtopts](https://github.com/apache/pekko/blob/main/.sbtopts#L1). While on CI there is an extra step that adds .jvmopts that raises the maximum size to [4GB](https://github.com/apache/pekko/blob/main/.jvmopts-ci#L5). If I do `cp .jvmopts-ci .jvmopts` locally and run `sbt -v` I get:

```
~/dev/github/apache/pekko on  main [!?] via 🆂 v2.13.13 22ms
🕙 21:38:23 ➜ sbt -v
[sbt_options] declare -a sbt_options='()'
[process_args] java_version = '11'
[copyRt] java9_rt = '/Users/avenderov/.sbt/1.0/java9-rt-ext-eclipse_adoptium_11_0_22/rt.jar'
# Executing command line:
java
-XX:+UseG1GC
-Xms4G
-Xmx4G
-Xss2M
-XX:+AlwaysActAsServerClassMachine
-XX:ReservedCodeCacheSize=256m
-XX:MaxGCPauseMillis=750
-XX:+UseCompressedOops
-XX:MetaspaceSize=512M
-XX:-ClassUnloadingWithConcurrentMark
-Djava.security.egd=file:/dev/./urandom
-Dpekko.ci-server=true
-Xmx3072M
-Xms1024M
-Dmultinode.XX:MetaspaceSize=128M
-XX:+AlwaysActAsServerClassMachine
-Dsbt.script=/Users/avenderov/.sdkman/candidates/sbt/current/bin/sbt
-Dscala.ext.dirs=/Users/avenderov/.sbt/1.0/java9-rt-ext-eclipse_adoptium_11_0_22
-jar
/Users/avenderov/.sdkman/candidates/sbt/1.9.7/bin/sbt-launch.jar
```
The command line has duplicated `-Xmx` values from both files `.sbtopts` and `.jvmopts`. Can it be that there is some indeterministic jvm version specific behaviour and this time the later lower max values was chosen? After reading the pipeline code I was expecting that 4Gb heap is used on CI. Unfortunately I do not have permissions to re-run the job to have another data point.  

I will continue the investigation tomorrow.

> 
> We'll need a CLA. Individual or Corporate.

My employer Gradle has a written agreement in place with the Apache Foundation. I don't know whether it covers such contributions or not. It is not a problem, I can sign iCLA meanwhile.

> 
> We do not have an access key to test with yet. I have requested one from ASF Infra team

I think it's because the PR is running from the fork. The secret should be available after the change is merged. Please see [Apache Beam pipeline](https://github.com/apache/beam/blob/master/.github/workflows/beam_PostCommit_Java.yml#L49) or [Apache Lucene pipeline](https://github.com/apache/lucene/blob/main/.github/workflows/run-checks-all.yml#L16). `GRADLE_ENTERPRISE_ACCESS_KEY` is the old name that we were using before renaming. Do you think Pekko still might need some additional configuration to get access to it?

He-Pin:
Double xmx is unexpected,  thanks for look into this.

pjfanning:
> I think it's because the PR is running from the fork. The secret should be available after the change is merged. Please see [Apache Beam pipeline](https://github.com/apache/beam/blob/master/.github/workflows/beam_PostCommit_Java.yml#L49) or [Apache Lucene pipeline](https://github.com/apache/lucene/blob/main/.github/workflows/run-checks-all.yml#L16). `GRADLE_ENTERPRISE_ACCESS_KEY` is the old name that we were using before renaming. Do you think Pekko still might need some additional configuration to get access to it?

Until the ASF Infra team respond to my JIRA issue, I don't know what we need to do. We will need to make some changes to our .github workflow files but without Infra input, we are blocked.

He-Pin:
@c00ler would you like to update the branch? The xmx issue is fixed now

project/PekkoDevelocityPlugin.scala
```diff
@@ -0,0 +1,65 @@
...
+              .withUrl(Some(ApacheDevelocityUrl))
+              .withAllowUntrusted(false))
+          .withBuildScan(
+            original.buildScan
+              .withPublishing(Publishing.onlyIf(_.authenticated))
+              .withBackgroundUpload(!isCI)
+              .withObfuscation(
+                original.buildScan.obfuscation
+                  .withIpAddresses(_.map(_ => ObfuscatedIPv4Address))))
+      if (isCI) {
```
He-Pin, line 56: So if it's not running inside a CI, eg running locally, nothing will happen?

pjfanning (changes requested):
just blocking merge until we get INFRA to set up our CI so that it supports an access key

project/PekkoDevelocityPlugin.scala
c00ler: Yes. There are two things in this configuration with regards local and CI runs. 

First is this condition `.withPublishing(Publishing.onlyIf(_.authenticated))`. It ensures that the Build Scan will not be published to ge.apache.org (this server is running on ASF infrastructure and Gradle does not have access to it) unless there is an authentication token. On CI the token is provided via the environment variable added in this PR. On CI the predicate will evaluate to `true` and the Build Scan will be published. Locally the predicate will evaluate to `false` unless the user explicitly authenticates with the Develocity instance. The standard way to authenticate is to execute `develocityProvisionAccessKey` task as described [here](https://docs.gradle.com/develocity/sbt-plugin/#authenticating). In order to be able to do so the user needs to have an account at ge.apache.org. Only Apache committers can have an account there. After that the Build Scan will be published from the local computer. To stop publishing from the local computer, users would need to delete the file with the access key.

The second is test retries. By default test retries are disabled. Yes, locally, even if the user is authenticated, due to this condition, the retries will be disabled. It make sense to enable retries on CI, because it will allow Develocity to detect whether the test is flaky or not. This is an [example](https://ge.apache.org/scans/tests?search.rootProjectNames=kafka&search.timeZoneId=Europe%2FBerlin&tests.container=org.apache.kafka.tools.MetadataQuorumCommandTest&tests.test=testDescribeQuorumReplicationSuccessful()%5B1%5D) test result for a test in Apache Kafka project. In the Build Scan it will be possible to see whether the test was really failing or it was flaky. We choose to provide a very conservative retry configuration. Only 1 retry attempt (does not add much to the build time) and the build outcome will still be `failed`. This approach allows us to detect and surface the flakyness, but the decision, whether to proceed with the failed build or not, is still on the team.

project/PekkoDevelocityPlugin.scala
```diff
@@ -0,0 +1,65 @@
...
+  private val ApacheDevelocityUrl = url("https://ge.apache.org")
+  private val PekkoProjectId = ProjectId("pekko")
+  private val ObfuscatedIPv4Address = "0.0.0.0"
+
+  override lazy val trigger: PluginTrigger = allRequirements
+  override lazy val requires: Plugins = DevelocityPlugin
+
+  override lazy val buildSettings: Seq[Setting[_]] = Def.settings(
+    develocityConfiguration := {
+      val isCI = insideCI.value
```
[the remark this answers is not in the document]
c00ler: Done, with a small difference. I use upper case i `isInsideCI` in order to make it aligned with the setting name.

gmcdonald:
> just blocking merge until we get INFRA to set up our CI so that it supports an access key

This is INFRA, confirming what has already been said. All projects already have the required access to the secret.
As is normal, all PRs coming from FORKS will never have such access.

pjfanning:
> > just blocking merge until we get INFRA to set up our CI so that it supports an access key
> 
> This is INFRA, confirming what has already been said. All projects already have the required access to the secret. As is normal, all PRs coming from FORKS will never have such access.

Thanks @gmcdonald. Can I confirm that the secret is called `GE_ACCESS_TOKEN`?

gmcdonald:
> > > just blocking merge until we get INFRA to set up our CI so that it supports an access key
> > 
> > 
> > This is INFRA, confirming what has already been said. All projects already have the required access to the secret. As is normal, all PRs coming from FORKS will never have such access.
> 
> Thanks @gmcdonald. Can I confirm that the secret is called `GE_ACCESS_TOKEN`?

It is indeed

pjfanning (approved):
Lgtm. Let's merge this to see if results are sent to ge.apache.org. We can revert if there are problems.

pjfanning:
@c00ler I couldn't find any pekko results on ge.apache.org. This PR merge led to some builds but they not seem to log to Develocity.

c00ler:
> @c00ler I couldn't find any pekko results on ge.apache.org. This PR merge led to some builds but they not seem to log to Develocity.

@pjfanning Yes, my bad. I was also monitoring the build. I put a wrong env var name.  `DEVELOCITY_ACCESS_TOKEN` must have been `DEVELOCITY_ACCESS_KEY` 🤦 I will create a quick PR. On a positive side, it showed that without a proper authentication nothing is being published.

PR #1292, opened 2024-04-25
Fix javadoc link to include module name

c00ler:
This PR fixes logback AsyncAppender javadoc link in the documentation.

Fixes https://github.com/apache/pekko/issues/1291

### DIFF
--- a/docs/src/main/paradox/typed/logging.md
+++ b/docs/src/main/paradox/typed/logging.md
@@ -199,7 +199,7 @@ A starting point for configuration of `logback.xml` for production:
 
 @@snip [logback.xml](/actor-typed-tests/src/test/resources/logback-doc-prod.xml)
 
-Note that the [AsyncAppender](https://logback.qos.ch/apidocs/ch/qos/logback/classic/AsyncAppender.html) may drop log events if the queue becomes full, which may happen if the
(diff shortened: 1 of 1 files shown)


Roiocam:
Thanks @c00ler, Do you know if the website for logback has been migrated or if it was mistake changed due to a certain iteration? I created an [issue](https://github.com/apache/pekko/issues/1291) but haven't received any reply. The slf4j under the same domain doesn't have a similar URL path.

He-Pin (approved):
lgtm

c00ler:
> Thanks @c00ler, Do you know if the website for logback has been migrated or if it was mistake changed due to a certain iteration? I created an [issue](https://github.com/qos-ch/logback-site/issues/13) but haven't received any reply. The slf4j under the same domain doesn't have a similar URL path.

@Roiocam I believe that the new url is an intentional and permanent change. The url is pointing to the latest javadoc that corresponds to version `1.5.x` of logback. The latest version is using java modules (see [module-info.java](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/module-info.java)). The module name `ch.qos.logback.classic` just happens to be the same as the package and this is why in the url it looks like a duplicate, but its different. It looks like that logback 1.5.x is using Java 21 toolchain to create javadoc and the new javadoc layout includes information about modules. I've checked javadocs for older logback versions and the latest version that uses the old layout (without modules) is `1.3.14` (latest version from `1.3.x` branch).

PR #1295, opened 2024-04-25
Fix Develocity access key environment variable name

c00ler:
Follow up PR for the change that was done in https://github.com/apache/pekko/pull/1288, where the wrong variable name was used. Due to the wrong name the Build Scan was not published. 

Develocity docs about environment variable can be found [here](https://docs.gradle.com/develocity/sbt-plugin/#via_environment_variable)

### DIFF
--- a/.github/workflows/scala3-build.yml
+++ b/.github/workflows/scala3-build.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Compile and run tests on Scala 3
         env:
-          DEVELOCITY_ACCESS_TOKEN: ${{ secrets.GE_ACCESS_TOKEN }}
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
         # note that this is not running any multi-jvm tests (yet) because multi-in-test=false
         run: |
           sbt \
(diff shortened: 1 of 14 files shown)


.github/workflows/scala3-build.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       - name: Compile and run tests on Scala 3
         env:
-          DEVELOCITY_ACCESS_TOKEN: ${{ secrets.GE_ACCESS_TOKEN }}
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
```
He-Pin, line 56: The parameter variable should be key too, better to keep the same

clayburn: The `GE_ACCESS_TOKEN` secret is controlled by ASF Infra and in use by many other projects. At some point, we will migrate the name, but it would be a larger effort to do that, given other projects use the secret.

pjfanning: secrets.GE_ACCESS_TOKEN is what is setup by the ASF Infra team - we have no control over this

DEVELOCITY_ACCESS_KEY is the name expected by Develocity - we have no control over this

pjfanning (approved):
lgtm

.github/workflows/scala3-build.yml
He-Pin: Thanks for the update , it's much cleaner now

PR #1299, opened 2024-04-27
feat: Add virtual thread support

He-Pin:
Motivation:
Add virtual thread support

Modification:
Add a virtual thread executor

Result:
Virtual thread support was added.

TODO:
I think it would be nice to add the blocker dispatcher with a virtual thread executor too.

Status:
I ran out of time, will update this when get time.


actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
```diff
@@ -0,0 +1,51 @@
...
+import java.lang.invoke.{ MethodHandles, MethodType }
+import java.util.concurrent.ThreadFactory
+
+@InternalApi
+private[dispatch] object VirtualThreadSupport {
+
+  /**
+   * Returns if the current Runtime supports virtual threads.
+   */
+  def isSupported: Boolean = create("testIsSupported") ne null
```
He-Pin, line 31: better with a lazy val, will update later.

actor/src/main/scala/org/apache/pekko/dispatch/NewVirtualThreadPerTaskExecutor.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  // TODO AS only this execute method is been used in `Dispatcher.scala`, so `submit` and other methods is not override.
+  override def execute(command: Runnable): Unit = {
+    if (state.get() >= SHUTDOWN) {
+      throw new RejectedExecutionException("Shutdown")
+    }
+    var started = false;
+    try {
+      val thread = threadFactory.newThread(Task(this, command))
+      virtualThreads.add(thread)
```
alexandru, line 117: This logic here is obviously thread-unsafe because `stage.get() >= SHUTDOWN` can be reordered with `virtualThreads.add(thread)`.

In other words, you can add start and add new threads that won't be waited on while the thread-pool is shutting down.

I don't know if this is a concern we need to have, or how the other implementations do it.

He-Pin: Thanks, would you like to continue the work? I will continue this when find time.

I have a isStarted check beblow, that would help? 
Another issue is, only execute method is been used, not sure what the origin design requires an ExecutionService.

actor/src/main/scala/org/apache/pekko/dispatch/NewVirtualThreadPerTaskExecutor.scala
```diff
@@ -0,0 +1,153 @@
...
+
+private[dispatch] class NewVirtualThreadPerTaskExecutor(threadFactory: ThreadFactory) extends AbstractExecutorService {
+  import NewVirtualThreadPerTaskExecutor._
+
+  /**
+   * 0 RUNNING
+   * 1 SHUTDOWN
+   * 2 TERMINATED
+   */
+  private val state = new AtomicInteger(RUNNING)
```
He-Pin, line 34: will replace with a single field and a fieldUpdater later

pjfanning:
This is probably useful but I would appreciate if we don't merge it until after 1.1.0-M1 is released. I think there are some changes that we need in the implementation and we have to work out how to test it.

I don't this dispatcher should be the default, even if you use JDK 21+. Users should only be able to opt in to using it.

Dispatchers are pluggable so users can create their own dispatcher for this while they are waiting for us to release one.

He-Pin:
Yes, I would like to make it in 1.1.0 M2 or M3 , Need more time to polish it.

actor/src/main/scala/org/apache/pekko/dispatch/NewVirtualThreadPerTaskExecutor.scala
```diff
@@ -0,0 +1,153 @@
...
+
+package org.apache.pekko.dispatch
+
+import java.util
+import java.util.Collections
+import java.util.concurrent._
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.locks.ReentrantLock
+
+private[dispatch] class NewVirtualThreadPerTaskExecutor(threadFactory: ThreadFactory) extends AbstractExecutorService {
```
He-Pin, line 26: I will use method handle to reflection create this instead of this class

He-Pin:
@pjfanning @alexandru @Roiocam @alexandru What do you think about the current state?

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
```diff
@@ -0,0 +1,75 @@
...
+  def newVirtualThreadFactory(prefix: String): ThreadFactory = {
+    require(isSupported, "Virtual thread is not supported.")
+    try {
+      val builderClass = ClassLoader.getSystemClassLoader.loadClass("java.lang.Thread$Builder")
+      val ofVirtualClass = ClassLoader.getSystemClassLoader.loadClass("java.lang.Thread$Builder$OfVirtual")
+      val ofVirtualMethod = lookup.findStatic(classOf[Thread], "ofVirtual", MethodType.methodType(ofVirtualClass))
+      var builder = ofVirtualMethod.invoke()
+      val nameMethod = lookup.findVirtual(ofVirtualClass, "name",
+        MethodType.methodType(ofVirtualClass, classOf[String], classOf[Long]))
+      // TODO support replace scheduler when we drop Java 8 support
```
He-Pin, line 49:
```scala
      if (executor != null) {
        val clazz = builder.getClass
        val privateLookup = MethodHandles.privateLookupIn(
          clazz,
          lookup
        )
        val schedulerFieldSetter = privateLookup
          .findSetter(clazz, "scheduler", classOf[Executor])
        schedulerFieldSetter.invoke(builder, executor)
      }
```

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.pekko.annotation.InternalApi
+import org.apache.pekko.util.JavaVersion
+
+import java.lang.invoke.{ MethodHandles, MethodType }
+import java.util.concurrent.{ ExecutorService, ThreadFactory }
+import scala.util.control.NonFatal
+
+@InternalApi
+private[dispatch] object VirtualThreadSupport {
+  private val lookup = MethodHandles.lookup
```
pjfanning, line 29: can you use [public lookup](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/MethodHandles.html#publicLookup--)?

actor/src/main/resources/reference.conf
```diff
@@ -539,6 +540,19 @@ pekko {
...
       }
 
+      # This will be used if you have set "executor = "virtual-thread-executor"
+      # This executor will execute the every task on a new virtual thread.
+      # Underlying thread pool implementation is java.util.concurrent.ForkJoinPool for JDK <= 22
+      # If the current runtime does not support virtual thread,
+      # then the executor configured in "fallback" will be used.
+      virtual-thread-executor {
+        #Please set the the underlying pool with system properties below:
+        #jdk.virtualThreadScheduler.parallelism
```
pjfanning, line 550: can these be supported as configs as opposed to relying on system properties?

He-Pin: Yes, but that will need to create a dedicated fork-join pool with a new factory for the `CarrierThread`( but which lives in java.internal.misc`...  we have that internal but not sure if that's a good idea for an opensource project.

pjfanning:
Could you add a basic unit test?

pjfanning:
@He-Pin would you have time to add a test that runs when we build with java 21+? And there is my comment about using MethodHandles publicLookup instead of lookup. Generally, I support adding this. I'm hoping to try to get the final changes merged so that we can press ahead with Pekko 1.1.0 or 1.1.0-M2 release.

He-Pin:
I'm a little busy at work but I would like to address this this weekend night, at day I need got the apache coc in HangZhou.

pjfanning:
#1436 was merged

PR #1304, opened 2024-04-30
chore: avoid the double evaluation of entityId in ClusterSharding

Roiocam:
Found something we could optimized it.

### DIFF
--- a/cluster-sharding-typed/src/main/scala/org/apache/pekko/cluster/sharding/typed/internal/ClusterShardingImpl.scala
+++ b/cluster-sharding-typed/src/main/scala/org/apache/pekko/cluster/sharding/typed/internal/ClusterShardingImpl.scala
@@ -173,10 +173,10 @@ import pekko.util.JavaDurationConverters._
       allocationStrategy: Option[ShardAllocationStrategy]): ActorRef[E] = {
 
     val extractorAdapter = new ExtractorAdapter(extractor)
-    val extractEntityId: ShardRegion.ExtractEntityId = {
-      // TODO is it possible to avoid the double evaluation of entityId
-      case message if extractorAdapter.entityId(message) != null =>
-        (extractorAdapter.entityId(message), extractorAdapter.unwrapMessage(message))
+    val extractEntityId: ShardRegion.ExtractEntityId = { message =>
+      extractorAdapter.entityId(message) match {
+        case eid => (eid, extractorAdapter.unwrapMessage(message))
+      }
     }
     val extractShardId: ShardRegion.ExtractShardId = { message =>
       extractorAdapter.entityId(message) match {
--- a/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ClusterSharding.scala
+++ b/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ClusterSharding.scala
@@ -425,15 +425,19 @@ class ClusterSharding(system: ExtendedActorSystem) extends Extension {
       allocationStrategy: ShardAllocationStrategy,
       handOffStopMessage: Any): ActorRef = {
 
+    val extractEntityId: ShardRegion.ExtractEntityId = { msg =>
+      messageExtractor.entityId(msg) match {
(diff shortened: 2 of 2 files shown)


cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ClusterSharding.scala
```diff
@@ -425,15 +425,19 @@ class ClusterSharding(system: ExtendedActorSystem) extends Extension {
       allocationStrategy: ShardAllocationStrategy,
       handOffStopMessage: Any): ActorRef = {
 
+    val extractEntityId: ShardRegion.ExtractEntityId = { msg =>
+      messageExtractor.entityId(msg) match {
+        case eid => (eid, messageExtractor.entityMessage(msg))
+      }
+    }
+    val extractShardId: ShardRegion.ExtractShardId = msg => messageExtractor.shardId(msg)
```
He-Pin, line 433: Should this be privare

Roiocam: Thanks, i think this PR wasn't correct. I will find another solution fix this.

laglangyue:
Seems like duplicate code, can it be abstracted

pjfanning (changes requested):
blocking - needs discussion on the dev list because of the ongoing release discussion

Roiocam:
Use the cacheable partial function, and then verify it works.

```scala

var calTime: Int = 0
def entityId(value: Any): String = {
  calTime = calTime + 1
  value match {
    case "wrong" => null
    case _ => String.valueOf(value)
  }
}

def extractEntityIdFromExtractor: PartialFunction[Any, Any] =
  new scala.runtime.AbstractPartialFunction[Any, (String, Any)] {
    var cache: String = _

    override def isDefinedAt(msg: Any): Boolean = {
      cache = entityId(msg)
      cache != null
    }

    override def apply(x: Any): (String, Any) = (cache, x)
  }

lazy val xx = extractEntityIdFromExtractor

lazy val yy: PartialFunction[Any, Any] = {
  case message if entityId(message) != null => (entityId(message), message)

}

assert(!xx.isDefinedAt("wrong"))
assert(!yy.isDefinedAt("wrong"))
calTime = 0
val health = "health"
println(xx.isDefinedAt(health))
println(xx(health))
println(s"caltime = $calTime")
println("=====")
calTime = 0
println(yy.isDefinedAt(health))
println(yy(health))
println(s"caltime = $calTime")



```

pjfanning (approved):
lgtm

pjfanning:
@Roiocam is it ok to get this merged now?

cluster-sharding-typed/src/main/scala/org/apache/pekko/cluster/sharding/typed/internal/ClusterShardingImpl.scala
```diff
@@ -173,10 +174,16 @@ import pekko.util.JavaDurationConverters._
...
+    // !!!important is only applicable if you know that isDefinedAt(x) is always called before apply(x) (with the same x)
+    val extractEntityId: ShardRegion.ExtractEntityId = new AbstractPartialFunction[Any, (String, Any)] {
+      var cache: String = _
+
+      override def isDefinedAt(msg: Any): Boolean = {
+        cache = extractorAdapter.entityId(msg)
+        cache != null
+      }
+
+      override def apply(x: Any): (DataCenter, Any) = (cache, extractorAdapter.unwrapMessage(x))
```
nvollmar, line 186: Shouldn't that be `(String, Any)`?

Roiocam: Yep, Looks like it was a mistake.

Roiocam: Fixed, thanks!

nvollmar (approved):
lgtm

PR #1313, opened 2024-05-07
add bcprov test dependency due to CVE

pjfanning:
see https://github.com/apache/pekko/security/dependabot/101

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -118,6 +118,7 @@ object Dependencies {
     }
 
     object TestDependencies {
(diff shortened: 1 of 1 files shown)


laglangyue:
the link is 404

nvollmar (approved):
lgtm

He-Pin (approved):
lgtm

raboof (commented):
Not sure we should care about all CVEs in test dependencies.

Should we add a comment specifying when we can remove this override again?

PR #1331, opened 2024-05-17
fix startsWith condition in nightly-builds.yml

pjfanning:
I accidentally changed the condition from `!=` to startsWith in https://github.com/apache/pekko/pull/1329

Needs to be 'not startsWith'.


He-Pin (approved):
lgtm

Issue #1327, opened 2024-05-15
EventSourcedBehavior/Durable State exposed to stack overflow when lots of read-only commands are in the stash


JacobF7:
Both EventSourcedBehavior and Durable state seem to be exposed to stack overflow exceptions when lots of read-only commands are in the stash. This bug has already been opened as an [issue](https://github.com/akka/akka/issues/29933) for Akka and a fix has also been recently merged. Are there any plans of fixing this in Pekko too?

pjfanning:
Legally, we can't take any code that is committed to Akka since its license changed. If someone wants to attempt a clean room fix, that would be great.

He-Pin:
@JacobF7 would you like to prepare a reproducer? I'm not using persistent at work.

pjfanning:
I've read the https://github.com/akka/akka/issues/29933 but don't want to look at the PR since the license is not compatible with Apache Pekko. It seems like the issue is that 'tryUnstashOne will, after a few steps, call the onCommand method that in turn will call tryUnstashOne'.

So it seems like we need to make tryUnstashOne non-recursive. We may want to allow a small number of recursive calls but after a certain depth, we will need to use something like sending a message to continue the unstashing. There is a good chance that we will need to wait for a response message that tells us the async unstashing has completed.

Alternatively, we could look at turning this into a loop instead of using recursion.

pjfanning:
@JacobF7 Can I get some background? I reproduced the issue but with a use case that is not real world. Have you hit this issue in the real world?

I have messed about with some changes but in all the honesty, they tend to break stuff.

If someone else wants to try their own solution, please feel free to have a look. Just a reminder, we can't accept any solutions that rely on examining the Akka changes.

One extra potential solution is to ignore the read only events if there are just too many of them. In my testing, it takes > 1000 such events to cause an issue. Is there a good reason not to limit the number of read only events?

He-Pin:
@JacobF7 Would you like to provide a reproducer? thanks.

JacobF7:
[...]

PR #1336, opened 2024-05-21
reproduce issue where stackoverflow happens in unstash

pjfanning:
see #1327 

This doesn't cause the test to fail or abort - for some reason testkit/scalatest logs the stackoverflow but then claims no tests were run.

```
[info] Run completed in 3 seconds, 635 milliseconds.
[info] Total number of tests run: 0
[info] Suites: completed 0, aborted 0
[info] Tests: succeeded 0, failed 0, canceled 0, ignored 0, pending 0
[info] No tests were executed.
[info] No tests to run for persistence-typed-tests / Test / testOnly
```


pjfanning:
I have a possible solution but it is breaking quite a few tests so it needs more debugging. It is only one approach but by passing back a behavior wrapping the unstash instead of running the unstash in place - this avoids stack overflow issues.
The broken tests show that it is not quite this simple and it may prove to the dead end.

pjfanning:
> I have a possible solution but it is breaking quite a few tests so it needs more debugging. It is only one approach but by passing back a behavior wrapping the unstash instead of running the unstash in place - this avoids stack overflow issues. The broken tests show that it is not quite this simple and it may prove to be a dead end.

This 'solution' isn't working out.

persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
```diff
@@ -55,10 +55,11 @@ object EventSourcedStashOverflowSpec {
     SteppingInmemJournal.config("EventSourcedStashOverflow").withFallback(ConfigFactory.parseString(s"""
        pekko.persistence {
          typed {
-           stash-capacity = 1000 # enough to fail on stack size
+           stash-capacity = 2000 # enough to fail on stack size
```
Roiocam, line 58: That's will cause more stack trace frame, that will event worst.

Maybe we could unstash with counter, after limitation, we will start another recursion(clear stack)

Roiocam: I see, thank you.

Issue #1341, opened 2024-05-24
With entity passivation enabled the shard region will cause a lot of unhandled ShardsUpdated during shutdown


nvollmar:
We encounter hundreds of these during a node shutdown
```
unhandled message from Actor[pekko://main/system/sharding/transactionsAggregates#-1585167465] to Actor[pekko://main/system/sharding/transactionsAggregates/4#-516417762]: ShardsUpdated(5)
```

Root cause is  to link  to link `ShardRegion` sending updates for each terminated shard when passivation strategy is used:
```
      if (settings.passivationStrategy != ClusterShardingSettings.NoPassivationStrategy) {
        shards.values.foreach(_ ! ShardsUpdated(shards.size))
      }
```
See [here](https://github.com/apache/pekko/blob/3fff73894bf339722b2e4d7a4a5491b1de4ca6a9/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ShardRegion.scala#L1006)

That message is not handled/ignored by the `Shard` during termination:
```
       context.become {
          case Terminated(ref) => receiveTerminated(ref)
        }
```
See [here](https://github.com/apache/pekko/blob/3fff73894bf339722b2e4d7a4a5491b1de4ca6a9/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/Shard.scala#L895)

There would be two ways to fix this:

1. At this time the `ShardRegion` property `gracefulShutdownInProgress` is already set to `true` - check for that in the if and don't send any more updates 
2. During termination in the `Shard` match and ignore the `ShardsUpdated`

PR #1342, opened 2024-05-24
Avoid unnecessary shard updates while shutting down

nvollmar:
Refs #1341


pjfanning:
looks sensible to me - let's see what other reviewers say

When we get a fix in, we might want to backport it to 1.0.x. I think we should make a 1.0.3 release in the next few weeks.

nvollmar:
Would be good to have that in 1.0.3

samueleresca (approved):
looks good to me.

He-Pin (approved):
lgtm

Roiocam (approved):
LGTM, Thanks.

`ShardUpdate` will only notify Shard to start detecting the passivation. This is seems like a harmless change.

Issue #325, opened 2023-05-22
Provide reliable/reproducible builds for Scala 3


mdedetrich:
Currently blocked by https://github.com/lampepfl/dotty/issues/17330. In the meantime should document that artifacts for Scala 3 are not reproducible.

mdedetrich:
Found another case of non determinism, issue filed at https://github.com/lampepfl/dotty/issues/18248

raboof:
https://github.com/scala/scala3/issues/20496 is now fixed, and we haven't seen https://github.com/scala/scala3/issues/18248 recently, so let's tentatively close this issue for now.

PR #1349, opened 2024-05-30
fix: improve workaround for Scala3 bug 18248

raboof:
Towards #325

Hopefully works around https://github.com/scala/scala3/issues/18248

(the diff looks larger than it is, it just moves the two inner functions to regular private functions)


pjfanning (approved):
lgtm

PR #1353, opened 2024-06-04
Avoid unnecessary shard updates while shutting down (#1342)

nvollmar:
* Avoid unnecessary shard updates while shutting down
* Avoid re-creating the same message for each shard

cherry pick 92b38e72d6a2701da06f483e8cd27953e403240e


pjfanning (approved):
lgtm

PR #1366, opened 2024-06-15
feat: Support adjust the scheduler's tick-duration.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1364

Modification:
Add another config `pekko.scheduler.error-on-tick-duration-verification-failed = on` by default and can be changed to `off` when needed.

Result:
Better experience

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
@@ -26,18 +26,26 @@ import scala.util.control.NonFatal
 import atomic.{ AtomicInteger, AtomicReference }
 import scala.annotation.nowarn
 import com.typesafe.config.{ Config, ConfigFactory }
+
 import language.postfixOps
 import org.scalatest.BeforeAndAfterEach
-
 import org.apache.pekko
 import pekko.pattern.ask
 import pekko.testkit._
+import pekko.util.Helpers
 
 object SchedulerSpec {
   val testConfRevolver =
     ConfigFactory.parseString("""
     pekko.scheduler.implementation = org.apache.pekko.actor.LightArrayRevolverScheduler
     pekko.scheduler.ticks-per-wheel = 32
+  """).withFallback(PekkoSpec.testConf)
+
+  val testConfAdjustment =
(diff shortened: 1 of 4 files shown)


jxnu-liguobin (approved):

actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
```diff
@@ -40,6 +41,13 @@ object SchedulerSpec {
     pekko.scheduler.ticks-per-wheel = 32
   """).withFallback(PekkoSpec.testConf)
 
+  val testConfAdjustment =
+    ConfigFactory.parseString("""
+    pekko.scheduler.implementation = org.apache.pekko.actor.LightArrayRevolverScheduler
+    pekko.scheduler.tick-duration = 1ns
```
jxnu-liguobin, line 47: ns？

He-Pin: Yes, I want to test that case

actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -54,18 +54,35 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
...
+    if (Helpers.isWindows && durationFromConfig < 10.millis) {
+      if (errorOnVerificationFailed) {
+        throw new IllegalArgumentException(
+          "requirement failed: minimum supported pekko.scheduler.tick-duration on Windows is 10ms")
+      } else {
+        log.warning(
+          "requirement failed: minimum supported pekko.scheduler.tick-duration on Windows is 10ms, adjusted to 10ms now.")
+        10.millis
+      }
+    } else if (durationFromConfig < 1.millis) {
```
pjfanning, line 75: Can you move this inside the other other check to improve performance - avoiding 2 checks in happy path?

appproximately
```
if <10ms {
  if (isWindows) {
    ... windows warning or throw 
  } else if < 1ms {
    ... 1ms warning or throw 
  } else durationFromConfig
}

```

PR #1372, opened 2024-06-17
fix: dependency-graph permissions

raboof:
apparently permissions must be at the job level, not the step level.

Follow-up on #1371


pjfanning (approved):
lgtm

PR #1377, opened 2024-06-25
feat: add QueueClosed singleton instance for Java API

Roiocam:
## Motivation

new API for Java users, used to detect queue completion.

```java
QueueOfferResult offerResult = queue.offer(params);

if (offerResult == QueueOfferResult.enqueued()) {
    // logic...
} else if (offerResult == QueueOfferResult.dropped()) {
    // logic...
} else if (offerResult instanceof QueueOfferResult.Failure) {
    // logic...
} else if (offerResult == QueueOfferResult.closed()) {
    // logic...
} else {
    // logic...
}
```

截屏2024-06-25 10 08 44

https://github.com/apache/pekko/blob/3fabc04acfc472777c85a1bef09bc7f2f777f2e0/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/BoundedSourceQueueSpec.scala#L98-L106

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/QueueOfferResult.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/QueueOfferResult.scala
@@ -76,4 +76,9 @@ object QueueOfferResult {
   case object QueueClosed extends QueueCompletionResult {
     override def isEnqueued: Boolean = false
   }
+
+  /**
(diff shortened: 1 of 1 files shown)


He-Pin (approved):

He-Pin:
You can use `isEqueued` too.

PR #1380, opened 2024-07-01
feat: publish SBOMs

raboof:
While clearly not perfect, I think we should start growing towards publishing valid SBOMs for our artifacts, and publishing these is a good first step in that direction.


pjfanning (approved):
lgtm

PR #1381, opened 2024-07-01
Simplify link validator config

raboof:


raboof:
hmm,

```
coursier.install.AppArtifacts$ScalaDependenciesNotFound: Can't find a scala version suffix for net.runne::site-link-validator:0.2.4 (likely a non existing module or version)
```

... though it works fine on my machine and is at https://repo1.maven.org/maven2/net/runne/site-link-validator_2.13/0.2.4/

raboof:
purged the cache and re-ran the action - not sure which of those fixed it :)

scripts/link-validator.conf
```diff
@@ -43,101 +43,32 @@ site-link-validator {
   ignore-missing-local-files-regex = ""
   // e.g. "^api/alpakka/snapshot/pekko/stream/alpakka/googlecloud/storage/impl/Formats.*"
 
+  ignore-files = [
+    # This file is generated, we can't do much about the invalid links here
+    "project/license-report.html"
```
He-Pin, line 48: Seems ok to me

He-Pin (approved):

pjfanning (approved):
lgtm

scripts/link-validator.conf
Roiocam: see more context: https://github.com/apache/pekko/pull/704

laglangyue:
lgtm

scripts/link-validator.conf
raboof: I'm not sure if I understand the context correctly. It seems the license report was disabled because the report generation is slow, not the link validation. So before we remove that logic, we should fix https://github.com/sbt/sbt-license-report/issues/87 ?

Roiocam: My bad, I thought the ISSUE could also solve the problem of report speed.

.github/workflows/link-validator.yml
```diff
@@ -42,4 +42,4 @@ jobs:
           sbt -Dpekko.genjavadoc.enabled=true -Dpekko.genlicensereport.enabled=true "Javaunidoc/doc; Compile/unidoc; docs/paradox"
```
Roiocam, line 42: if we don't validate license report links, then do we need to generate them?

PR #1386, opened 2024-07-03
feat: support the flattening syntax for supervising

Roiocam:
## MOTIVATION

Continue #680

The original api wasn't flatening.

```java
Behaviors.supervise(
        Behaviors.supervise(behavior)
            .onFailure(IllegalStateException.class, SupervisorStrategy.restart()))
    .onFailure(IllegalArgumentException.class, SupervisorStrategy.stop());
```

## RELATED WORK

- [x] scaladsl
- [x] javadsl
- [x] documentation
- [x] made unit test passed
- [x] binary compatibility

### DIFF
--- a/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
+++ b/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
@@ -133,7 +133,7 @@ public class RouterTest extends JUnitSuite {
               Routers.pool(
                   poolSize,
                   // make sure the workers are restarted if they fail
-                  Behaviors.supervise(Worker.create()).onNotFatalFailure(SupervisorStrategy.restart()));
+                  Behaviors.supervise(Worker.create()).onException(SupervisorStrategy.restart()));
           ActorRef<Worker.Command> router = context.spawn(pool, "worker-pool");
 
           for (int i = 0; i < 10; i++) {
--- a/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
+++ b/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
@@ -133,7 +133,7 @@ public class RouterTest extends JUnitSuite {
               Routers.pool(
                   poolSize,
                   // make sure the workers are restarted if they fail
-                  Behaviors.supervise(Worker.create()).onFailure(SupervisorStrategy.restart()));
+                  Behaviors.supervise(Worker.create()).onNotFatalFailure(SupervisorStrategy.restart()));
           ActorRef<Worker.Command> router = context.spawn(pool, "worker-pool");
 
           for (int i = 0; i < 10; i++) {
--- a/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
+++ b/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
@@ -133,7 +133,7 @@ public class RouterTest extends JUnitSuite {
               Routers.pool(
                   poolSize,
                   // make sure the workers are restarted if they fail
-                  Behaviors.supervise(Worker.create()).onAnyFailure(SupervisorStrategy.restart()));
+                  Behaviors.supervise(Worker.create()).onFailure(SupervisorStrategy.restart()));
           ActorRef<Worker.Command> router = context.spawn(pool, "worker-pool");
 
           for (int i = 0; i < 10; i++) {
--- a/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
+++ b/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
@@ -133,7 +133,7 @@ public class RouterTest extends JUnitSuite {
               Routers.pool(
                   poolSize,
                   // make sure the workers are restarted if they fail
-                  Behaviors.supervise(Worker.create()).onException(SupervisorStrategy.restart()));
+                  Behaviors.supervise(Worker.create()).onAnyFailure(SupervisorStrategy.restart()));
           ActorRef<Worker.Command> router = context.spawn(pool, "worker-pool");
 
           for (int i = 0; i < 10; i++) {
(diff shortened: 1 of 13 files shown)


actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/RouterTest.java
```diff
@@ -133,7 +133,7 @@ static Behavior<Void> showPoolRouting() {
               Routers.pool(
                   poolSize,
                   // make sure the workers are restarted if they fail
-                  Behaviors.supervise(Worker.create()).onFailure(SupervisorStrategy.restart()));
+                  Behaviors.supervise(Worker.create()).onNotFatalFailure(SupervisorStrategy.restart()));
```
Roiocam, line 136: This method signature is because the overload issue between java API and scala API

actor-typed/src/main/scala/org/apache/pekko/actor/typed/Behavior.scala
```diff
@@ -71,6 +69,43 @@ abstract class Behavior[T](private[pekko] val _tag: Int) { behavior =>
...
+  }
+
+  /**
+   * Java API:
+   * Specify the [[SupervisorStrategy]] to be invoked when the wrapped behavior throws.
+   *
+   * Only exceptions of the given type (and their subclasses) will be handled by this supervision behavior.
+   */
+  def onNotFatalFailure[Thr <: Throwable](strategy: SupervisorStrategy): SuperviseBehavior[T] = {
+    onFailure(classOf[Exception], strategy)
```
He-Pin, line 103: Better with NonFatal apply？

pjfanning: onNonFatalFailure would be better

Roiocam: What a about `onException`?

actor-typed/src/main/mima-filters/1.0.x.backwards.excludes/issue-3413.excludes
```diff
@@ -0,0 +1,6 @@
+ProblemFilters.exclude[MissingFieldProblem]("org.apache.pekko.actor.typed.scaladsl.Behaviors.Supervise")
+ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.actor.typed.scaladsl.Behaviors$Supervise$")
+ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.actor.typed.scaladsl.Behaviors$Supervise")
+ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.actor.typed.javadsl.Behaviors$Supervise")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.actor.typed.scaladsl.Behaviors.supervise")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.actor.typed.javadsl.Behaviors.supervise")
```
raboof, line 6: I really think we should not just add things to mima backwards excludes, but also add comments why the exclusions are justified.

In this case it looks, to me, like the mima errors were actually legitimate and this PR breaks our binary compatibility promises documented at https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html and https://github.com/apache/pekko/blob/main/CONTRIBUTING.md#binary-compatibility

To make things explicit: suppose we have an application 'A' which uses a library 'L' which depends on Pekko 1.0.0 . The point of the binary compatibility rules is that it should be safe for 'A' to update to Pekko 1.1.0 without breaking.

In this case, if `L` contained:

```
val s: Behaviors.Supervise = Behaviors.supervise(SupervisedActor.create())
```

... AFAICS this code would throw a runtime exception when executed with any Pekko version that has the change in this PR, because `Behaviors.Supervise` no longer exists.

Roiocam: > I really think we should not just add things to mima backwards excludes, but also add comments why the exclusions are justified.

I totally agree with that.

> In this case it looks, to me, like the mima errors were actually legitimate

Yes. 


What should we do when we face such API changes?

raboof: > What should we do when we face such API changes?

It depends on the situation. I haven't looked closely, but in this case, it looks like the SuperviseBehaviour was introduced to reduce code duplication between the two `Behavior.Supervise` classes, is that correct? I wonder if we could restore binary compatibility while still avoiding code duplication by restoring the `Behavior.Supervise` classes, but having them `extends SuperviseBehaviour`. If that does not work, possibly we should just choose to accept some duplication to achieve compatibility.

Roiocam: I has been fix these binary compatibility via some duplicate code. would you take a look?

actor-typed/src/main/mima-filters/1.0.x.backwards.excludes/issue-3413.excludes
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Change the return type of `Behaviors.supervise` to support flattened supervision
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.actor.typed.javadsl.Behaviors#Supervise.onFailure")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.actor.typed.scaladsl.Behaviors#Supervise.onFailure")
```
raboof, line 20: So these are OK because the old signatures returned `Behavior`, and the new ones return `SuperviseBehavior`, and since `SuperviseBehavior extends Behavior` every `SuperviseBehavior` is a `Behavior`? Sounds reasonable.

actor-typed/src/main/mima-filters/1.0.x.backwards.excludes/issue-3413.excludes
```diff
@@ -0,0 +1,21 @@
...
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.actor.typed.scaladsl.Behaviors#Supervise.onFailure$extension")
```
raboof, line 21: This one as well (checked with `javap`)

raboof (approved):
seems like a nice syntax

Roiocam:
Thanks @raboof, let we merged it.

PR #1388, opened 2024-07-04
chore: migrate tests to flatten supervise API

Roiocam:
continue #1386, separate PR for test migration

### DIFF
--- a/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
+++ b/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
@@ -209,7 +209,6 @@ class StubbedSupervisionSpec extends AnyWordSpec with Matchers with LogCapturing
       supervise(targetBehavior(inboxRef))
         .onFailure[Exc2](SupervisorStrategy.resume)
         .onFailure[Exc3](SupervisorStrategy.restart)
-        .onFailure[Exc1](SupervisorStrategy.stop)
     }
 
     "not catch fatal error" in {
@@ -283,8 +282,7 @@ class StubbedSupervisionSpec extends AnyWordSpec with Matchers with LogCapturing
   }
 }
 
-class SupervisionSpec extends ScalaTestWithActorTestKit(
-      """
+class SupervisionSpec extends ScalaTestWithActorTestKit("""
     pekko.log-dead-letters = off
     """) with AnyWordSpecLike with LogCapturing {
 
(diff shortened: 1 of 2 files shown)


actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
```diff
@@ -1232,34 +1222,33 @@ class SupervisionSpec extends ScalaTestWithActorTestKit("""
     "work with nested supervisions and defers" in {
       val strategy = SupervisorStrategy.restart.withLimit(3, 1.second)
       val probe = TestProbe[AnyRef]("p")
-      val beh = supervise[String](setup(_ =>
-        supervise[String](setup { _ =>
-          probe.ref ! Started
-          Behaviors.empty[String]
-        }).onFailure[RuntimeException](strategy))).onFailure[Exception](strategy)
+      val behv = supervise[String](setup { _ =>
```
pjfanning, line 1225: Would it make sense to also test the old way too? We should have regression testing of old behavior too.

Roiocam: That makes sense, I will add additional tests.

Roiocam: the PR https://github.com/apache/pekko/pull/1386 will keep the same behavior as the old way, it is just syntax sugar.

i will make extra test for it, in case we want change it in the future.

actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/SupervisionSpec.scala
```diff
@@ -401,40 +411,16 @@ class SupervisionSpec extends ScalaTestWithActorTestKit("""
...
       })
-      val behv = Behaviors.supervise(failedSetup).onFailure[Throwable](SupervisorStrategy.stop)
+      val behv = supervise(failedSetup).onFailure[Throwable](SupervisorStrategy.stop)
       LoggingTestKit.error[Exc3].expect {
         spawn(behv)
       }
     }
 
-    "support nesting exceptions with different strategies" in {
+    def throwIOExceptionThenIllegalArgumentException(behvFactory: ActorRef[Event] => Behavior[Command]): Unit = {
```
Roiocam, line 420: `"support nesting exceptions with outer restart and inner backoff strategies" in {`
and 
`"support nesting exceptions with different strategies" in {`

almost same, Therefore, it is compressed into the same method here

pjfanning (approved):
lgtm

PR #1398, opened 2024-07-18
chore: Test 3.3.4-RC1

He-Pin:
Motivation:
Test with 3.3.4-RC1

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -43,7 +43,7 @@ object Dependencies {
 
   val scala212Version = "2.12.19"
   val scala213Version = "2.13.14"
-  val scala3Version = "3.3.3"
+  val scala3Version = "3.3.4-RC1"
   val allScalaVersions = Seq(scala213Version, scala212Version, scala3Version)
 
   val reactiveStreamsVersion = "1.0.4"


He-Pin:
@WojciechMazur Why it now reports:
```scala
[error] -- [E030] Match case Unreachable Error: /home/runner/work/pekko/pekko/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala:3132:13 
[error] 3132 |        case t: (A, U) @unchecked           => t
[error]      |             ^^^^^^^^^^^^^^^^^^^^
[error]      |             Unreachable case
[error] one error found
```
https://github.com/apache/pekko/actions/runs/9985318652/job/27595952763?pr=1398#step:6:492

WojciechMazur:
It seems like we unfortunately backported a change that caused a regression, it can be minimized to 
```scala
def Test[U, A](thisElem: A, thatElem: U) = {
  case object passedEnd
  val any: Seq[Any] = ???
  any.zip(any)
    .map {
      case (`passedEnd`, r: U @unchecked) => (thisElem, r)
      case (l: A @unchecked, `passedEnd`) => (l, thatElem)
      case t: (A, U) @unchecked           => t
    }
}
```
The 3rd case is reachable so it's a bug.
I'm doing a bisect on the compiler right now before I fill in an issue in the compiler, but it seems like the regression was present in the Scala Next series since 3.4.0 (approximately introduced ~ 02.09.2023) so I'm astonished I was not reported previously. 
RC2 would either revert the change that caused the regression or would include a fix

PR #1406, opened 2024-07-21
Fix uncurried Pekko Stream ops in javadsl

mdedetrich:
So I discovered this when using the javadsl within a kotlin project but it appears that many of the stream functions in javadsl are using their curried equivalents (i.e. `limitWeighted(n: Long)(costFn: function.Function[Out, java.lang.Long])` vs `limitWeighted(n: Long, costFn: function.Function[Out, java.lang.Long])`,

This appears to be a massive oversight, as not only is this currently entirely inconsistent in the codebase (see fold vs foldWhile in Flow javadsl or fold in Source vs fold in Flow javadsl) but Java has no real concept curried functions and so applying curried functions written in Scala is not idiomatic. For Java the standard practice is to just have a single parameter list with the function as the last arg (and Kotlin also expects this).

There is also the case of `watchTermination` which had a zero arg curried implementation i.e. `def watchTermination[M]()(matF: function.Function2[Mat, CompletionStage[Done], M])`. I don't know if the zero arg was doing anything meaningful (afaik curried function application is strict so its not suspending the future unless you partially apply it), in any case I added a `def watchTermination[M](matF: function.Function2[Mat, CompletionStage[Done], M])` variant so check if this is indeed correct and I am not messing a behaviour (if the future is meant to suspended in a thunk then we should just use java function).

@He-Pin @jrudolph @raboof Maybe you want to have a look at this PR since you are more intimate with the codebase, particularly wrt `watchTermination` and if its arg needs to be suspended or not

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
@@ -742,7 +742,7 @@ object GraphDSL extends GraphCreate {
     new GenericGraph(s, gbuilder.delegate.result(s))
   }
 
-  final class Builder[+Mat]()(private[stream] implicit val delegate: scaladsl.GraphDSL.Builder[Mat]) { self =>
+  final class Builder[+Mat](private[stream] implicit val delegate: scaladsl.GraphDSL.Builder[Mat]) { self =>
     import pekko.stream.scaladsl.GraphDSL.Implicits._
 
     /**
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
@@ -1254,7 +1254,7 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
    *
    * '''Cancels when''' downstream cancels
    */
-  def groupedWeighted(minWeight: Long)(
+  def groupedWeighted(minWeight: Long,
       costFn: java.util.function.Function[Out, java.lang.Long]): javadsl.Flow[In, java.util.List[Out], Mat] =
     new Flow(delegate.groupedWeighted(minWeight)(costFn.apply).map(_.asJava)) // TODO optimize to one step
 
@@ -1311,7 +1311,7 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
    *
    * See also [[Flow.take]], [[Flow.takeWithin]], [[Flow.takeWhile]]
    */
-  def limitWeighted(n: Long)(costFn: function.Function[Out, java.lang.Long]): javadsl.Flow[In, Out, Mat] = {
+  def limitWeighted(n: Long, costFn: function.Function[Out, java.lang.Long]): javadsl.Flow[In, Out, Mat] = {
     new Flow(delegate.limitWeighted(n)(costFn.apply))
   }
 
@@ -1355,7 +1355,7 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
    *
    * '''Cancels when''' downstream cancels
    */
-  def scan[T](zero: T)(f: function.Function2[T, Out, T]): javadsl.Flow[In, T, Mat] =
+  def scan[T](zero: T, f: function.Function2[T, Out, T]): javadsl.Flow[In, T, Mat] =
     new Flow(delegate.scan(zero)(f.apply))
 
   /**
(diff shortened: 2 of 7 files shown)


He-Pin:
When using it from Java, it works, does it cause problems when using it from Kotlin? IDEA/Comping error? Does that affect the kotlin's trailing lambda?

mdedetrich:
So TIL, curried functions written in Scala appear as single parameter lists within Java so there isn't any actual problem its just that it really messes syntax highlighting/code completion in cross Scala/Java projects. Updating PR so that we only have a single version of the function without any deprecation's aside from `watchTermination`

mdedetrich:
> When using it from Java, it works, does it cause problems when using it from Kotlin? IDEA/Comping error? Does that affect the kotlin's trailing lambda?

So its not a compile issue (since both versions are the same after erasure and the curried version is compiled into a single parameter list) but it does mess with code completion/syntax highlighting in IDE.

In any case I think this change is valuable since it makes it a lot more clear

jxnu-liguobin:
I agree with this. It's better to make changes as soon as possible.

mdedetrich:
Pull request is ready, since MiMa has passed we have verification that generated bytecode is exactly the same, rather this feature improves quality of life for users (i.e. reading code, reading snippets taken from source code, generated docs from source etc etc)

stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
```diff
@@ -742,7 +742,7 @@ object GraphDSL extends GraphCreate {
     new GenericGraph(s, gbuilder.delegate.result(s))
   }
 
-  final class Builder[+Mat]()(private[stream] implicit val delegate: scaladsl.GraphDSL.Builder[Mat]) { self =>
+  final class Builder[+Mat](private[stream] implicit val delegate: scaladsl.GraphDSL.Builder[Mat]) { self =>
```
pjfanning, line 745: How do implicit args work when calling from Java? Is it possible that this constructor is not meant to be used directly by Java users, that is for use by internal Pekko code?

mdedetrich: implicit args also get flattened to flat parameter lists since under the hood they are also implemented via currying (with only Scala having the ability to pass those arguments implicitly).

pjfanning: But do you think that this constructor is to meant to be called by external Java users? It looks like an internal Pekko use only constructor.

mdedetrich: On the surface you seem to be right although it is referred to in multiple places in the docs.

In any case even if its only meant for internal Pekko usage, this change is harmless.

He-Pin: This change is fine, but it does break some binary compatibility, I'm using Javadsl only in Java, so I think that should also apply to others.

He-Pin (commented):
Does this affect the library which compiles with pekko 1.0.x but be used where only pekko .1.1x in classpath? eg a runtime `NoSuchMethodException`.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
mdedetrich: If it breaks binary compatibility MiMa would detect it

pjfanning:
> Does this affect the library which compiles with pekko 1.0.x but be used where only pekko .1.1x in classpath? eg a runtime `NoSuchMethodException`.

In theory no - because the Mima checks are happy that the change is binary compatible.

mdedetrich:
For those that are concerned about MiMa, breaking binary compatible or proof that any combination of implicits/currying in Scala just gets converted to a flat parameter list in JVM bytecode, with the current version of

```scala
final class Builder[+Mat]()(private[stream] implicit val delegate: scaladsl.GraphDSL.Builder[Mat])
```

This `.java` source file compiles

```java
package org.apache.pekko.stream.javadsl;

public class Test {
    void raw() {
        org.apache.pekko.stream.scaladsl.GraphDSL.Builder<Void> delegate = new org.apache.pekko.stream.scaladsl.GraphDSL.Builder<>();

        new GraphDSL.Builder<>(delegate);
    }
}
```

And if you change `final class Builder` to

```scala
final class Builder[+Mat](private[stream] implicit val delegate: scaladsl.GraphDSL.Builder[Mat])
```

It also compiles without any changes to the `.java` source. As mentioned before if there was an issue MiMa would have picked this up anyways

He-Pin (approved):
lgtm , as a daily java developer, I think this is fine now.

He-Pin:
btw, why you don't use the Kotlin flow in a Kotlin project:)

mdedetrich:
> btw, why you don't use the Kotlin flow in a Kotlin project:)

We are, I actually noticed this when using Pekko javadsl in a Kotlin project

He-Pin:
I'm using pekko stream because kotlin flow lack some operators:)

PR #1407, opened 2024-07-21
Fix uncurried stream ops in javadsl

mdedetrich:
(cherry picked from commit f9ad446a58dae77ccdec52719c5235e02ce8c770)

Since this change has been determined to be completely safe (see https://github.com/apache/pekko/pull/1406#issuecomment-2241528879) I think its a good candidate to cherry pick back to the 1.0.x series since if we ever do a release in the 1.0.x series it will mean that the docs/source/syntax completion will be better

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
@@ -1040,7 +1040,7 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
    *
    * '''Cancels when''' downstream cancels
    */
-  def groupedWeighted(minWeight: Long)(
+  def groupedWeighted(minWeight: Long,
       costFn: java.util.function.Function[Out, java.lang.Long]): javadsl.Flow[In, java.util.List[Out], Mat] =
     new Flow(delegate.groupedWeighted(minWeight)(costFn.apply).map(_.asJava)) // TODO optimize to one step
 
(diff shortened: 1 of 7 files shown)


He-Pin:
I think we better keep it in 1.0.x, @pjfanning ,wdyt?

pjfanning:
I'd prefer to hold off on this. Can we focus on getting 1.1.0 released?

mdedetrich:
> I'd prefer to hold off on this. Can we focus on getting 1.1.0 released?

Well reviewing/merging this isn't going to have an impact on 1.1.0 one way or another

Regarding 1.1.0, I am not sure if there was a discussion on the release plan (I have been a bit out of the loop as of late), or more specifically is the idea that we are meant to wait for a bit just to be sure that its stable (we are currently using 1.1.0-M1 in production), if so its best to wait to get feedback from a critical amount of users.

pjfanning:
> > I'd prefer to hold off on this. Can we focus on getting 1.1.0 released?
> 
> Well reviewing/merging this isn't going to have an impact on 1.1.0 one way or another
> 
> Regarding 1.1.0, I am not sure if there was a discussion on the release plan (I have been a bit out of the loop as of late), or more specifically is the idea that we are meant to wait for a bit just to be sure that its stable (we are currently using 1.1.0-M1 in production), if so its best to wait to get feedback from a critical amount of users.

My point is that we have been very conservative with 1.0.x changes. I would prefer not to merge this to 1.0.x. We can have the 1.1.0 release discussion on the mailing list. There are a few open issues but we could probably consider releasing 1.1.0-M2 soon or even doing 1.1.0 instead.

mdedetrich:
> My point is that we have been very conservative with 1.0.x changes. I would prefer not to merge this to 1.0.x.

Sure, but conservative doesn't mean "none at all" and at least currently without these changes, for cross Java/Scala projects the effected parts javadsl is currently broken when it comes to IDE's syntax highlighting and considering that it has zero impact on bytecode (and people are using 1.0.x) I think that this should pass the bar, we have certainly merged things into 1.0.x on a lot less than this.

Lets keep the 1.1.0 release discussion out of this issue, there wasn't any reason to bring it up and its just a distraction.

He-Pin:
I'm not using Kotlin with pekko-stream, but the current one works ok with Java, even with Java21, do you have a screenshot where is cause problem when work with Kotlin @mdedetrich

He-Pin:
@mdedetrich should this retarget 2.0.0？

He-Pin:
@pjfanning @mdedetrich what's the status of this?

PR #1408, opened 2024-07-21
Remove redundant params from defn bodies

mdedetrich:
I noticed that the codebase has a lot of cases of redundant/useless braces, this scalafmt configuration removes all of these braces so that.

In a future PR I will add this commit to `.git-blame-ignore-revs`


mdedetrich:
Seems like there is a bug in scalafmt, reported upstream at https://github.com/scalameta/scalafmt/issues/4108

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/BehaviorTestKitImpl.scala
```diff
@@ -49,14 +49,12 @@ private[pekko] final class BehaviorTestKitImpl[T](
...
   private var currentUncanonical = _initialBehavior
-  private var current = {
+  private var current =
     try {
       context.setCurrentActorThread()
       Behavior.validateAsInitial(Behavior.start(_initialBehavior, context))
-    } finally {
+    } finally
       context.clearCurrentActorThread()
-    }
```
He-Pin, line 58: I think the old one is begtter

Issue #1417, opened 2024-07-29
Deprecate `org.apache.pekko.dispatch.Futures`


raboof:
I think this class was introduced when Akka still supported Java versions before Java 8, which introduced `CompletionStage`.

We should probably try and deprecate `org.apache.pekko.dispatch.Futures` and any Java APIs that still produce/expect `Future`s, and introduce ones that work with `CompletionStage`.

raboof:
I didn't intend to suggest Java users should use `Future$.MODULE$.failed` and such: what I was thinking is that Java users shouldn't need to touch the Scala `Future` at all, and we should make sure that where they currently must use `Future`, we provide equivalent APIs using the more Java-native `CompletionStage`.

Something along the lines of (sketch) https://github.com/apache/pekko/compare/main...raboof:pekko:no-scala-futures-in-java?expand=1

pjfanning:
90% of the internal Pekko usages of org.apache.pekko.dispatch.Futures are in Java unit tests.
I guess we could move the code to a test class and deprecate the org.apache.pekko.dispatch.Futures class.
Some of the tests are linked in the Paradox documentation though.

FutureTimeoutSupport has one usage that could moved internal to that class.

PR #1418, opened 2024-07-29
reduce use of org.apache.pekko.dispatch.Futures

pjfanning:
see #1417 

Some of the methods in org.apache.pekko.dispatch.Futures don't have simple equivalents but these methods are not needed:
* promise
* failed
* successful


pjfanning:
Even promise and failed are a bit messy to replace - so I'm going to close this.

PR #1419, opened 2024-07-30
Don't use Scala Futures in Java APIs

raboof:
Sketching out #1417 - incomplete and notably not bothering with binary compatibility yet, just to illustrate the idea.


He-Pin:
@raboof Would you like to continue this? `AsyncRecoveryPlugin` and etcs should use Java's CompletionStage.

pjfanning:
#2092 took most of what was in this PR.

PR #1420, opened 2024-07-31
Clarify how our binary compatibility applies to external dependencies

raboof:


pjfanning (approved):
lgtm - may be worth backporting to 1.0.x branch too so we can built it into our 1.0 and 'current' docs.

raboof:
> may be worth backporting to 1.0.x branch too so we can built it into our 1.0 and 'current' docs.

(I think it's fine when this appears when we release 1.1.0)

PR #1424, opened 2024-08-04
feat: Add Pattern timeout support

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1359

still wip, test is missing

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/pattern/PatternSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/pattern/PatternSpec.scala
@@ -13,10 +13,8 @@
 
 package org.apache.pekko.pattern
 
-import scala.concurrent.{ Await, Future, Promise }
-import scala.concurrent.ExecutionContextExecutor
+import scala.concurrent.{ Await, ExecutionContextExecutor, Future, Promise, TimeoutException }
 import scala.concurrent.duration._
-
 import org.apache.pekko
 import pekko.actor.{ Actor, Props }
 import pekko.testkit.{ PekkoSpec, TestLatch }
@@ -76,4 +74,23 @@ class PatternSpec extends PekkoSpec {
       intercept[IllegalStateException] { Await.result(r, remainingOrDefault) }.getMessage should ===("Mexico")
     }
   }
+
+  "pattern.timeout" must {
+    "be completed successfully eventually" in {
+      val f = pekko.pattern.timeout(100.millis, using = system.scheduler)(Future.successful(5))
+      Await.result(f, remainingOrDefault) should ===(5)
+    }
+
+    "be completed abnormally eventually" in {
+      val f =
+        pekko.pattern.timeout(100.millis, using = system.scheduler)(Future.failed(new IllegalStateException("ABC")))
+      intercept[IllegalStateException] { Await.result(f, remainingOrDefault) }.getMessage should ===("ABC")
+    }
+
+    "be completed with a TimeoutException if not completed within the specified time" in {
+      val f = pekko.pattern.timeout(100.millis, using = system.scheduler)(Future.never)
+      intercept[TimeoutException] { Await.result(f, remainingOrDefault) }
+    }
+  }
+
 }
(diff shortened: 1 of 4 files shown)


He-Pin:
My wife just come to hangzhou, and I may don't have enough time to polish this , any feed back is welcome, I will try to get it done soon.

pjfanning:
Can this wait till 1.1.1 release?

actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -87,4 +86,64 @@ trait FutureTimeoutSupport {
...
+      catch {
+        case NonFatal(t) => Future.failed(t)
+      }
+    future.value match {
+      case Some(_) => future
+      case None => // not completed yet
+        val p = Promise[T]()
+        val timeout = using.scheduleOnce(duration) {
+          p.tryFailure(new TimeoutException(s"Timeout of $duration expired"))
+          if (future.isInstanceOf[CompletableFuture[T]]) {
```
Roiocam, line 107: cool feature, but seem like it doesn't support scala future, because it haven't cancel/interrrupt method

He-Pin: that's true.

Roiocam:
what is the plan for the shipping of this PR? it looks like it can be help to implement Hedged requests.

He-Pin:
@Roiocam Chinese Spring Festival is coming, I would like to have this done before that.

actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -87,4 +86,64 @@ trait FutureTimeoutSupport {
       }
       p
     }
+
+  /**
+   * Returns a [[scala.concurrent.Future]] that will be completed with a [[TimeoutException]]
+   * if the provided value is not completed within the specified duration.
+   */
+  def timeout[T](duration: FiniteDuration, using: Scheduler)(value: => Future[T])(
```
nvollmar, line 94: would rename `using` to `scheduler`

He-Pin: but it was `using` in the code above.

actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -87,4 +86,64 @@ trait FutureTimeoutSupport {
...
+          p.tryComplete(result)
+        }(pekko.dispatch.ExecutionContexts.parasitic)
+        p.future
+    }
+  }
+
+  /**
+   * Returns a [[java.util.concurrent.CompletionStage]] that will be completed with a [[TimeoutException]]
+   * if the provided value is not completed within the specified duration.
+   */
```
pjfanning, line 124: can you add `@since 1.2.0` to new public methods?

actor-tests/src/test/scala/org/apache/pekko/pattern/PatternSpec.scala
```diff
@@ -76,4 +74,22 @@ class PatternSpec extends PekkoSpec {
...
+      Await.result(f, remainingOrDefault) should ===(5)
+    }
+
+    "be completed abnormally eventually" in {
+      val f = pekko.pattern.timeout(1.second, using = system.scheduler)(Future.failed(new IllegalStateException("ABC")))
+      intercept[IllegalStateException] { Await.result(f, remainingOrDefault) }.getMessage should ===("ABC")
+    }
+
+    "be completed with a TimeoutException if not completed within the specified time" in {
+      val f = pekko.pattern.timeout(1.second, using = system.scheduler)(Future.never)
```
pjfanning, line 90: Can you use a shorter timeout? I don't see why the test needs to always take 1 second (or more). I think you should also check that the elapsed time for the test is greater than or equal to the timeout - maybe with a small delta (allow a few milliseconds leeway).

He-Pin:
@pjfanning @Roiocam I think this one should be ready.

pjfanning (approved):
lgtm

Issue #1433, opened 2024-08-08
jdk9 tests do not run


pjfanning:
See https://github.com/apache/pekko/pull/1432 - I deliberately modified 2 classes (1 Java, 1 Scala) to see if they were even compiled. The PR did fail because scalafmt and the doc build failed but the main part of the build where we compile code and run tests - this all ran ok.

Roiocam:
I guess `sbt dependWalkerCheck` didn't want to verify JDK9+ classes compile, but I don't remember why it was on purpose.

pjfanning:
#1434 merged

PR #1434, opened 2024-08-08
run extra tests that need Java 9+

pjfanning:
related to #1433


raboof (approved):
LGTM, good observation that we were missing those.

(FWIW I'd also be OK with running all the tests under Java 21: any bugs that'd only pop up when running under Java 8 should be picked up by the nightlies)

pjfanning:
I'd prefer to keep the Java 21 run separate, for now. We can optimise the build later. I suspect at some stage, we will be requested to cut down on CI build time.

Roiocam:
Thanks for this, It looks like a new case that we didn't cover it. 
we haven't jdk9 test before, only exists in documentation.

PR #1443, opened 2024-08-22
Add alsoTo/alsoToContext to FlowWithContext/SourceWithContext

mdedetrich:
This PR adds `alsoTo`/`alsoToContext` onto `FlowWithContext`/`SourceWithContext` which just wrap the underlying `delegate` methods, this is analogous to the current `map`/`mapContext` methods that already exist.

Note that `delegate.alsoTo(Flow.fromFunction { in: (Out, Ctx) => in._2 }.toMat(that)(Keep.right))` is copied from the implementation of `contramap`. I would have ideally liked to use `contramap` however the method is not available on a `Graph[SinkShape[_]]`

I didn't add tests due to the sheer triviality of the implementation but can add them if asked.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/FlowWithContextOps.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/FlowWithContextOps.scala
@@ -89,6 +89,22 @@ trait FlowWithContextOps[+Out, +Ctx, +Mat] {
       combine: (Mat, Mat2) => Mat3): ReprMat[Out2, Ctx2, Mat3]
 
   /**
+   * Data variant of [[pekko.stream.scaladsl.FlowOps.alsoTo]]
+   *
+   * @see [[pekko.stream.scaladsl.FlowOps.alsoTo]]
+   * @since 1.1.0
+   */
+  def alsoTo(that: Graph[SinkShape[Out], _]): Repr[Out, Ctx]
+
+  /**
+   * Context variant of [[pekko.stream.scaladsl.FlowOps.alsoTo]]
+   *
+   * @see [[pekko.stream.scaladsl.FlowOps.alsoTo]]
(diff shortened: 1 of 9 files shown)


mdedetrich:
PR is now ready to review, the link-validator failure is not related

pjfanning:
I would like a couple of trivial test cases - even if for regression purposes

pjfanning (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/scaladsl/FlowWithContextOps.scala
```diff
@@ -88,6 +88,20 @@ trait FlowWithContextOps[+Out, +Ctx, +Mat] {
   def viaMat[Out2, Ctx2, Mat2, Mat3](flow: Graph[FlowShape[(Out, Ctx), (Out2, Ctx2)], Mat2])(
       combine: (Mat, Mat2) => Mat3): ReprMat[Out2, Ctx2, Mat3]
 
+  /**
+   * Data variant of [[pekko.stream.scaladsl.FlowOps.alsoTo]]
+   *
+   * @see [[pekko.stream.scaladsl.FlowOps.alsoTo]]
+   */
+  def alsoTo(that: Graph[SinkShape[Out], _]): Repr[Out, Ctx]
```
pjfanning, line 96: Can you add `@since 1.1.0` on this and other new methods?

mdedetrich: Will do

mdedetrich: Added and pushed

mdedetrich:
Thanks for the review, everything is passing and I have added the `@since`, merging.

PR #1446, opened 2024-08-23
Add wiretap/wiretapContext to FlowWithContext/SourceWithContext

mdedetrich:
Analogous to https://github.com/apache/pekko/pull/1443 but with `wireTap`/`wireTapContext`

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-1.1.md
+++ b/docs/src/main/paradox/release-notes/releases-1.1.md
@@ -43,6 +43,7 @@ The Stream API has been updated to add some extra functions.
 * added extra retry operators that allow users to provide a predicate to decide whether to retry based on the exception ([PR1269](https://github.com/apache/pekko/pull/1269))
 * add optionalVia/unsafeOptionalDataVia operators ([PR1422](https://github.com/apache/pekko/pull/1422))
 * add alsoTo/alsoToContext operators to `SourceWithContext`/`FlowWithContext` ([PR-1443](https://github.com/apache/pekko/pull/1443))
+* add wireTap/wireTapContext operators to `SourceWithContext`/`FlowWithContext` ([PR-1446](https://github.com/apache/pekko/pull/1446))
 
 The Stream Testkit Java DSL has some extra functions.
 
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowWithContextSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowWithContextSpec.scala
@@ -126,6 +126,50 @@ class FlowWithContextSpec extends StreamSpec {
         }
     }
 
+    "pass through all data when using wireTap" in {
+      val listBuffer = new ListBuffer[String]()
+      Source(Vector(Message("A", 1L), Message("B", 2L), Message("D", 3L), Message("C", 4L)))
+        .asSourceWithContext(_.offset)
+        .via(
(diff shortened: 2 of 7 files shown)


docs/src/main/paradox/release-notes/releases-1.1.md
```diff
@@ -43,6 +43,7 @@ The Stream API has been updated to add some extra functions.
 * added extra retry operators that allow users to provide a predicate to decide whether to retry based on the exception ([PR1269](https://github.com/apache/pekko/pull/1269))
 * add optionalVia/unsafeOptionalDataVia operators ([PR1422](https://github.com/apache/pekko/pull/1422))
 * add alsoTo/alsoToContext operators to `SourceWithContext`/`FlowWithContext` ([PR-1443](https://github.com/apache/pekko/pull/1443))
+* add wireTap/wireTapContext operators to `SourceWithContext`/`FlowWithContext` ([PR-1446](https://github.com/apache/pekko/pull/1446))
```
pjfanning, line 46: This is ok but I think when we put together the release notes, we should have separate sections for 1.1.0 and 1.1.0-M1 releases. We can consolidate them again later but until we get all the modules like HTTP and Connectors 1.1.0 releases, it is useful for 1.1.0-M1 users to know which changes are in that release.
We don't need to repeat the changes. 1.1.0 users can be told that that release contains everything in 1.1.0 plus the extra listed changes.

mdedetrich: Personally I think this process is a bit heavy and not giving much benefit considering that we are following SemVer it shouldn't make a difference about whether something is in M1 or in M2 or the first milestone.

We can discuss this on the mailing list if you want to push for it.

pjfanning: The alternative is that we hijack the 1.1.0-M1 section of the release notes (where you have added this change) - and just annotate the additional changes that were not in 1.1.0-M1.

mdedetrich: After thinking about it I have just moved the 3 added api's since M1 to the appropriate section

mdedetrich:
So the tests in the PR are now working, I ended up having to use `atLeastOneElementOf` since the feature of `wireTap` is that it will drop elements if the downstream consumer is not ready (rather than back pressuring which is what `alsoTo` does).

In other words its possible that in certain test runs one of the elements may be missing, which is intentional behaviour of wireTap.

mdedetrich:
@pjfanning Do you mind reviewing this PR? You have reviewed the previous version and its the last thing from my side for the next RC

docs/src/main/paradox/release-notes/releases-1.1.md
```diff
@@ -5,6 +5,11 @@
 Release notes for Apache Pekko 1.1.0-M1. See [GitHub Milestone](https://github.com/apache/pekko/milestone/2?closed=1) for fuller list of changes.
 As with all milestone releases, this release is not recommended for production use - it is designed to allow users to try out the changes in a test environment.
 
+### Additional APIs
```
pjfanning, line 8: why separate these from the other 'Additional APIs' below?

mdedetrich: I thought thats what you were asking for, I will move it back

mdedetrich: Resolved and pushed.

pjfanning (approved):
lgtm

PR #1456, opened 2024-08-27
release notes for 1.1.0

pjfanning:
1.1.0 release has been approved. Need to get this merged so I can announce the release.


He-Pin (approved):
lgtm, btw, I as the we have migrate Netty 3 to Netty 4, which need user test it more carefully, I think we should mention this in the released note.

pjfanning:
> lgtm, btw, I as the we have migrate Netty 3 to Netty 4, which need user test it more carefully, I think we should mention this in the released note.

I'll merge this as is - so as not to delay the release - but I'll do a new PR in the next few days to put more emphasis on the netty 4 change.

Issue #1463, opened 2024-09-06
Cluster sharding delivers message to the wrong entity


jchapuis:
Sorry to bring some bad news, I have been investigating failing tests in https://github.com/endless4s/endless-transaction/pull/48, a PR that upgrades Pekko from 1.0.3 to 1.1.0 and I think I found a serious issue. 

The failing test suite is stress-testing event-sourced entities using the persistence test toolkit, and I have identified that a command sometimes gets delivered to the wrong entity. 

I have bisected the problem to [this optimization](https://github.com/apache/pekko/pull/1304) that was introduced after the 1.1.0-M1 release. That new code makes use of a `var cache` and it doesn't seem thread-safe. Could it be that we introduced races?

pjfanning:
@jchapuis @Roiocam I looked at the PR that seems to be the issue and I think if we try to change the `var cache` to some sort of AtomicObject or lazy val that the benefits of the change will vanish. I suggest that we revert it.

@jchapuis would you have any idea how we could minimise a reproducible test that could be used for regression purposes?

jchapuis:
@pjfanning I haven't yet had the time to look into this. I decided to report the problem as soon as I had a certain degree of confidence that this was the issue, due to the release still being young, I got worried. I would say some tests sending commands to multiple sharded entities in quick succession, verifying that each command gets delivered to the proper destination.

pjfanning:
@raboof it would be nice to be able to start on the 1.0.1 RC in the next few days.

What do you think of this course of action?
1. revert #1304 
2. hopefully @jchapuis will have time to test endless4s with a Pekko snapshot with this change
3. at some point early next week, we get an RC together
4. at some point, we try to fill in the test gap in Pekko so that we don't get a future bug in this area - including tests that try to force multithreaded evaluation of the entityId 
5. we avoid further changes to cluster sharding until we are happy that the test coverage has improved

jchapuis:
@pjfanning sure happy to run endless4s tests as soon as the revert is merged

Roiocam:
> @jchapuis @Roiocam I looked at the PR that seems to be the issue and I think if we try to change the `var cache` to some sort of AtomicObject or lazy val that the benefits of the change will vanish. I suggest that we revert it.

after investigating, i think because the extractEntityId  instance was shared by both `ShardRegion` Actor and multiple `Shard` Actors.

[...]

PR #1467, opened 2024-09-08
add some warnings to release notes

pjfanning:
see #1463 and https://github.com/apache/pekko/pull/1456#pullrequestreview-2276084772

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-1.1.md
+++ b/docs/src/main/paradox/release-notes/releases-1.1.md
@@ -6,6 +6,10 @@ Apache Pekko 1.1.x releases support Java 8 and above.
 
 Release notes for Apache Pekko 1.1.0. See [GitHub Milestone for 1.1.0-M1](https://github.com/apache/pekko/milestone/2?closed=1) and [GitHub Milestone for 1.1.0](https://github.com/apache/pekko/milestone/7?closed=1) for a fuller list of changes.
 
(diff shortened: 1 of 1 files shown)


raboof (approved):

raboof:
(link validator is about 
https://cowtowncoder.medium.com/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062 , that link seems fine (so either intermittent or paywall issue)

docs/src/main/paradox/release-notes/releases-1.1.md
```diff
@@ -99,6 +103,7 @@ Most of the dependency changes are small patch level upgrades. Some exceptions i
 
 ### Known Issues
 
+* A race condition has been introduced into pekko-cluster-sharding which needs to be fixed in a Pekko 1.1.1 release. Users who use Cluster Sharding should skip this release. The problem does not appear in v1.1.0-M1. ([#1463](https://github.com/apache/pekko/pull/1463))
```
Roiocam, line 106: It also can be temporarily fixed by using another API in classic ClusterSharding(can not fix on typed pekko):  https://github.com/apache/pekko/blob/09486097e7115a0a173faf2fca929f9d62f80394/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ClusterSharding.scala#L267-L273

pjfanning: We should have 1.1.1 released in a couple of days so it is probably not very important to put this is in the docs but if anyone gets in touch in the mean time, then we can share this info with them. Thanks.

PR #1468, opened 2024-09-09
Add overridden duration timeout to StreamTestKit

mdedetrich:
The context of this change is https://github.com/apache/pekko-http/pull/590#pullrequestreview-2281514043. It turns out its extremely cumbersome to manually override the duration so I am adding this public method

@raboof @jrudolph Let me know if this line of thinking is on the right lines.

### DIFF
--- a/stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
+++ b/stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
@@ -15,5 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# Add overridden duration timeout to StreamTestKit
+# StreamTestKit.assertNoChildren is an internal API and can be changed
 ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.testkit.scaladsl.StreamTestKit.assertNoChildren")
--- a/stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
+++ b/stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
@@ -1,0 +1,1 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.testkit.scaladsl.StreamTestKit.assertNoChildren")
--- a/stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
+++ b/stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
@@ -1,1 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Add overridden duration timeout to StreamTestKit
 ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.testkit.scaladsl.StreamTestKit.assertNoChildren")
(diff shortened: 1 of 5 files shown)


stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/scaladsl/StreamTestKit.scala
```diff
@@ -35,12 +35,30 @@ object StreamTestKit {
    * This assertion is useful to check that all of the stages have
    * terminated successfully.
    */
+  def assertAllStagesStopped[T](block: => T, overrideTimeout: FiniteDuration)(implicit materializer: Materializer): T =
+    materializer match {
```
mdedetrich, line 39: There is code duplication here, I can clean it up if the premise of this PR is acceptable.

stream-testkit/src/test/scala/org/apache/pekko/stream/testkit/StreamSpec.scala
mdedetrich, line 1: Not sure if this is the best way to configure the timeout for scalatest, looking to see if there are more idiomatic ways.

stream-testkit/src/test/scala/org/apache/pekko/stream/testkit/StreamSpec.scala
```diff
@@ -27,10 +26,26 @@ import pekko.testkit.{ PekkoSpec, TestProbe }
...
 import org.scalatest.Failed
+import org.scalatest.time.{ Millis, Span }
 
 import com.typesafe.config.{ Config, ConfigFactory }
 
+import java.util.concurrent.TimeUnit
+
 abstract class StreamSpec(_system: ActorSystem) extends PekkoSpec(_system) {
+
+  final case class StreamConfig(allStagesStoppedTimeout: Span = Span({
```
mdedetrich, line 37: This might look a bit un-orthodox by Pekko standards, but it is the idiomatic way to override a configuration when in ScalaTest (which is what `StreamSpec` extends).

The defaault behaviour is still the same (i.e. getting the value from `pekko.stream.testkit.all-stages-stopped-timeout`) but if you decide to override it then you can do it the ScalaTest specific way.

pjfanning:
including in 1.1.2 milestone - if 1.1.1-RC1 vote fails, we can consider adding this to a 1.1.1-RC2

mdedetrich:
Adding some reviewers to this PR to get an idea whether its even on the right track, @jrudolph @raboof Maybe you have some comments?

stream-testkit/src/test/scala/org/apache/pekko/stream/testkit/StreamConfiguration.scala
```diff
@@ -0,0 +1,40 @@
...
+
+package org.apache.pekko.stream.testkit
+
+import org.apache.pekko.testkit.TestKitBase
+import org.scalatest.time.{ Millis, Span }
+
+import java.util.concurrent.TimeUnit
+
+trait StreamConfiguration extends TestKitBase {
+  case class StreamConfig(allStagesStoppedTimeout: Span = Span({
```
mdedetrich, line 26: The premise here is that by default we get the timeout value from Pekko typesafe config (as is expected when using Pekko) however if a user wants to override this then its done the ScalaTest idiomatic way (which is overriding a `def` in a trait, similar to how `PatienceConfig` works).

jxnu-liguobin: It is possible to omit units and support more flexibility, but the code is slightly complex and for reference only.
```
import scala.jdk.DurationConverters._
import org.scalatest.time.Span._
system.settings.config.getDuration("pekko.stream.testkit.all-stages-stopped-timeout").toScala.convertDurationToSpan
```
I just saw this `{}` and wanted to make some changes. I looked at the Span code and found that it has convertDurationoSpan. You can do whatever you want.

stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/scaladsl/StreamTestKit.scala
```diff
@@ -53,10 +71,15 @@ object StreamTestKit {
   }
 
   /** INTERNAL API */
-  @InternalApi private[testkit] def assertNoChildren(sys: ActorSystem, supervisor: ActorRef): Unit = {
+  @InternalApi private[testkit] def assertNoChildren(sys: ActorSystem, supervisor: ActorRef,
```
He-Pin, line 74: add an overloaded method here to avoid the mima issue?

stream-testkit/src/main/mima-filters/1.1.1.backwards.excludes/pr-1468-override-timeout-stream-testkit.backwards.excludes
```diff
@@ -0,0 +1 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.testkit.scaladsl.StreamTestKit.assertNoChildren")
```
raboof, line 1: When adding mima excludes, please add the reason why it is acceptable. In this case, the reason might be we don't promise binary compatiblity for -testkit artifacts (https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html), and that this method is `private[testkit]`

raboof (approved):
This approach seems reasonable to me!

He-Pin:
Cool, nice to see @pjfanning carry it on

pjfanning (approved):
lgtm - I committed some small changes to the mima excludes file to address some issues (license and a description of why the change is ok)

He-Pin (approved):

PR #1472, opened 2024-09-09
KEYS: add second public key for Arnout Engelen

raboof:
If you want you can verify this key is plausible because it's been used to sign Pekko and Akka commits since way back.


pjfanning (approved):
lgtm

Issue #1484, opened 2024-09-17
consider marking some functions as `@noinline` to fix issues with using Kamon


pjfanning:
Pekko 1.1 enables Scala 2 inlining.

https://www.baeldung.com/scala/inline-noinline-annotations

This has caused issues with using Kamon for monitoring Pekko 1.1. https://github.com/kamon-io/Kamon/issues/1352

One option might be to mark some of the functions that Kamon relies on instrumenting as `@noinline`.

mdedetrich:
To me this is acceptable as long as the number of methods we have to mark as `@noinline` is reasonable (as a rough ballback I would say < 10) and that ontop of this we also have a path forward to provide official instrumentation/hooks.

The thing I want to avoid is that we make a release with a some methods marked `@noinline` and then at some point in time this isn't enough and then we end up adding even more methods with `@noinline` creating a roundabout back and forth circus.

pjfanning:
It seems like the Scala 2 compiler can inline any functions that are marked as `final`. Does anyone know what happens if a class is marked as `final`, does that mean all of its function are regarded by the inliner as being final and therefore inline friendly?

mdedetrich:
> It seems like the Scala 2 compiler can inline any functions that are marked as `final`. Does anyone know what happens if a class is marked as `final`, does that mean all of its function are regarded by the inliner as being final and therefore inline friendly?

@lrytz Can you quickly answer this one?

lrytz:
yes, the answer is yes :)

Roiocam:
I agree with _"One option might be to mark some of the functions that Kamon relies on instrumenting as @noinline."_

Creating an un-instrumentation library is not what we want, but we need to consider what kind of method can not be inline.

pjfanning:
I've opened https://github.com/apache/pekko/discussions/1487 to discuss a long term solution. I still think this `@noinline` solution is something that we should do for Pekko 1.1.2. We have broken kamon-pekko by enabling Scala 2 compiler inlining in Pekko 1.1. Maintaining these `@noinline` annotations indefinitely will be an issue but I think we can maintain them on the 1.1 branch and find a better solution for future releases.

pjfanning:
@lrytz Thanks for your assistance. If you have time, would you be able my observation below?

[...]

PR #1485, opened 2024-09-17
mark stop() method as noinline

pjfanning:
relates to #1484 

I have similar issues in https://github.com/pjfanning/micrometer-pekko and this change fixes the broken test I ran into there when upgrading to Pekko 1.1.

I have tested this locally and it fixes the issue in micrometer-pekko.

More changes will be needed for Kamon but I'm raising this to show a path forward and to see if this acceptable for Pekko 1.1.2. We can add more changes in more PRs.

I won't merge this until a few people have reviewed it and there are no objections.

fyi @hughsimpson

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/dungeon/Dispatch.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/dungeon/Dispatch.scala
@@ -159,7 +159,7 @@ private[pekko] trait Dispatch { this: ActorCell =>
     catch handleException
 
   // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
-  final def stop(): Unit =
+  @noinline final def stop(): Unit =
     try dispatcher.systemDispatch(this, Terminate())
     catch handleException
 
--- a/actor/src/main/scala/org/apache/pekko/actor/dungeon/Dispatch.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/dungeon/Dispatch.scala
@@ -159,6 +159,8 @@ private[pekko] trait Dispatch { this: ActorCell =>
     catch handleException
 
   // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
+  // this function is marked `noinline` because Kamon and other metrics libraries
+  // instrument this function
   @noinline final def stop(): Unit =
     try dispatcher.systemDispatch(this, Terminate())
     catch handleException


actor/src/main/scala/org/apache/pekko/actor/dungeon/Dispatch.scala
```diff
@@ -159,7 +159,7 @@ private[pekko] trait Dispatch { this: ActorCell =>
     catch handleException
 
   // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
-  final def stop(): Unit =
+  @noinline final def stop(): Unit =
```
Roiocam, line 162: I wonder why this method has been inline. Does us toggle something on? If does, for those method like `start`, `resume`, `suspend` shouldn't be inline too.

jrudolph:
It would be good to document these unwritten compatibility constraints. In Akka, we used `@InternalStableApi` but any kind of simple documentation will help understand the reasoning behind adding something like `@noinline` later (which otherwise is very uncommon in Scala and has potential performance impacts in the future).

actor/src/main/scala/org/apache/pekko/actor/dungeon/Dispatch.scala
jrudolph: The Scala compiler is free to inline method when it is sound. This is the case, when the final receiver of a call can be determined. Whether or not that has any performance impact is always hard to say. In general, the JVM JIT compilation will also do heavy inlining and may in many cases do the same kind of optimizations at runtime. In other cases, like e.g. for higher-order methods, the Scala compiler might be able to generate significantly less code by earlier inlining, which also helps JIT compilations.

That said, the point of adding `@noinline` here is that some telemetry tools like Kamon instrument internal APIs to do their work. This has mainly historic reasons. In the future, it could be better to introduce official hooks for these kinds of use cases, so that it is easier to track what the public telemetry API is instead of relying on keeping some internal APIs stable.

pjfanning: Thanks @Roiocam - there will need to be more changes. This PR is just to show one change and to highlight that the change doesn't break anything (tests or binary compatibility checks).

pjfanning: #1489 adds more annotations

pjfanning:
> It would be good to document these unwritten compatibility constraints. In Akka, we used `@InternalStableApi` but any kind of simple documentation will help to later understand the reasoning behind adding something like `@noinline` (which otherwise is very uncommon in Scala and has potential performance impacts in the future).

@jrudolph I've summarised the situation in intro to https://github.com/apache/pekko/discussions/1487 and the aim of that discussion is to discuss how we can remove the `@noinline` annotations at some point in the future.

pjfanning:
https://github.com/apache/pekko/pull/1489 was merged - would it be possible for any Pekko PMC members to review this? This change is more useful for micrometer-pekko than Kamon.

nvollmar (approved):
lgtm

raboof (approved):

PR #1486, opened 2024-09-17
simplify scala 3 build in CI

pjfanning:
The complicated CI for Scala 3 which runs many separate compile and test subjobs - while Scala 2 build is not like this at all - I think this might have been useful when Scala 3 support was being added but it is now stable.

The multiple subjobs will have to do a lot of duplicated work. The single subjob will take longer but the PR validation subjob takes ages any way so this won't slow down the build - and makes us better GitHub citizens because we use fewer concurrent VMs and take less compute time in total.


raboof (approved):
Nice! Perhaps it can be integrated further in the regular tests, but this is a step in the right direction in any case

PR #1490, opened 2024-09-20
Support primitives in Flow#collectType

JD557:
`Flow#collectType` was not working as expected for primitives. This PR replaces the use of `ClassTag#runtimeClass#instanceOf` with `ClassTag#unapply`.

While the default implementation of `ClassTag#unapply` simply defers to `ClassTag#runtimeClass#instanceOf`, this is not the case for the `ClassTag`s of primitives, which only override `unapply`.

See the `IntManifest` for an example:

https://github.com/scala/scala/blob/v2.13.14/src/library/scala/reflect/Manifest.scala#L223-L236

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowCollectTypeSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowCollectTypeSpec.scala
@@ -25,7 +25,7 @@ class FlowCollectTypeSpec extends StreamSpec {
 
   "A CollectType" must {
 
-    "collectType" in {
+    "collectType with references" in {
       val fruit = Source(List(Orange, Apple, Apple, Orange))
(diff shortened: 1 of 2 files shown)


raboof (approved):
nice!

pjfanning (approved):
lgtm - merging, thanks @JD557

pjfanning:
Is this something that we should consider backporting to the 1.0.x relaeses?

JD557:
> Is this something that we should consider backporting to the 1.0.x relaeses?

That's obviously up to you (the maintainers) to decide, but my gut feeling says that maybe not.

Mostly because it changes the behavior from Akka (I doubt anyone depended on this behavior, though) and it's not that hard to fix with a normal `collect`.

I actually imagine most people just use `collect` and don't even notice that this method exists (at least that's how I stumbled on the bug 😅)

Issue #1488, opened 2024-09-20
Use of japi.Util.immutableSeq(j.u.List) is inefficient because it iterates over the list to create a new Seq


pjfanning:
https://github.com/apache/pekko/blob/22a738511375baa23c5b84c87d44a065c0401e0e/actor/src/main/scala/org/apache/pekko/japi/JavaAPI.scala#L269-L281

It's used in a lot of places internally but the most performance sensitive area is the pekko-stream code (eg Flow, Source, etc).
The obvious solution of using pekko.util.ccompat.JavaConverters to convert using `.asScala.toSeq` doesn't compile in Scala 2.12 because toSeq gives a scala.Seq instead of a scala.immutable.Seq.

We could use `.asScala.toList` but that is likely to be slower. We could also a new Util.immutableWrappingSeq function that has different implementations for Scala 2.12 (possibly delegate to existing immutableSeq) and Scala 2.13+ that uses `.asScala.toSeq`.

pjfanning:
Some changes are in Pekko 1.1.2.

PR #1491, opened 2024-09-20
issue with full list copies in pekko-stream javadsl

pjfanning:
see #1488

### DIFF
--- a/stream/src/main/scala-2.12/org/apache/pekko/stream/javadsl/CollectionUtil.scala
+++ b/stream/src/main/scala-2.12/org/apache/pekko/stream/javadsl/CollectionUtil.scala
@@ -26,10 +26,10 @@ import pekko.util.ccompat.JavaConverters._
 /**
  * INTERNAL API
  */
-private[javadsl] object CollectionUtil {
-  def toSeq[T](jlist: java.util.List[T]): immutable.Seq[T] =
+private[javadsl] final object CollectionUtil {
+  @inline def toSeq[T](jlist: java.util.List[T]): immutable.Seq[T] =
     jlist.asScala.toList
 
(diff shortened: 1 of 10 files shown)


samueleresca (approved):
Few questions:
- should we create a follow-up issue to update the rest of the code to use the new `CollectionUtil`?
- Is it worth updating the existing documentation of the `immutableSeq` method to point to the new CollectionUtil object?

Otherwise looks good to me.

pjfanning:
> Few questions:
> 
> * should we create a follow-up issue to update the rest of the code to use the new `CollectionUtil`?
> * Is it worth updating the existing documentation of the `immutableSeq` method to point to the new CollectionUtil object?
> 
> Otherwise looks good to me.

CollectionUtil is package private. I think there is a tradeoff between having centralised code but then also meaning that many Pekko libs require a minimum version of a shared Pekko lib like pekko-actor.
I looked at the other usages of Util.immutableSeq(java.util.List) in this repo and in pekko-http and they don't really seem to be in performance critical paths. Some of the changes in this PR are in performance critical paths for Java DSL users.

samueleresca:
> > Few questions:
> > 
> > * should we create a follow-up issue to update the rest of the code to use the new `CollectionUtil`?
> > * Is it worth updating the existing documentation of the `immutableSeq` method to point to the new CollectionUtil object?
> > 
> > Otherwise looks good to me.
> 
> CollectionUtil is package private. I think there is a tradeoff between having centralised code but then also meaning that many Pekko libs require a minimum version of a shared Pekko lib like pekko-actor. I looked at the other usages of Util.immutableSeq(java.util.List) in this repo and in pekko-http and they don't really seem to be in performance critical paths. Some of the changes in this PR are in performance critical paths for Java DSL users.

Thanks for clarifying.

PR #1510, opened 2024-10-01
Fix CollectionUtil `toSeq` methods

mdedetrich:
This PR fixes the the newly added `.toSeq` methods, more specifically

* Adds the pekko  `@InternalApi` annotation
* Moves the `scala-2.13+` to just `scala-2.13` and adds an implementation bespoke to `scala-3`. This is because `@inline` doesn't do anything in Scala 3, instead you have to use the new `inline` keyword.


pjfanning (approved):
lgtm

PR #1515, opened 2024-10-03
release 1.1.2

pjfanning:


docs/src/main/paradox/release-notes/releases-1.1.md
```diff
@@ -2,9 +2,27 @@
...
+* Support primitives in Flow#collectType ([PR1490](https://github.com/apache/pekko/pull/1490))
+
+### Changes
+
+* Some updates to fix issues seen by Kamon users ([#1484](https://github.com/apache/pekko/issues/1484))
+
+### Dependency Changes
+
+* pekko-protobuf-v3 is now based on protobuf-java 3.25.5
+* netty 4.1.114.Final
```
raboof, line 21: (might be interesting to also list the 'previous' version, but not essential)

raboof (approved):

docs/src/main/paradox/release-notes/releases-1.1.md
pjfanning: I've avoided linking PRs for dependency upgrades. I find that style of listing every PR for such upgrades to be noisy. Many releases can have far more commits for upgrades than other code changes and many dependencies have loads of upgrades.

I guess that I could summarise the version change but in practice, I prefer to think of release notes as a quick summary and people who are really interested can do their own research looking at commit history or pom files. This one in particular is an upgrade from 4.1.113.Final to 4.1.114.Final - you would hope that a minor version number like this would come with few wrinkles.

PR #1524, opened 2024-10-09
Update circuitbreaker.md

azarboon:
I've added some challenges about using circuit breaker pattern and some workarounds. So readers can take more informed decision.

### DIFF
--- a/docs/src/main/paradox/common/circuitbreaker.md
+++ b/docs/src/main/paradox/common/circuitbreaker.md
@@ -61,6 +61,10 @@ The Apache Pekko library provides an implementation of a circuit breaker called
 
 ![circuit-breaker-states.png](../images/circuit-breaker-states.png)
 
+## Challenges
+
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/common/circuitbreaker.md
```diff
@@ -61,6 +61,10 @@ The Apache Pekko library provides an implementation of a circuit breaker called
 
 ![circuit-breaker-states.png](../images/circuit-breaker-states.png)
 
+## Challenges
```
pjfanning, line 64: Thanks for the contribution. I must admit that I don't think this new paragraph adds anything. The text is not specific to Pekko. I'm not against updating the docs but I'd prefer to see high level info about the Circuit Breaker pattern appear in external docs and we can link to those as a 'Suggested Reading' section.

azarboon: Thanks for feedback. Sounds good but can you please suggest some external docs where I can make this change? Because I'm not aware of any authoritative and editable docs that explains circuit breaker. Many specific-technology pages explain this pattern but don't tell about its challenge. So editing pages like this are my effort to address this issue.

FYI, I've already added this to Wikipedia page and soon I'll make PR[ to Azure's ](https://learn.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker)

PR #1533, opened 2024-10-15
Implement EventsBySliceQuery in PersistenceTestKitReadJournal 

ptrdom:
Related to https://github.com/apache/pekko/issues/1513, reasoning same as https://github.com/apache/pekko/issues/1527 - it would be useful if testkit had this capability.

`TimestampOffset` could be implemented here, because only real persistence layer implementation that uses this that I am aware of - r2dbc - does use `TimestampOffset`, but I am not sure if it is worth the hassle.


pjfanning:
@ptrdom would you be able to resolve the conflicts when you have time?

ptrdom:
Yes, of course!

pjfanning (approved):
lgtm

PR #1552, opened 2024-11-07
fix: link to GitHub for Hyperic Sigar

raboof:
sigar seems dormant: the website is not on 'https', and has become entirely unreachable recently. The repo hasn't seen an update since 9 years.

Perhaps we should consider whether it's responsible to still link to it at all - but for now let's at least link to the GitHub repo instead of the defunct website.


pjfanning (approved):
lgtm

PR #1554, opened 2024-11-07
exclude provided and optional dependencies from submission

### LINKED ISSUE #1553: dependabot: guava
our GitHub 'security' tab claims that we depend on a version of guava that is vulnerable to GHSA-mvr2-9pj6-7w5j .

I think this is a false positive: I think this is the guava that comes in as a transitive dependency of leveldb, but this is an `optional;provided` dependency. This apparently ends up in the `compile-internal` and `optional` scopes.

I think we should probably exclude the `compile-internal` and `optional` scopes, and use the dependabot security report for artifacts that actually come in as transitive dependencies for our users. The chance that an advisory for an optional/provided/test/built-time dependency actually impacts our build seems to small to justify the noise it adds.

raboof:
This makes it easier to use the dependabot security report for artifacts that actually come in as transitive dependencies for our users. The chance that an advisory for an optional/provided/test/built-time dependency actually impacts our build seems to small to justify the noise it adds.

Fixes #1553


raboof:
Possibly we should make sure those get marked correctly as 'development' dependencies in sbt-dependency-submission (https://github.com/scalacenter/sbt-dependency-submission/issues/234).

Since we don't care about advisories in development dependencies, though, it still makes sense to exclude them here entirely.

raboof:
LOL this is actually a duplicate of my own earlier #1392 :rofl:

Issue #578, opened 2023-08-17
Clustering issues leading to all nodes being downed


fredfp:
I'm reopening here an issue that I reported at the time under the akka repo.

We had a case where an issue on a single node lead to the whole akka-cluster being taken down.

### Here's a summary of what happened: 
1. Healthy cluster made of 20ish nodes, running on k8s
2. Node A: encounters issues, triggers CoordinatedShutdown
3. Node A: experiences high CPU usage, maybe GC pause 
4. Node A: sees B as unreachable, broadcasts it (B is certainly reachable, but detected as such because of high CPU usage, GC pause, or similar issues)
5. Cluster state: A Leaving, B seen unreachable by A, all the other nodes are Up
6. Leader can currently not perform its duties (remove A), reachability status (B seen unreachable by A)
7. Node A: times out some coordinated shutdown phases. Hypothesis: timed out because leader could not remove A.
8. Node A: finishes coordinated shutdown nonetheless.
9. hypothesis - Node A: quarantined associations to other cluster nodes
10. Nodes B, C, D, E: SBR took decision DownSelfQuarantinedByRemote and is downing [...] including myself
11. hypothesis - Node B, C, D, E: quarantined associations to other cluster nodes
12. in a few steps, all remaining cluster nodes down themselves: SBR took decision DownSelfQuarantinedByRemote
13. the whole cluster is down

### Discussions, potential issues:

Considering the behaviour of CoordinatedShutdown (phases can time out and shutdown continues), shouldn't the leader ignore unreachabilities added by a Leaving node and be allowed to perform its duties?
At step 6 above, the Leader was blocked from removing A, but A still continued its shutdown process. The catastrophic ending could have been stopped here.

DownSelfQuarantinedByRemote: @patriknw 's [comment](https://github.com/akka/akka/pull/29737#discussion_r515906571) seems spot on.
At step 9, nodes B, C, D, E should probably not take into account the `Quarantined` from a node that is Leaving.

DownSelfQuarantinedByRemote: another case where Patrik's [comment](https://github.com/akka/akka/pull/29737#discussion_r515906571) also seems to apply, `Quarantined` from nodes downing themselves because of DownSelfQuarantinedByRemote should probably not be taken into account.

At steps 10 and 12. Any cluster singletons running on affected nodes wouldn't be gracefully shutdown using the configured termination message. This is probably the right thing to do but I'm adding this note here nonetheless.

fredfp:
I have extra logs that may be useful:

[...]

PR #1555, opened 2024-11-08
add non-default config that allows InboundQuarantineCheck to ignore 'harmless' quarantine events

pjfanning:
* relates to #578
* basic tests that now watch for the quarantine event and with an experimental change to try to suppress that quarantine event when harmless=true
* this suppression is non-default and can be enabled with the config `pekko.remote.artery.propagate-harmless-quarantine-events = off`

### DIFF
--- a/remote/src/main/mima-filters/1.1.x.backwards.excludes/quarantine.backwards.excludes
+++ b/remote/src/main/mima-filters/1.1.x.backwards.excludes/quarantine.backwards.excludes
@@ -1,0 +1,23 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# changes made due to issues with downing during harmless quarantine
+# https://github.com/apache/pekko/issues/578
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.copy")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.apply")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.unapply")
--- a/remote/src/main/mima-filters/1.1.x.backwards.excludes/quarantine.backwards.excludes
+++ b/remote/src/main/mima-filters/1.1.x.backwards.excludes/quarantine.backwards.excludes
@@ -21,3 +21,4 @@ ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.remote.arte
 ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.this")
 ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.apply")
 ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.remote.artery.AssociationState#QuarantinedTimestamp.unapply")
+ProblemFilters.exclude[MissingTypesProblem]("org.apache.pekko.remote.artery.AssociationState$QuarantinedTimestamp$")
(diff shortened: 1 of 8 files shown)


fredfp:
> we may need to modify the test harmless=true test to send a message from one cluster member to the other to cause the shutdown issue

Without an active SBR, no node will be shutdown: it is the SBR that downs itself when receiving `ThisActorSystemQuarantinedEvent`. Without a cluster setup in the test (and as such without SBR running), we need to watch the `ThisActorSystemQuarantinedEvent` event instead, which is what I did to check the bug exists (test passes if the bug exists):
```scala
"eliminate quarantined association when not used (harmless=true)" in withAssociation {
  (remoteSystem, remoteAddress, _, localArtery, localProbe) =>
    remoteSystem.eventStream.subscribe(testActor, classOf[ThisActorSystemQuarantinedEvent]) // event to watch out for, indicator of the issue

    val remoteEcho = remoteSystem.actorSelection("/user/echo").resolveOne(remainingOrDefault).futureValue

    val localAddress = RARP(system).provider.getDefaultAddress

    val localEchoRef = remoteSystem.actorSelection(RootActorPath(localAddress) / localProbe.ref.path.elements).resolveOne(remainingOrDefault).futureValue
    remoteEcho.tell("ping", localEchoRef)
    localProbe.expectMsg("ping")

    val association = localArtery.association(remoteAddress)
    val remoteUid = futureUniqueRemoteAddress(association).futureValue.uid
    localArtery.quarantine(remoteAddress, Some(remoteUid), "HarmlessTest", harmless = true)
    association.associationState.isQuarantined(remoteUid) shouldBe true

    eventually {
      remoteEcho.tell("ping", localEchoRef) // trigger sending message from remote to local, which will trigger local to wrongfully notify remote that it is quarantined
      expectMsgType[ThisActorSystemQuarantinedEvent] // this is what remote emits when it learns it is quarantined by local. This is not correct and is what (with SBR enabled) triggers killing the node.
    }
}
```

fredfp:
What would it add to move the test to the cluster or cluster-tests projects? To me this is a bug of the `remote` module and is better tested here. For sure, you could test consequences of the bug in cluster, but the root cause is here.

pjfanning:
I've added a change to InboundQuarantineCheck based on https://github.com/apache/pekko/issues/578#issuecomment-2115847363. This may not be the best solution but it seems to help in this one test case.

fredfp:
It seems good to me like that, thank you!

pjfanning:
@raboof @mdedetrich @jrudolph what do you think about the runtime change? We could add a config to users to control if the new runtime check is enabled.

raboof (commented):
I finally got a chance to review this change.

The fact that 'harmless' quarantines are not propagated, but are returned in `InboundQuarantineCheck`, indeed looks like a bug.

I think there's a good possibility that this caused the scenario in #578 and that it'd be worth releasing this new behavior. I'd even lean towards making the new behavior the default, but I'm also OK with being conservative and first testing with the reporters of #578.

Whether this fix completely prevents situations like the one described in #578 is harder to say with confidence: there's still quite a few situations that can cause non-harmless quarantines, and it's possible there's still situations where a single misbehaving node may 'take out' its surrounding peers.

pjfanning:
Thanks for the review. I was running the tests a few weeks ago and I'm not sure that they work. @fredfp tested this a few weeks ago and said it helped but I don't yet think this is ready - until the new tests are improved.

He-Pin:
A workmate told me this ,but as I am not using the cluster at work, so thanks for take care of this hard task.

I knew a team is using akka/pekko cluster with a centralized nodes server, we called it vipserver,  which will decided which nodes are on the same cluster as a single source of truth instead of this gossip thing, where a wrong message spreading the cluster can take the whole cluster down.

Another reason I'm not using clustering at work is because the chaos monkey,  where the sre team will schedule some randomly network partition, and I think that will always require a reboot by the application owner, I am a little lazy here.

pjfanning:
The tests seem to be working for me today. If reviewers are amenable, we could merge this to main and 1.1.x branches and document that the config exists to enable an experimental fix for #578.

raboof (approved):

He-Pin:
@pjfanning @raboof I think we should backport this to 1.1.0, for people who can't upgrade netty but need this.

pjfanning:
I can live with this being backported. It is a reasonably big change for a patch release but it is disabled by default so users who upgrade to 1.1.6 won't get this change unless they update their config settings.

pjfanning:
@He-Pin doesn't this only affect users who use Artery remoting - and you say elsewhere that you are blocked by Netty issues?

He-Pin:
@pjfanning Yes, she is using Artery, transport, but to fix this, she has to upgrade to pekko 1.2, and then Netty 4.2 cause another problem :(

He-Pin:
@fredfp Did you try this release, would you like to share the result?

fredfp:
We didn't try the fix (as it's only included in a milestone releases). I'd be very happy to turn the needed flag on if this made it into 1.1.6. However only time will tell as we don't get that kind of cluster crash very often.

He-Pin:
@fredfp Thanks, there is a backport to 1.1.6 and 1.2.0 will be released soon too.

fredfp:
Over the last couple of days, we've successfully deployed pekko 1.2.0 with `pekko.remote.artery.propagate-harmless-quarantine-events = off` in 2 clusters. Both seem stable, and more stable than before i.e., no more nodes downed with `DownSelfQuarantinedByRemote`. A great success so far, thank you!

He-Pin:
@fredfp Thank you for that udpate~~, great sharing.

He-Pin:
@fredfp Are you using the pekko-management at the sametime?

fredfp:
Yes, we rely on pekko-management to create and then join the cluster.

He-Pin:
One of my workmates said, her cluster node is killed by k8s, not sure why :(, k8s say the management port is unreachable, and the kill the node.

pjfanning:
> One of my workmates said, her cluster node is killed by k8s, not sure why :(, k8s say the management port is unreachable, and the kill the node.

@He-Pin this code has nothing to do with kubernetes. If you have pekko-management issues, it would be better to raise issues in the pekko-management project.

He-Pin:
@pjfanning I knew, I just asking @fredfp if he knew that issue too.

PR #1556, opened 2024-11-09
jackson 2.18 with max-token-count support

pjfanning:
* supports https://github.com/FasterXML/jackson-core/pull/1296
* Jackson lock-free buffer recyclers turned out to have major trouble and have been deprecated - https://github.com/trinodb/trino/issues/21356

### DIFF
--- a/serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala
+++ b/serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala
@@ -160,8 +160,6 @@ object JacksonObjectMapperProvider extends ExtensionId[JacksonObjectMapperProvid
   private def getBufferRecyclerPool(cfg: Config): RecyclerPool[BufferRecycler] = {
     cfg.getString("buffer-recycler.pool-instance") match {
       case "thread-local"            => JsonRecyclerPools.threadLocalPool()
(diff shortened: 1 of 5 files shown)


serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala
```diff
@@ -159,11 +160,10 @@ object JacksonObjectMapperProvider extends ExtensionId[JacksonObjectMapperProvid
   private def getBufferRecyclerPool(cfg: Config): RecyclerPool[BufferRecycler] = {
     cfg.getString("buffer-recycler.pool-instance") match {
       case "thread-local"            => JsonRecyclerPools.threadLocalPool()
-      case "lock-free"               => JsonRecyclerPools.newLockFreePool()
-      case "shared-lock-free"        => JsonRecyclerPools.sharedLockFreePool()
```
He-Pin, line 163: So there is been removed and should we mention that in the release note?

He-Pin (approved):
lgtm, but seems need be included in the release note.

pjfanning:
We add release notes for every release. Generally, they are created around release time.

serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala
pjfanning: yes

PR #1567, opened 2024-11-27
Allow cluster formation with mixed protocols

sadekmunawar:
Rolling upgrade from Akka to Pekko requires cluster formation with mixed protocol nodes. If the config `accept-protocol-names` contains both "akka" and "pekko", nodes using either protocol should be able join the same cluster.

Relates to [#765](https://github.com/apache/pekko/pull/765)

### DIFF
--- a/cluster/src/test/scala/org/apache/pekko/cluster/MixedProtocolClusterSpec.scala
+++ b/cluster/src/test/scala/org/apache/pekko/cluster/MixedProtocolClusterSpec.scala
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.cluster
+
+import com.typesafe.config.{ Config, ConfigFactory }
+
+import org.apache.pekko.testkit.{ LongRunningTest, PekkoSpec }
+
+object MixedProtocolClusterSpec {
+
+  val baseConfig: Config =
+    ConfigFactory.parseString("""
+     pekko.actor.provider = "cluster"
+     pekko.coordinated-shutdown.terminate-actor-system = on
+
+     pekko.remote.classic.netty.tcp.port = 0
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
thanks @sadekmunawar - this looks like a necessary change

raboof (approved):
Another great observation, and thanks for also adding a test!

raboof:
(I'll backport)

pjfanning:
@sadekmunawar do you think doing a release with your changes is a good idea or are you in the middle of testing mixed clusters and feel like you might find more issues?

sadekmunawar:
> @sadekmunawar do you think doing a release with your changes is a good idea or are you in the middle of testing mixed clusters and feel like you might find more issues?

@pjfanning I had to make one more change for the migration. I'll create a PR by the end of the week. Once that's merged, I think a release can be done.

pjfanning:
@sadekmunawar any news on progress? If you don't have time to create a PR for the remaining issue, could you give us some pointers so someone else could create the PR instead?

sadekmunawar:
@pjfanning There two missing pieces left:
1. Deserialization of Akka cluster messages. I created the PR for this one #1578 
2. Supporting Akka config for JoinConfigCompatChecker. An exception is thrown [here](https://github.com/apache/pekko/blob/main/cluster/src/main/scala/org/apache/pekko/cluster/JoinConfigCompatCheckCluster.scala#L40-L42), when a Pekko node tries to find the downing path in the Akka config. I solved it by simply checking if the path exists in the config before doing the comparison. 

```
    val downingProviderResult = {
      if (toCheck.hasPath(DowningProviderPath) && actualConfig.hasPath(DowningProviderPath))
        JoinConfigCompatChecker.checkEquality(List(DowningProviderPath), toCheck, actualConfig)
      else Invalid(im.Seq("Downing path not found. Set configuration-compatibility-check.enforce-on-join to off to enable cluster joining."))
    }
```

While this all that's necessary for the rolling upgrade support, I tried to find a better approach for the second problem. See this [commit](https://github.com/sadekmunawar/pekko/commit/60f208f7c9690c98356466094ec97516cc0f8230). I haven't had the chance to integration test this approach, yet. It should fix the exception issue, but it likely sends configs with Pekko paths to Akka nodes [here](https://github.com/sadekmunawar/pekko/commit/60f208f7c9690c98356466094ec97516cc0f8230#diff-93928721f2ceeb4acf73b9c1582977f8c1157feefbf64e6d5f8a694e5ae40a5aR647-R654)

pjfanning:
I am going to revert this due to #1586 

A build of #1587 with the nightly builds temporarily enabled shows that this PR is the cause.

We can try to fix this issue again but we need a working build first.

cluster/src/test/scala/org/apache/pekko/cluster/MixedProtocolClusterSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.pekko.cluster
+
+import com.typesafe.config.{ Config, ConfigFactory }
+
+import org.apache.pekko.testkit.{ LongRunningTest, PekkoSpec }
+
+object MixedProtocolClusterSpec {
```
pjfanning, line 24: this test doesn't work in the nightly build CI job and the change breaks many other cluster tests too

PR #1578, opened 2024-12-10
Enable deserialization of old Akka cluster messages (mixed pekko/akka cluster)

sadekmunawar:
Forming a cluster with Akka nodes requires the deserialization of cluster messages sent by the Akka. This commit fixes the following exception that occurs during deserialization.

```
[akka://HybridCluster@127.0.0.1:2551] with serializer id [5] and manifest [akka.cluster.InternalClusterAction$InitJoinAck].
java.lang.IllegalArgumentException: Unknown manifest [akka.cluster.InternalClusterAction$InitJoinAck]
    at org.apache.pekko.cluster.protobuf.ClusterMessageSerializer.fromBinary(ClusterMessageSerializer.scala:156)
```

### DIFF
--- a/cluster/src/main/scala/org/apache/pekko/cluster/protobuf/ClusterMessageSerializer.scala
+++ b/cluster/src/main/scala/org/apache/pekko/cluster/protobuf/ClusterMessageSerializer.scala
@@ -125,35 +125,43 @@ final class ClusterMessageSerializer(val system: ExtendedActorSystem)
       throw new IllegalArgumentException(s"Can't serialize object of type ${obj.getClass} in [${getClass.getName}]")
   }
 
-  def fromBinary(bytes: Array[Byte], manifest: String): AnyRef = manifest match {
-    case HeartbeatManifest         => deserializeHeartBeat(bytes)
-    case HeartbeatRspManifest      => deserializeHeartBeatResponse(bytes)
-    case GossipStatusManifest      => deserializeGossipStatus(bytes)
-    case GossipEnvelopeManifest    => deserializeGossipEnvelope(bytes)
-    case InitJoinManifest          => deserializeInitJoin(bytes)
-    case InitJoinAckManifest       => deserializeInitJoinAck(bytes)
-    case InitJoinNackManifest      => deserializeInitJoinNack(bytes)
-    case JoinManifest              => deserializeJoin(bytes)
-    case WelcomeManifest           => deserializeWelcome(bytes)
-    case LeaveManifest             => deserializeLeave(bytes)
-    case DownManifest              => deserializeDown(bytes)
-    case ExitingConfirmedManifest  => deserializeExitingConfirmed(bytes)
-    case ClusterRouterPoolManifest => deserializeClusterRouterPool(bytes)
-    // needs to stay in Akka 2.6.5 to be able to talk to an Akka 2.5.{3,4} node during rolling upgrade
-    case HeartBeatManifestPre2523     => deserializeHeartBeatAsAddress(bytes)
-    case HeartBeatRspManifest2523     => deserializeHeartBeatRspAsUniqueAddress(bytes)
-    case OldGossipStatusManifest      => deserializeGossipStatus(bytes)
-    case OldGossipEnvelopeManifest    => deserializeGossipEnvelope(bytes)
-    case OldInitJoinManifest          => deserializeInitJoin(bytes)
-    case OldInitJoinAckManifest       => deserializeInitJoinAck(bytes)
-    case OldInitJoinNackManifest      => deserializeInitJoinNack(bytes)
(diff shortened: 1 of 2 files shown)


cluster/src/main/scala/org/apache/pekko/cluster/protobuf/ClusterMessageSerializer.scala
```diff
@@ -125,35 +125,43 @@ final class ClusterMessageSerializer(val system: ExtendedActorSystem)
...
-    case OldJoinManifest              => deserializeJoin(bytes)
-    case OldWelcomeManifest           => deserializeWelcome(bytes)
-    case OldLeaveManifest             => deserializeLeave(bytes)
-    case OldDownManifest              => deserializeDown(bytes)
-    case OldExitingConfirmedManifest  => deserializeExitingConfirmed(bytes)
-    case OldClusterRouterPoolManifest => deserializeClusterRouterPool(bytes)
-    case _                            => throw new IllegalArgumentException(s"Unknown manifest [$manifest]")
+  def fromBinary(bytes: Array[Byte], manifest: String): AnyRef = {
+    val updatedManifest = {
+      if (manifest.startsWith("akka"))
```
He-Pin, line 130: this is only needed when migrating from Akka, may be better under a boolean guard.

He-Pin: IIRC, @mdedetrich  once added one, and then is can be `if (guard && manifest.startsWith("akka"))`, WDYT

pjfanning: I can provide a PR for this based on checking `pekko.remote.accept-protocol-names` config. That config is an array value and if "akka" is in the array then we can allow this check. We only need to do this config once so the boolean result can be stored as a val.

cluster/src/test/scala/org/apache/pekko/cluster/protobuf/ClusterMessageSerializerSpec.scala
```diff
@@ -177,6 +177,14 @@ class ClusterMessageSerializerSpec extends PekkoSpec("pekko.actor.provider = clu
         ClusterMessageSerializer.OldWelcomeManifest)
     }
 
+    "be de-serializable with class manifests from Akka nodes" in {
+      val oldAkkaJoinAckManifest = s"org.apache.pekko.cluster.InternalClusterAction$$InitJoinAck"
```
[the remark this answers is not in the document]
pjfanning: @sadekmunawar this class only needs to be supported for Akka prior to v2.6.4.

Do we really need to support clusters that are running with very old Akka releases? Ideally we would only support pretty recent Akka releases. It is a pity that Akka changed the cluster messages in a patch release (2.6.5).

https://github.com/apache/pekko/blob/5d4e8283c3b530cdec56357d0f5f3bfb50effbfd/cluster/src/main/scala/org/apache/pekko/cluster/protobuf/ClusterMessageSerializer.scala#L44-L58

sadekmunawar: I upgraded from Akka-2.6.2, so this change was necessary. But updating the documentation to advise users to perform the rolling upgrade from a version higher than 2.6.4 may be sufficient.

An alternative approach would be to revert the old manifest back to Akka (e.g.  ` val OldJoinManifest = s"akka.cluster.InternalClusterAction$$Join"`). No version of Pekko is expected to use the old manifest in the cluster messages, right?

cluster/src/main/scala/org/apache/pekko/cluster/protobuf/ClusterMessageSerializer.scala
sadekmunawar: I agree. If we decide to keep the changes in this PR, then having boolean guard would be better.

pjfanning:
I have updated https://cwiki.apache.org/confluence/display/PEKKO/Pekko+Akka+Compatibility and include the fact the we only support forming clusters with Akka nodes of version 2.6.5 and above.

raboof:
Does that mean we can close this PR?

pjfanning:
I would prefer not to use this because I think trying to support Akka before v2.6.5 almost certainly will lead to us having to add extra fixes. One that I suspect that we'd need is that we might need to also allow Pekko nodes to optionally send cluster messages that look like the old Akka format and this PR does not address that.

raboof:
> I would prefer not to use this because I think trying to support Akka before v2.6.5 almost certainly will lead to us having to add extra fixes.

Sounds reasonable to me

PR #1587, opened 2024-12-16
revert #1568 due to nightly build issues

pjfanning:
see #1586

PR #1588, opened 2024-12-16
Revert "Allow cluster formation with mixed protocols"

pjfanning:
Reverts apache/pekko#1568

see #1587 

I raised #1590 to revisit


He-Pin (approved):
Lgtm

He-Pin:
A problem is the cluster test is only enabled  when nightly，we may want to enable it in pr validation

PR #1589, opened 2024-12-16
Update pins in .scala-steward.conf

pjfanning:
* Mockito seems to be a transitive dependency so we probably don't need to pin it.
* see #1582 also
* also allow for patch version updates on agrona

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -11,6 +11,10 @@ updates.pin = [
   { groupId = "com.google.protobuf", artifactId = "protobuf-java", version = "3." }
   # aeron 1.46 requires Java 17
   { groupId = "io.aeron", version = "1.45." }
+  # agrona 1.23 requires Java 17
+  { groupId = "org.agrona", artifactId = "agrona" version = "1.22." }
+  # bndlib 7 requires Java 17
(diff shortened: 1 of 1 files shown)


.scala-steward.conf
```diff
@@ -11,6 +11,10 @@ updates.pin = [
   { groupId = "com.google.protobuf", artifactId = "protobuf-java", version = "3." }
   # aeron 1.46 requires Java 17
   { groupId = "io.aeron", version = "1.45." }
+  # agrona 1.23 requires Java 17
+  { groupId = "org.agrona", artifactId = "agrona" version = "1.22." }
```
raboof, line 15: I seem to remember the reason we dodn't let scala steward update agrona is because aeron and agrona are somewhat tightly coupled, so we only update agrona when we update aeron (to the exact version that aeron expects).

pjfanning: the problem is that with scala steward, we might accept the aeron update and forget the agrona one - with this change, we will be warned about a new jar - we don't need to merge it but we can research if the change is a good idea

raboof: The comments around the versions in `project/Dependencies.scala` seem fairly clear - but I see at https://github.com/real-logic/aeron/blob/1.45.1/build.gradle#L59 it might be less tightly coupled than I thought, so :+1: I suppose

PR #1593, opened 2024-12-16
feat: Add SchedulerTask which will be notified once cancelled.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1579

Tests is still missing

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
@@ -131,6 +131,17 @@ trait SchedulerSpec extends BeforeAndAfterEach with DefaultTimeout with Implicit
         task.isCancelled should ===(true)
       }
 
+      "notify callback if cancel is performed before execution" taggedAs TimingTest in {
+        val latch = new CountDownLatch(1)
+        val task = system.scheduler.scheduleOnce(100 millis,
+          new SchedulerTask {
+            override def run(): Unit = ()
+            override def cancelled(): Unit = latch.countDown()
+          })
+        task.cancel()
+        latch.await(100, TimeUnit.MILLISECONDS) should ===(true)
+      }
+
       "not be canceled if cancel is performed after execution" taggedAs TimingTest in {
         val latch = TestLatch(1)
         val task = collectCancellable(system.scheduler.scheduleOnce(10.millis)(latch.countDown()))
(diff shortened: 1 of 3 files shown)


actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
```diff
@@ -390,7 +395,19 @@ object LightArrayRevolverScheduler {
...
+        true
+      case _ => true
+    }
+
+    private def notifyCancellation(task: SchedulerTask): Unit = {
+      try {
+        task.cancelled()
+      } catch {
+        case Throwable =>
+        // TODO how to logging here?
```
He-Pin, line 409: Can't access the `loggingAdapter` here :(

He-Pin: I think these method should run on a dedicated executor?

pjfanning:
I'm not sure that we could add this in a v1.1 patch. I'd prefer if we delay this till v1.2.0.

He-Pin:
I'm fine with this, but as the current method is add as a default method, it should be fine too.

pjfanning (approved):
lgtm

actor-tests/src/test/scala/org/apache/pekko/actor/SchedulerSpec.scala
```diff
@@ -131,6 +131,17 @@ trait SchedulerSpec extends BeforeAndAfterEach with DefaultTimeout with Implicit
         task.isCancelled should ===(true)
       }
 
+      "notify callback if cancel is performed before execution" taggedAs TimingTest in {
+        val latch = new CountDownLatch(1)
+        val task = system.scheduler.scheduleOnce(10 seconds,
```
pjfanning, line 136: can we have a shorter schedule? to save time on the test run

pjfanning:
I'm quite worried that these tests will be brittle in CI. The CI agents can often suffer from performance issues and tests can take longer to run than on laptops, etc.

actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala
```diff
@@ -498,6 +503,20 @@ trait Scheduler {
 // this one is just here so we can present a nice AbstractScheduler for Java
 abstract class AbstractSchedulerBase extends Scheduler
 
+/**
+ * A Task that will be notified when it is cancelled.
+ *
+ * @since 1.2.0
```
He-Pin, line 509: add the since annotation

PR #1611, opened 2024-12-17
bug: fix occasional ordering issue in result in unsafeOptionalDataVia

He-Pin:
refs: https://github.com/apache/pekko/issues/1610

Motivation:
Fix flaky test in unsafeOptionalDataVia

Modification:
1. rewrite with `mergeSequence`

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/SourceWithContext.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/SourceWithContext.scala
@@ -57,36 +57,44 @@ object SourceWithContext {
     SourceWithContext.fromTuples(Source.fromGraph(GraphDSL.createGraph(source, viaFlow)(combine) {
       implicit b => (s, viaF) =>
         import GraphDSL.Implicits._
-        val broadcast = b.add(Broadcast[(Option[SOut], Ctx)](2))
-        val merge = b.add(Merge[(Option[FOut], Ctx)](2))
 
-        val unzip = b.add(Unzip[SOut, Ctx]())
-        val zipper = b.add(Zip[FOut, Ctx]())
+        case class IndexedCtx(idx: Long, ctx: Ctx)
+        val partition = b.add(Partition[(Option[SOut], IndexedCtx)](2,
+          {
+            case (None, _)    => 0
+            case (Some(_), _) => 1
+          }))
 
-        val filterAvailable = Flow[(Option[SOut], Ctx)].collect {
-          case (Some(f), ctx) => (f, ctx)
+        val sequence = Flow[(Option[SOut], Ctx)].zipWithIndex
+          .map {
+            case ((opt, ctx), idx) => (opt, IndexedCtx(idx, ctx))
+          }
+
+        val unzip = b.add(Unzip[Option[SOut], IndexedCtx]())
+        val zipper = b.add(Zip[FOut, IndexedCtx]())
+        val mergeSequence = b.add(MergeSequence[(Option[FOut], IndexedCtx)](2)(_._2.idx))
+        val unwrapSome = b.add(Flow[Option[SOut]].map {
+          case Some(elem) => elem
+          case _          => throw new IllegalStateException("Only expects Some")
+        })
+        val unwrap = Flow[(Option[FOut], IndexedCtx)].map {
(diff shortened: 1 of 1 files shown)


mdedetrich (approved):
lgtm

raboof (commented):
shouldn't `unsafeOptionalDataVia` keep the order?

I thought the 'unsafe' aspect was the `Flow` parameter must be a flow that produces 1 output per input without reordering (and we can't 'safely' express that in the typesystem), but as long as that's the case (as it is here) it should keep order.

Shouldn't we change the implementation rather than the test?

He-Pin:
@raboof That's what's I'm wondering, because the current design ,it can not keep order, not sure if that's how @mdedetrich expected it works.

mdedetrich:
Let me retract my approval, I didn't pay attention to the change.

Indeed the ordering of the elements should not change, @raboof is correct where the unsafe part pertains to the fact that you could provide a Flow that reorders/drops elements and there isn't any check for this but assuming you provide a flow that only transforms elements than ordering should be retained.

Funnily enough I haven't experienced this in production, ill have a look at it.

He-Pin:
@mdedetrich Can you give some more context of the current design？

mdedetrich (changes requested):
Blocking for now as the ordering is meant to be determinstic, prob the implementation needs to be fixed.

mdedetrich:
> @mdedetrich Can you give some more context of the current design？

So the context is primarily when using kafka cursors with FlowWithContext/SourceWithContext. This feature lets you provide a flow that only transforms the data portion of the stream while leaving the context part untouched (which is critical when dealing with Kafka and committing cursors).

He-Pin:
Let me update it, I can workout a solution.

He-Pin:
@raboof @mdedetrich @pjfanning I think it should be OK now, it now track the origin order with an index and then merge back with the same index.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/SourceWithContext.scala
```diff
@@ -57,36 +57,44 @@ object SourceWithContext {
...
-
-        unzip.out0 ~> viaF ~> zipper.in0
-        unzip.out1 ~> zipper.in1
-
-        zipper.out ~> mapIntoOption ~> merge.in(0)
-
-        broadcast.out(1) ~> filterUnavailable ~> merge.in(1)
+        //format: off
+        s ~> sequence ~> partition.in
+        partition.out(0).asInstanceOf[Outlet[(Option[FOut], IndexedCtx)]] ~> mergeSequence.in(0)
```
He-Pin, line 90: cast here to avoid allocation, it will always be `None` because of partition

pjfanning:
> Do we know if the test runs in PR CI builds? I did a quick check of the test logs and didn't find a match.

I found `must Apply a viaFlow with optional elements using unsafeOptionalVia (52 milliseconds)` so the test is running.

pjfanning (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/scaladsl/SourceWithContext.scala
```diff
@@ -57,36 +57,44 @@ object SourceWithContext {
...
-        val unzip = b.add(Unzip[SOut, Ctx]())
-        val zipper = b.add(Zip[FOut, Ctx]())
-
-        val filterAvailable = Flow[(Option[SOut], Ctx)].collect {
-          case (Some(f), ctx) => (f, ctx)
-        }
-
-        val filterUnavailable = Flow[(Option[SOut], Ctx)].collect {
-          case (None, ctx) => (Option.empty[FOut], ctx)
+        case class IndexedCtx(idx: Long, ctx: Ctx)
```
He-Pin, line 61: extra allocation here

mdedetrich (approved):

He-Pin:
Thanks for the review

pjfanning:
It is in v1.1.3-RC1.
Have a look at https://github.com/apache/pekko/commit/7184dadf4a1d57ac07cd4eb26b67b3f8b5ccf47d
and the tags that it is in (v1.1.3-RC1 and v1.2.0-M0).

He-Pin:
Thanks  , seems there are some problem on my local history

PR #1612, opened 2024-12-18
feat: Add required reflection/resource metadata for native image generation

utkuaydn:
I hope the title is self-explanatory. Providing reflection metadata to Graal's native image generation would be a nice addition to the library. The Akka counterparts of the artifacts already had these in place, so the contents were trivial.

### DIFF
--- a/actor-typed/src/main/resources/META-INF/native-image/org.apache.pekko/pekko-actor-typed/native-image.properties
+++ b/actor-typed/src/main/resources/META-INF/native-image/org.apache.pekko/pekko-actor-typed/native-image.properties
@@ -1,0 +1,1 @@
+Args = --enable-monitoring=jfr
--- a/serialization-jackson/src/main/resources/META-INF/native-image/org.apache.pekko/pekko-serialization-jackson/reflect-config.json
+++ b/serialization-jackson/src/main/resources/META-INF/native-image/org.apache.pekko/pekko-serialization-jackson/reflect-config.json
@@ -1,0 +1,234 @@
+[
+  {
+    "name": "org.apache.pekko.serialization.jackson.PekkoJacksonModule",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "org.apache.pekko.serialization.jackson.PekkoTypedJacksonModule",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "org.apache.pekko.serialization.jackson.PekkoStreamJacksonModule",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "com.fasterxml.jackson.module.paramnames.ParameterNamesModule",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "com.fasterxml.jackson.datatype.jdk8.Jdk8Module",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "com.fasterxml.jackson.datatype.jsr310.JavaTimeModule",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "com.fasterxml.jackson.module.scala.DefaultScalaModule",
+    "methods": [
+      {
+        "name": "<init>"
+      }
+    ]
+  },
+  {
+    "name": "org.apache.pekko.serialization.jackson.CborSerializable",
+    "fields": [
+      {
+        "name": "MODULE$"
+      }
+    ]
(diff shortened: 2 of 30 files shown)


raboof:
> The Akka counterparts of the artifacts already had these in place, so the contents were trivial.

Does that mean you derived these artifacts from the Akka artifacts? What license were those Akka artifacts under? If they were BSL-licensed we cannot accept them in Pekko.

I'm a little concerned how we're going to keep those maintained as the library evolves. I remember there was an initiative to maintain such configurations outside of the main Akka tree, https://github.com/vmencik/akka-graal-config . If we could adapt such an approach, the Graal definitions could evolve independently from Pekko 'itself'. Would that be a feasible approach?

pjfanning:
I guess the idea is that Graal users want these files included in our jars. And an outside group can still repackage the Pekko jars and add these files independently.

I haven't done a full review (currently using a mobile phone) but I didn't notice anything to continuously test these config s so that our CI could police changes.

And to reiterate @raboof 's point that we only accept PRs when the work is the submitter's original work. We will not accept code that is taken from Akka unless the submitter to Pekko is the original submitter to Akka. The code writer retains the copyright on the code and can choose to submit it Pekko.

utkuaydn:
About the contents of the changes, they are simply JSON files that were generated via the [native image agent](https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/#tracing-agent) running on the tests, that are in a schema that the native image executable accepts. Correct me if I'm wrong, but I am not sure if this could be considered copyright material or "original work" as it merely contains class paths, field types, etc. which reference the classes in the Pekko org. TBH, I did not double check if all the classes mentioned in the JSON files exist in the artifacts. But we can do due diligence if it is still a concern.

I didn't include the full workflow to keep these configurations updated because firstly, I wanted gauge what everyone thinks about it, and secondly, it would be a lot of work and I thought it could be done in parts.

I would also be fine with having these in a separate project, but I think having them in Pekko itself can help with the release process.

pjfanning:
Fair enough about adding regression tests in a 2nd pass.

Could you write up how to generate these JSON files? Ideally someone else would be able to generate them too. We may even need to regenerate if we make certain changes in Pekko. If we know how the generate them that removes any lingering worries that these files were copied directly from Akka.

slf4j/src/main/resources/META-INF/native-image/ch.qos.logback/logback-classic/reflect-config.json
```diff
@@ -0,0 +1,194 @@
+[
```
pjfanning, line 1: Why would Pekko need this? Shouldn't this be in logback-classic jar?

If it doesn't matter what, these json files go in then, we could just have standalone jars that have these files instead of having the Pekko jars contain then.

pjfanning: logback-classic is an optional dependency - with slf4j, you should be able to add your own compliant logger jar

serialization-jackson/src/main/resources/META-INF/native-image/org.apache.pekko/pekko-serialization-jackson/reflect-config.json
```diff
@@ -0,0 +1,234 @@
+[
```
pjfanning, line 1: pekko-serialization-jackson is configurable and users can choose to add extra Jackson modules - what happens in that case (in terms of the native image)? Will those users have to create their own reflect-config.json.

actor-typed/src/main/resources/META-INF/native-image/org.apache.pekko/pekko-actor-typed/native-image.properties
```diff
@@ -0,0 +1 @@
+Args = --enable-monitoring=jfr
```
pjfanning, line 1: is it a good idea to always enable JFR monitoring?

He-Pin:
the file is generated with tools,  but it's hard to cover every code paths, i think this better be a plug in for automatic generating instead of a static file

He-Pin (commented):
Adding GraalVM native-image reflection/resource metadata is a valuable contribution for users running Pekko in native image environments.

1. **Completeness**: The PR covers all major modules (actor, actor-typed, cluster, persistence, remote, stream, serialization-jackson, etc.) with comprehensive reflect-config.json files.

2. **Maintenance concern**: These JSON files are large and will need to be kept in sync as new classes are added. Consider adding CI validation or documentation about how to regenerate them when new reflective classes are introduced.

3. **Staleness**: Open since December 2024. The reviews from @pjfanning appear to be informational comments. Has this been tested against recent versions of GraalVM native-image? The metadata format has evolved over time. Would be good to verify compatibility with the latest GraalVM before merging.

PR #1614, opened 2024-12-21
feat: Add Sink.none operator

He-Pin:
Motivation:

refs: https://github.com/apache/pekko/issues/1613
refs https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.10.0 too.

Result:
1. add `Sink#none` operator
2. The doc renders OK

image

### DIFF
--- a/docs/src/main/paradox/stream/operators/Sink/none.md
+++ b/docs/src/main/paradox/stream/operators/Sink/none.md
@@ -29,10 +29,10 @@ when the input stream ends, or completed with `Failure` if there is a failure si
 This example tests all elements in the stream is `<=` 100.
 
 Scala
-:   @@snip [ForAll.scala](/docs/src/test/scala/docs/stream/operators/sink/NoneMatch.scala) { #none }
+:   @@snip [NoneMatch.scala](/docs/src/test/scala/docs/stream/operators/sink/NoneMatch.scala) { #none }
 
 Java
-:   @@snip [ForAll.java](/docs/src/test/java/jdocs/stream/operators/sink/NoneMatch.java) { #none }
+:   @@snip [NoneMatch.java](/docs/src/test/java/jdocs/stream/operators/sink/NoneMatch.java) { #none }
 
 ## Reactive Streams Semantics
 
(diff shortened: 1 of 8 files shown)


pjfanning (commented):
it is not usually acceptable to add new features in a patch (in semver)

He-Pin:
If there will be a 1.2.0 release, I suggest we delete some deprecated code and maybe drop Java 8 and Scala 2.12 too.

He-Pin:
But if we only delivery features in minor release , that will slow us down
I agree that we should only delete things in minor release.

docs/src/main/paradox/stream/operators/Sink/none.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Example
+
+This example tests all elements in the stream is `<=` 100.
+
+Scala
+:   @@snip [ForAll.scala](/docs/src/test/scala/docs/stream/operators/sink/NoneMatch.scala) { #none }
+
+Java
+:   @@snip [ForAll.java](/docs/src/test/java/jdocs/stream/operators/sink/NoneMatch.java) { #none }
```
raboof, line 35:
```suggestion
:   @@snip [NoneMatch.scala](/docs/src/test/scala/docs/stream/operators/sink/NoneMatch.scala) { #none }

Java
:   @@snip [NoneMatch.java](/docs/src/test/java/jdocs/stream/operators/sink/NoneMatch.java) { #none }
```

raboof:
Seems like a nice utility.

I agree we shouldn't slow ourselves down.

In this case following semver strictly just means the 'since' tag should be 1.2.0 and the next release will be 1.2.0, right? That seems fine to me?

pjfanning:
> Seems like a nice utility.
> 
> I agree we shouldn't slow ourselves down.
> 
> In this case following semver strictly just means the 'since' tag should be 1.2.0 and the next release will be 1.2.0, right? That seems fine to me?

Having a 1.2.0 release does come with an overhead of adding extra testing in our other repos to test with 1.0, 1.1, 1.2 jars but it is achievable.

He-Pin:
Maybe we can relax that, which means only removing things in the minor version, this way, we can move faster, otherwise , any contributor will waiting a little longer to have their contributions been released.

pjfanning (approved):
lgtm

PR #1615, opened 2024-12-21
chore: null out ref for closed stream.

He-Pin:
Motivation:
reduce hard ref to help gc
refs: https://github.com/apache/pekko/discussions/1566

Modification:
Null out the ref when the `Inlet` will never be used.


He-Pin:
Okay, just forgot it's immutable...

PR #1616, opened 2024-12-21
chore: avoid create an empty array everytime.

He-Pin:
Motivation:
Code clean up and performance.

Modification:
I found this todo when read the code, and we should avoid creating the empty array everytime.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
@@ -689,14 +689,21 @@ private final case class SavedIslandData(
 /**
  * INTERNAL API
  */
+@InternalApi
+private[pekko] object GraphStageIsland {
+  // used as a type hint when creating the array of logics
+  private final val emptyLogicArray = Array.empty[GraphStageLogic]
+}
+
+/**
+ * INTERNAL API
+ */
 @InternalApi private[pekko] final class GraphStageIsland(
     effectiveAttributes: Attributes,
     materializer: PhasedFusingActorMaterializer,
     islandName: String,
     subflowFuser: OptionVal[GraphInterpreterShell => ActorRef])
     extends PhaseIsland[GraphStageLogic] {
-  // TODO: remove these
-  private val logicArrayType = Array.empty[GraphStageLogic]
   private[this] val logics = new util.ArrayList[GraphStageLogic](16)
 
   private var connections = new Array[Connection](16)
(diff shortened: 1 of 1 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
```diff
@@ -686,6 +686,15 @@ private final case class SavedIslandData(
  */
 @InternalApi private[pekko] object GraphStageTag extends IslandTag
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[pekko] object GraphStageIsland {
+  // used as a type hint when creating the array of logics
+  private final val logicArrayType = Array.empty[GraphStageLogic]
```
raboof, line 695: `Array` is mutable, so doesn't this mean everyone shares the same `logics`? Wouldn't that be problematic? (though it's interesting that the tests succeed)

stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
```diff
@@ -695,8 +704,6 @@ private final case class SavedIslandData(
     islandName: String,
     subflowFuser: OptionVal[GraphInterpreterShell => ActorRef])
     extends PhaseIsland[GraphStageLogic] {
-  // TODO: remove these
```
raboof, line 698: does 'these' suggests the `logics` should also be removed here?

stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
He-Pin: You can't, it's size 0

pjfanning: This seems safe enough because it is an empty array. Can we rename it as emptyArray? I would still be very surprised if this improves performance. The hotspot compiler should quickly deal with this if is a hotspot.

stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
He-Pin: it's been used in `materializeAtomic` method.

pjfanning (approved):
lgtm - main is now dev branch for 1.2.0

jxnu-liguobin (approved):

PR #1617, opened 2024-12-21
chore: tidy up implementation of PhasedFusingActorMaterializer

He-Pin:
Motivation:
Code clean-up with IDEA

Modification:
remove `s` and make fields private.


raboof (approved):
Removing `s` and making return values explicit is nice.

I'm not sure about adding `private`: the advantage is that it reduces the scope of the variable, but it's already in a `@InternalApi private[pekko] final` class so I'm not sure it's worth the noise. I'm OK with it though.

PR #1619, opened 2024-12-21
chore: Remove SupervisedGraphStageLogic

He-Pin:
Motivation:
refs:https://github.com/apache/pekko/issues/1618

Modification:
remove SupervisedGraphStageLogic

### DIFF
--- a/stream/src/main/mima-filters/1.0.x.backwards.excludes/pr-1619-remove-supervised-group-stage-logic.backwards.excludes
+++ b/stream/src/main/mima-filters/1.0.x.backwards.excludes/pr-1619-remove-supervised-group-stage-logic.backwards.excludes
@@ -15,5 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 
-ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.stream.scaladsl.FlowWithContextOps.wireTap")
-ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.stream.scaladsl.FlowWithContextOps.wireTapContext")
+# Remove SupervisedGraphStageLogic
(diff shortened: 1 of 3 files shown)


stream/src/main/mima-filters/1.0.x.backwards.excludes/pr-1619-remove-supervised-group-stage-logic.backwards.excludes
```diff
@@ -0,0 +1 @@
+ ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.stream.impl.fusing.SupervisedGraphStageLogic")
```
pjfanning, line 1: we can't do this in Pekko 1.1 - this will need to be in Pekko 1.2 or even 2.0

raboof: Anything in the `impl` package is internal implementation and not public API and the class is `private[pekko]`, so it might be fine?

raboof (approved):

stream/src/main/mima-filters/1.0.x.backwards.excludes/pr-1619-remove-supervised-group-stage-logic.backwards.excludes
pjfanning: @He-Pin can you add a license header (the standard Apache one, not the one that mentions Akka)? Can you add a new line at the end of the file?

He-Pin: I missed that:(

PR #1622, opened 2024-12-22
fix: Fix leak in FlatMapPrefix operator.

He-Pin:
extracted from https://github.com/apache/pekko/pull/1621

Motivation:
In https://github.com/apache/pekko/discussions/1566 , @queimadus points out a leak

Modification:
1. change the builder from a buffer to a vector builder.
2. free that builder to null after the prefix is generated.

Result:
leak fixed.

```scala
import org.apache.pekko.actor.ActorSystem
import org.apache.pekko.stream.scaladsl.{Flow, Sink, Source}
import org.apache.pekko.util.ByteString

import scala.concurrent.Await


object PekkoQuickstart extends App {
  private implicit val system: ActorSystem = ActorSystem()

  val s = Source
    .repeat(())
    .map(_ => ByteString('a' * 400000))
    .take(1000000)
    .flatMapPrefix(50000) { _ => Flow[ByteString] }

  val r = Source.empty
    .concatAllLazy(List.tabulate(30000)(_ => s): _*)
    .runWith(Sink.ignore)

  Await.result(r, scala.concurrent.duration.Duration.Inf)
  println(r.value)

//  Source
//    .repeat(s)
//    .take(30000)
//    .flatMapConcat(x => x)
//    .runWith(Sink.ignore)
//    .onComplete(println(_))

//  Source.empty
//    .concatAllLazy(List.tabulate(30000)(_ => Source.lazySource(() => s)): _*)
//    .runWith(Sink.ignore).onComplete(println(_))
}
```

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/FlatMapPrefix.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/FlatMapPrefix.scala
@@ -43,8 +43,10 @@ import pekko.util.OptionVal
         .mandatoryAttribute[Attributes.NestedMaterializationCancellationPolicy]
         .propagateToNestedMaterialization
     val matPromise = Promise[M]()
-    val logic = new GraphStageLogic(shape) with InHandler with OutHandler {
-      val accumulated = collection.mutable.Buffer.empty[In]
+    object FlatMapPrefixLogic extends GraphStageLogic(shape) with InHandler with OutHandler {
+      private var left = n
+      private var builder = Vector.newBuilder[In]
+      builder.sizeHint(left)
 
       private var subSource = OptionVal.none[SubSourceOutlet[In]]
       private var subSink = OptionVal.none[SubSinkInlet[Out]]
@@ -65,11 +67,12 @@ import pekko.util.OptionVal
         subSource match {
           case OptionVal.Some(s) => s.push(grab(in))
           case _ =>
-            accumulated.append(grab(in))
-            if (accumulated.size == n) {
+            builder += grab(in)
+            left -= 1
+            if (left == 0) {
               materializeFlow()
             } else {
-              // gi'me some more!
+              // give me some more!
               pull(in)
             }
         }
@@ -98,12 +101,12 @@ import pekko.util.OptionVal
             // delegate to subSink
             s.pull()
           case _ =>
-            if (accumulated.size < n) pull(in)
-            else if (accumulated.size == n) {
+            if (left > 0) pull(in)
+            else if (left == 0) {
               // corner case for n = 0, can be handled in FlowOps
               materializeFlow()
             } else {
-              throw new IllegalStateException(s"Unexpected accumulated size: ${accumulated.size} (n: $n)")
+              throw new IllegalStateException(s"Unexpected accumulated size, left : $left (n: $n)")
             }
         }
       }
@@ -114,7 +117,7 @@ import pekko.util.OptionVal
           case _ =>
             if (propagateToNestedMaterialization) {
               downstreamCause = OptionVal.Some(cause)
-              if (accumulated.size == n) {
+              if (left == 0) {
                 // corner case for n = 0, can be handled in FlowOps
                 materializeFlow()
               } else if (!hasBeenPulled(in)) { // if in was already closed, nested flow would have already been materialized
@@ -128,8 +131,8 @@ import pekko.util.OptionVal
 
       def materializeFlow(): Unit =
         try {
-          val prefix = accumulated.toVector
-          accumulated.clear()
+          val prefix = builder.result()
+          builder = null // free for GC
           subSource = OptionVal.Some(new SubSourceOutlet[In]("FlatMapPrefix.subSource"))
           val theSubSource = subSource.get
           subSink = OptionVal.Some(new SubSinkInlet[Out]("FlatMapPrefix.subSink"))
@@ -196,6 +199,6 @@ import pekko.util.OptionVal
           case NonFatal(ex) => failStage(ex)
         }
     }
-    (logic, matPromise.future)
+    (FlatMapPrefixLogic, matPromise.future)
   }
 }


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/FlatMapPrefix.scala
```diff
@@ -128,8 +129,8 @@ import pekko.util.OptionVal
 
       def materializeFlow(): Unit =
         try {
-          val prefix = accumulated.toVector
-          accumulated.clear()
+          val prefix = builder.result()
+          builder = null // free for GC
```
He-Pin, line 133: @queimadus would you like to give this patch a try?

queimadus: I'll give it a try but I'm not sure this will work because `prefixAndTail` uses a similar implementation and has the same problem.

He-Pin: @queimadus There is a fix for prefixAndTail too https://github.com/apache/pekko/pull/1623

He-Pin: There do have leaks in the interpreter, eg the logic is still being refs.
I did a try in https://github.com/apache/pekko/pull/1621, which needs changing more code to clean up the logics.

queimadus: This change does seem to help when running the provided example. This crash by lack of memory. Previous run would crash out at around the 2 minute mark.
Screenshot 2024-12-22 at 11 34 34

He-Pin: image

queimadus: My original code before I simplified it with what looked to be the MRE, was more of a:

```scala
  def myLogic(prefix: Seq[ByteString]): Flow[ByteString, ByteString, NotUsed] =
    Flow[ByteString]

  val s = Source
    .repeat(())
    .map(_ => ByteString('a' * 400000))
    .take(1000000)
    .prefixAndTail(50000)
    .flatMapConcat { case (prefix, tail) =>
      Source(prefix).concatLazy(tail).via(myLogic(prefix))
    }

  Source.empty
    .concatAllLazy(List.tabulate(30000)(_ => s): _*)
    .runWith(Sink.ignore).onComplete(println(_))
```

Where I would need to use the prefix to generate the remaining flow logic and then emit the prefix back again. It still runs out of heap in this case, but it may be a different problem?

Nonetheless, I think this change makes sense.

He-Pin: `prefixAndTail` fix is in another PR, could you verify that too?
@queimadus https://github.com/apache/pekko/pull/1623 this one

queimadus: For reference, this would be the `flatMapPrefix` equivalent that wouldn't work with `Concat`s but work with `flatMapConcat` (the commented code):

```scala
  val s = Source
    .repeat(())
    .map(_ => ByteString('a' * 400000))
    .take(1000000)
    .flatMapPrefix(50000) { prefix =>
      Flow[ByteString].prepend(Source(prefix))
    }

  Source.empty
    .concatAllLazy(List.tabulate(30000)(_ => s): _*)
    .runWith(Sink.ignore).onComplete(println(_))

//  Source
//    .repeat(s)
//    .take(30000)
//    .flatMapConcat(x => x)
//    .runWith(Sink.ignore)
//    .onComplete(println(_))
```

He-Pin: I think the concat should be reimplemented to reduce memory usage somehow.

He-Pin:
@pjfanning Would you like to give this a review?

pjfanning:
> @pjfanning Would you like to give this a review?

Would it be possible to unit test this? We probably don't want anything that takes too much time or memory to run but it might be feasible to add some extra test coverage.

He-Pin:
@pjfanning I'm not sure how to test it. This will only be an issue when the graph is been attached to a long-lived large graph.

He-Pin:
I think I can't test it because the builder is protected inside the logic, otherwise, I have to extract the code to a dedicated class and expose it to tests.

Another way is using the MBean, but that is too much.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/FlatMapPrefix.scala
```diff
@@ -128,8 +129,8 @@ import pekko.util.OptionVal
 
       def materializeFlow(): Unit =
         try {
-          val prefix = accumulated.toVector
-          accumulated.clear()
```
He-Pin, line 132: which holds an `null` array.
```scala
  /** Reduce length to `n`, nulling out all dropped elements */
  private def reduceToSize(n: Int): Unit = {
    mutationCount += 1
    Arrays.fill(array, n, size0, null)
    size0 = n
  }
```
Now fixed, @raboof @pjfanning

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/FlatMapPrefix.scala
```diff
@@ -43,8 +43,10 @@ import pekko.util.OptionVal
         .mandatoryAttribute[Attributes.NestedMaterializationCancellationPolicy]
         .propagateToNestedMaterialization
     val matPromise = Promise[M]()
-    val logic = new GraphStageLogic(shape) with InHandler with OutHandler {
-      val accumulated = collection.mutable.Buffer.empty[In]
+    object logic extends GraphStageLogic(shape) with InHandler with OutHandler {
+      private var left = if (n < 0) 0 else n
```
raboof, line 47: we already have a `require(n >= 0)` above, I don't think we need to add code to deal with the `n < 0` case here.

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/FlatMapPrefix.scala
```diff
@@ -98,12 +101,10 @@ import pekko.util.OptionVal
             // delegate to subSink
             s.pull()
           case _ =>
-            if (accumulated.size < n) pull(in)
-            else if (accumulated.size == n) {
+            if (left > 0) pull(in)
+            else if (left == 0) {
               // corner case for n = 0, can be handled in FlowOps
               materializeFlow()
-            } else {
```
raboof, line 105: why not keep this? it should never happen, but that's what the error says

raboof (approved):

He-Pin:
@queimadus Thanks for the reporting, this do spend me a weekend to work out:)

PR #1623, opened 2024-12-22
fix: Fix a leak in PrefixAndTail operator.

He-Pin:
Motivation:
Fix @queimadus's report in https://github.com/apache/pekko/discussions/1566

Modification:
When the upstream finishes quickly but the downstream is not pulled, a leak can happen, so clean up that.

Result:
I think the leak should be fixed now.


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -237,7 +239,9 @@ import pekko.util.ccompat.JavaConverters._
     override def onUpstreamFinish(): Unit = {
       if (!prefixComplete) {
         // This handles the unpulled out case as well
-        emit(out, (builder.result(), Source.empty), () => completeStage())
+        val prefix = builder.result();
+        builder = null // free for GC
+        emit(out, (prefix, Source.empty), () => completeStage())
```
He-Pin, line 244: it will leak before.

pjfanning: looks good - if @queimadus can try this change too that would be very useful

queimadus: I actually can't seem to reproduce the issue with the [example I mentioned](https://github.com/apache/pekko/discussions/1566) with `flatMapPrefix` with `prefixAndTail`.

E.g. running the following runs fine with and without these changes
```scala
  val s = Source
    .repeat(())
    .map(_ => ByteString('a' * 4000000))
    .take(1000000)
    .prefixAndTail(50000)
    .flatMapConcat { case (prefix, tail) => tail }

  Source.empty
    .concatAllLazy(List.tabulate(30000)(_ => s): _*)
    .runWith(Sink.ignore).onComplete(println(_))
```

What doesn't run fine is the the [following snippet](https://github.com/apache/pekko/pull/1622#discussion_r1894903082), but it may be a different problem.

```scala
 def myLogic(prefix: Seq[ByteString]): Flow[ByteString, ByteString, NotUsed] =
    Flow[ByteString]

  val s = Source
    .repeat(())
    .map(_ => ByteString('a' * 400000))
    .take(1000000)
    .prefixAndTail(50000)
    .flatMapConcat { case (prefix, tail) =>
      Source(prefix).concatLazy(tail).via(myLogic(prefix))
    }

  Source.empty
    .concatAllLazy(List.tabulate(30000)(_ => s): _*)
    .runWith(Sink.ignore).onComplete(println(_))
```

That said, these changes still seem to make sense.

He-Pin: This problem is a kind of AsyncCallback leak, https://github.com/apache/pekko/pull/408
where many instances consume memories.

To total fix this, we will lost some support for debug I think, eg the `org.apache.pekko.stream.impl.fusing.GraphInterpreter#toSnapshot`

He-Pin: @queimadus Yes, different problem, the new problem is the interpreter keeps many things in memory

He-Pin:
@queimadus Hi, I realized that the origin issue of `concatAllLazy` can not be fixed with the current implementation of Interpreter.
So I have to add a new operator for this.

```scala
  def concatAllLazy[U >: Out](those: Graph[SourceShape[U], _]*): Repr[U] =
    internalConcatAll(those.toArray, detached = false)

vs:

  def concatAllDeferred[U >: Out](those: Graph[SourceShape[U], _]*): Repr[U] =
    concatGraph(
      Source.lazySource(
        () => Source(those).flatMapConcat(ConstantFun.scalaIdentityFunction)),
      detached = false)
```

@pjfanning wdyt

He-Pin:
@queimadus continue here https://github.com/apache/pekko/issues/1652

pjfanning (approved):
lgtm - but would prefer if this was not yet backfit to the 1.1.x branch

We've had problems with changes in the stream internals causing unforeseen issues.

He-Pin:
Okay, let me remove the backport label from it

PR #1628, opened 2024-12-28
perf: Reduce loops in when clean queue in BroadcastHub

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1627

Modification:
Reduce the loop count

Result:
Reduce the loop count

@andreaslochbihler-da cc as he is the origin reporter

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
@@ -605,15 +605,26 @@ private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: I
             else if (head != finalOffset) {
               // If our final consumer goes away, we roll forward the buffer so a subsequent consumer does not
               // see the already consumed elements. This feature is quite handy.
-              while (head != finalOffset) {
-                queue(head & Mask) = null
-                head += 1
-              }
+              cleanQueueInRange(head, finalOffset)
               head = finalOffset
               tryPull()
             }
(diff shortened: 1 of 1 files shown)


He-Pin:
it's better not to change the code in `unblockIfPossible` to 
```scala
        var upToOffset = head
        while (consumerWheel(upToOffset & WheelMask).isEmpty) {
          upToOffset += 1
          unblocked = true
        }
        if (upToOffset != head) {
          cleanQueueInRange(head, upToOffset)
          head = upToOffset
        }
```

pjfanning:
Can we keep this till v1.2.0? We've had a couple of breaks due to optimisations like this.

He-Pin:
I'm not sure if this should be marked as a bug or not.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
```diff
@@ -605,10 +605,7 @@ private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: I
             else if (head != finalOffset) {
               // If our final consumer goes away, we roll forward the buffer so a subsequent consumer does not
               // see the already consumed elements. This feature is quite handy.
-              while (head != finalOffset) {
-                queue(head & Mask) = null
-                head += 1
-              }
```
He-Pin, line 611: Which can loop many times

pjfanning (approved):
lgtm

He-Pin:
@pjfanning I want to backport to 1.1.x too, wdyt

PR #1631, opened 2024-12-28
add license headers to CI files

pjfanning:
different header based on whether I think file is of Pekko origin or if it is based on Akka files that we inherited at fork time


He-Pin (approved):
lgtm

PR #1635, opened 2024-12-28
fix: Fix ByteBuf leak in remoting transport.

He-Pin:
Motivation:

Modification:
Call `byteBuf.release` after bytes read.

Result:
No leaks.

Sorry for everyone.

It's using `retainedSlice` in `LengthFieldBasedFrameDecoder`
```
    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
        return buffer.retainedSlice(index, length);
    }
```

### DIFF
--- a/multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
+++ b/multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
@@ -56,8 +56,13 @@ private[pekko] class ProtobufEncoder extends MessageToMessageEncoder[Message] {
 private[pekko] class ProtobufDecoder(prototype: Message) extends MessageToMessageDecoder[ByteBuf] {
 
   override def decode(ctx: ChannelHandlerContext, msg: ByteBuf, out: java.util.List[AnyRef]): Unit = {
-    val bytes = ByteBufUtil.getBytes(msg)
-    out.add(prototype.getParserForType.parseFrom(bytes))
+    try {
+      val bytes: Array[Byte] = new Array[Byte](msg.readableBytes())
+      msg.readBytes(bytes)
+      out.add(prototype.getParserForType.parseFrom(bytes))
+    } catch {
+      case NonFatal(e) => ctx.pipeline().fireExceptionCaught(e)
+    }
   }
 }
 
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/TcpSupport.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/TcpSupport.scala
@@ -14,10 +14,8 @@
 package org.apache.pekko.remote.transport.netty
 
 import java.net.InetSocketAddress
-
 import scala.annotation.nowarn
 import scala.concurrent.{ Future, Promise }
-
 import org.apache.pekko
 import pekko.actor.Address
 import pekko.event.LoggingAdapter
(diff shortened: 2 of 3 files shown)


multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
```diff
@@ -56,8 +56,15 @@ private[pekko] class ProtobufEncoder extends MessageToMessageEncoder[Message] {
...
 
   override def decode(ctx: ChannelHandlerContext, msg: ByteBuf, out: java.util.List[AnyRef]): Unit = {
-    val bytes = ByteBufUtil.getBytes(msg)
-    out.add(prototype.getParserForType.parseFrom(bytes))
+    try {
+      val bytes = ByteBufUtil.getBytes(msg)
+      out.add(prototype.getParserForType.parseFrom(bytes))
+    } catch {
+      case NonFatal(e) =>
+        throw new RuntimeException("Error while decoding protobuf message", e)
```
pjfanning, line 64: why not rethrow the e? I dislike using the super generic RuntimeException here. If we need to wrap the exception, can't we find something more specific or create a new class like NettyDecodeException.

pjfanning: Can you suggest to Flink team to revert their upgrade to Pekko 1.1 if you think that this is definietly broken.

He-Pin: OK, I will change it then .

remote/src/main/scala/org/apache/pekko/remote/transport/netty/TcpSupport.scala
```diff
@@ -56,8 +58,16 @@ private[remote] trait TcpHandlers extends CommonHandlers {
...
-    if (bytes.length > 0) notifyListener(ctx.channel(), InboundPayload(ByteString(bytes)))
+    try {
+      val bytes: Array[Byte] = ByteBufUtil.getBytes(msg)
+      if (bytes.length > 0) notifyListener(ctx.channel(), InboundPayload(ByteString(bytes)))
+    } catch {
+      case NonFatal(e) =>
+        log.warning("Error while handling message from [{}]: {}", ctx.channel().remoteAddress(), e)
+        onException(ctx, e)
+    } finally {
+      msg.release()
```
He-Pin, line 69: Sorry for not checking the `LengthFieldBasedFrameDecoder` in detail, we need to turn on the leak detector in nightly testing I think.

multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
He-Pin: @pjfanning I have notified with an Email.

jxnu-liguobin (approved):

remote/src/main/scala/org/apache/pekko/remote/transport/netty/TcpSupport.scala
He-Pin: I can't call release, because it's a `retainedSlice` not a `readRetainedSlice`, will dig it deeper

pjfanning:
I would prefer if the exception handling changes were tracked separately - separate issue with a description of why the existing exception handling is wrong - and separate PR.

I think this PR should just add the finally block with the buffer release. That is all that highlighted as wrong in the issue.

He-Pin:
OK, Will split

PR #1636, opened 2024-12-28
fix: Fix ByteBuf leak in remoting transport.

He-Pin:
Motivation:

Modification:
Call byteBuf.release after bytes read.

Result:
No leaks.

Sorry for everyone.

It's using retainedSlice in LengthFieldBasedFrameDecoder

    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
        return buffer.retainedSlice(index, length);
    }


pjfanning:
changed back to main branch - the PR was messed up when it was set to merge to 1.1.x branch and would have brought lots of unrelated changes

pjfanning:
This should not be attempted while #1636. We need 1 PR to review, not 2.

PR #1637, opened 2024-12-28
chore: Turn on Netty's leak detector in build.

He-Pin:
Motivation:
refs:
Turn on  Netty's leak detector  to avoid for https://github.com/apache/pekko/issues/1634

Modification:
turn on the leak detector.

Result:
Will get logged in the building log.


pjfanning (approved):
lgtm

PR #1638, opened 2024-12-28
chore: Try to fix Netty leak with explicit read.

He-Pin:
Motivation:

Modification:
Do explicit reading instead.

Result:
Expect No leaks.

NOTE:
The code was right.
```scala
private[netty] abstract class NettyChannelHandlerAdapter extends SimpleChannelInboundHandler[ByteBuf]
```
will do auto-release.

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyHelpers.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyHelpers.scala
@@ -22,7 +22,6 @@ import org.apache.pekko
 import pekko.PekkoException
 import pekko.util.unused
 
-import io.netty.buffer.ByteBuf
 import io.netty.channel.{ ChannelHandlerContext, SimpleChannelInboundHandler }
 
 /**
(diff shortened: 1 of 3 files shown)


He-Pin:
Do local run with `cluster/MultiJvm/test` and `   -Dio.netty.leakDetection.level=PARANOID` to find if it leaked.

remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyHelpers.scala
```diff
@@ -36,7 +35,7 @@ private[netty] trait NettyHelpers {
 
   protected def onOpen(@unused ctx: ChannelHandlerContext): Unit = ()
 
-  protected def onMessage(@unused ctx: ChannelHandlerContext, @unused msg: ByteBuf): Unit = ()
+  protected def onMessage(@unused ctx: ChannelHandlerContext, @unused msg: Array[Byte]): Unit = ()
```
jxnu-liguobin, line 38: Is there anything special about ByteBuf ? (;I don't know about Netty

He-Pin: This PR is just a Try, and will not be needed. the old code is actually right.

He-Pin:
This should not be needed, the current code is right, thanks god :)

PR #1641, opened 2024-12-28
chore: Add netty leak detector in multi-node-testing.

He-Pin:
Motivation:
Add the leak detector to multi-node-testing too.

Modification:
Add the options to the MultiNode CommandLine Options

Result:
image


raboof (approved):
https://netty.io/wiki/reference-counted-objects.html#leak-detection-levels mentions enabling this during testing can be reasonable, so I don't have an objection

PR #1643, opened 2024-12-30
perf: Reduce loops in when clean queue in BroadcastHub (#1628)

He-Pin:
(cherry picked from commit 9596ea4c1827701ab7f540897bba9f6209ceef15)


pjfanning:
I don't think we should backport this. It is not semver friendly to be putting so many changes in patch releases.
We are taking too many risks in the patch releases as it is.

He-Pin:
Ok, let's only get it in the next release.

He-Pin:
@pjfanning Btw, this is a kind of bug too, because the loop can take seconds.

He-Pin:
Okay, let me close it, anyway ,ship it in next version will be safer.

PR #1642, opened 2024-12-30
Update scala-steward.yml to latest ubuntu image.

counter2015:
This PR makes scala-steward github action to run on ubuntu-latest and install sbt before running.

see:
https://github.com/actions/runner-images/issues/10636
https://github.com/scala-steward-org/scala-steward-action/pull/644

PR #1644, opened 2024-12-30
chore: Update scala-steward.yml (#1642)

He-Pin:
(cherry picked from commit 136319fbceb69a142a785251b2caf0bb80ce3d13)


pjfanning:
-1 from me. We don't need to backport every change. Ubuntu 22 will be supported for years.

PR #1649, opened 2024-12-31
fix: avoid call finalizeStage more times than once.

He-Pin:
Motivation:
fix https://github.com/apache/pekko/issues/1646

Modification:
Avoid call `finalizeStage(logic)` many times.

Result:
No NPE on 1.0.x

need forward port, there is no NPE on 1.1.x , but the `finalizeStage` still can be called many times.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
@@ -239,6 +239,9 @@ import pekko.stream.stage._
     logics(i).handlers.length
   }
 
+  // Marks whether a stage has been finalized (finalizeStage been called) or not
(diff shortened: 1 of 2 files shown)


He-Pin:
forward port to https://github.com/apache/pekko/pull/1650

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -239,6 +239,9 @@ import pekko.stream.stage._
     logics(i).handlers.length
   }
 
+  // Marks for if a stage has been finalized (finalizeStage been called) or not
```
He-Pin, line 242:
```suggestion
  // Marks for whether a stage has been finalized (finalizeStage been called) or not
```

pjfanning (changes requested):
I think this should be closed and #1654 backported after that is merged

PR #1650, opened 2024-12-31
fix: avoid call finalizeStage more times than once.

He-Pin:
Motivation:
fix https://github.com/apache/pekko/issues/1646

Modification:
Avoid calling `finalizeStage(logic)` many times.

Result:
No NPE on 1.1.x

forward port, there is no NPE on 1.1.x, but the `finalizeStage` can still be called many times.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
@@ -240,6 +240,9 @@ import pekko.stream.stage._
     logics(i).handlers.length
   }
 
+  // Marks whether a stage has been finalized (finalizeStage been called) or not
+  private[this] val finalizedMark = Array.fill(logics.length)(false)
+
   private[this] var _subFusingMaterializer: Materializer = _
   def subFusingMaterializer: Materializer = _subFusingMaterializer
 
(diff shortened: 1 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -240,6 +240,9 @@ import pekko.stream.stage._
     logics(i).handlers.length
   }
 
+  // Marks for if a stage has been finalized (finalizeStage been called) or not
```
pjfanning, line 243:
```suggestion
  // Marks whether a stage has been finalized (finalizeStage been called) or not
```

pjfanning (approved):
lgtm

samueleresca:
Overall, it looks good. Shouldn't the change also be applied to the `finish` finalizeStage call?

https://github.com/apache/pekko/blob/c8810cd73567a69edc35e605d9557545e789b63e/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala#L338

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -332,7 +335,10 @@ import pekko.stream.stage._
     var i = 0
     while (i < logics.length) {
       val logic = logics(i)
-      if (!isStageCompleted(logic)) finalizeStage(logic)
+      if (!isStageCompleted(logic) && !isStageFinalized(logic)) {
+        markStageFinalized(logic)
+        finalizeStage(logic)
+      }
```
He-Pin, line 341: updated , thanks @samueleresca

PR #1654, opened 2024-12-31
fix: avoid call finalizeStage more times than once. (#1650)

He-Pin:
(cherry picked from commit 7fef5b75f3a4e435f572d870758f364e1d65a84c)


pjfanning:
You keep raising PRs against the wrong branch.

image

I don't think this should be merged to 1.1 yet. I am not convinced that this issue is causing major issues and therefore would not justify being added in a patch.

He-Pin:
It will cause problems in 1.0.x, but still, it may call twice `GraphStageLogic#postStop` in 1.1.x.

pjfanning:
Can we hold off merging to patch branches for a few days so that the changes can be more widely reviewed?

He-Pin:
@pjfanning Okay, Let's do this next weekend.

pjfanning:
My preference is not to backfit changes that are not definitely causing major issues.
With the main branch, we could start doing milestone releases, eg 1.2.0-M1. We just need more people to get involved in voting on releases.

He-Pin:
let's keep this open to see if a user requests it.

PR #1657, opened 2024-12-31
chore: Implement zipWithIndex with statefulMap

He-Pin:
Motivation:
Performance, but it was reverted in https://github.com/apache/pekko/issues/1525

Modification:
Reimplemented it in the right way.

Result:
Better performance

Root cause:
The attributes should be set on the inner graph.

@remyhaemmerle-da sorry ~


stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowZipWithIndexSpec.scala
```diff
@@ -57,6 +57,7 @@ class FlowZipWithIndexSpec extends StreamSpec {
       // #zip-with-index
     }
 
+
```
He-Pin, line 60:
```suggestion
```

He-Pin:
use https://github.com/apache/pekko/pull/1669 for performance

Issue #1525, opened 2024-10-09
zipWithIndex is broken for org.apache.pekko.stream.scaladsl.GraphDSL.Implicits.PortOps


remyhaemmerle-da:
On the one hand,  [org.apache.pekko.stream.scaladsl.GraphDSL.Implicits.PortOps](https://github.com/apache/pekko/blob/95e8355c4c6f2041587e4d64590764e008d1f2b2/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Graph.scala#L1839) inherits from [org.apache.pekko.stream.scaladsl.Flow](https://github.com/apache/pekko/blob/95e8355c4c6f2041587e4d64590764e008d1f2b2/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala#L61). 

On the other hand, as of https://github.com/apache/pekko/pull/591 the implementation of [FlowOps#zipWithIndex](https://github.com/apache/pekko/blob/95e8355c4c6f2041587e4d64590764e008d1f2b2/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala#L3307) calls [FlowOps#withAttributes](https://github.com/apache/pekko/blob/95e8355c4c6f2041587e4d64590764e008d1f2b2/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala#L3310)
```
 def zipWithIndex: Repr[(Out, Long)] =
    statefulMap(() => 0L)((index, out) =>
        (index + 1L, (out, index)), _ => None)
      .withAttributes(DefaultAttributes.zipWithIndex)
```
However `PortOpsImpl` the implementation of `PortOps` does not support [`withAttributes`](https://github.com/apache/pekko/blob/95e8355c4c6f2041587e4d64590764e008d1f2b2/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Graph.scala#L1848)
```
  override def withAttributes(attr: Attributes): Repr[Out] = throw settingAttrNotSupported
```
 
Here is a small program that works with pekko 1.0.2 but crashes with 1.1.1

```
// adapted from https://doc.akka.io/docs/akka/current/stream/stream-graphs.html
import org.apache.pekko
import pekko.actor.ActorSystem
import pekko.stream._
import pekko.stream.scaladsl._

object Example extends App {
  import GraphDSL.Implicits._

  implicit val system = ActorSystem("GraphAndShapeExampleSystem")
  implicit val materializer = ActorMaterializer()

[...]

PR #1658, opened 2024-12-31
chore: Tweak withAttributes in Flow

He-Pin:
Motivation:

To avoid https://github.com/apache/pekko/issues/1525 , I checked all the `Flow#withAttributes` usage.

Modification:
Move the `.withAttributes()` inside the `via(...)`

Result:
1. consistent style
2. safer when using it with graphdsl.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
@@ -24,10 +24,7 @@ import scala.util.control.NoStackTrace
 
 import org.apache.pekko
 import pekko.Done
-import pekko.stream.AbruptStageTerminationException
-import pekko.stream.ActorAttributes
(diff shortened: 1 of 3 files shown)


He-Pin:
@pjfanning I think we should have this in 1.2.0-M1 , it's some kind of bug fix.

pjfanning:
Is there any way to add test coverage? Something like https://github.com/apache/pekko/blob/main/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowZipWithIndexSpec.scala#L60-L92 - which was added for #1525

He-Pin:
Feel free to add it to m2, but actually the current fix not harm even without test.

It will cause problem when using in graphdsl as the zipwithIndex was implemented.

pjfanning:
I am uncomfortable with adding this in M1. There will be an M2 in a month or 2 and hopefully this can be added with some tests.

He-Pin:
@pjfanning I have updated the code with test

pjfanning (approved):
lgtm

PR #1594, opened 2024-12-16
reimplement fix for akka/pekko cluster

pjfanning:
Run I the tests locally they pass but the nightly tests were failing until I removed this change.

see #1590

```
sbt \
            -Dpekko.cluster.assert=on \
            -Dpekko.log.timestamps=true \
            -Dpekko.test.timefactor=2 \
            -Dpekko.actor.testkit.typed.timefactor=2 \
            -Dpekko.test.tags.exclude=gh-exclude,timing \
            -Dpekko.test.multi-in-test=false \
"cluster/testOnly *JoinConfigCompatCheckerRollingUpdateSpec"
```

PR #1664, opened 2025-01-02
reimplement fix for akka/pekko cluster (#1594)

pjfanning:
cherry pick ef212e9a98614ba13cd65cde3cd8140702f1f35b #1594

### DIFF
--- a/.github/workflows/build-test-prValidation.yml
+++ b/.github/workflows/build-test-prValidation.yml
@@ -101,6 +101,53 @@ jobs:
           -Dpekko.log.timestamps=true \
           validatePullRequest
 
+  pekko-classic-remoting-tests:
+    name: Pekko Classic Remoting Tests
+    runs-on: ubuntu-22.04
+    if: github.repository == 'apache/pekko'
+    strategy:
+      fail-fast: false
+      matrix:
+        command:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test-prValidation.yml
```diff
@@ -101,6 +101,53 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    if: github.repository == 'apache/pekko'
+    strategy:
+      fail-fast: false
+      matrix:
+        command:
+          - cluster/test distributed-data/test cluster-tools/test cluster-metrics/test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
He-Pin, line 115: needs branch

He-Pin (approved):
lgtm

.github/workflows/build-test-prValidation.yml
pjfanning: when the PR targets a branch, GitHub CI seems intelligent enough to checkout the PR branch

Have a look at https://github.com/apache/pekko/blob/1.0.x/.github/workflows/build-test-prValidation.yml -- no refs for any checkouts.

The nightly runs are different because we have the ymls in the main branch so if we want to test a branch, we need to adjust the checkouts in that case.

PR #1666, opened 2025-01-02
chore: Add FutureOps which with an await style.

He-Pin:
Motivation:
I found the `Await.result` is everywhere.

Modification:
Add an `await` operator to the future.

Result:
Fluent code.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/Helpers.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/Helpers.scala
@@ -31,9 +31,13 @@ import java.time.format.DateTimeFormatter
 import java.util.{ Comparator, Locale }
 import java.util.concurrent.TimeUnit
 import java.util.regex.Pattern
+
 import scala.annotation.tailrec
 import scala.concurrent.duration.{ Duration, FiniteDuration }
+
 import com.typesafe.config.{ Config, ConfigRenderOptions }
+
+import scala.concurrent.{ blocking, Future }
 
 object Helpers {
 
@@ -195,4 +199,19 @@ object Helpers {
       Duration(config.getDuration(path, unit), unit)
   }
 
+  /**
+   * INTERNAL API
+   */
+  private[pekko] final implicit class FutureOps[T](val future: Future[T]) extends AnyVal {
+
+    /**
+     * Wait for the future to complete and return the result, or throw an exception if the future failed.
+     * Optimize for the case when the future is already completed.
+     * @since 1.2.0
+     */
+    def await(atMost: Duration = Duration.Inf): T = future.value match {
+      case Some(value) => value.get
+      case None        => blocking(future.result(atMost)(null))
+    }
+  }
 }
--- a/actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/LightArrayRevolverScheduler.scala
@@ -19,7 +19,7 @@ import java.util.concurrent.atomic.{ AtomicLong, AtomicReference }
 
 import scala.annotation.{ nowarn, tailrec }
 import scala.collection.immutable
-import scala.concurrent.{ Await, ExecutionContext, Future, Promise }
+import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.concurrent.duration._
 import scala.util.control.NonFatal
 
@@ -178,7 +178,8 @@ class LightArrayRevolverScheduler(config: Config, log: LoggingAdapter, threadFac
       }
     }
 
-    Await.result(stop(), getShutdownTimeout).foreach {
+    import pekko.util.Helpers._
+    stop().await(getShutdownTimeout).foreach {
       case task: Scheduler.TaskRunOnClose =>
         runTask(task)
       case holder: TaskHolder => // don't run
--- a/actor/src/main/scala/org/apache/pekko/pattern/CircuitBreaker.scala
+++ b/actor/src/main/scala/org/apache/pekko/pattern/CircuitBreaker.scala
@@ -19,7 +19,7 @@ import java.util.concurrent.atomic.{ AtomicBoolean, AtomicInteger, AtomicLong }
 import java.util.function.BiFunction
 import java.util.function.Consumer
 import scala.annotation.nowarn
-import scala.concurrent.{ Await, ExecutionContext, Future, Promise }
+import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
 import scala.util.{ Failure, Success, Try }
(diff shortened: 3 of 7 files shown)


pjfanning:
This code is internal. Does it matter if it is fluent?

He-Pin:
So does performance, I think

pjfanning:
If there is a perf issue in Await.result, I would prefer it was fixed in Scala runtime.

The Scala 2 impl seems to try to optimise for the completed case.
```
    final def result[T](awaitable: Awaitable[T], atMost: Duration): T = awaitable match {
      case f: Future[T] if f.isCompleted => f.result(atMost)(AwaitPermission)
      case _ => blocking(awaitable.result(atMost)(AwaitPermission))
    }
```

He-Pin:
The code seems need more adjustment,  as we don't need to test if it's a future anymore

He-Pin:
??? Why its closed?

actor/src/main/scala/org/apache/pekko/util/Helpers.scala
```diff
@@ -195,4 +199,19 @@ object Helpers {
...
+  private[pekko] final implicit class FutureOps[T](val future: Future[T]) extends AnyVal {
+
+    /**
+     * Wait for the future to complete and return the result, or throw an exception if the future failed.
+     * Optimize for the case when the future is already completed.
+     * @since 1.2.0
+     */
+    def await(atMost: Duration = Duration.Inf): T = future.value match {
+      case Some(value) => value.get
+      case None        => blocking(future.result(atMost)(null))
```
He-Pin, line 214: avoid the double checking now

He-Pin:
@pjfanning I have attached the result

actor/src/main/scala/org/apache/pekko/util/Helpers.scala
He-Pin:
```scala
    @throws(classOf[TimeoutException])
    @throws(classOf[InterruptedException])
    final def resultOption[T](awaitable: Awaitable[T], atMost: Duration): T = awaitable match {
      case CompletedFuture(v)  => v.get
      case _ => blocking(awaitable.result(atMost)(AwaitPermission))
    }

    private final object CompletedFuture {
      def unapply[T](f: Future[T]): Option[Try[T]] = f.value
    }
```

pjfanning:
> @pjfanning I have attached the result

I would still like this discussed with the Scala team to understand if the Await.result can be improved. Everyone benefits if you have discovered a perf issue there and that it then gets fixed.

He-Pin:
@pjfanning https://github.com/scala/scala/pull/10972 , submitted to scala

raboof:
> [scala/scala#10972](https://github.com/scala/scala/pull/10972) , submitted to scala

Awesome that you have proposed this improvement upstream, thanks!

I see two motivations for this change: the performance improvement, and the fact that the code 'looks nicer'. I agree that if this performance improvement could make it upstream that would be much better.

For the idea that this change makes the code looks nicer, I'm not sure I agree: it replaces well-known Scala idiom with a custom helper. While the custom helper is perhaps nicer, it's also 'one more thing to learn'. Also, because asynchronicity is at the heart of Pekko, we should use `Await` sparingly and with caution. For this reason, having to use the arguably-'clumsy' Scala idiom might not be so bad, as it adds some friction to consider whether `Await` is really needed.

He-Pin:
1. this is just following the style in gears.
```scala
  extension [T](src: Source[scala.util.Try[T]])
    /** Waits for an item to arrive from the source, then automatically unwraps it. Suspends until an item returns.
      * @see
      *   [[Source!.awaitResult awaitResult]] for non-unwrapping await.
      */
    inline def await(using Async) = src.awaitResult.get
```
and then with usage:
```scala
  test("Constant returns") {
    Async.blocking:
      for (i <- -5 to 5)
        val f1 = Future { i }
        val f2 = Future.now(Success(i))
        assertEquals(f1.await, i)
        assertEquals(f1.await, f2.await)
  }
```

2. The current implementation by @viktorklang will double-getting on `java.util.concurrent.atomic.AtomicReference#get`, have no idea, why would we do that.
3. This optimization fix point 2.
4. I try to avoid the lifting to `Option` with 
```scala
  def unwrap: Try[T]
```
in Future.scala, and then 
```scala
    @throws(classOf[TimeoutException])
    @throws(classOf[InterruptedException])
    final def result[T](awaitable: Awaitable[T], atMost: Duration): T = awaitable match {
      case f: Future[T] @unchecked =>
        val r = f.unwrap
        if (r ne null) r.get
        else blocking(awaitable.result(atMost)(AwaitPermission))
      case _ => blocking(awaitable.result(atMost)(AwaitPermission))
    }
```

But the result doesn't have much difference in gc pressure.

He-Pin:
One advantage of having this inside is it works on all scala versions.

pjfanning:
@He-Pin could you add the CompleteFutureBenchmark to this PR? I think we have some benchmarks in this repo where this could be added.

He-Pin:
I will add that tomorrow,

bench-jmh/src/main/scala/org/apache/pekko/util/FutureOpsBenchmark.scala
```diff
@@ -0,0 +1,33 @@
...
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 1000)
+@Measurement(iterations = 10000)
+class FutureOpsBenchmark {
+  private val completedFuture: Future[Int] = Future.successful(1)
+  // jmh:run -i 11 -wi 11 -f1 -t1 org.apache.pekko.util.FutureOpsBenchmark
+  //    [info] Benchmark                               Mode  Cnt       Score      Error   Units
+  //    [info] FutureOpsBenchmark.awaitWithAwaitable  thrpt   11  706198.499 ± 8185.983  ops/ms
+  //    [info] FutureOpsBenchmark.awaitWithFutureOps  thrpt   11  766901.781 ± 9741.792  ops/ms
```
He-Pin, line 18: @pjfanning Attached.

pjfanning:
I would like us to take a little bit of time over this just to see what the Scala team and other users have to say.
I am not against merging a change to Pekko but I think that change must not negatively impact the awaits for incomplete futures. The existing benchmark only checks complete futures. I know it is harder to measure the impact of a change on incomplete futures because benchmarks are mainly measuring the wait and not the code around getting the result from the awaitable after it completes.

He-Pin:
@pjfanning I think that will not make much difference, the old implementation will always do the `AtomicReference.get` so does this, one. and when the future is not completed, this will return a `None` (no allocation), which was returning a boolean.

He-Pin:
And I think there is a limitation of the current Scala implementation, were we can't save the `value` into a slot to avoid double `get`, the value is on the stack, which is very cheap,but we can't do it with current scala.

https://contributors.scala-lang.org/t/inline-convertion-in-pattern-matchings-if-guard/6962/12

He-Pin:
merged upstream

PR #1667, opened 2025-01-03
perf: optmize NettyChannelHandlerAdapter with explict extends.

He-Pin:
Motivation:
Performance, it was extended to the `SimpleChannelInboundHandler` which will check if the type is a `ByteBuf`, now, we just test it explicitly.

Modification:
Make it extends the `ChannelInboundHandlerAdapter` and `release` explicitly , which was released by the `SimpleChannelInboundHandler`

Result:
Simple code and performance a little better.

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyHelpers.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyHelpers.scala
@@ -23,7 +23,7 @@ import pekko.PekkoException
 import pekko.util.unused
 
 import io.netty.buffer.ByteBuf
-import io.netty.channel.{ ChannelHandlerContext, SimpleChannelInboundHandler }
+import io.netty.channel.{ ChannelHandlerContext, ChannelInboundHandlerAdapter }
 
 /**
  * INTERNAL API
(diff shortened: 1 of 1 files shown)


remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyHelpers.scala
```diff
@@ -53,10 +53,19 @@ private[netty] trait NettyHelpers {
...
-    onMessage(ctx, msg)
+
+  final override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef): Unit = {
+    msg match {
+      case buf: ByteBuf =>
+        try {
+          onMessage(ctx, buf)
+        } catch {
+          case ex: Throwable => transformException(ctx, ex)
+        } finally buf.release() // ByteBuf must be released explicitly
```
He-Pin, line 66: it was released by the `finally` block in `SimpleChannelInboundHandler`

raboof (approved):
the upstream code is at https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/SimpleChannelInboundHandler.java

unrolling the generic code that checks for types dynamically makes some sense

He-Pin:
I ran the cluster test locally and it passed too, we are using this kind of trick at work.

He-Pin:
@raboof @pjfanning Do you think we should backport this? as Flink will use this pekko too.

pjfanning:
I don't want to pre-empt anything. If Flink team want to test the 1.2.0 snapshot and find that works better for them - we could then consider backporting this.

PR #1671, opened 2025-01-04
chore: Add identity function.

He-Pin:
Motivation:
Make life easier for Java developers without the `i -> i`

Modification:
Add an `identity` helper function.

Result:
Clean code.

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/util/ConstantFunJavaTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/util/ConstantFunJavaTest.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


mdedetrich (changes requested):
This is pointless? The Java identity function already exists in JDK 1.8 as `java.util.function.Function.identity`.

This might be useful for versions of JDK before 1.8 but Pekko doesn't support those JDK's

He-Pin:
@mdedetrich Yes and no, see below:
image

pjfanning:
Can you translate the IDE issue into English? I prefer not to have images - if you paste the text, I can get a tool to translate it. If you paste images, I need an OCR tool and I don't have one available.

He-Pin:
Sorry, the `java.util.function.Function#identity` returns a `java.util.function.Function`, even these methods share the same name, but returns two different types.

pjfanning:
I'm ok with adding this if it helps. I would like to still have some of tests use `ConstantFun.javaIdentityFunction()` for regression purposed though.

actor-tests/src/test/java/org/apache/pekko/util/ConstantFunJavaTest.java
```diff
@@ -0,0 +1,46 @@
...
+  @Test
+  public void alwaysReturnFunctionInstance() {
+    final Function<Object, Object> funObj = ConstantFun.javaIdentityFunction();
+    final Function<String, String> funStr = ConstantFun.javaIdentityFunction();
+    try {
+      Assert.assertTrue(funObj.equals(funStr));
+    } catch (Exception e) {
+      Assert.fail("Exception thrown: " + e.getMessage());
+    }
+  }
```
He-Pin, line 45: added a dedicated test @pjfanning

He-Pin:
@mdedetrich would you like talking another look, thanks

raboof (approved):
unfortunate to have all those function types, but given those this seems like a nice utility

He-Pin:
@mdedetrich Would you like to unblock this, thanks.

mdedetrich (approved):
lgtm

PR #1686, opened 2025-01-05
chore: Use more lambda instread in stream module.

He-Pin:
Motivation:
More lambda

Modification:
use lambda

Result:
Clean code


pjfanning (approved):
lgtm

PR #1687, opened 2025-01-05
chore: Avoid forwarding method on ArrayDequeue in stream module.

He-Pin:
Motivation:
Call methods directly on ArrayDequeue to avoid forwarding.

Modification:
Call methods directly on ArrayDequeue to avoid forwarding.

Result:
performant better


pjfanning (approved):
lgtm

PR #1688, opened 2025-01-05
chore: Avoid forwarding method on ArrayDequeue in BatchingExecutor.

He-Pin:
Motivation:
Call methods directly on ArrayDequeue to avoid forwarding.

Modification:
Call methods directly on ArrayDequeue to avoid forwarding.

Result:
performant better

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/BatchingExecutor.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/BatchingExecutor.scala
@@ -16,7 +16,6 @@ package org.apache.pekko.dispatch
 import org.apache.pekko
 import pekko.annotation.InternalApi
 
-import java.util.ArrayDeque
 import java.util.concurrent.Executor
 import scala.annotation.tailrec
 import scala.concurrent._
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/BatchingExecutor.scala
```diff
@@ -67,10 +67,10 @@ private[pekko] trait BatchingExecutor extends Executor {
   // invariant: if "_tasksLocal.get ne null" then we are inside Batch.run; if it is null, we are outside
   private[this] val _tasksLocal = new ThreadLocal[AbstractBatch]()
 
-  private[this] abstract class AbstractBatch extends ArrayDeque[Runnable](4) with Runnable {
+  private[this] abstract class AbstractBatch extends util.ArrayDeque[Runnable](4) with Runnable {
     @tailrec final def processBatch(batch: AbstractBatch): Unit =
       if ((batch eq this) && !batch.isEmpty) {
-        batch.poll().run()
+        batch.pollFirst().run()
```
pjfanning, line 73: isn't this the only line that needs to change?

He-Pin: yes it is

He-Pin: The others are lint problem

He-Pin: Do you suggest me remove this?

He-Pin: I think current chagne should be fine to include the lint fix .

actor/src/main/scala/org/apache/pekko/dispatch/BatchingExecutor.scala
```diff
@@ -67,10 +67,10 @@ private[pekko] trait BatchingExecutor extends Executor {
   // invariant: if "_tasksLocal.get ne null" then we are inside Batch.run; if it is null, we are outside
   private[this] val _tasksLocal = new ThreadLocal[AbstractBatch]()
 
-  private[this] abstract class AbstractBatch extends ArrayDeque[Runnable](4) with Runnable {
+  private[this] abstract class AbstractBatch extends util.ArrayDeque[Runnable](4) with Runnable {
```
raboof, line 70: (the reasoning here is to disambiguate between `scala.collection.mutable.ArrayDequeu` and `java.util.ArrayDeque` - that's fair I suppose. It might be even more readable to have the full `java.util.ArrayDeque` here instead of importing `java.util`)

raboof (approved):

PR #1693, opened 2025-01-06
chore: ignore even more configs for dependency submission

raboof:
confirmed ignoring `runtime-internal` suppresses the really old guava, the `scala-doc-tool` and `scala-tool` configs currently don't pull in any problematic dependencies but they should not be relevant to our users either.

Follow-up on https://github.com/apache/pekko/pull/1689 and https://github.com/apache/pekko/pull/1392


He-Pin (approved):
Thanks

PR #1698, opened 2025-01-10
perf: optmize NettyChannelHandlerAdapter with explict extends. (#1667)

He-Pin:
(cherry picked from commit 03712a927e67723e9ffeadaa0b774dae45cbea88)

For Better performance.

refs: https://github.com/apache/flink/pull/25866


pjfanning:
I don't think we should backport small perf changes.

He-Pin:
We will not have 1.2.x any soon I think, it doesn't harm.

pjfanning:
We have already discussed doing a 1.2.0-M1 in the next few weeks.

He-Pin:
OK, that seems fine now.

PR #1699, opened 2025-01-10
chore: Require breadth >= 1 in FlattenMerge

He-Pin:
Motivation:
Check `breadth` >= 1 in FlattenMerge

Modification:
Add a requirement for that.


raboof (approved):

He-Pin:
@pjfanning This is extract of a pending PR,  I can't keep 100% code coverage at $Work too.

raboof:
I think checks like this are helpful because they make it easier to read the implementation of `FlattenMerge` and be confident you don't need to consider the case where breadth < 1

> * is it really possible to end up in a position where the requirement fails?

That's indeed not super obvious, but I suspect it can when the user chooses a <1 parallelism somewhere.

>     * if so, can we add a test?

I guess so, though I'm not sure it's worth it to test error checking so explicitly

>     * also, what would users of older versions see in this case? an exception, a nonsensical result?

I suspect they'd see a flow which stalls (fails to pull) after accepting its first element.

PR #1700, opened 2025-01-10
chore: Require parallelism >= 1 in MapAsync and MapAsyncUnordered

He-Pin:
Motivation:
require parallelism >= 1 in MapAsync and MapAsyncUnordered

Modification:
Add requirements


pjfanning:
* is it really possible to end up in a position where the requirement fails?
* if so, can we add a test?
* also, what would users of older versions see in this case? an exception, a nonsensical result?

He-Pin:
@pjfanning The code is the same as it is in `MapAsyncPartitioned`

pjfanning:
> @pjfanning The code is the same as it is in `MapAsyncPartitioned`

this is still a change that reduces performance by adding a check that needs evaluation

pjfanning (approved):
lgtm - I guess it is better to have to check even if it has a small perf impact

He-Pin:
@pjfanning Thanks.

PR #1701, opened 2025-01-10
feat: Add TraversalBuilder.getValuePresentedSource method for further optimization

He-Pin:
Motivation:
Add helper method to extract the value presented graph which will enable future optimization.
which is needed in https://github.com/apache/pekko/issues/1672

Modification:
Add TraversalBuilder.getValuePresentedSource method

Result:
TraversalBuilder.getValuePresentedSource internal method added

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/impl/TraversalBuilderSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/impl/TraversalBuilderSpec.scala
@@ -17,8 +17,13 @@ import org.apache.pekko
 import pekko.NotUsed
 import pekko.stream._
 import pekko.stream.impl.TraversalTestUtils._
+import pekko.stream.impl.fusing.IterableSource
+import pekko.stream.impl.fusing.GraphStages.{ FutureSource, SingleSource }
 import pekko.stream.scaladsl.{ Keep, Source }
+import pekko.util.OptionVal
 import pekko.testkit.PekkoSpec
+
+import scala.concurrent.Future
 
 class TraversalBuilderSpec extends PekkoSpec {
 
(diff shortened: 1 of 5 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/FailedSource.scala
```diff
@@ -22,7 +22,7 @@ import pekko.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
 /**
  * INTERNAL API
  */
-@InternalApi private[pekko] final class FailedSource[T](failure: Throwable) extends GraphStage[SourceShape[T]] {
+@InternalApi private[pekko] final class FailedSource[T](val failure: Throwable) extends GraphStage[SourceShape[T]] {
```
He-Pin, line 25: needed for optimization.

stream/src/main/scala/org/apache/pekko/stream/impl/TraversalBuilder.scala
```diff
@@ -380,12 +380,53 @@ import pekko.util.unused
...
+    }
+  }
+
   /**
    * Test if a Graph is an empty Source.
    */
   def isEmptySource(graph: Graph[SourceShape[_], _]): Boolean = graph match {
     case source: scaladsl.Source[_, _] if source eq scaladsl.Source.empty => true
     case source: javadsl.Source[_, _] if source eq javadsl.Source.empty() => true
+    case EmptySource                                                      => true
```
pjfanning, line 429: this doesn't look related to this change - is this a bug that needs to be fixed in its own PR?

stream-tests/src/test/scala/org/apache/pekko/stream/impl/TraversalBuilderSpec.scala
```diff
@@ -447,4 +452,93 @@ class TraversalBuilderSpec extends PekkoSpec {
...
+    val singleSourceA = new SingleSource("a")
+    TraversalBuilder.getValuePresentedSource(singleSourceA) should be(OptionVal.Some(singleSourceA))
+
+    TraversalBuilder.getValuePresentedSource(Source.single("c").async) should be(OptionVal.None)
+    TraversalBuilder.getValuePresentedSource(Source.single("d").mapMaterializedValue(_ => "Mat")) should be(
+      OptionVal.None)
+  }
+
+  "find Source.empty via TraversalBuilder with getValuePresentedSource" in {
+    val emptySource = EmptySource
```
He-Pin, line 478: @pjfanning the tests coverage is here for the `EmptySource`

stream/src/main/scala/org/apache/pekko/stream/impl/TraversalBuilder.scala
He-Pin: It's not a bug, just a missing optimization case.

pjfanning: I thought that the cases were evaluated in order so this would not change performance in that case and it would actually lead to a different result if EmptySource was matched but that neither of the first 2 cases matched.

He-Pin: these methods are internal methods acutually.

He-Pin:
After this pr get merged, we can do some optimization for `flatmapMerge` too.

stream/src/main/scala/org/apache/pekko/stream/impl/TraversalBuilder.scala
pjfanning: I still don't understand why this can't be in its own PR. If this is an issue, it looks like it should be backported to 1.1.x.

He-Pin: You said no performance optimization should be backported.

He-Pin: I can extract it in a separated pr, I will back my hometown tomorrow

pjfanning: this is not a perf change - it is a behaviour change

Before this change, `case EmptySource` would not be there and then EmptySource would lead to a return value of false.

He-Pin: But this method is only been used for avoid optimization

pjfanning: I am lost. The only way this is an optimisation is if EmptySource appears before one or more the existing cases.

He-Pin: Should  I  extract this, I just turn on my computer, Tomorrow this time I will at the airport:)

pjfanning:
I think this looks ok but I would prefer to see some more reviews but it gets merged.

He-Pin:
Yeah, really hope we get more reviewer:)

pjfanning:
There will an M2 in a month or 2. I'm not going to delay M1 for a perf tweak.

He-Pin:
@pjfanning @mdedetrich would you like take this a look

pjfanning (approved):
lgtm

PR #1707, opened 2025-01-13
chore: Add support for controlling the NettyTransport's byteBuf allocator type.

He-Pin:
Motivation:
As a response to https://github.com/apache/flink/pull/25866 
@davidradl  want to control the bytebuf allocator type.

Modification:
Add `bytebuf-allocator-type` selector , with `pooled` by default.

Result:
The bytebuf allocator can be configured now.

PR #1709, opened 2025-01-14
chore: Add support for controlling the NettyTransport's byteBuf alloc…

He-Pin:
…ator type. (#1707)

* chore: Add support for controlling the NettyTransport's byteBuf allocator type.

* chore: extract deriveByteBufAllocator method

(cherry picked from commit dbc9ed3a99b0d97a697b705ceb927c48d72eea90)

related thread: https://lists.apache.org/thread/6lj5lbg18hz4n9zckn0b89p9mv9nz14g

### DIFF
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -621,6 +621,14 @@ pekko {
         # "off-for-windows" of course means that it's "on" for all other platforms
         tcp-reuse-addr = off-for-windows
 
+        # Used to control the Netty 4's ByteBufAllocator. The default is "pooled".
+        # pooled          : use a PooledByteBufAllocator.DEFAULT
+        # unpooled        : use an UnpooledByteBufAllocator.DEFAULT
+        # unpooled-heap   : use an UnpooledByteBufAllocator with prefer direct `false`
+        # adaptive        : use an AdaptiveByteBufAllocator
+        # adaptive-heap   : use an AdaptiveByteBufAllocator with prefer direct `false`
+        bytebuf-allocator-type = "pooled"
+
         # Used to configure the number of I/O worker threads on server sockets
         server-socket-worker-pool {
           # Min number of threads to cap factor-based number to
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettyTransport.scala
@@ -24,6 +24,7 @@ import scala.util.Try
 import scala.util.control.{ NoStackTrace, NonFatal }
 
 import com.typesafe.config.Config
+
 import org.apache.pekko
 import pekko.ConfigurationException
 import pekko.OnlyCauseStackTrace
@@ -38,7 +39,13 @@ import pekko.util.Helpers.Requiring
 import pekko.util.{ Helpers, OptionVal }
 
 import io.netty.bootstrap.{ Bootstrap => ClientBootstrap, ServerBootstrap }
-import io.netty.buffer.Unpooled
+import io.netty.buffer.{
+  AdaptiveByteBufAllocator,
+  ByteBufAllocator,
+  PooledByteBufAllocator,
+  Unpooled,
+  UnpooledByteBufAllocator
+}
 import io.netty.channel.{
   Channel,
   ChannelFuture,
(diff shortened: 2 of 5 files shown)


pjfanning:
This is grey area for me. Whether a change like this be in a patch. If it is really useful then I'd support merging it. That said, I'd prefer if the Flink team test with a 1.2 snapshot and then if they confirm that they get the behaviour that they want then we can merge this.

He-Pin:
image

He-Pin:
@pjfanning I still think we should merge this and make sure we get it in the next 1.1.4 to help the Flink community.

He-Pin:
refs: https://github.com/apache/flink/pull/26042 too

pjfanning:
I am -1 on this. Let the Flink team approach us and convince us. I have seen noone say that this is a must for Flink.

afedulov:
@pjfanning A brief explanation of why this is generally needed: Flink uses Netty not only for RPC but also for data shuffling. Migrating to Netty 4 for RPC caused some tests to become flaky, primarily due to differences in memory allocation in the newer version. The safest option for us would be to set Pekko allocation to `unpooled`, as this would roughly replicate the memory requirements of the previous version and thereby reduce the probability of tipping containers at the memory limit into OOM in production. The problem is that we cannot apply this as a global setting, because other parts of the framework, such as the core data exchange network stack, also use Netty 4 and would consequently be affected by the change.

pjfanning:
@afedulov have you tested with the pekko 1.2 snapshot jars that already support this?

ferenc-csaky:
@pjfanning I just created a local Pekko build (more details in https://github.com/apache/flink/pull/25955) from the 1.1.x branch with this PR applied on top and executed our failing test with setting the `bytebuf-allocator-type` to `unpooled`. With that setting, at least on my machine the failing test was successful from 5/5 runs.

So we believe this config option would be useful for us if you guys can get it into 1.1.4. From our side it is not something that is urgent, but it could be useful for use if in case there will be a 1.1.4 release, this change would be part of it.

pjfanning:
This PR is not merged. You can only get this support by using the latest 1.2 snapshots.

He-Pin:
> This PR is not merged. You can only get this support by using the latest 1.2 snapshots.
> 
> See https://github.com/apache/pekko/pull/1709#issuecomment-2599698240

And I built 1.1.x after I applied your commit from

Seems he did a local patch

ferenc-csaky:
Yes, I built Pekko myself locally after I applied this patch to 1.1.x locally, and built Flink against that version with the added setting to run the test.

pjfanning:
@raboof @nvollmar What do you think? This looks not too bad for a patch - not ideal to have a non-bug fix in a patch but the change is small enough and doesn't have an API impact. The default behaviour remains unchanged and you have to opt in via a config change to see any behaviour change.

mdedetrich:
PR looks fine to me but I would prefer to have it released in a non patch update

He-Pin:
@mdedetrich it's already in the main, here is to decide if we include it in 1.1.4

mdedetrich:
Then let's do a 1.20 release, we can backport strictly fixes from main to 1.1.x branch if needed.

pjfanning:
> I'm not sure if I can have all my features ready in 1.2.0, Chinese Spring Festival is coming. I still have some features in mind but not yet there.

I'm thinking 1.2.0-M1, 1.2.0-M2, maybe some more milestonees and later this year, a full 1.2.0 release.

Flink team should be fine using 1.2.0-M1 in production if we don't add any extra changes to what is already on main branch.

He-Pin:
I think pekko's M1 release should be very stable too, @afedulov @ferenc-csaky wdyt about it.

ferenc-csaky:
Does the `1.2` branch includes a Netty version update compared to `1.1.x`? Cause if yes, it might be not as straightforward for Flink to jump to it.

pjfanning:
> Does the `1.2` branch includes a Netty version update compared to `1.1.x`? Cause if yes, it might be not as straightforward for Flink to jump to it.

netty is a provided dependency of pekko-remote - so choose your own patch version of netty as long at is 4.1.x.

ferenc-csaky:
@pjfanning in that case it should be no problem on our end to wait for 1.2, thanks! 👍

ferenc-csaky:
@afedulov To be clear, I did not mean to bump Netty for the currently in-flight patch releases, but for the next one against Flink 1.20. :)

He-Pin:
@pjfanning Do you think we should have this in 1.1.4? seems we need more time to publish 1.2.0

pjfanning:
I don't think this should go into 1.1.4.

He-Pin:
As Pekko 1.2.0 is out, so I can close this now.
@ferenc-csaky ping, Pekko 1.2.0 is released.

ferenc-csaky:
@He-Pin  Thank you for the reminder!

He-Pin:
@ferenc-csaky, btw, will Flink move to Java 17+ later is there any timeline?

ferenc-csaky:
@He-Pin we already support Java 21, and there is already a ticket to add Java 25 support once it gets released ([FLINK-37719](https://issues.apache.org/jira/browse/FLINK-37719)).

PR #1710, opened 2025-01-14
Add info about port change in cluster-management

gregor-i:
pekko uses a different port than akka for the module cluster-mangement. This is not documented in the migration guide. This PR adds it.

Sources:
* [Pekko Doc](https://nightlies.apache.org/pekko/docs/pekko-management/main-snapshot/docs/pekko-management.html#basic-configuration)
* [Akka Doc](https://doc.akka.io/libraries/akka-management/current/akka-management.html#basic-configuration)


pjfanning:
pekko-cluster-management is not part of Pekko Core. It is part of Pekko Management - a separate module and separate repo. We have the port change in its docs.

https://pekko.apache.org/docs/pekko-management/current/migration.html#1-0

pjfanning:
I can add a change to the core migration docs to direct people to the other module migration guides.

raboof:
I think #1711 closes this issue, right?

gregor-i:
That works for me, thanks!

PR #1722, opened 2025-01-16
[DO NOT MERGE] test with scala 3.6.2

pjfanning:


He-Pin:
IIRC， we need to use `?` instead of `[_]` in type parameter

He-Pin:
@xuwei-k Do you have some drop-in rules to support this, thanks.

pjfanning:
> IIRC， we need to use `?` instead of `[_]` in type parameter

So far, this is not causing us an issue. We still support Scala 2.12 builds, so we can't make massive code changes. We can make some small changes if they help us getter closer to Scala 3.6 support.

If it comes to making large PRs, then I think we need to abandon this until we start Pekko 2.0.0 work.

xuwei-k:
- https://github.com/typelevel/cats-effect/pull/4236
- https://github.com/scala/scala/pull/10005

scalafmt support rewrite `[_]` to `[?]`.

latest Scala 2.12.x and 2.13.x support new wildcard syntax.

PR #1743, opened 2025-01-24
chore: Add the missing EmptySource case to TraversalBuilder

He-Pin:
Motivation:
The EmptySource should be added to the cases too.
extracted from https://github.com/apache/pekko/pull/1701#discussion_r1929190034

Modification:
Add EmptySource to match


pjfanning (approved):
lgtm - this is a bug fix because this test fails without the main source change

```
[01-24 22:30:06.562] [info] - find Source.empty via TraversalBuilder with 
isEmptySource *** FAILED *** (1 millisecond)
[01-24 22:30:06.562] [info]   false was not true (TraversalBuilderSpec.scala:452)
```

Issue #1749, opened 2025-01-28
Possibly replace ThreadLocalRandom


pjfanning:
ThreadLocalRandom may not be a great idea if you are using Virtual Threads.

Java 17+ has a new RandomGenerator and we could allow users to choose via config which generator to use.

```
RandomGenerator random = RandomGenerator.of("L64X128MixRandom")
```

List of algorithms:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/package-summary.html#algorithms

We could have a centralised util function to get a generator. Java 8-16 users would get a ThreadLocalRandominstance when this functio was called. Java 17+ users - we could use Reflection or MethodHandles to create a generator based on config.

He-Pin:
Yes, recently Netty has been making changes to work well for virtual thread.

PR #1757, opened 2025-01-30
[WIP] New RandomNumberGenerator API

pjfanning:
* relates to #1749
* experimental new RandomNumberGenerator API that allows us to plug in different RandomGenerators based on the config (to be done)
* so far, I'm just plugging in the API to catch compile issues
* later, I will add the new configs and RandomNumberGenerator.get() will react to the config
* for performance reasons, the init of RandomNumberGenerator.get() will happen once so the config will be read just once
* not necessarily going to be merged

### DIFF
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/StubbedActorContext.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/StubbedActorContext.scala
@@ -19,12 +19,11 @@ import pekko.actor.typed._
 import pekko.actor.typed.internal._
 import pekko.actor.{ ActorPath, ActorRefProvider, InvalidMessageException }
 import pekko.annotation.InternalApi
-import pekko.util.Helpers
+import pekko.util.{ Helpers, RandomNumberGenerator }
 import pekko.{ actor => classic }
 import org.slf4j.{ Logger, Marker }
 import org.slf4j.helpers.{ MessageFormatter, SubstituteLoggerFactory }
 
-import java.util.concurrent.ThreadLocalRandom.{ current => rnd }
 import scala.collection.immutable.TreeMap
 import scala.concurrent.ExecutionContextExecutor
 import scala.concurrent.duration.FiniteDuration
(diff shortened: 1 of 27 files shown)


He-Pin:
Another thing is with the current implementation, the virtual thread is never pooled, which means the ThreadLocalRandom will not leak the thread locals

pjfanning:
So far, it seems that the performance is lower if you do not use the ThreadLocalRandom. The speed is still good and I don't think we are in a hotspot area where the small perf loss here will make a big diff. I don't think anyone should use this option unless they have a good reason to worry about their virtual thread usage leading to an excessive amount of object instance creation (one ThreadLocalRandom instance per virtual thread).

He-Pin (commented):
This is marked as [WIP] and the author notes it's 'not necessarily going to be merged'. A few observations:

1. **Purpose**: The new `RandomNumberGenerator` API provides configurability for plugging in different RNG implementations, which addresses #1749. This is useful for cases where security or reproducibility requirements differ.

2. **WIP status**: The PR explicitly states it's experimental and only plugged in to catch compile issues. Config-driven selection is noted as a future step.

3. **Scope**: The changes touch many modules (actor, cluster, remote, stream, etc.) which indicates this would be a significant architectural change.

4. **Staleness**: Open since January 2025 with no reviews and no recent activity. Since the author marked it as potentially not being merged, this should either be completed or closed. If the team wants this functionality, someone needs to take ownership of finishing the config-driven selection logic.

pjfanning:
@He-Pin the reason that I haven't pushed to merge this, is that I am not happy with having the new RandomNumberGenerator instance being an object that reads a static config.
I'm thinking that I could add make add a randomNumberGenerator() method to the ActorSystem and that it returns a Java 17+ RandomGenerator instance.
This randomNumberGenerator is not intended to cryptographically strong and where you need that, you should still use SecureRandom directly.
The new randomNumberGenerator would be a wrapper for ThreadLocalRandom by default and via the system config, it can be made to return a wrapper for Java 21+ https://openjdk.org/jeps/356 compatible RandomGenerator instead.
With an ActorSystem instance that has a hot path requiring random numbers, I would expect the Java compiler to optimise the code because the config would only be read once at ActorSystem creation.
The new code would only be used in places where it is easy to access the underlying ActorSystem instance.
It might be best to only have this accessible on ExtendedActorSystem.

pjfanning:
I'm working on a new PR

PR #1768, opened 2025-02-08
Java API in PersistenceQuery - add new methods that don't require the unused class param

pjfanning:


pjfanning:
closing as the unused param seems to help in deriving the return type

PR #1773, opened 2025-02-16
Persistence API: add CurrentLastSequenceNumberByPersistenceIdQuery

pjfanning:
Based on https://github.com/apache/pekko-persistence-jdbc/pull/267

### DIFF
--- a/persistence-query/src/main/scala/org/apache/pekko/persistence/query/javadsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
+++ b/persistence-query/src/main/scala/org/apache/pekko/persistence/query/javadsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.persistence.query.javadsl
+
+import java.util.Optional
+import java.util.concurrent.CompletionStage
+
+/**
+ * A trait that enables querying the current last known sequence number for a given `persistenceId`.
+ * @since 1.2.0
+ */
+trait CurrentLastKnownSequenceNumberByPersistenceIdQuery extends ReadJournal {
+
+/**
+   * Returns the last known sequence number for the given `persistenceId`. Empty if the `persistenceId` is unknown.
+   *
+   * @param persistenceId The `persistenceId` for which the last known sequence number should be returned.
+   * @return Some sequence number or None if the `persistenceId` is unknown.
+   */
+  def currentLastKnownSequenceNumberByPersistenceId(persistenceId: String): CompletionStage[Optional[Long]]
+}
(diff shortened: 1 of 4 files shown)


persistence-query/src/main/scala/org/apache/pekko/persistence/query/javadsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
```diff
@@ -0,0 +1,36 @@
...
+import java.util.concurrent.CompletionStage
+
+/**
+ * A trait that enables querying the current last known sequence number for a given `persistenceId`.
+ * @since 1.2.0
+ */
+trait CurrentLastKnownSequenceNumberByPersistenceIdQuery extends ReadJournal {
+
+  /**
+   * Returns the last known sequence number for the given `persistenceId`. Empty if the `persistenceId` is unknown.
```
He-Pin, line 30: Optional should use `the value is absent` instead of `Empty`?

persistence-query/src/main/scala/org/apache/pekko/persistence/query/scaladsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package org.apache.pekko.persistence.query.scaladsl
+
+import scala.concurrent.Future
+
+/**
+ * A trait that enables querying the current last known sequence number for a given `persistenceId`.
+ * @since 1.2.0
+ */
+trait CurrentLastKnownSequenceNumberByPersistenceIdQuery extends ReadJournal {
```
He-Pin, line 26: CurrentLastKnownSequenceNumberQuery ?

janjaali: Just to highlight the intent behind the `Current` in name: It seemed to me that "final" queries like `CurrentEventsByPersistenceIdQuery` are prefixed with 'Current', while continuous corresponding queries like `EventsByPersistenceIdQuery` drop this prefix. The `ByPersistenceId` clarifies the scope of query. Even though a `LastKnownSequenceNumberByPersistenceIdQuery` doesn't exist yet, one could imagine a query that continuously emits the last known sequence number for persistence id in future and thus I had preferred to 1. not block the name already 2. align to existing naming schema.

He-Pin: I just think the name is too long, maybe `CurrentLastKnownSequenceNumberQuery`?

He-Pin (approved):
Overall look good to me

persistence-query/src/main/scala/org/apache/pekko/persistence/query/scaladsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
ptrdom: How about dropping the `Known` from the name - `Current` already implies it in a way. `EventSourcedBehavior.lastSequenceNumber` already exist, so naming would align.

persistence-query/src/main/scala/org/apache/pekko/persistence/query/javadsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
pjfanning: I think 'empty' makes since this is the terminology used by scala.Option and java Optional.

persistence-query/src/main/scala/org/apache/pekko/persistence/query/scaladsl/CurrentLastKnownSequenceNumberByPersistenceIdQuery.scala
pjfanning: Would CurrentLastSequenceNumberQuery work?

ptrdom: I think `CurrentLastSequenceNumberByPersistenceIdQuery` is good if we are sticking to the style of having a trait per least amount of methods, which does make sense - for example each journal might want to implement `CurrentLastSequenceNumberByPersistenceIdQuery`, but not necessarily `CurrentLastSequenceNumberBySliceQuery` or `CurrentLastSequenceNumberByTagQuery` if we ever implement them.

Roiocam (approved):
LGTM, but the function name kind of too long

pjfanning:
Any objections to me merging this over the coming days?

PR #1775, opened 2025-02-22
chore: optimize recoverWith to avoid some materialization

He-Pin:
Motivation:
refs： https://github.com/apache/pekko/issues/1663

Result:
Avoid some materialization.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowRecoverWithSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowRecoverWithSpec.scala
@@ -14,6 +14,7 @@
 package org.apache.pekko.stream.scaladsl
 
 import scala.annotation.nowarn
+import scala.concurrent.Future
 import scala.util.control.NoStackTrace
 
 import org.apache.pekko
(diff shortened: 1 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -2171,14 +2177,29 @@ private[pekko] object TakeWithin {
...
+                        case Some(Failure(ex))   => failStage(ex)
+                        case None =>
+                          switchTo(source)
+                          attempt += 1
+                      }
+                    case iterableSource: IterableSource[T @unchecked] =>
+                      emitMultiple(out, iterableSource.elements, () => completeStage())
+                    case javaStreamSource: JavaStreamSource[T @unchecked, _] =>
+                      emitMultiple(out, javaStreamSource.open().iterator(), () => completeStage())
+                    case maybeEmpty if TraversalBuilder.isEmptySource(maybeEmpty) => completeStage()
```
pjfanning, line 2196: Is this going to be slower than the old code? Old code called `TraversalBuilder.isEmptySource(source)` before calling `TraversalBuilder.getSingleSource`. The new code calls `TraversalBuilder.getValuePresentedSource` before checking `TraversalBuilder.isEmptySource(source)` .

He-Pin: I think the empty case should not be that common, or we can remove the `maybeEmpty` test and let it left in a `_ =>` case , but that way, we need to keep the changes in `getValuePresentedSource(source)` in sync with here.

what do you think

He-Pin:
@pjfanning @Roiocam I think this one is ready.

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowRecoverWithSpec.scala
```diff
@@ -23,6 +23,8 @@ import pekko.stream.testkit.StreamSpec
 import pekko.stream.testkit.Utils._
 import pekko.stream.testkit.scaladsl.TestSink
 
+import scala.concurrent.Future
```
pjfanning, line 26: can you keep the scala imports together?

pjfanning (dismissed):
lgtm

He-Pin:
> lgtm

I think we can do a sortimports globally..

pjfanning (approved):

PR #1776, opened 2025-02-22
feat: Add emitMulti with Spliterator support

He-Pin:
Motivation:
Add Spliterator support for emitMulti, reducing overhead when working with java stream.

Result:
less overhead

I would like rebase https://github.com/apache/pekko/pull/1775 on this.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/impl/GraphStageLogicSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/impl/GraphStageLogicSpec.scala
@@ -121,6 +121,20 @@ class GraphStageLogicSpec extends StreamSpec with GraphInterpreterSpecKit with S
     override def toString = "GraphStageLogicSpec.emitEmptyIterable"
   }
 
+  object EmitSplitIterator extends GraphStage[SourceShape[Int]] {
+    val out = Outlet[Int]("out")
+    override val shape = SourceShape(out)
+    override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+      setHandler(out,
+        new OutHandler {
+          override def onPull(): Unit = emitMultiple(
+            out,
+            java.util.stream.Stream.of(1, 2, 3).spliterator(), () => emit(out, 42, () => completeStage()))
+        })
+    }
+    override def toString = "GraphStageLogicSpec.emitEmptyIterable"
+  }
+
   private case class ReadNEmitN(n: Int) extends GraphStage[FlowShape[Int, Int]] {
     override val shape = FlowShape(Inlet[Int]("readN.in"), Outlet[Int]("readN.out"))
 
@@ -193,6 +207,12 @@ class GraphStageLogicSpec extends StreamSpec with GraphInterpreterSpecKit with S
     "emit properly after empty iterable" in {
 
       Source.fromGraph(emitEmptyIterable).runWith(Sink.seq).futureValue should ===(List(42))
+
+    }
+
+    "emit properly when using split iterator" in {
+
+      Source.fromGraph(EmitSplitIterator).runWith(Sink.seq).futureValue should ===(List(1, 2, 3, 42))
 
     }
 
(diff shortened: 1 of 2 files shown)


He-Pin:
@pjfanning Would you like to take a look at this, thank.

pjfanning:
> @pjfanning Would you like to take a look at this, thank.

Can you give some background on who needs this? Is it intended that anyone can use this or do you just want it for https://github.com/apache/pekko/pull/1775?

He-Pin:
> > @pjfanning Would you like to take a look at this, thank.
> 
> Can you give some background on who needs this? Is it intended that anyone can use this or do you just want it for https://github.com/apache/pekko/pull/1775?


For now, only in #1775 , but it's generally

pjfanning:
the code is protected - so other users won't easily be accessed by other users

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -979,6 +980,26 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
     } else andThen()
 
+  /**
+   * Emit a sequence of elements through the given outlet and continue with the given thunk
+   * afterwards, suspending execution if necessary.
+   * This action replaces the [[OutHandler]] for the given outlet if suspension
+   * is needed and reinstalls the current handler upon receiving an `onPull()`
+   * signal (before invoking the `andThen` function).
+   */
+  final protected def emitMultiple[T](out: Outlet[T], elems: Spliterator[T], andThen: () => Unit): Unit = {
```
He-Pin, line 990: @pjfanning It the same access level as the other `emitMultiple` method, its been used in a sub class of `GraphStage`

stream-tests/src/test/scala/org/apache/pekko/stream/impl/GraphStageLogicSpec.scala
```diff
@@ -121,6 +121,20 @@ class GraphStageLogicSpec extends StreamSpec with GraphInterpreterSpecKit with S
     override def toString = "GraphStageLogicSpec.emitEmptyIterable"
   }
 
+  object emitSplitIterator extends GraphStage[SourceShape[Int]] {
```
pjfanning, line 124: Can we call this EmitSpliterator ?

pjfanning: Class names and Object names should start with capital letters

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -1118,6 +1139,19 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
     }
   }
 
+  private final class EmittingSpliterator[T](_out: Outlet[T], elems: Spliterator[T], _previous: OutHandler,
```
pjfanning, line 1142: * the advantage of a Spliterator over an iterator is that you can use trySplit() to partition the workload so you can get multiple threads to work on the data
* if you just use tryAdvance, you might as well be just using the standard iterator]
* so what do we gain here?

He-Pin: you can avoid the `hasNext` and `next` call when you already have a `Spliterator`

pjfanning: I'm not sure this will make a big diff to performance. I'll approve this but without jmh bechmarks, I am very sceptical about whether this extra code is worth the extra maintenance overhead.

pjfanning (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
He-Pin: Thanks

PR #1788, opened 2025-03-05
remove incomplete section in cluster-singleton.md

pjfanning:
* relates to https://github.com/akka/akka/issues/27705 - closed but the docs were not updated
* and https://github.com/apache/pekko/discussions/1786
* the link to the issue is broken because it generates one pointing to Pekko issues not Akka issues
* I don't see a need to raise an issue to provide actual documentation here - if someone wants to provide it, they can but I don't think logging an issue will encourage that


He-Pin (approved):
lgtm

PR #1794, opened 2025-03-09
upgrade to Netty 4.2.0.Final

pjfanning:
full release is out

### DIFF
--- a/multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
+++ b/multi-node-testkit/src/main/scala/org/apache/pekko/remote/testconductor/RemoteConnection.scala
@@ -16,6 +16,7 @@ package org.apache.pekko.remote.testconductor
 import java.net.InetSocketAddress
 import java.util.concurrent.TimeUnit
 
+import scala.annotation.nowarn
 import scala.util.control.NonFatal
 
 import io.netty.bootstrap.{ Bootstrap, ServerBootstrap }
@@ -114,6 +115,7 @@ private[pekko] trait RemoteConnection {
 /**
  * INTERNAL API.
  */
+@nowarn("msg=deprecated")
 private[pekko] object RemoteConnection {
   def apply(
       role: Role,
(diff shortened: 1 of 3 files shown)


pjfanning:
Our build fails if we use any deprecated code and Netty's NioEventLoopGroup is now deprecated.

He-Pin:
Yes, in Netty 4.2, the IOHandler is seperated with the EventLoop

He-Pin:
I think we should schedule this to 2.0.x and keep using Netty 4.1.x in 1.2.x

pjfanning:
I wouldn't delay the 1.2.0 release for this but if 4.2.0 final release is made before we make the 1.2.0 final release then I think we should upgrade.

He-Pin:
@pjfanning That will require flink to upgrade netty too.

mkurz:
> I wouldn't delay the 1.2.0 release for this but if 4.2.0 final release is made before we make the 1.2.0 final release then I think we should upgrade.

+1 for upgrading to Netty 4.2 in Pekko 1.2.0 already (I guess Pekko 2.0 is far away still, so IMHO upgrading now would be much better). Should not do any harm, [according to their release notes](https://netty.io/news/2025/03/07/4-2-0-RC4.html):

> Everyone using netty 4.1.x should be able to upgrade to 4.2.0.RC4 without any API breakage. The only new requirement is JDK8 or later.

We will upgrade in Play as soon as 4.2 is out (and adding the new io_uring native transport):
- https://github.com/playframework/playframework/pull/13200

He-Pin:
@ferenc-csaky wdyt , we plan to upgrade to netty 4.2.0.final in next version, Will flink upgrade to Netty 4.2.x too?

pjfanning:
@He-Pin @mkurz I had a quick look at Pekko code and it doesn't immediately look like we use the Netty abstractions that now support io_uring. I may have missed something. I would prefer to do any uptake of Netty 4.2 specific features in a later PR. Still, it's interesting to consider if there is something that we could support via a config option.

mkurz:
I agree. Just upgrading the dependency is definitely enough for now.

He-Pin:
@pjfanning Yes, we need some abstraction to do that, currently, it's bound to NioSocket, the Netty 4.2 is pretty cool about separate the iohandler with eventloop, I think that can be done more elegantly.

ferenc-csaky:
> @ferenc-csaky wdyt , we plan to upgrade to netty 4.2.0.final in next version, Will flink upgrade to Netty 4.2.x too?

Sorry for the late answer, AFAIK there is no push ATM to do that. Although IMO the Flink dependency should probably not be a blocker for this on the Pekko side. Requiring the Netty upgrade might drive the project a bit to do that sooner than later.

PR #1795, opened 2025-03-09
protobuf-java 4.29.3

pjfanning:
* Move to protobuf 4.
* The generated code differs quite a bit but not in any ways that affect our code


pjfanning:
GeneratedMessageV3 is deprecated

pjfanning:
* Seems to work - tests pass
* Big binary compatibility diffs for the generated classes

Do we expect this to cause issues for users? I don't think they'll be coding against the generated classes.
This would be target at the 1.2.0 release or if it's too big for there, we might need to consider delaying to 2.0.0.

@raboof @kerr @Roiocam what do you think?

pjfanning:
Pekko gRPC uses the io.grpc libs and has a direct dependency on the google protobuf jars. Pekko core depends on our shaded jar. So we can update independently. Pekko gRPC has a PR but it is blocked waiting on a new scalapb release. A new alpha release of scalapb 1.0.0 is due soon. https://github.com/apache/pekko-grpc/pull/371

He-Pin:
I checked https://github.com/grpc/grpc-java/blob/master/examples/pom.xml, I think we should keep using protobuf 3.x in 1.2.0

pjfanning:
> I checked https://github.com/grpc/grpc-java/blob/master/examples/pom.xml, I think we should keep using protobuf 3.x in 1.2.0

This is covered in https://github.com/apache/pekko/pull/1795#issuecomment-2717874334

pjfanning:
> > I checked https://github.com/grpc/grpc-java/blob/master/examples/pom.xml, I think we should keep using protobuf 3.x in 1.2.0
> 
> This is covered in [#1795 (comment)](https://github.com/apache/pekko/pull/1795#issuecomment-2717874334)
> 
> @raboof has suggested that maybe we could merge this and see what happens in the pekko-grpc build. I expect that it will have no effect but if it does, I will be happy to revert this.

@raboof do you think this is ok to merge? If it causes any downstream issues, I will revert it. I think this is reasonably unlikely to happen.

raboof (approved):
This seems like a worthwhile experiment to perform.

In retrospect it is now a bit sad that the shaded package is called `org.apache.pekko.protobufv3` and the artifact `protobuf-v3`, but I think changing that to v4 might cause trouble... in any case we can consider that separately. I think this is an experiment worth doing.

PR #1802, opened 2025-03-18
Introduce Task monad

### LINKED ISSUE #1801: Task monad based on `Materializer`
In a pekko Java project I'm working on, I've been creating a `Task` class whose abstraction looks something like:
```java
public abstract class Task<T> {
    public static <T> Task<T> of(Supplier<T> fn) {}
    public static <T> Task<T> connect(Source<A, ? extends KillSwitch> source, Sink<A,CompletionStage<T>> sink) {}
    /* ... */
    public abstract TaskControl<T> run(Materializer mat);
}
public class TaskControl<T> {
    public void cancel() { /* ... */ }
    public CompletableFuture<T> future() { /* ... */ }
}
```
`Task` is a description of running something, but also has a unified cancellation API (across futures and connected `RunnableGraph`s). It is inspired heavily from ZIO's API. A lot of combinators are written.

Would this be an interesting abstraction to have in Pekko itself? The fact that streams need to be materialized brings them close to a `Task` monad already, and in my experience having a unified API between running streams and futures is very valuable.

We'd of course add a nice Scala API as well (which should be trivial).

jypma:
Relates to #1801 

This commit introduces `Task`, a data structure that represents a recipe, or program, for producing a value of type T (or failing with an exception). It is similar in semantics to `RunnableGraph[T]`, but intended as first-class building block.

It has the following properties:
- A task can have resources associated to it, which are guaranteed to be released if the task is cancelled or fails
- Tasks can be forked so multiple ones can run concurrently
- Such forked tasks can be cancelled

A Task can be created from a `RunnableGraph` which has a `KillSwitch`, by connecting a `Source` and a `Sink` through a `KillSwitch`, or by direct lambda functions.

### Open discussion points and TODOs (in order of highest architectural impact):

- [X] Current cancellation of a graph through `KillSwitch` doesn't communicate back a trigger when the graph is actually done cancelling. However, since we also have the `Future[T]` that the graph's sink materializes to, that's enough to synchronize on.
- [X] Resource safety, which can then guarantee cleanup even under cancellation signals.
- [X] Background processing wth cancellation (fibers, `forkDaemon`, `forkResource`). No child fibers (yet).
- [X] Add many more combinators (plain `zip`, `andThen`, `before`, `raceAll`)
- [x] Add more combinators (`asResource`, ...)
- [x] Add `.delay()` and general scheduling by hooking in the scheduling features of `Materializer`
- [x] `Sink.forEachTask` and `Flow.mapTask`
- [x] `Resource.ofAutoCloseable`
- [x] `Task.never()`
- [ ] `Task.runAsMain` (or perhaps `Application` base classes / traits), which runs a task as `main()`, capturing Ctrl+C to interrupt the task (and, hence, clean up resources).
- [ ] Move to `task` module. Don't depend on `stream`. Streams should depend on `task`, so we can use tasks directly in stream combinators.
- [ ] Scala DSL

### Needing an eye from reviewers:
- [ ] More test cases to exercise typical edge cases and concurrency race conditions

### To be handled in later pull requests:
- `TestClock` (implementation of `Clock` for unit tests, where tests can advance the clock, so time-driven code can run deterministically and faster than realtime)
- Fun `Task`-friendly concurrency primitives like a queue, mutex, and a mutable ref with stream subscription support.
- (Potentially) Another monad `Effect[E,A]` which introduces modelled errors (in addition to exceptions). This can be easily built on top of a `TaskDef[Either[E,A]]`.

Fixes #1801 .

### DIFF
--- a/stream-tests/src/test/java/org/apache/pekko/task/javadsl/TaskTest.java
+++ b/stream-tests/src/test/java/org/apache/pekko/task/javadsl/TaskTest.java
@@ -1,0 +1,270 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.task.javadsl;
+
+import org.apache.pekko.stream.StreamTest;
+import org.apache.pekko.testkit.PekkoJUnitActorSystemResource;
+import org.apache.pekko.testkit.PekkoSpec;
+import org.apache.pekko.stream.Materializer;
+import org.apache.pekko.Done;
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import org.apache.pekko.japi.function.Creator;
+import org.apache.pekko.stream.javadsl.Sink;
+import org.apache.pekko.stream.javadsl.Source;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicLong;
+
+import java.util.Optional;
+import java.time.Duration;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TaskTest extends StreamTest {
+  private final Runtime runtime = Runtime.create(Materializer.createMaterializer(system));
+
+  public TaskTest() {
+    super(actorSystemResource);
+  }
+
+  @ClassRule
+  public static PekkoJUnitActorSystemResource actorSystemResource =
+      new PekkoJUnitActorSystemResource("TaskTest", PekkoSpec.testConf());
+
+  private <T> T run(Task<T> task) throws Throwable {
+    return runtime.run(task.timeout(Duration.ofSeconds(2)));
+  }
+
+  @Test
+  public void can_run_task_from_lambda() throws Throwable {
+    assertEquals("Hello", run(Task.run(() -> "Hello")));
+  }
+
+  @Test
+  public void can_map() throws Throwable {
+    assertEquals(25, run(Task.run(() -> "25").map(Integer::parseInt)).intValue());
+  }
+
+  @Test
+  public void can_flatMap_to_run() throws Throwable {
+    assertEquals(
+        25, run(Task.run(() -> "25").flatMap(s -> Task.run(() -> Integer.parseInt(s)))).intValue());
+  }
+
+  @Test
+  public void can_zipPar_two_tasks() throws Throwable {
+    Task<String> task =
+        Task.run(
+            () -> {
+              return "Hello";
+            });
+    assertEquals("HelloHello", run(task.zipPar(task, (s1, s2) -> s1 + s2)));
+  }
+
+  @Test
+  public void zipPar_interrupts_first_on_error_in_second() throws Throwable {
+    AtomicLong check = new AtomicLong();
+    Task<String> task1 =
+        Task.succeed("A").delayed(Duration.ofMillis(100)).before(Task.run(check::incrementAndGet));
+    Task<String> task2 = Task.fail(new RuntimeException("simulated failure"));
+    org.junit.Assert.assertThrows(
+        RuntimeException.class, () -> run(task1.zipPar(task2, (a, b) -> a + b)));
+    assertEquals(0, check.get());
+  }
+
+  @Test
+  public void zipPar_interrupts_second_on_error_in_first() throws Throwable {
+    AtomicLong check = new AtomicLong();
+    Task<String> task1 =
+        Task.succeed("A").delayed(Duration.ofMillis(100)).before(Task.run(check::incrementAndGet));
+    Task<String> task2 = Task.fail(new RuntimeException("simulated failure"));
+    org.junit.Assert.assertThrows(
+        RuntimeException.class, () -> run(task2.zipPar(task1, (a, b) -> a + b)));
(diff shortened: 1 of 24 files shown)


stream/src/main/java/org/apache/pekko/task/javadsl/Fiber.java
```diff
@@ -0,0 +1,22 @@
+package org.apache.pekko.task.javadsl;
```
pjfanning, line 1: new source files need a license header like the one in https://github.com/apache/pekko/blob/main/project/AddMetaInfLicenseFiles.scala

stream/src/main/java/org/apache/pekko/task/javadsl/Task.java
```diff
@@ -0,0 +1,116 @@
+package org.apache.pekko.task.javadsl;
+
+import org.apache.pekko.japi.function.Effect;
```
pjfanning, line 3: can you order the imports?

stream/src/main/scala/org/apache/pekko/task/Task.scala
```diff
@@ -0,0 +1,176 @@
...
+        val (killswitch, future) = mat.materialize(graph)
+        // This returns a Fiber, which when cancelled should cancel the graph as well.
+        val childFiber = new FiberImpl[g.Res] {
+          override def onCancel() = {
+            killswitch.shutdown()
+          }
+        }
+        future.onComplete { res =>
+          childFiber.complete(res)
+        }(scala.concurrent.ExecutionContext.global)
```
pjfanning, line 138: * we need to avoid hardcoding the global context
* even if comes to setting up a config in the reference.conf that defines an executioncontext created specifically for this
* might be nicer if it was controlled by an implicit param though

jypma: Sure thing, this one was temporary, that's why it's explicit :) I'll replace it with `executor`. The intent is to make the executor pluggable (probably by picking a dispatcher from the `Materializer`), I just haven't decided for the API on that yet.

pjfanning:
@jypma if you thinking copying some of the pekko-connectors-kafka code into your task package helps, go ahead. Let's keep any additions specific to the task package for now and we can later discuss if it is worth taking some of reusable bits out and making them available as public APIs or to be used by other Pekko classes.

jypma:
I've decided to just introduce a `RunningGraph` interface to represent a graph that will eventually complete (with a `Task`), and has the ability to be interrupted (with another `Task`). That covers both `KillSwitch` (where interruption is immediate, but followed by awaiting the sink's completion stage) and `DrainingControl` (where interruption has a clear "done" signal).

stream/src/main/scala/org/apache/pekko/task/ClockDef.scala
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.task
+
+import java.time.Duration
+import java.time.Instant
+import scala.jdk.DurationConverters._
```
pjfanning, line 22: we need to support Scala 2.12 - please use Pekko's own DurationConverters

stream/src/main/scala/org/apache/pekko/task/scaladsl/Sinks.scala
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.task.scaladsl
+
+import org.apache.pekko.task.AbstractTask
```
pjfanning, line 20: can you order the imports and use `import org.apache.pekko` so you can shorten the pekko imports

stream-tests/src/test/java/org/apache/pekko/task/javadsl/TaskTest.java
```diff
@@ -0,0 +1,242 @@
...
+        Task<String> task1 = Task.succeed("A").delayed(Duration.ofMillis(100)).before(Task.run(check::incrementAndGet));
+        Task<String> task2 = Task.fail(new RuntimeException("simulated failure"));
+        org.junit.Assert.assertThrows(RuntimeException.class, () -> run(task2.zipPar(task1, (a,b) -> a + b)));
+        assertEquals(0, check.get());
+    }
+
+  @Test
+  public void can_interrupt_forked_task() throws Throwable {
+    AtomicLong check = new AtomicLong();
+    Task<Long> task = Task.run(() -> check.incrementAndGet()).delayed(Duration.ofMillis(100));
```
pjfanning, line 105: can you format the file? `sbt javafmtAll`

He-Pin:
@jypma Seems there are some conflicts. I would like to learn more once this is ready

He-Pin (commented):
This is an ambitious and substantial addition to Pekko -- introducing a Task monad with resource safety, cancellation, and concurrency primitives. A few observations:

1. **Architectural significance**: This introduces a significant new abstraction layer. The TODO list in the PR body acknowledges important open items: Scala DSL completeness, moving to a dedicated `task` module, and `Task.runAsMain`. These should be resolved before merge.

2. **Module placement**: Currently in the `stream` module, but the TODO notes that streams should eventually depend on `task`, not the other way around. This architectural decision should be settled before merging.

3. **Testing gap**: The PR acknowledges needing more test cases for edge cases and concurrency race conditions. This is critical for a concurrency primitive.

4. **Staleness**: Open since March 2025 with activity stopping in May 2025. The concept is interesting but the PR is clearly incomplete (several unchecked TODO items). @jypma -- is this something you plan to complete? Otherwise consider closing and revisiting when there's bandwidth to finish the full implementation.

jypma:
I don't have any plans to complete this, unfortunately. I've concluded that Pekko doesn't currently have the resource safeties needed for a proper Task monad, particularly with respect to graph cancellation and shutdown. Hence, we'd have to look at the internals of Pekko first, e.g. tightening the semantics of callbacks within a graph when a shutdown happens, to get precise control there.

For those reading along: what I've been working on a bit on instead, is to take a framework that does give these guarantees, but only in Scala (ZIO), and create a "nice" fluent Java API on top of that. Pekko streams could slot in there as well. My initial approach for that is here: [JIO](https://github.com/jypma/jio), although development there has also stalled a bit (but only due to shifting interests). Send me a message if anyone is curious about that direction.

For this PR on Pekko, feel free to close it. Thanks for all the thoughtful discussion and comments!

PR #1804, opened 2025-03-22
chore: Avoid materialize an empty source in switchMap

He-Pin:


He-Pin:
I tried to avoid materialization as flatmapConcat, but I have not finished it yet, I will do that when I find time.

pjfanning (approved):
lgtm

PR #1805, opened 2025-03-22
fix: Materialize sub stream before pull it.

He-Pin:
Motivation:
The sub stream should be first materialized before pulling it.


He-Pin:
Will dig it later,hum.

PR #1806, opened 2025-03-22
Re-enable github discussions

mdedetrich:
As per the https://lists.apache.org/thread/908mycqopxnm7cggqok3g0vmwt06m2v6 , github discussions now need to be explicitly enabled in other to be viewable.


pjfanning (approved):
Honestly feels like this should be enabled by default. ASF are supposed to encourage community involvement.

PR #1817, opened 2025-04-18
chore: update Scala 3 to 3.3.6

He-Pin:
Test scala 3.3.6

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/ActorLifeCycleSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/ActorLifeCycleSpec.scala
@@ -37,6 +37,8 @@ object ActorLifeCycleSpec {
     def receive = { case "status" => sender() ! message("OK") }
   }
 
+  final case class Become(recv: ActorContext => Receive)
+
 }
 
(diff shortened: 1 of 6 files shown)


He-Pin:
It reports that the local type can not be checked.

@WojciechMazur So I have to move those local classes to the object.

WojciechMazur:
Bisect points to this change that was backport to 3.3.6 https://github.com/scala/scala3/pull/22099 

It seems to be fixing the soundness warning, seems like previously it was warning for bind in pattern matching but not for extractor 

```scala
case class RequestLogsOuter(streamId: Int)
@main def Test = {
  case class RequestLogs(streamId: Int)
  var x: List[Any] = List(RequestLogsOuter(12), RequestLogs(36), RequestLogs(42), 42)
  x.collect{
    case outer @ RequestLogsOuter(id) => println(1)
    case RequestLogs(42) => println(2) // warns in 3.3.6
    case v: RequestLogs => println(3) // warned in 3.3.5 if above is commented out
  }
}

```

fyi @tgodzik

tgodzik:
https://github.com/scala/scala3/issues/23028

pjfanning (approved):
lgtm

PR #1819, opened 2025-04-19
feat: Add invokeWithFeedbackCompletionStage for javadsl.

He-Pin:
Motivation:
Add java api for `invokeWithFeedback` method.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
@@ -1683,6 +1683,7 @@ trait AsyncCallback[T] {
    *
    * The handling of the returned future incurs a slight overhead, so for cases where it does not matter
    * to the invoking logic see [[AsyncCallback#invoke]]
+   * @since 1.2.0
    */
   def invokeWithFeedbackCompletionStage(t: T): CompletionStage[Done] = {
     import pekko.util.FutureConverters._
--- a/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
@@ -13,7 +13,7 @@
 
 package org.apache.pekko.stream.stage
 
-import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.{ CompletionStage, ConcurrentHashMap }
 import java.util.concurrent.atomic.AtomicReference
 
 import scala.annotation.nowarn
@@ -21,6 +21,7 @@ import scala.annotation.tailrec
 import scala.collection.{ immutable, mutable }
 import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration.FiniteDuration
+
 import org.apache.pekko
 import pekko.{ Done, NotUsed }
 import pekko.actor._
@@ -1668,6 +1669,25 @@ trait AsyncCallback[T] {
    * to the invoking logic see [[AsyncCallback#invoke]]
    */
   def invokeWithFeedback(t: T): Future[Done]
+
+  /**
+   * Java API
+   *
+   * Dispatch an asynchronous notification. This method is thread-safe and
+   * may be invoked from external execution contexts.
+   *
+   * The method returns directly and the returned future is then completed once the event
+   * has been handled by the operator, if the event triggers an exception from the handler the future
+   * is failed with that exception and finally if the operator was stopped before the event has been
+   * handled the future is failed with `StreamDetachedException`.
+   *
+   * The handling of the returned future incurs a slight overhead, so for cases where it does not matter
+   * to the invoking logic see [[AsyncCallback#invoke]]
+   */
+  def invokeWithFeedbackCompletionStage(t: T): CompletionStage[Done] = {
+    import pekko.util.FutureConverters._
+    invokeWithFeedback(t).asJava
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -1668,6 +1672,25 @@ trait AsyncCallback[T] {
...
+   * is failed with that exception and finally if the operator was stopped before the event has been
+   * handled the future is failed with `StreamDetachedException`.
+   *
+   * The handling of the returned future incurs a slight overhead, so for cases where it does not matter
+   * to the invoking logic see [[AsyncCallback#invoke]]
+   */
+  def invokeWithFeedbackCompletionStage(t: T): CompletionStage[Done] = {
+    import pekko.util.FutureConverters._
+    invokeWithFeedback(t).asJava
+  }
```
He-Pin, line 1693: It would be nice if we could split this method into something else, but as the graphdsl is not separated with two `javadsl/scaladsl` , I have to add it here.
cons: 
1. scala developer will see this
2. name is too long.

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -1643,7 +1644,10 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
  *
  * Typical use cases are exchanging messages between stream and substreams or invoking from external world sending
  * event to a stream
+ *
+ * Not for user extension
  */
+@DoNotInherit
```
pjfanning, line 1650: I don't understand this. `def invokeWithFeedback(t: T): Future[Done]` is not implemented so this trait needs to be extended.

pjfanning: That's my point. It can be inherited. Can we remove this annotation and maybe change the to say that this is for Pekko Internal Usage?

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
pjfanning: is it possible to add a basic unit test?

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
He-Pin: Do you mean add an @InternalApi annotation.

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
He-Pin: test added

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
pjfanning: Can we just undo the new comment and annotation then?

mdedetrich: Not sure if applicable but ‘@DoNotInherit` doesn't indicate that it can never be inherit, just that it shouldn't be inherited by users of Pekko. The Pekko library itself can inherit it as much as it wants.

The reason this annotation exists is because of limitations in Java, i.e. Java has no concept of sealed like Scala/Kotlin has

pjfanning: The trait has been as is for a while. I am struggling to understand why this PR needs to change the docs and annotations about whether this trait is extensible.

mdedetrich: @He-Pin what is the reason for not making this user extensible, are you relying on some specific behaviour which could be redefined by a user of Pekko?

mdedetrich: So I am not seeing why this trait per se would be problematic to extend, it does seem generic enough users could plausibly want to provide their own implementation.

The newly added method however does seem like it shouldn't be extended, in which case you would make it `final`.

Or am I missing something?

pjfanning: Why not let users extend and even override this trait? I have no problem with the code changes - just the doc and annotation changes.

He-Pin: I will remove that tomorrow but I think this is and only should be called but never be implemented by user

stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -1668,6 +1669,25 @@ trait AsyncCallback[T] {
...
+   * may be invoked from external execution contexts.
+   *
+   * The method returns directly and the returned future is then completed once the event
+   * has been handled by the operator, if the event triggers an exception from the handler the future
+   * is failed with that exception and finally if the operator was stopped before the event has been
+   * handled the future is failed with `StreamDetachedException`.
+   *
+   * The handling of the returned future incurs a slight overhead, so for cases where it does not matter
+   * to the invoking logic see [[AsyncCallback#invoke]]
+   */
```
He-Pin, line 1686:
```suggestion
   * @since 1.2.0
   */
```

pjfanning (approved):
Lgtm

PR #1824, opened 2025-04-20
feat: Add BufferUntilChanged functionality for Java and Scala streams

BH-M87:


pjfanning:
@BH-M87 thanks for the PR. Would it be possible to provide a description of the PR in the PR description. It is not appropriate to leave it blank.

stream-tests/src/test/java/org/apache/pekko/stream/javadsl/BufferUntilChangedTest.java
```diff
@@ -0,0 +1,116 @@
+/*
```
pjfanning, line 1: this is the wrong header unless you have copied this code from somewhere else - the right header for 100% new code is https://github.com/apache/pekko/blob/main/project/AddMetaInfLicenseFiles.scala#L1-L16

stream/src/main/scala/org/apache/pekko/stream/javadsl/SubFlow.scala
```diff
@@ -714,6 +714,40 @@ class SubFlow[In, Out, Mat](
   def sliding(n: Int, step: Int = 1): SubFlow[In, java.util.List[Out @uncheckedVariance], Mat] =
     new SubFlow(delegate.sliding(n, step).map(_.asJava)) // TODO optimize to one step
 
+  /**
+   * Collect subsequent repetitions of an element (that is, if they arrive right after one another) into multiple
+   * `List` buffers that will be emitted by the resulting SubFlow.
+   *
+   * @return a SubFlow that buffers elements until they change
```
pjfanning, line 721: add `@since 1.2.0` to new public methods

stream/src/main/java/org/apache/pekko/stream/javadsl/BufferUntilChanged.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param <T> the element type
+     * @return a Flow that buffers elements until they change
+     */
+    @ApiMayChange
+    public static <T> Flow<T, List<T>, NotUsed> flow() {
+        return Flow.<T>create().asScala().bufferUntilChanged().asJava();
+    }
+
+    /**
+     * Collect subsequent repetitions of an element (that is, if they arrive right after one another), as compared by a key
```
pjfanning, line 39: lines are too long

He-Pin:
Hi @pjfanning THIS PR is totally done with Ai llm, so we need take careful about merge it.

pjfanning (changes requested):
-1 from me if this is AI generated. It doesn't even compile. I would need proof that any AI generated code has not used a LLM trained on copyrighted code without permission. If you are going to use AI, you could at least test the code before submitting it and provide a proper justification for why the PR is needed.

He-Pin:
let's close this to keep a clean room

raboof:
for reference https://www.apache.org/legal/generative-tooling.html

PR #1841, opened 2025-05-09
Avoid register/unregister race condition that can wipe elements from BroadcastHup

andreaslochbihler-da:
Refs https://github.com/apache/pekko/issues/1836.

The first commit adds a test case to demonstrate that BroadcastHub actually drops elements. The second commit fixes BroadcastHub so that the test case passes.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
@@ -561,6 +561,49 @@ class HubSpec extends StreamSpec {
       out.expectComplete()
     }
 
+    "handle unregistration concurrent with registration" in {
+
+      var sinkProbe1: TestSubscriber.Probe[Int] = null
+
+      def registerConsumerCallback(id: Long): Unit = {
+        if (id == 1) {
+          sinkProbe1.cancel()
+          Thread.sleep(10)
+        }
+      }
+
+      val in = TestPublisher.probe[Int]()
+      val hubSource = Source
+        .fromPublisher(in)
+        .runWith(Sink.fromGraph(new BroadcastHub[Int](0, 2, registerConsumerCallback)))
+
+      // Put one element into the buffer
+      in.sendNext(15)
+
+      // add a consumer to receive the first element
+      val sinkProbe0 = hubSource.runWith(TestSink.probe[Int])
+      sinkProbe0.request(1)
+      sinkProbe0.expectNext(15)
+      sinkProbe0.cancel()
+
+      // put more elements into the buffer
+      in.sendNext(16)
+      in.sendNext(17)
+      in.sendNext(18)
+
+      // Add another consumer and kill it during registration
+
+      sinkProbe1 = hubSource.runWith(TestSink.probe[Int])
+      Thread.sleep(100)
+
+      // Make sure that the element 16 isn't lost by reading it with a third consumer
+      val sinkProbe2 = hubSource.runWith(TestSink.probe[Int])
+      sinkProbe2.request(1)
+      sinkProbe2.expectNext(16)
+
+      in.sendComplete()
+      sinkProbe2.cancel()
+    }
   }
 
   "PartitionHub" must {
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
@@ -476,7 +476,8 @@ object BroadcastHub {
 /**
  * INTERNAL API
  */
-private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: Int)
+private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: Int,
+    registrationPendingCallback: Long => Unit = _ => ())
     extends GraphStageWithMaterializedValue[SinkShape[T], Source[T, NotUsed]] {
   require(startAfterNrOfConsumers >= 0, "startAfterNrOfConsumers must >= 0")
   require(bufferSize > 0, "Buffer size must be positive")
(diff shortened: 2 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
```diff
@@ -476,7 +476,8 @@ object BroadcastHub {
 /**
  * INTERNAL API
  */
-private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: Int)
+private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: Int,
+    registrationPendingCallback: Long => Unit = _ => ())
```
He-Pin, line 480: Add a doc about the `registrationPendingCallback` 's function input and expected output?

andreaslochbihler-da: Added both in [4596113](https://github.com/apache/pekko/pull/1841/commits/4596113cd124f72f3000bd2bd899605506089bd9)

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
```diff
@@ -561,6 +561,49 @@ class HubSpec extends StreamSpec {
...
+
+      // put more elements into the buffer
+      in.sendNext(16)
+      in.sendNext(17)
+      in.sendNext(18)
+
+      // Add another consumer and kill it during registration
+
+      sinkProbe1 = hubSource.runWith(TestSink.probe[Int])
+      Thread.sleep(100)
```
pjfanning, line 597: * the reliance on short sleeps will make this test brittle, especially in CI
* would it be possible to use Scalatest's eventually support to retry the test if it fails?

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
```diff
@@ -561,6 +561,49 @@ class HubSpec extends StreamSpec {
...
     }
 
+    "handle unregistration concurrent with registration" in {
+
+      var sinkProbe1: TestSubscriber.Probe[Int] = null
+
+      def registerConsumerCallback(id: Long): Unit = {
+        if (id == 1) {
+          sinkProbe1.cancel()
+          Thread.sleep(10)
```
pjfanning, line 571: is this sleep useful?

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
andreaslochbihler-da: > the reliance on short sleeps will make this test brittle

I don't think that the test is going to be flaky. It tests normal behavior of the `BroadcastHub` and therefore should pass even if you remove all the sleeps. The delays are only there to make a certain interleaving more likely. So you can argue that these short sleeps make the test brittle as a regression test because there is a chance that it will still pass even though a regression is introduced at some point in the future.

If I wanted to make this test rigorous, then I would have to add more instrumentation hooks to `BroadcastHub`. However, I could not find any prior art on how Pekko's unit tests reliably trigger race conditions in implementations. Do you have a pointer how this is done normally?

I have added this test primarily to demonstrate that there is a problem with the existing implementation. I'd also be fine with completely removing it if such race conditions are normally not covered by unit tests.

> would it be possible to use Scalatest's eventually support to retry the test if it fails?

I think we want the opposite: if this test fails, then there's something wrong with the implementation, and you do not want to hide this by retrying the test.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
```diff
@@ -854,7 +861,12 @@ private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: I
           }
 
           override def postStop(): Unit = {
-            if (hubCallback ne null)
+            // If `postStop` is called before the consumer has processed the `RegistrationPending`'s `Initialize` event,
+            // then the `Initialize` message will fail with a `StreamDetachedException`,
+            // upon which the `RegistrationPending` logic itself unregisters this consumer.
+            // In particular, this client must not send the `Unregister` event itself because the values in
+            // `previousPublishedOffset` and `offset` are wrong.
+            if ((hubCallback ne null) && offsetInitialized)
```
He-Pin, line 869: I see, which means there will be no leak in such scenario .

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
```diff
@@ -475,13 +475,18 @@ object BroadcastHub {
...
  */
-private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: Int)
+private[pekko] class BroadcastHub[T](startAfterNrOfConsumers: Int, bufferSize: Int,
+    registrationPendingCallback: Long => Unit)
     extends GraphStageWithMaterializedValue[SinkShape[T], Source[T, NotUsed]] {
   require(startAfterNrOfConsumers >= 0, "startAfterNrOfConsumers must >= 0")
   require(bufferSize > 0, "Buffer size must be positive")
   require(bufferSize < 4096, "Buffer size larger then 4095 is not allowed")
   require((bufferSize & bufferSize - 1) == 0, "Buffer size must be a power of two")
+  def this(startAfterNrOfConsumers: Int, bufferSize: Int) = this(startAfterNrOfConsumers, bufferSize, _ => ())
```
He-Pin, line 489: `_ => ()` can be replaced by `ConstantFun.anyToUnit` IIRC

pjfanning: @andreaslochbihler-da could you apply this change?

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
andreaslochbihler-da: It was probably for my local test case in earlier versions. I can now reproduce the broken behavior without the sleep on the old implementation. I'll remove it.

pjfanning (approved):
lgtm

He-Pin (commented):
lgtm

PR #1845, opened 2025-05-12
support jackson enum features

pjfanning:
A new of feature enum added a while back in Jackson.
https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.18.4/com/fasterxml/jackson/databind/cfg/EnumFeature.html

### DIFF
--- a/serialization-jackson/src/main/resources/reference.conf
+++ b/serialization-jackson/src/main/resources/reference.conf
@@ -92,6 +92,16 @@ pekko.serialization.jackson {
     FAIL_ON_UNKNOWN_PROPERTIES = off
   }
 
+  # Configuration of the ObjectMapper deserialization features.
+  # See com.fasterxml.jackson.databind.cfg.EnumFeature
+  # Enum values corresponding to the EnumFeature and their
+  # boolean values, for example:
+  #
+  # enum-features {
+  #   READ_ENUM_KEYS_USING_INDEX = off
+  # }
+  enum-features {}
+
   # Configuration of the ObjectMapper mapper features.
   # See com.fasterxml.jackson.databind.MapperFeature
   # Enum values corresponding to the MapperFeature and their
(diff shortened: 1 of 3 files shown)


serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonObjectMapperProvider.scala
```diff
@@ -512,6 +521,24 @@ class JacksonObjectMapperFactory {
...
+   *
+   * @param bindingName bindingName name of this `ObjectMapper`
+   * @param configuredFeatures the list of `DateTimeFeature` that were configured in
+   *                           `pekko.serialization.jackson3.enum-features`
+   */
+  def overrideConfiguredEnumFeatures(
+      @unused bindingName: String,
+      configuredFeatures: immutable.Seq[(EnumFeature, Boolean)])
+      : immutable.Seq[(EnumFeature, Boolean)] =
+    configuredFeatures
```
He-Pin, line 540: then the java api should be a Pair

jxnu-liguobin: Java doesn't have a pair, right? Then why not Map?

pjfanning: This code is consistent with the existing code in the class - ie there are similar methods already and this is basically a copy/paste of one of them.

serialization-jackson/src/main/resources/reference.conf
```diff
@@ -92,6 +92,16 @@ pekko.serialization.jackson {
...
 
+  # Configuration of the ObjectMapper deserialization features.
+  # See com.fasterxml.jackson.databind.cfg.EnumFeature
+  # Enum values corresponding to the EnumFeature and their
+  # boolean values, for example:
+  #
+  # enum-features {
+  #   READ_ENUM_KEYS_USING_INDEX = off
+  # }
+  enum-features {}
```
He-Pin, line 103: Should we include a Jackson version required in the comment?

He-Pin (approved):
lgtm, and there are no java api, but there is none before.

serialization-jackson/src/main/resources/reference.conf
pjfanning: thing is that this EnumFeature class exists since Jackson 2.14 and even Pekko 1.0 requires Jackson 2.14

Issue #1854, opened 2025-05-20
Seed nodes in a cluster using classic remoting


regiskuckaertz:
Hello 👋  while migrating a cluster that is still using classic remoting, I came across a blocker. Here's the config:

```
pekko {
  cluster {
    seed-nodes = [ "akka.ssl.tcp://node1", "akka.ssl.tcp://node2" ]
  }

  remote {
    artery.enabled = false
    accept-protocol-names = ["pekko", "akka"]
    protocol-name = "akka"
    classic {
      enabled-transports = ["pekko.remote.classic.netty.ssl"]
      netty.ssl { ... }
    }
  }
}
```

At startup, the cluster fails to start immediately, rejecting the protocol "akka.ssl.tcp" because only "akka.tcp" and "pekko.tcp" are valid. There's a few references to "pekko.ssl.tcp" in the documentation but not much so I wonder if maybe there is something missing.

I think the issue starts here: https://github.com/apache/pekko/blob/e308cf5e10e0614a7e887e90ce15dbdc3332ff97/remote/src/main/scala/org/apache/pekko/remote/RemoteSettings.scala#L213-L225 which is used in the cluster daemon to join members (I have also seen rejected members due to unaccepted protocols when I changed the seed nodes protocol to "akka.tcp" for testing). Maybe somewhere there should be a check to see if at least one enabled transport has `enable-ssl` turned on, and if yes then `{protocolName}.ssl.tcp` should be accepted? I'm not sure what the repercussions would be.

pjfanning:
@regiskuckaertz would you have logs or stack traces? This would help us to find the code where the rejection happens. The code that you highlight is meant to just support akka and pekko. The protocol suffix like `.tcp` or `.ssl.tcp` - we want to allow those but the lines you are highlighting are not the cause of the issue.

pjfanning:
I'm working on a test case and am getting a warning like this:

```
[WARN] [05/20/2025 14:11:21.255] [MixedProtocolClusterSpec-pekko.actor.internal-dispatcher-11] [Cluster(pekko://MixedProtocolClusterSpec)] Cluster Node [pekko.ssl.tcp://MixedProtocolClusterSpec@192.168.0.2:62635] - Trying to join member with wrong protocol, but was ignored, expected any of Set(pekko, akka, pekko.tcp, akka.tcp) but was [pekko.ssl.tcp]
```

[...]

PR #1857, opened 2025-05-20
Issue forming mixed Akka/Pekko cluster when classic remoting with SSL/TLS is used

pjfanning:
relates to #1854 

the aim would be to backport this the 1.1.x branch too because we appear to have broken support for classic remoting using SSL/TLS altogether.


pjfanning:
* test currently fails with `java.net.BindException: Address already in use`
* will need to find out why the ports are clashing or force different ports to be used in the test

raboof (approved):
ouch, looks convincing, great with the targeted integration tests!

PR #1858, opened 2025-05-21
CI: github actions use sha instead of version

pjfanning:
see https://github.com/apache/pekko-management/pull/429


raboof (approved):
not checked, but assuming the infra tooling will take care of that

PR #1868, opened 2025-05-25
feat: Add dropRepeated stream operator.

He-Pin:
Motivation:
Add dropRepeated stream operator.

refs: https://github.com/apache/pekko/issues/1662

### DIFF
--- a/docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
+++ b/docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
@@ -58,16 +58,12 @@ import java.util.*;
 // #zip
 
 // #log
-import org.apache.pekko.event.LogMarker;
-import org.apache.pekko.stream.Attributes;
 
 // #log
 
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
 class SourceOrFlow {
   private static ActorSystem system = null;
@@ -606,6 +602,21 @@ class SourceOrFlow {
     // eiusmod
     // incididunt
     // #filterNot
+  }
+
+  void dropRepeatedExample() {
+    // #dropRepeated
+    Source.from(Arrays.asList(1, 2, 2, 3, 3, 1, 4))
+        .dropRepeated()
+        .runForeach(System.out::println, system);
+    // prints:
+    // 1
+    // 2
+    // 3
+    // 1
+    // 4
+
+    // #dropRepeated
   }
 
   void dropExample() {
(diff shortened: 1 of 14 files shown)


pjfanning:
Is there a reason not to just call this `distinct`? I don't get what the `UntilChanged` bit adds to this - ie it seems unnecessary and potentially confusing.

docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
```diff
@@ -608,6 +608,20 @@ void filterNotExample() {
...
+  void distinctUntilChangedExample() {
+    // #distinctUntilChanged
+    Source.from(Arrays.asList(1, 2, 2, 3, 3, 4))
+        .distinctUntilChanged()
+        .runForeach(System.out::println, system);
+    // prints:
+    // 1
+    // 2
+    // 3
+    // 4
```
raboof, line 620: Can we add another '2' to the example to show what happens when a value shows up that has been seen before, but was not the most recent value?

raboof:
> Is there a reason not to just call this `distinct`? I don't get what the `UntilChanged` bit adds to this - ie it seems unnecessary and potentially confusing.

I think we should try and avoid confusion with https://www.scala-lang.org/api/3.7.0/scala/collection/SeqOps.html#distinct-0

I understand the name follows https://rxjs.dev/api/operators/distinctUntilChanged . I'm not sure it makes sense to me: the results aren't 'distinct' 'until' they are the same? I don't want to bikeshed too much, but something like `dropRepeated' might make more sense? We can mention in the docs that this is called distinctUntilChanged in rx so people can find it.

The docs for 'statefulMap' have this as an example (https://pekko.apache.org/docs/pekko/current/stream/operators/Source-or-Flow/statefulMap.html). There's something to be said for having it as a separate construct, but perhaps the docs should link to each other?

docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
He-Pin: Do you mean `1,2,2,3,3,2,4`?

raboof: Yes, for example!

He-Pin:
@pjfanning @pjfanning And in Kotlin flow, it's `distinctUntilChanged` ,but in `FS2`, it's `filterWithPrevious`.

raboof:
I could live with `changes` or `dropRepeated`. (I don't think we need an equivalent of `changesWith`, people can use `statefulMap` for more advanced use cases)

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/DropRepeated.scala
```diff
@@ -0,0 +1,76 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.impl.fusing
+
+import org.apache.pekko.stream.ActorAttributes.SupervisionStrategy
```
pjfanning, line 20: ideally, this should `import org.apache.pekko` and base the other imports as relative to that

He-Pin:
@pjfanning Thanks

PR #1869, opened 2025-05-25
chore: Fix two stream operator name typos

He-Pin:
Motivation:
Fix typos


pjfanning (changes requested):
I do not see how these code behaviour changes can be described as typos. A typo refers to a misspelling in docs.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
```diff
@@ -1402,7 +1402,7 @@ trait FlowOps[+Out, +Mat] {
     } else {
       new MapAsyncPartitioned(parallelism, orderedOutput = false, partitioner, f)
     }
-    via(graph.withAttributes(DefaultAttributes.mapAsyncPartition and SourceLocation.forLambda(f)))
+    via(graph.withAttributes(DefaultAttributes.mapAsyncPartitionUnordered and SourceLocation.forLambda(f)))
```
pjfanning, line 1405: my main point about these changes is that they could change behaviour - why are the changes better than what we have before?

He-Pin: And the commit message is changed too

pjfanning (approved):
lgtm

PR #1876, opened 2025-05-29
fix: Change aggregateWithBoundary operator in javadsl to use Optional.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1875

Which should use the `Optional` in javadsl.


stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
```diff
@@ -4628,15 +4628,18 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
   def aggregateWithBoundary[Agg, Emit](allocate: java.util.function.Supplier[Agg],
       aggregate: function.Function2[Agg, Out, Pair[Agg, Boolean]],
       harvest: function.Function[Agg, Emit],
-      emitOnTimer: Pair[java.util.function.Predicate[Agg], java.time.Duration]): javadsl.Flow[In, Emit, Mat] =
+      emitOnTimer: Optional[Pair[java.util.function.Predicate[Agg], java.time.Duration]])
```
pjfanning, line 4631: Is there any way that we could try method overloading? Keep the existing aggregateWithBoundary function and add an extra aggregateWithBoundary function that uses `Optional<Pair<???>>`.

He-Pin: Yes, but I think that's unneeded, because,  the api is marked `ApiMayChange`

pjfanning: we could deprecate the old version and move the ApiMayChange to the new version

pjfanning (approved):
lgtm

PR #1534, opened 2024-10-15
[EXPERIMENT] test with java 23

pjfanning:
temporary test run but may be refactored later to run nightly

PR #1887, opened 2025-06-07
[EXPERIMENT] try Java 25-ea

pjfanning:
relates to #1534

### DIFF
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -117,7 +117,7 @@ jobs:
         # full version from it.
         scalaVersion: ["2.12.x", "2.13.x", "3.3.x"]
         # temp remove 8, 11, 17, 21
-        javaVersion: [23]
+        javaVersion: [24]
     env:
       DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
     steps:
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -1,6 +1,7 @@
 name: Nightly Builds
 
 on:
+  pull_request:
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
@@ -115,7 +116,8 @@ jobs:
         # binary version is required and Pekko build will set the right
         # full version from it.
         scalaVersion: ["2.12.x", "2.13.x", "3.3.x"]
-        javaVersion: [8, 11, 17, 21]
+        # temp remove 8, 11, 17, 21
+        javaVersion: [23]
     env:
       DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
     steps:
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Java 11
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: oracle
           java-version: 11
 
       - name: Install sbt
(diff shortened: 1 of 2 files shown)


He-Pin:
Cool, we should make 2.0.0 release jdk25 ready.

mkurz:
Java 25 officially released: https://mail.openjdk.org/pipermail/announce/2025-September/000360.html

EmadAlblueshi:
> Cool, we should make 2.0.0 release jdk25 ready.

That would be great 👍🏻

pjfanning:
The open issues are at https://github.com/apache/pekko/issues?q=state%3Aopen%20label%3Ajava24%2F25

The TLDR is that Java 24 and 25 compilers seem to compile Pekko code but it doesn't work as normal at runtime. If we build Pekko with an older Java compiler (eg Java 17), the compiled classes work at runtime with Java 24 and 25 - as far as I can see.

If anyone has time to investigate why the Java compiler is behaving like this, feel free to take this on - it would be very useful.

He-Pin:
it should be -release 17

pjfanning:
> it should be -release 17

We have flags like that in the build but still we end up with invalid classes that cause pattern matching issues (as far as I recall).

pjfanning:
with Java 25-ea (GitHub CI does seem to have a Temurin distro for Java 25 yet), we get Scala code gen issues (similar issues in pekko-http build).

```
[info] Generating 'ZipLatestWith.scala'
[info] Generating 'UnzipWith.scala'
[info] Generating 'ZipWith.scala'
[info] Generating 'GraphCreate.scala'
[info] Generating 'FanInShapeN.scala'
[info] Generating 'FanOutShapeN.scala'
[info] Generating 'GraphApply.scala'
[info] Generating 'ZipLatestWithApply.scala'
[info] Generating 'UnzipWithApply.scala'
[info] Generating 'ZipWithApply.scala'
[info] Generating 'Tuples.scala'
[info] Generating 'Functions.scala'
[info] compiling 203 Scala sources and 27 Java sources to /home/runner/work/pekko/pekko/actor/target/scala-2.13/classes ...
[error] Error while emitting SameThreadExecutionContext.scala
[error] Unsupported class file major version 69
[error] Error while emitting ScalaBatchable.scala
[error] Unsupported class file major version 69
[error] Error while emitting ByteIterator.scala
[error] assertion failed: ClassBType.info not yet assigned: Ljava/lang/Object;
```

He-Pin:
Should we retry this？

pjfanning:
https://github.com/apache/pekko/issues/1998 still seems to be an issue - 44 tests failed for pekko-remote with Artery in latest run here.
Otherwise, there are some failures in other tests but not as bad as before.

.github/workflows/nightly-builds.yml
```diff
@@ -160,8 +160,7 @@ jobs:
             clean "++ ${{ matrix.scalaVersion }} test" checkTestsHaveRun
 
       - name: Docs
-        # Docs generation requires JDK 11.
-        if: ${{ matrix.javaVersion == 11 }}
```
raboof, line 164: huh ;)

.github/workflows/nightly-builds.yml
```diff
@@ -170,8 +169,7 @@ jobs:
             "++ ${{ matrix.scalaVersion }} doc"
 
       - name: Publish
-        # Publish (osgi bundle) not working with JDK 17, issue #31132
-        if: ${{ matrix.javaVersion == 11 }}
+        if: ${{ matrix.javaVersion == 17 }}
```
raboof, line 172: so this was fixed? I guess we'll see :)

.github/workflows/nightly-builds.yml
pjfanning: docs build ok with Java 17 now - after Paradox fixes

.github/workflows/nightly-builds.yml
pjfanning: if this fails, we'll have to fix it or drop this testing altogether - it is effectively disabled prior to this change because Java 17 is the min version used in the build

Issue #1891, opened 2025-06-07
remove use of Unsafe in AbstractActorCell ?


pjfanning:
sun.misc.Unsafe is deprecated and Java 23+ users will be warned when we use it and generally, the class is strongly discouraged.

https://github.com/apache/pekko/blob/696b210348936901446dddf2ce57ca7f571b7f23/actor/src/main/java/org/apache/pekko/actor/dungeon/AbstractActorCell.java#L19-L46

The Unsafe.instance.objectFieldOffset method doesn't seem to have a replacement.

These offsets in AbstractActorCell are then accessed and manipulated using more methods from sun.misc.Unsafe.
Examples:
* https://github.com/apache/pekko/blob/696b210348936901446dddf2ce57ca7f571b7f23/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala#L41
* https://github.com/apache/pekko/blob/696b210348936901446dddf2ce57ca7f571b7f23/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala#L117-L118
* https://github.com/apache/pekko/blob/696b210348936901446dddf2ce57ca7f571b7f23/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala#L125

Could we consider just writing package private functions in AbstractActorCell to access and change these values referenced by these positions? We could wrap the variables in java.util.concurrent.atomic wrappers if we need to protect atomicity.
Is the performance boost of the Unsafe code really worth it?

He-Pin:
We can do this after 1.2.0 is released,  Netty is migrating too

pjfanning:
this is not for pekko 1.2 - probably 2.0 - but we need to get onto 2.0 work soon

PR #1892, opened 2025-06-07
don't use unsafe to update nextName

pjfanning:
* relates to #1891 
* there are probably other ways to do this
* from my benchmarking in https://github.com/pjfanning/pekko-bench, this change is only a tad slower than the unsafe approach and much easier to maintain - and gets away from the warnings about using sun.misc.Unsafe in one use case

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
@@ -147,8 +147,8 @@ private[pekko] trait Children { this: ActorCell =>
 
   @nowarn private def _preventPrivateUnusedErasure = {
     _childrenRefsDoNotCallMeDirectly
-    _nextNameDoNotCallMeDirectly
     _functionRefsDoNotCallMeDirectly
+    _nextNameDoNotCallMeDirectly
   }
 
   /**


actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
```diff
@@ -120,13 +120,13 @@ private[pekko] trait Children { this: ActorCell =>
     refs.valuesIterator.foreach(_.stop())
   }
 
-  @nowarn @volatile private var _nextNameDoNotCallMeDirectly = 0L
+  private val _nextName = new java.util.concurrent.AtomicLong()
```
He-Pin: better go with AtomicLongFieldUpdater.

pjfanning: I tried AtomicLongFieldUpdater but it doesn't work with private fields.

```
[error] Caused by: java.lang.IllegalAccessException: class example.AbstractActorCell cannot access a member of class example.ActorCell with modifiers "private volatile"
```

I also found in benchmarks that AtomicLong is nearly as fast as the existing code and the code is much simpler to maintain. No reflection, no sun.misc.Unsafe.

https://github.com/pjfanning/pekko-bench

pjfanning: I actually got AtomicLongFieldUpdater to work by moving the code to be in the class that has the private field. The performance in my bench testing is no better than the simpler AtomicLong.

He-Pin:
Track upstream: https://github.com/scala/scala3/issues/9013

He-Pin:
User can use `--sun-misc-unsafe-memory-access=allow` for now

raboof (approved):

actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
He-Pin: AtomicLongFieldUpdater can save memory

actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
```diff
@@ -120,13 +120,13 @@ private[pekko] trait Children { this: ActorCell =>
     refs.valuesIterator.foreach(_.stop())
   }
 
-  @nowarn @volatile private var _nextNameDoNotCallMeDirectly = 0L
+  private val _nextName = new java.util.concurrent.atomic.AtomicLong()
```
He-Pin: We should use the *Updater here, see http://normanmaurer.me/blog/2013/10/28/Lesser-known-concurrent-classes-Part-1/

pjfanning: this article is 12 years old. Java runtimes and their memory allocation have improved a lot since then.

actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
pjfanning: AtomicLongFieldUpdater is slower in my testing

actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
pjfanning: even netty uses AtomicLongs

https://github.com/search?q=repo%3Anetty%2Fnetty%20atomiclong&type=code

It also uses the FieldUpdater in a few places but the Netty team seem to believe that the FieldUpdater doesn't need to be used in every case.

He-Pin: That's not true, it's still valid, you can test it with jol.

pjfanning:
I've tried a few ways to get AtomicLongFieldUpdater to work but so far no look. AtomicLongFieldUpdater does runtime checks and can only access fields in the same class or public variables - but we can't make the long public because then it could be manipulated by any code and it isn't good to have a AtomicLongFieldUpdater declared in the Children trait because you get one instance per ActorCell. I tried using a companion object but in Java that's a separate class and you can't then access a private field in another class.
It's easier in Java because you can declare static and non-static vars in the same class. It tried creating a Java abstract class that does this and have it inherited by Children trait. This causes ActorCell not to compile.

actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
```diff
@@ -147,8 +147,8 @@ private[pekko] trait Children { this: ActorCell =>
 
   @nowarn private def _preventPrivateUnusedErasure = {
     _childrenRefsDoNotCallMeDirectly
-    _functionRefsDoNotCallMeDirectly
     _nextNameDoNotCallMeDirectly
+    _functionRefsDoNotCallMeDirectly
```
mdedetrich: Any reason behind this reordering?

mdedetrich (approved):
lgtm (pending CI pass). Left one additional comment, not critical but it looks a bit sus.

actor/src/main/scala/org/apache/pekko/actor/dungeon/Children.scala
pjfanning: I updated this (to revert the order change)

He-Pin:
Netty recently using varhandle 
https://github.com/netty/netty/pull/15486/

PR #1893, opened 2025-06-07
use varhandle for nextName

pjfanning:
equivalent of #1892 but needs us to abandon Java 8


He-Pin:
Yes, At work, we have moved to Java 21

pjfanning:
I did a benchmark on this and it seems to be much slower than using Unsafe (existing code) or AtomicLong.

PR #1899, opened 2025-06-13
tidy up metrics exchange

pjfanning:
* small adjustments in pekko-cluster-metrics code
* turns out to not be enough as some number classes serialize in formats that include refs to byte arrays and/or java.math classes (RoundingMode and MathContext) - #1912 is for this

### DIFF
--- a/cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
+++ b/cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
@@ -23,7 +23,7 @@ import org.apache.pekko
 import pekko.annotation.InternalApi
 
 /**
- * A special ObjectInputStream that loads will only load built-in primitives or
+ * A special ObjectInputStream that will only load built-in primitives or
  * Number classes.
  * <p>
  * This is for internal Pekko use only, and is not intended for public use.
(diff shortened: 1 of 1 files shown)


cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
```diff
@@ -0,0 +1,56 @@
...
+  InputStream,
+  ObjectInputStream,
+  ObjectStreamClass
+}
+
+/**
+ * A special ObjectInputStream that loads will only load built-in primitives or
+ * Number classes.
+ * <p>
+ * This is for internal Pekko use only, and is not intended for public use.
```
raboof, line 29: Maybe mark it `@InternalApi`?

raboof (approved):
neat!

cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
```diff
@@ -0,0 +1,57 @@
...
+ * </p>
+ */
+@InternalApi
+private[protobuf] class NumberInputStream(inputStream: InputStream) extends ObjectInputStream(inputStream) {
+
+  private val Loader = classOf[NumberInputStream].getClassLoader()
```
He-Pin: Why we can't use the `system.dynamicAccess.classLoader` here?

pjfanning: I've changed the code to use the  system.dynamicAccess.classLoader

danischroeter (approved):
lgtm - nice

PR #1901, opened 2025-06-13
try making TLS v1.3 the default

pjfanning:
possibly for v1.2 but may need to be delayed till v2.0

### DIFF
--- a/docs/src/main/paradox/remoting-artery.md
+++ b/docs/src/main/paradox/remoting-artery.md
@@ -292,7 +292,7 @@ pekko.remote.artery {
     key-password = ${SSL_KEY_PASSWORD}
     trust-store-password = ${SSL_TRUST_STORE_PASSWORD}
 
-    protocol = "TLSv1.2"
+    protocol = "TLSv1.3"
 
     enabled-algorithms = [TLS_DHE_RSA_WITH_AES_128_GCM_SHA256]
   }
--- a/docs/src/main/paradox/remoting.md
+++ b/docs/src/main/paradox/remoting.md
@@ -467,7 +467,7 @@ pekko {
         key-password = ${SSL_KEY_PASSWORD}
         trust-store-password = ${SSL_TRUST_STORE_PASSWORD}
 
-        protocol = "TLSv1.2"
+        protocol = "TLSv1.3"
 
         enabled-algorithms = [TLS_DHE_RSA_WITH_AES_128_GCM_SHA256]
       }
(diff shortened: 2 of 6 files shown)


pjfanning:
Seeing
```
Caused by: javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate)
	at sun.security.ssl.HandshakeContext.<init>(HandshakeContext.java:171)
	at sun.security.ssl.ClientHandshakeContext.<init>(ClientHandshakeContext.java:103)
	at sun.security.ssl.TransportContext.kickstart(TransportContext.java:227)
	at sun.security.ssl.SSLEngineImpl.beginHandshake(SSLEngineImpl.java:97)
	at org.apache.pekko.stream.impl.io.TLSActor.<init>(TLSActor.scala:176)
	at org.apache.pekko.stream.impl.io.TLSActor$.$anonfun$props$1(TLSActor.scala:49)
```

This issue also happens when we test with Java 24. I think we'll need to review the cipher suites that we use.

pjfanning:
@He-Pin @raboof I still have 1 test case to fix, so this isn't quite ready.
I still wonder if it's about time for us to encourage people off TLS 1.2 and onto 1.3. The biggest diff is that the cipher suites for 1.3 are bit different.

One of the reasons that I started looking at this is that with Java 24 and 25, we get a lot of test failures related to the TLSActor. I'm not sure whether Java 24/25 have changed their TLS defaults. Some of the test failures with Java 24 and 25 are also related to TLS Renegotiation which is definitely one feature that was removed in TLS 1.3.

raboof:
> I still wonder if it's about time for us to encourage people off TLS 1.2 and onto 1.3

This makes sense to me. For example https://csrc.nist.gov/pubs/sp/800/52/r2/final requires it to be 'supported', so it seems sensible to also default to it.

raboof (approved):

mkurz:
Why did you not switch to TLSv1.3 here:
https://github.com/apache/pekko/blob/1a8decd9a30ced8ba6daafbcb0a873f9a78c419f/stream/src/main/resources/reference.conf#L183-L186

?

pjfanning:
@mkurz that was probably an oversight - I created #2358 and I'll see how that gets on but it may need test changes

I haven't yet changed pekko-management because that attempt ran into issues but I did make it configurable there. https://github.com/apache/pekko-management/pull/456

Issue #1911, opened 2025-06-19
Scala 2.12 nightly tests: MessageSerializerSpec issue with BigInteger


pjfanning:
* may be related to #1899 
* does not happen with newer Java versions it seems

```
[06-19 01:12:04.847] [info] ClusterMessages
[WARN] [06/19/2025 01:12:05.013] [MessageSerializerSpec-pekko.actor.default-dispatcher-4] [pekko://MessageSerializerSpec/system/cluster/core/daemon] unhandled message from Actor[pekko://MessageSerializerSpec/deadLetters]: Leave(pekko://MessageSerializerSpec@10.1.0.4:38107)
[06-19 01:12:05.024] [info] - must be serializable *** FAILED *** (137 milliseconds)
[06-19 01:12:05.025] [info]   java.io.StreamCorruptedException: invalid type code: 7F
[06-19 01:12:05.025] [info]   at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1793)
[06-19 01:12:05.025] [info]   at java.base/java.io.ObjectInputStream.readArray(ObjectInputStream.java:2161)
[06-19 01:12:05.026] [info]   at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1750)
[06-19 01:12:05.026] [info]   at java.base/java.io.ObjectInputStream$FieldValues.<init>(ObjectInputStream.java:2618)
[06-19 01:12:05.026] [info]   at java.base/java.io.ObjectInputStream.readFields(ObjectInputStream.java:722)
[06-19 01:12:05.026] [info]   at java.base/java.math.BigInteger.readObject(BigInteger.java:4839)
[06-19 01:12:05.026] [info]   at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
[06-19 01:12:05.027] [info]   at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[06-19 01:12:05.027] [info]   at java.base/java.io.ObjectStreamClass.invokeReadObject(ObjectStreamClass.java:1102)
[06-19 01:12:05.028] [info]   at java.base/java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2444)
[06-19 01:12:05.028] [info]   at java.base/java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2284)
[...]

PR #1912, opened 2025-06-19
try to fix broken cluster metrics test

pjfanning:
* see #1911 
* reworking the newly added NumberInputStream to allow arrays of primitives (eg arrays of bytes) fixes the issue - that oddly only occurs with Scala 2.12 runtime

### DIFF
--- a/cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
+++ b/cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
@@ -17,10 +17,11 @@
 
 package org.apache.pekko.cluster.metrics.protobuf
 
-import java.io.{ InputStream, ObjectInputStream, ObjectStreamClass }
+import java.io.{ InputStream, ObjectStreamClass }
 
 import org.apache.pekko
 import pekko.annotation.InternalApi
+import pekko.util.ClassLoaderObjectInputStream
 
 /**
  * A special ObjectInputStream that will only load built-in primitives or
@@ -32,7 +33,7 @@ import pekko.annotation.InternalApi
 @InternalApi
 private[protobuf] class NumberInputStream(
     classLoader: ClassLoader,
-    inputStream: InputStream) extends ObjectInputStream(inputStream) {
+    inputStream: InputStream) extends ClassLoaderObjectInputStream(classLoader, inputStream) {
 
   /**
    * Resolve a class specified by the descriptor using the provided classloader
@@ -44,7 +45,7 @@ private[protobuf] class NumberInputStream(
    * @throws ClassNotFoundException if the Class cannot be found (or is rejected)
    */
   override protected def resolveClass(objectStreamClass: ObjectStreamClass): Class[_] = {
-    val clazz = Class.forName(objectStreamClass.getName(), false, classLoader)
+    val clazz = super.resolveClass(objectStreamClass)
     if (clazz.isPrimitive() || classOf[Number].isAssignableFrom(clazz)) {
       clazz
     } else {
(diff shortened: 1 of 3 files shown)


cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
```diff
@@ -32,20 +33,21 @@ import pekko.annotation.InternalApi
...
    * @param objectStreamClass  descriptor of the class
    * @return the Class object described by the ObjectStreamClass
    * @throws ClassNotFoundException if the Class cannot be found (or is rejected)
    */
   override protected def resolveClass(objectStreamClass: ObjectStreamClass): Class[_] = {
-    val clazz = Class.forName(objectStreamClass.getName(), false, classLoader)
-    if (clazz.isPrimitive() || classOf[Number].isAssignableFrom(clazz)) {
+    val clazz = super.resolveClass(objectStreamClass)
+    if (clazz.isPrimitive() || (clazz.isArray() && clazz.getComponentType.isPrimitive) ||
+      classOf[Number].isAssignableFrom(clazz)) {
```
He-Pin: 1. we should fix the error messages too, at line 54
2. add a test for BigDecimal and BigInteger cases?

pjfanning: Good points. I added extra tests and found another issue. Now fixed.

pjfanning: @He-Pin would you be able to review this again if you have time?

He-Pin (approved):
lgtm

cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
```diff
@@ -32,24 +33,28 @@ import pekko.annotation.InternalApi
...
    * @param objectStreamClass  descriptor of the class
    * @return the Class object described by the ObjectStreamClass
    * @throws ClassNotFoundException if the Class cannot be found (or is rejected)
    */
   override protected def resolveClass(objectStreamClass: ObjectStreamClass): Class[_] = {
-    val clazz = Class.forName(objectStreamClass.getName(), false, classLoader)
-    if (clazz.isPrimitive() || classOf[Number].isAssignableFrom(clazz)) {
+    val clazz = super.resolveClass(objectStreamClass)
+    if (clazz.isPrimitive() || (clazz.isArray() && clazz.getComponentType.isPrimitive) ||
+      classOf[Number].isAssignableFrom(clazz) || clazz.getPackage.getName == "java.math") {
```
He-Pin, line 51: I think we should just convert the Scala's BigInt and Java's BigInteger and bigdecimal to primitive value, not sure why Scala's BigInt and BigDecimal is not a problem

cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/protobuf/NumberInputStream.scala
```diff
@@ -32,24 +33,28 @@ import pekko.annotation.InternalApi
...
       clazz
     } else {
       throw new ClassNotFoundException(
-        s"Class rejected: ${objectStreamClass.getName()} (only primitive types and subclasses of java.lang.Number are allowed)")
+        s"Class rejected: ${objectStreamClass.getName()} " +
+        "(only primitive types, arrays of primitive types, subclasses of java.lang.Number " +
```
He-Pin, line 56: How about the Scala number (ScalaNumber), @pjfanning

pjfanning: I am not familiar with that class. I can't find it in the Scala docs.

He-Pin: `scala.math.ScalaNumber` 
image


I see, it extends the java.lang.numbers.

He-Pin: And it's been used in `Metric#convertNumber`

PR #1914, opened 2025-06-20
try fieldupdater

pjfanning:


mdedetrich:
@pjfanning Should we close this since we know that `VarHandle`'s work (and its slightly faster)?

PR #1916, opened 2025-06-22
chore: bump scalafmt to 3.9.7

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1906

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version                                  = 3.9.3
+version                                  = 3.9.7
 runner.dialect                           = scala213
 project.git                              = true
 style                                    = defaultWithAlign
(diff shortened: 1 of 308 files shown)


.git-blame-ignore-revs
```diff
@@ -79,3 +79,6 @@ c44c0b7cbdab11d85176cfe062288fdcba16c56a
 
 # Scala Steward: Reformat with scalafmt 3.8.4
 f48453e60a8d11f74acc7a10d3a9da90b8acd4af
+
+# Reformat with scalafmt 3.9.7
+dca3609b79c2c2e7dcef82e112fd8636f5e90466
```
raboof: I don't think this will work, since we have 'create a merge commit' disabled for this repo

He-Pin: So I have do merge this first and then add a new commit?

He-Pin: Will add a new commit to fix this, dropped this commit.

raboof: I think so

.git-blame-ignore-revs
```diff
@@ -79,3 +79,6 @@ c44c0b7cbdab11d85176cfe062288fdcba16c56a
 
 # Scala Steward: Reformat with scalafmt 3.8.4
 f48453e60a8d11f74acc7a10d3a9da90b8acd4af
+
+# Reformat with scalafmt 3.9.7
+b152e056b5fd22e25283e48a7a96e6e6c8caf884
```
raboof: this is still here?

He-Pin: Thanks, fixed now.

PR #1917, opened 2025-06-22
chore: Set vector builder to null after stage completed to avoid leak.

He-Pin:
Motivation:
Tweak memory usage same as : https://github.com/apache/pekko/pull/1623


pjfanning (approved):
lgtm

PR #1919, opened 2025-06-23
Fix typed persistence stack overflow with many read only commands

Tomassino-ibm:
relates to #1327

### DIFF
--- a/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
+++ b/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
@@ -55,10 +55,11 @@ object EventSourcedStashOverflowSpec {
     SteppingInmemJournal.config("EventSourcedStashOverflow").withFallback(ConfigFactory.parseString(s"""
        pekko.persistence {
          typed {
-           stash-capacity = 1000 # enough to fail on stack size
+           stash-capacity = 2000 # enough to fail on stack size
            stash-overflow-strategy = "drop"
          }
        }
+       pekko.jvm-exit-on-fatal-error = off
    """))
 }
 
--- a/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
+++ b/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
@@ -57,7 +57,7 @@ object EventSourcedStashOverflowSpec {
          typed {
            stash-capacity = 20000 # enough to fail on stack size
            stash-overflow-strategy = "drop"
-           break-recursive-calls-when-unstashing-read-only-commands = true # Necessary to enable the fix
+           recurse-when-unstashing-read-only-commands = false
          }
        }
        pekko.jvm-exit-on-fatal-error = off
--- a/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
+++ b/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
@@ -55,8 +55,9 @@ object EventSourcedStashOverflowSpec {
     SteppingInmemJournal.config("EventSourcedStashOverflow").withFallback(ConfigFactory.parseString(s"""
        pekko.persistence {
          typed {
-           stash-capacity = 2000 # enough to fail on stack size
+           stash-capacity = 20000 # enough to fail on stack size
            stash-overflow-strategy = "drop"
+           break-recursive-calls-when-unstashing-read-only-commands = true # Necessary to enable the fix
          }
        }
        pekko.jvm-exit-on-fatal-error = off
@@ -86,8 +87,8 @@ class EventSourcedStashOverflowSpec
       for (_ <- 0 to (stashCapacity * 2)) {
         es.tell(EventSourcedStringList.DoNothing(probe.ref))
       }
-      // capacity + 1 should mean that we get a dropped last message when all stash is filled
-      // while the actor is stuck in replay because journal isn't responding
+      // capacity * 2 should mean that we get many dropped messages when all stash is filled
+      // while the actor is stuck in replay because journal isn't responding (checking only one)
       droppedMessageProbe.receiveMessage()
       implicit val classicSystem: pekko.actor.ActorSystem =
         testKit.system.toClassic
--- a/persistence-typed/src/main/mima-filters/1.1.x.backwards.excludes/eventsourcedbehavior-stackoverflow.excludes
+++ b/persistence-typed/src/main/mima-filters/1.1.x.backwards.excludes/eventsourcedbehavior-stackoverflow.excludes
@@ -1,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# EventSourcedBehavior/DurableState Read-Only Events Stackoverflow
+ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.persistence.typed.internal.Running#*")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.persistence.typed.internal.Running#*")
--- a/persistence-typed/src/main/mima-filters/1.1.x.backwards.excludes/eventsourcedbehavior-stackoverflow.excludes
+++ b/persistence-typed/src/main/mima-filters/1.1.x.backwards.excludes/eventsourcedbehavior-stackoverflow.excludes
@@ -18,3 +18,5 @@
 # EventSourcedBehavior/DurableState Read-Only Events Stackoverflow
 ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.persistence.typed.internal.Running#*")
 ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.persistence.typed.internal.Running#*")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.persistence.typed.state.internal.Running#*")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.persistence.typed.state.internal.Running#*")
--- a/persistence-typed/src/main/resources/reference.conf
+++ b/persistence-typed/src/main/resources/reference.conf
@@ -47,6 +47,12 @@ pekko.persistence.typed {
   # this can be changed by setting this to 'true' in which case the internal logging is sent to
   # the actor context logger.
   use-context-logger-for-internal-logging = false
+
+  # Commands get stashed during persistence and un-stashed afterward, causing functions to be called
+  # recursively if commands are read-only (i.e. do not cause any persistence). This might cause a stack
+  # overflow in case there are many messages to unstash. Setting this to true enables a fix that turns
+  # recursion into a plain cycle, thus preventing stack overflows
+  break-recursive-calls-when-unstashing-read-only-commands = false
 }
 
 pekko.reliable-delivery {
--- a/persistence-typed/src/main/resources/reference.conf
+++ b/persistence-typed/src/main/resources/reference.conf
@@ -48,11 +48,12 @@ pekko.persistence.typed {
   # the actor context logger.
   use-context-logger-for-internal-logging = false
 
-  # Commands get stashed during persistence and un-stashed afterward, causing functions to be called
-  # recursively if commands are read-only (i.e. do not cause any persistence). This might cause a stack
-  # overflow in case there are many messages to unstash. Setting this to true enables a fix that turns
-  # recursion into a plain cycle, thus preventing stack overflows
-  break-recursive-calls-when-unstashing-read-only-commands = false
+  # Commands get stashed during persistence and un-stashed afterward. Normally, when many read-only
+  # (i.e. that do not cause any persistence) commands are present in the stash, the functions are called
+  # in a loop to unstash all of them. Setting this to true, will cause functions to be called
+  # recursively (that was the default before this setting was introduced). That might cause a stack
+  # overflow in case there are many messages to unstash.
+  recurse-when-unstashing-read-only-commands = false
 }
 
 pekko.reliable-delivery {
--- a/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
+++ b/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
@@ -264,10 +264,10 @@ private[pekko] object Running {
 
     def onMessage(msg: InternalProtocol): Behavior[InternalProtocol] = msg match {
       case IncomingCommand(c: C @unchecked) =>
-        // The configuration flag is only checked here: if it is false, we don't set
+        // The configuration flag is only checked here: if it is true, we don't set
         // state.unstashRecurrenceState.recOnCommandParams and onCommand will behave as if
         // the fix was not implemented at all
-        if (setup.settings.breakRecursiveCallsWhenUnstashingReadOnlyCommands && state.unstashRecurrenceState.inOnCommandCall) {
+        if (!setup.settings.recurseWhenUnstashingReadOnlyCommands && state.unstashRecurrenceState.inOnCommandCall) {
           state.unstashRecurrenceState.recOnCommandParams = Some((state, c))
           this // This will be ignored in onCommand
         } else {
--- a/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
+++ b/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
@@ -97,21 +97,35 @@ private[pekko] object Running {
     def currentSequenceNumber: Long
   }
 
-  final case class RunningState[State](
+  // This is part of the fix for https://github.com/apache/pekko/issues/1327 and it's necessary to break
+  // recursion between the `onCommand` and `onMessage` functions in Running[C, E, S]#HandlingCommands. We
+  // set the flag `inOnCommandCall` to true in `onCommand` before calling functions that will in turn call
+  // `onMessage`. In `onMessage`, if the flag is set and we would recursively call `onCommand`, we instead
+  // save the parameters we would use to call `onCommand` and return. When back in `onCommand`, we check if
+  // `recOnCommandParams` is not empty and, if not, act as if `onMessage` called us directly. In function
+  // `onCommand` we have a while loop replacing recursive calls so that we don't use all the stack space in
+  // case a lot of read-only message are stashed
+  final class UnstashRecurrenceState[State, Command] {
+    var inOnCommandCall: Boolean = false
+    var recOnCommandParams: Option[(RunningState[State, Command], Command)] = None
+  }
+
+  final case class RunningState[State, Command](
       seqNr: Long,
       state: State,
       receivedPoisonPill: Boolean,
       version: VersionVector,
       seenPerReplica: Map[ReplicaId, Long],
-      replicationControl: Map[ReplicaId, ReplicationStreamControl]) {
+      replicationControl: Map[ReplicaId, ReplicationStreamControl],
+      unstashRecurrenceState: UnstashRecurrenceState[State, Command] = new UnstashRecurrenceState[State, Command]) {
 
-    def nextSequenceNr(): RunningState[State] =
+    def nextSequenceNr(): RunningState[State, Command] =
       copy(seqNr = seqNr + 1)
 
-    def updateLastSequenceNr(persistent: PersistentRepr): RunningState[State] =
+    def updateLastSequenceNr(persistent: PersistentRepr): RunningState[State, Command] =
       if (persistent.sequenceNr > seqNr) copy(seqNr = persistent.sequenceNr) else this
 
-    def applyEvent[C, E](setup: BehaviorSetup[C, E, State], event: E): RunningState[State] = {
+    def applyEvent[E](setup: BehaviorSetup[Command, E, State], event: E): RunningState[State, Command] = {
       val updated = setup.eventHandler(state, event)
       copy(state = updated)
     }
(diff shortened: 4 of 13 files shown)


persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
```diff
@@ -55,10 +55,11 @@ object EventSourcedStashOverflowSpec {
     SteppingInmemJournal.config("EventSourcedStashOverflow").withFallback(ConfigFactory.parseString(s"""
        pekko.persistence {
          typed {
-           stash-capacity = 1000 # enough to fail on stack size
+           stash-capacity = 20000 # enough to fail on stack size
```
pjfanning, line 58: 2000 should be enough to reproduce the issue - if 20000 doesn't cause the test to run for too long then this is ok
I would regard 10 seconds to be a long time for a unit test

pjfanning:
* thanks @Tomassino-ibm for the PR and I know this is still in progress
* no hurry but could you run scalafmt on your code at some point because that check is failing for
`persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala`

pjfanning:
> * thanks @Tomassino-ibm for the PR and I know this is still in progress
> * no hurry but could you run scalafmt on your code at some point because that check is failing for
>   `persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala`

I pushed fdfa3f5f8d545d72ea480627609fafbdc27b6043 to your branch to fix the scalafmt and mima binary compat checks

persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
ghost: I increased that number because in one attempt the test was no longer failing (I had reduced the amount of functions in the "recursion cycle" and so the stack overflow was no longer happening). That example runs in approximately 1.5 seconds on my machine, but I can change it to 2000 if you want, that was definitely enough to trigger the failure in the original code

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
```diff
@@ -97,30 +97,44 @@ private[pekko] object Running {
...
-  final case class RunningState[State](
+  // This is part of the fix for https://github.com/apache/pekko/issues/1327 and it's necessary to break
+  // recursion between the `onCommand` and `onMessage` functions in Running[C, E, S]#HandlingCommands. We
+  // set the flag `inOnCommandCall` to true in `onCommand` before calling functions that will in turn call
+  // `onMessage`. In `onMessage`, if the flag is set and we would recursively call `onCommand`, we instead
+  // save the parameters we would use to call `onCommand` and return. When back in `onCommand`, we check if
+  // `recOnCommandParams` is not empty and, if not, act as if `onMessage` called us directly. In function
+  // `onCommand` we have a while loop replacing recursive calls so that we don't use all the stack space in
+  // case a lot of read-only message are stashed
+  final class UnstashRecurrenceState[State, Command] {
```
ghost, line 108: Naming is always hard, so please let me know if you have a suggestion for a better name for this class :-)

ghost:
> I pushed [fdfa3f5](https://github.com/apache/pekko/commit/fdfa3f5f8d545d72ea480627609fafbdc27b6043) to your branch to fix the scalafmt and mima binary compat checks

Hi @pjfanning, thanks a lot for the early review and the fixes :-). I don't know if this solution is acceptable, maybe it would have been possible to have a better and more elegant implementation by touching more parts of the code base. However, considering that I know very little of pekko internals, I tried to limit changes to the smallest possible portion of code.

I haven't marked the pull request as ready yet because I have a few questions (in addition to the other comments above):
- as you've probably noticed, the message of the second commit starts with "TMP AMEND" because I was planning to amend it instead of pushing more commits for other changes. I would still like to change it, but that would force you to reset your local branch. Is that ok for you?
- regarding binary compatibility, I think we can reduce breaking changes if I remove the new `Command` type parameter of `RunningState`. That was added to avoid a type cast in `onCommand` (we would need to store the command as `Any` in `recOnCommandParams`), but it's not critical to the solution. Please let me know if I should do that.
- reading again the initial [issue description](https://github.com/apache/pekko/issues/1327), I see that the same problem could be present in `DurableState` also. Should I also attempt to fix it? Is there a test to reproduce the issue there already? I've never used it, but I had a quick look at pekko code and it seems that porting the solution there is relatively easy.

Of course, please let me know if there's anything else I should change in code or if I should try a different approach

persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedStashOverflowSpec.scala
pjfanning: 1.5s should be ok

pjfanning:
> > I pushed [fdfa3f5](https://github.com/apache/pekko/commit/fdfa3f5f8d545d72ea480627609fafbdc27b6043) to your branch to fix the scalafmt and mima binary compat checks
> 
> Hi @pjfanning, thanks a lot for the early review and the fixes :-). I don't know if this solution is acceptable, maybe it would have been possible to have a better and more elegant implementation by touching more parts of the code base. However, considering that I know very little of pekko internals, I tried to limit changes to the smallest possible portion of code.
> 
> I haven't marked the pull request as ready yet because I have a few questions (in addition to the other comments above):
> 
> * as you've probably noticed, the message of the second commit starts with "TMP AMEND" because I was planning to amend it instead of pushing more commits for other changes. I would still like to change it, but that would force you to reset your local branch. Is that ok for you?
> * regarding binary compatibility, I think we can reduce breaking changes if I remove the new `Command` type parameter of `RunningState`. That was added to avoid a type cast in `onCommand` (we would need to store the command as `Any` in `recOnCommandParams`), but it's not critical to the solution. Please let me know if I should do that.
> * reading again the initial [issue description](https://github.com/apache/pekko/issues/1327), I see that the same problem could be present in `DurableState` also. Should I also attempt to fix it? Is there a test to reproduce the issue there already? I've never used it, but I had a quick look at pekko code and it seems that porting the solution there is relatively easy.
> 
> Of course also please let me know if there's anything else I should change in code or if I should try a different approach

feel free to make changes to the PR branch including squashing or removing commits

pjfanning:
@Tomassino-ibm if we are worried that this change might cause unforeseen issues in production, one option is to add a config that allows this to be enabled/disabled. I'm not sure this is required but it might be worth considering.

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
```diff
@@ -268,15 +288,43 @@ private[pekko] object Running {
...
+        val effect = setup.commandHandler(rs.state, c)
+
+        applyEffects(c, rs, effect.asInstanceOf[EffectImpl[E, S]]) // TODO can we avoid the cast?
+      }
+
+      var applyEffectsRetval: (Behavior[InternalProtocol], Boolean) = callApplyEffects(state, cmd)
+
+      var retVal: Option[Behavior[InternalProtocol]] = None
+
+      while (retVal.isEmpty) {
```
pjfanning: * is there any reason to worry about this loop spinning? - continuously trying unstash and getting empty returns
* would it be worth considering some sort of backoff after we have tried unstash a few times and received empty values in each case?

ghost:
> @Tomassino-ibm if we are worried that this change might cause unforeseen issues in production, one option is to add a config that allows this to be enabled/disabled. I'm not sure this is required but it might be worth considering.

That's a good idea, I'll try to implement it

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
ghost: I'm not sure I understand the question: this loop basically replaces the recursion, so if the problem you expect is to keep a thread busy for too long, then that was probably already happening before (with recursion). The `retVal` variable is the final behavior returned by `onCommand`, it is an option so that it can be used as the exit condition of the loop: it is only set to `Some` if either `doUnstash` is false or the call to `tryUnstashOne` is not recursive (i.e. `state.unstashRecurrenceState` is not set in `onMessage`)

persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/state/scaladsl/DurableStateStashOverflowSpec.scala
```diff
@@ -0,0 +1,99 @@
+/*
```
pjfanning: this is the correct header for new source files:
https://github.com/apache/pekko/blob/main/project/AddMetaInfLicenseFiles.scala#L1-L16

Unless this code is copied from somewhere else and is not necessarily new

ghost: Thanks, I'll change it. I copied the stash overflow test to try to reproduce it with `DurableStateBehavior`

ghost: Mmm, or maybe I must not change the header... What should I do?

pjfanning: ok - if this is a copy of another file - then the headers are ok as is

ghost:
Hi, I pushed some new commits: I added a feature flag to enable the fix and also ported the fix to DurableStateBehavior. I have, however, two questions.

The first one is: when I added the configuration parameter to DurableStateSettings, I noticed that they are the same (I mean, also same config path) as the ones of EventSourcedSettings. Is this expected?

The other one might be important, but I need help from somebody that is more familiar than me with pekko internals. The fix I added breaks the recursion by using a while loop in `onCommand` and storing recursive `onCommand` call parameters during the call to `onMessage`, which are then used when the control gets back to `onCommand`. There is, however, one function in the recursion chain, that calls another function _after_ the call leading to the recursion: `org.apache.pekko.actor.typed.Behavior$.interpret` calls `start` on the beavior returned by `onMessage`. Since it's not clear to me what `start` actually does and what "starting a behavior" means, my fear is that the recursive and non-recursive versions might do different things, at least in some edge cases. Can this be a problem?

persistence-typed/src/main/mima-filters/1.1.x.backwards.excludes/eventsourcedbehavior-stackoverflow.excludes
```diff
@@ -0,0 +1,22 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# EventSourcedBehavior/DurableState Read-Only Events Stackoverflow
+ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.persistence.typed.internal.Running#*")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.persistence.typed.internal.Running#*")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("org.apache.pekko.persistence.typed.state.internal.Running#*")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.persistence.typed.state.internal.Running#*")
```
ghost, line 22: Let me know if adding these two filters regarding binary compatibility is acceptable or not

pjfanning: internal classes - can be changed, especially as the target is 1.2.0 as opposed to a patch release

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
mdedetrich: I am looking at this briefly and maybe it makes sense to write this code using recursion style with the `@tailrec` annotation (which confirms at compile time whether the scala compiler can unroll the recursion into a loop)?

persistence-typed/src/main/resources/reference.conf
```diff
@@ -47,6 +47,12 @@ pekko.persistence.typed {
   # this can be changed by setting this to 'true' in which case the internal logging is sent to
   # the actor context logger.
   use-context-logger-for-internal-logging = false
+
+  # Commands get stashed during persistence and un-stashed afterward, causing functions to be called
+  # recursively if commands are read-only (i.e. do not cause any persistence). This might cause a stack
+  # overflow in case there are many messages to unstash. Setting this to true enables a fix that turns
+  # recursion into a plain cycle, thus preventing stack overflows
+  break-recursive-calls-when-unstashing-read-only-commands = false
```
pjfanning: I'd prefer to default this to true. We would have to have a reasonable level of confidence but if the unit tests pass. It is useful to have the config in case someone finds an issue but in a 1.2.0 release, we can change the behaviour in Pekko if it fixes issues.

pjfanning: can we call this `recurse-when-unstashing-read-only-commands` - opposite meaning to your flag (recurse=true is the old behaviour)? - I'd like it to default to false so that the new non-recursive code is the default

ghost: Ok, I'll change the name and the default behavior

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
ghost: Ok, I'll try and push a new commit

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/state/internal/Running.scala
```diff
@@ -100,11 +122,37 @@ private[pekko] object Running {
...
+        val effect = setup.commandHandler(rs.state, c)
+
+        applyEffects(c, rs, effect.asInstanceOf[EffectImpl[S]]) // TODO can we avoid the cast?
+      }
+
+      var applyEffectsRetval: (Behavior[InternalProtocol], Boolean) = callApplyEffects(state, cmd)
+
+      var retVal: Option[Behavior[InternalProtocol]] = None
+
+      while (retVal.isEmpty) {
```
ghost: I'll also try to change this from a loop to a tail recursive function

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
ghost: Changed

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/state/internal/Running.scala
ghost: Also changed here

pjfanning:
> Hi, I pushed some new commits: I added a feature flag to enable the fix and also ported the fix to DurableStateBehavior. I have, however, two questions.
> 
> The first one is: when I added the configuration parameter to DurableStateSettings, I noticed that they are the same (I mean, also same config path) as the ones of EventSourcedSettings. Is this expected?
> 
> The other one might be important, but I need help from somebody that is more familiar than me with pekko internals. The fix I added breaks the recursion by using a while loop in `onCommand` and storing recursive `onCommand` call parameters during the call to `onMessage`, which are then used when the control gets back to `onCommand`. There is, however, one function in the recursion chain, that calls another function _after_ the call leading to the recursion: `org.apache.pekko.actor.typed.Behavior$.interpret` calls `start` on the beavior returned by `onMessage`. Since it's not clear to me what `start` actually does and what "starting a behavior" means, my fear is that the recursive and non-recursive versions might do different things, at least in some edge cases. Can this be a problem?

* Regarding the configs, they do look a bit strange to me, sharing paths where you would expect them to have their own paths. Changing this is a breaking change and would be better discussed as a standalone issue. We don't seem to get too many complaints so the issue doesn't seem to be affecting people too much in real world scenarios.
* reordering the code to allow tailrec to work: this does come with risks. All we really have is our test coverage and if that passes then maybe we should be happy enough. But yes, the test coverage could be better. We have a config that controls whether the new code is applied. I wonder if the code path for when the config is set to keep the Pekko 1.0/1.1 (and Akka 2.6) behaviour, that we keep that code free of the tailrec refactor - ie we only try to make the new code path tail recursive and the old code path remains as similar as possible to the code that exists in the main branch prior to this PR.

ghost:
> Regarding the configs, they do look a bit strange to me, sharing paths where you would expect them to have their own paths. Changing this is a breaking change and would be better discussed as a standalone issue. We don't seem to get too many complaints so the issue doesn't seem to be affecting people too much in real world scenarios.

Ok, I agree we can leave it like that in this pull request

> reordering the code to allow tailrec to work: this does come with risks. All we really have is our test coverage and if that passes then maybe we should be happy enough. But yes, the test coverage could be better. We have a config that controls whether the new code is applied. I wonder if the code path for when the config is set to keep the Pekko 1.0/1.1 (and Akka 2.6) behaviour, that we keep that code free of the tailrec refactor - ie we only try to make the new code path tail recursive and the old code path remains as similar as possible to the code that exists in the main branch prior to this PR.

I have to admit that the code looks quite different anyway, so it is not evident that the "original" code path is maintained unchanged. I can try to make this more explicit, I'll try to push a new commit tomorrow

pjfanning:
@Tomassino-ibm I would like the code that tries to replicate the existing behaviour (before this PR) to be as similar as possible to the existing code and the refactors to support the new path only apply to the new path. This doesn't need be 100% - we can discuss having some of the refactor on the old path where it is hard to avoid or causes excessive code duplication.

mdedetrich:
> Regarding the configs, they do look a bit strange to me, sharing paths where you would expect them to have their own paths. Changing this is a breaking change and would be better discussed as a standalone issue. We don't seem to get too many complaints so the issue doesn't seem to be affecting people too much in real world scenarios.

I would avoid messing with configs in a backwards breaking way if possible. While its still technically allowed under SemVer as its a behavioural change (at least if this is released for 1.2.x) it can still be surprising for end users. I agree this should be discussed and done separately

> reordering the code to allow tailrec to work: this does come with risks. All we really have is our test coverage and if that passes then maybe we should be happy enough. But yes, the test coverage could be better. We have a config that controls whether the new code is applied. I wonder if the code path for when the config is set to keep the Pekko 1.0/1.1 (and Akka 2.6) behaviour, that we keep that code free of the tailrec refactor - ie we only try to make the new code path tail recursive and the old code path remains as similar as possible to the code that exists in the main branch prior to this PR.

> @Tomassino-ibm I would like the code that tries to replicate the existing behaviour (before this PR) to be as similar as possible to the existing code and the refactors to support the new path only apply to the new path. This doesn't need be 100% - we can discuss having some of the refactor on the old path where it is hard to avoid or causes excessive code duplication.

Regarding the tailrec, the intention of writing it as tailrec was actually to maintain the original style so its easier to read and hence make sure that its behaving more correctly. While its true that the current implementation doesn't have `@tailrec`, thats because the current implementation is so simple that there aren't any loops to begin with.

I just compared the tailrec version to the older non-tailrec version and the tailrec version is far easier simpler, so my preference would be to stick to the tailrec version (this is why I pushed for it).

ghost:
> Regarding the tailrec, the intention of writing it as tailrec was actually to maintain the original style so its easier to read and hence make sure that its behaving more correctly. While its true that the current implementation doesn't have `@tailrec`, thats because the current implementation is so simple that there aren't any loops to begin with.
> 
> I just compared the tailrec version to the older non-tailrec version and the tailrec version is far easier simpler, so my preference would be to stick to the tailrec version (this is why I pushed for it).

I think @pjfanning was referring to the original code (before this pull request) versus the pull request implementation of `onMessage` and `onCommand`: since it's possible that the new code introduces some subtle changes that might result in different outcomes in some edge cases, it's better to make it as explicit as possible that, if the configuration flag (`recurse-when-unstashing-read-only-commands`) is enabled, code works like before.

Regarding the tailrec function replacing the loop, I also agree that it looks nicer and more idiomatic.

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
```diff
@@ -249,7 +263,15 @@ private[pekko] object Running {
...
-      case IncomingCommand(c: C @unchecked)          => onCommand(state, c)
+      case IncomingCommand(c: C @unchecked) if setup.settings.recurseWhenUnstashingReadOnlyCommands =>
+        onCommand(state, c)
+      case IncomingCommand(c: C @unchecked) =>
+        if (state.unstashRecurrenceState.inOnCommandCall) {
+          state.unstashRecurrenceState.recOnCommandParams = Some((state, c))
+          this // This will be ignored in onCommand
+        } else {
+          onCommand(state, c)
+        }
```
ghost, line 274: I separated the cases for IncomingCommand for the two possible values of the flag.

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
```diff
@@ -268,15 +290,60 @@ private[pekko] object Running {
         else Behaviors.unhandled
     }
 
-    def onCommand(state: RunningState[S], cmd: C): Behavior[InternalProtocol] = {
-      val effect = setup.commandHandler(state.state, cmd)
-      val (next, doUnstash) = applyEffects(cmd, state, effect.asInstanceOf[EffectImpl[E, S]]) // TODO can we avoid the cast?
-      if (doUnstash) tryUnstashOne(next)
-      else next
+    def onCommand(state: RunningState[S, C], cmd: C): Behavior[InternalProtocol] = {
```
ghost, line 293: The `onCommand` function is made up of three internal functions:
- `callApplyEffects` is in common and is needed to reduce code replication. It contains the first two rows of the original `onCommand` function;
- `recursiveUnstashImpl` is the old version and corresponds to the third and fourth line of the original `onCommand` function.
- `nonRecursiveUnstashImpl` is the new version of code

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
```diff
@@ -268,15 +290,60 @@ private[pekko] object Running {
...
+        loop(applyEffectsRetval)
+      }
+
+      val r = callApplyEffects(state, cmd)
+
+      if (setup.settings.recurseWhenUnstashingReadOnlyCommands) {
+        recursiveUnstashImpl(r)
+      } else {
+        nonRecursiveUnstashImpl(r)
+      }
```
ghost, line 342: If the flag is true (old behavior), the two inner functions that are called in sequence (`callApplyEffects` and `resursiveUnstashImpl`) are equivalend to the previous implementation of `onCommand`. If the flag is false, the new code path is used

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
pjfanning: seems fine to me

pjfanning (commented):
* this PR looks good to me - thanks @Tomassino-ibm for all your work on it
* is there anything else that needs to be done or is it about time to look into getting it merged
* I'm hoping to get a 1.2.0-M2 release together in the next few weeks and it would be great to include this

ghost:
@pjfanning I don't have anything else to change in this pull request

@mdedetrich thanks for marking the pull request as ready to review :slightly_smiling_face:

ghost:
Sorry for the new comment, but it's the first time I contribute to pekko and, while re-reading the contribution guidelines, a few questions came to my mind:

- I see I forgot to add the ticket number in commit messages, but I've also read that all commits will be squashed. Should I squash commits and add the ticket number now?
- Do I have to sign the CLA? If so, should I sign it using my personal email account, as https://www.apache.org/licenses/contributor-agreements.html suggests? I'm asking because all commits have my work e-mail, not sure if the CLA and commit email must be the same...

P.S.: I know this is slightly off-topic, let me know if I should remove this comment and maybe ask on https://github.com/apache/pekko/discussions

pjfanning:
@Tomassino-ibm we squash the commits when merging so no need to squash the commits in your branch.
We don't need iCLAs for every PR.

PR #1932, opened 2025-07-04
fix spelling of Algorithm

pjfanning:
these are internal classes


serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/JacksonSerializer.scala
```diff
@@ -162,10 +162,10 @@ import pekko.util.OptionVal
       JacksonObjectMapperProvider(system).getOrCreate(bindingName, Some(new CBORFactory)))
 
 @InternalApi object Compression {
-  sealed trait Algoritm
-  object Off extends Algoritm
-  final case class GZip(largerThan: Long) extends Algoritm
-  final case class LZ4(largerThan: Long) extends Algoritm
+  sealed trait Algorithm
```
He-Pin, line 165: I think we can just add a Algorithm trait and make Algoritm extends it, and in 2.0 we can do the migration.

pjfanning: it's an internal class - I don't see why we need to worry about renaming it now

pjfanning: `@InternalApi` marked in line 164

He-Pin (approved):
lgtm

PR #1937, opened 2025-07-06
feat: Add groupedAdjacentBy and GroupedAdjacentByWeighted operators.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/1821

Modification:
Add groupedAdjacentBy and GroupedAdjacentByWeighted operators.

### DIFF
--- a/docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
+++ b/docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
@@ -467,6 +467,32 @@ class SourceOrFlow {
     // #groupedWeighted
   }
 
+  void groupedAdjacentByExample() {
+    // #groupedAdjacentBy
+    Source.from(Arrays.asList("Hello", "Hi", "Greetings", "Hey"))
+        .groupedAdjacentBy(str -> str.charAt(0))
+        .runForeach(System.out::println, system);
+    // prints:
+    // [Hello, Hi]
+    // [Greetings]
+    // [Hey]
+    // #groupedAdjacentBy
+  }
+
+  void groupedAdjacentByWeightedExample() {
+    // #groupedAdjacentByWeighted
+    Source.from(Arrays.asList("Hello", "HiHi", "Hi", "Hi", "Greetings", "Hey"))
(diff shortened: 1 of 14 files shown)


docs/src/test/java/jdocs/stream/operators/SourceOrFlow.java
```diff
@@ -467,6 +467,32 @@ void groupedWeightedExample() {
...
 
+  void groupedAdjacentByExample() {
+    // #groupedAdjacentBy
+    Source.from(Arrays.asList("Hello", "Hi", "Greetings", "Hey"))
+        .groupedAdjacentBy(str -> str.charAt(0))
+        .runForeach(System.out::println, system);
+    // prints:
+    // [Hello, Hi]
+    // [Greetings]
+    // [Hey]
```
He-Pin, line 478: This operator works like the fs2, but use `Seq` instead of `(Key, Seq)`

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
```diff
@@ -1779,6 +1779,46 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Backpressures when''' a chunk has been assembled and downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   *
+   * @since 1.2.0
+   */
+  def groupedAdjacentByWeighted[T](f: Out => T, maxWeight: Long)(costFn: Out => Long): Repr[immutable.Seq[Out]] =
```
pjfanning, line 1819: is there a reason not to support methods on the scaladsl Source, SubFlow and SubSource like we have for javadsl?

He-Pin: scaladsl'SubFlow extends FlowOps, so it's visible to scaladsl's Source/SubFlow/SubSource.

pjfanning (approved):
lgtm

He-Pin:
@pjfanning, thanks for the review. I think this can be used in AI-related things.

PR #1940, opened 2025-07-11
chore: Make calculateDelay a public method.

He-Pin:
Motivation:

This method is useful.
btw, it maybe a  problem to use the `ThreadLocalRandom.current()` when working with virtual thread.


actor/src/main/scala/org/apache/pekko/pattern/BackoffOptions.scala
```diff
@@ -255,6 +258,22 @@ object BackoffOpts {
       maxBackoff: java.time.Duration,
       randomFactor: Double): BackoffOnStopOptions =
     onStop(childProps, childName, minBackoff.asScala, maxBackoff.asScala, randomFactor)
+
+  /**
+   * Calculates an exponential back off delay.
+   */
```
pjfanning: could you add the `@since 1.2.0` here?

pjfanning:
mima issue:
`ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.pattern.BackoffSupervisor.calculateDelay")`

actor/src/main/scala/org/apache/pekko/pattern/BackoffOptions.scala
He-Pin: I see, totally missed that.

He-Pin: Do you think we should put it in RetrySupport or here? @pjfanning

He-Pin:
thanks for take care of this

pjfanning:
@He-Pin this is almost certainly the cause of https://github.com/apache/pekko-persistence-cassandra/issues/305

https://github.com/apache/pekko-persistence-cassandra/blob/c5e25aad476b2ba9a28d7be9aa11a213a6ca2004/core/src/main/scala/org/apache/pekko/persistence/cassandra/Retries.scala#L61

I think we'll need to add back BackoffSupervisor.calculateDelay for 1.2.1 and delegate it the new method.

PR #1943, opened 2025-07-12
Release notes for 1.2.0-M2

pjfanning:
release under discussion

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-1.2.md
+++ b/docs/src/main/paradox/release-notes/releases-1.2.md
@@ -2,6 +2,41 @@
 
 Apache Pekko 1.2.x releases support Java 8 and above.
 
+# 1.2.0-M2
+
+Pekko 1.2.0-M2 has some new features, performance updates and dependency upgrades. See the [GitHub Milestone](https://github.com/apache/pekko/milestone/15?closed=1) for a fuller list of changes.
+
+This is milestone release and is aimed at testing some new changes. This release should ideally not be used in production.
+
+### Bug Fixes
+
+* Issue forming mixed Akka/Pekko cluster when classic remoting with SSL/TLS is used ([PR1857](https://github.com/apache/pekko/pull/1857))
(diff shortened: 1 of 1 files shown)


He-Pin (approved):
lgtm

docs/src/main/paradox/release-notes/releases-1.2.md
```diff
@@ -2,6 +2,63 @@
 
 Apache Pekko 1.2.x releases support Java 8 and above.
 
+# 1.2.0-M2
+
+Pekko 1.2.0-M2 has some new features, performance updates and dependency upgrades. See the [GitHub Milestone](https://github.com/apache/pekko/milestone/15?closed=1) for a fuller list of changes.
+
+This is milestone release and is aimed at testing some new changes. This release should ideally not be used in production.
```
He-Pin, line 9: Should we include the eta of the final 1.2.0 release?

mkurz: Yes, I would be interested in the ETA. Thanks!

pjfanning: * there is no ETA for 1.2.0 and under ASF rules, releases need to be discussed on the Pekko dev mailing list
* there is no point discussing 1.2.0 until 1.2.0-M2 is released and we still need at least 1 more PMC member to vote for it
* there is little point in a milestone release if we immediately follow up with a full release - so let's say a min of 1 month before a 1.2.0 release but it could be 2 or 3 or even more months.
* Play Framework does not need anything that is in 1.2.0 as far as I am aware - I would recommend that Play Framework ignores 1.2.0 and releases 3.1.0 based on Pekko 1.1.x
* If I was just using pekko-streams, I don't see why you couldn't use Pekko 1.2.0-M2 in production, especially if you're willing to do some careful testing before going into production
* there are still 1 or 2 PRs around that I think at least some PMC members want to get into 1.2.0
* once 1.2.0 is out, then I hope we are looking at dev for a 2.0 release after that so it could be a while and any enhancements that miss 1.2.0 will be waiting a while for 2.0.0 to come out - and 2.0.0 is likely to include breaking changes (removing deprecated code, dropping Java 8 support, possibly Scala 2.12 removed, etc.)

PR #1949, opened 2025-07-14
Allow overriding dispatcher in mapWithResource

### LINKED ISSUE #1948: mapWithResource does not allow overriding the dispatcher
When using `FlowOps.mapWithResource`, I would like to override the dispatcher used to run it, because I have a dedicated bulkheading dispatcher for specific types of workloads.

By default the dispatcher is set to `default-blocking-io-dispatcher`. The scaladoc of the method says this:

> You can configure the default dispatcher for this Source by changing the pekko.stream.materializer.blocking-io-dispatcher or set it for a given Source by using ActorAttributes.

But I can't get this to work.

Reproducible example:

https://scastie.scala-lang.org/rg11wr5oSGK3aov3PZ9RmQ

I have a very limited understanding of Pekko Streams internals, so my diagnosis may be very wrong. But, from what I can tell, this is because here we apply `withAttributes` to the `StatefulMap` stage, not the wrapping `via`:

https://github.com/apache/pekko/blob/5ca1b3846e0846cdc0c745c7551b8ffdaeee1266/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala#L1187-L1193

As far as I understand, because the stage with applied attributes is wrapped in a `via`, we have no way of overriding its settings, as the attributes set on nested-most stage take precedence.

I can walk around this issue simply by using `statefulMap` instead, so it's not a huge deal, aside from being a headache to figure out. I will try to contribute a fix for this.

Ostrzyciel:
Closes #1948

See the issue.

Added a regrtest for this that fails with the current version of Pekko.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
@@ -1190,7 +1190,7 @@ trait FlowOps[+Out, +Mat] {
         create,
         (resource, out) => (resource, f(resource, out)),
         resource => close(resource))
-        ).withAttributes(DefaultAttributes.mapWithResource)
+    ).withAttributes(DefaultAttributes.mapWithResource)
 
   /**
    * Transform each stream element with the help of an [[AutoCloseable]] resource and close it when the stream finishes or fails.


He-Pin:
@pjfanning Should this be backported to 1.1.x too?

pjfanning:
> @pjfanning Should this be backported to 1.1.x too?

I'm not against backporting this and tagging for a possible 1.1.6 release in he future.

PR #1958, opened 2025-07-22
remove some deprecated code from streams module

pjfanning:
* tip of the iceberg - there is a lot more - I can follow up with extra PRs but I want to avoid making the individual PRs too big
* aimed at 2.0


He-Pin (approved):
lgtm

PR #1966, opened 2025-07-27
persistence deprecations

pjfanning:
* low hanging fruit
* there is more that is deprecated but the code is more heavily wired in


He-Pin (approved):
lgtm

PR #1967, opened 2025-07-28
build with java 17

pjfanning:
* build for main branch with Java 17 minimum
* all the java 8 build support can be removed in another PR

Issue #1730, opened 2025-01-18
Drop Java 8 support


He-Pin:
Much pain on Java 8

pjfanning:
java 17 is now the min on main branch

Issue #1237, opened 2024-03-28
Declare Java 8 support deprecated


raboof:
If we declare Java 8 support deprecated in/before Pekko 1.1.0, we give ourselves the option to remove it in Pekko 1.2.0 (in the spirit of https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#when-will-a-deprecated-method-be-removed-entirely).

Depending on feedback we might still decide to keep supporting Java 8 in 1.2.0, but I think deprecating it sends the right message, and it's good to have the option.

He-Pin:
Maybe after 1.2.0 ? IIRC, @mdedetrich said they are using Java 8.

We can starting remove deprecated things in 1.2.0 and drop Java 8 （Scala 2.12) support in 1.3.0

mdedetrich:
The original idea was to drop Java 8 support in 2.0.x, as it is both considered a breaking change and there are also some significant users of Pekko that rely on it.

We can bring it up again on the mailing list, but last time we did it we got pushback

He-Pin:
I don't know much about others, but at my company, there is only 13% on Java 11, so we have to publish our internal libraries with `<release>8</release>`.

raboof:
> The original idea was to drop Java 8 support in 2.0.x, as it is both considered a breaking change and there are also some significant users of Pekko that rely on it.
> 
> We can bring it up again on the mailing list, but last time we did it we got pushback

Could you link to this pushback on https://lists.apache.org/list.html?dev@pekko.apache.org or similar? I searched but couldn't find this thread.

mdedetrich:
> Could you link to this pushback on https://lists.apache.org/list.html?dev@pekko.apache.org or similar? I searched but couldn't find this thread.

https://lists.apache.org/thread/83xmcls29opo3o8q3mwdhdpqc6bvf69c

mdedetrich:
[...]

PR #1968, opened 2025-07-28
Remove java8 home and scala steward pins related to java 11 and java 17

### LINKED ISSUE #259: Remove explicit handling of `java8Home` entirely in Pekko sbt build
With a result of the changes in https://github.com/apache/incubator-pekko/pull/258 that removes the `-release`/`-target` workaround it appears that all of this special case handling for JDK 1.8 in Pekko (i.e. https://github.com/apache/incubator-pekko/blob/main/project/JdkOptions.scala#L26 and all of the relevant places where its referenced) is no longer necessary.

pjfanning:
* fixes #259
* see also #1967 and #1730 and #1237 
* I plan to follow up with an extra PR to move the Jdk9 code into the main code directories (which looks like a fairly big change because there is a lot of build config associated with this)

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,17 +1,6 @@
 updates.pin = [
-  # Pin logback to v1.3.x because v1.4.x needs JDK11
-  { groupId = "ch.qos.logback", version="1.3." }
-  # Pin sbt-paradox to v0.9.x because 0.10.x needs JDK 11
-  { groupId = "com.lightbend.paradox", artifactId = "sbt-paradox-project-info", version = "0.9." }
-  { groupId = "com.lightbend.paradox", artifactId = "sbt-paradox", version = "0.9." }
   # Scala 3.3 is the latest LTS version
   { groupId = "org.scala-lang", artifactId = "scala3-library", version = "3.3." }
-  # aeron 1.46 requires Java 17
-  { groupId = "io.aeron", version = "1.45." }
-  # agrona 1.23 requires Java 17
-  { groupId = "org.agrona", artifactId = "agrona", version = "1.22." }
-  # bndlib 7 requires Java 17
-  { groupId = "biz.aQute.bnd", artifactId = "biz.aQute.bndlib", version = "6." }
 ]
 
 updates.ignore = [
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/LoggerClass.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/LoggerClass.scala
@@ -13,6 +13,7 @@
 
 package org.apache.pekko.actor.typed.internal
 
+import scala.annotation.nowarn
 import scala.util.control.NonFatal
 
 import org.apache.pekko
@@ -26,6 +27,7 @@ import pekko.util.OptionVal
 private[pekko] object LoggerClass {
 
   // just to get access to the class context
+  @nowarn("msg=deprecated")
   private final class TrickySecurityManager extends SecurityManager {
     def getClassStack: Array[Class[_]] = getClassContext
   }
--- a/project/Jdk9.scala
+++ b/project/Jdk9.scala
@@ -15,14 +15,11 @@ import sbt.Keys._
 import sbt._
 
 object Jdk9 extends AutoPlugin {
-  import JdkOptions.notOnJdk8
   import JdkOptions.JavaVersion._
 
-  // The version 9 is special for any Java versions >= 9
-  // and the version 11 is special for any Java versions >= 11
-  // and the version 17 is special for any Java versions >= 17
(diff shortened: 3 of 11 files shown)


.scala-steward.conf
```diff
@@ -1,17 +1,6 @@
 updates.pin = [
-  # Pin logback to v1.3.x because v1.4.x needs JDK11
```
raboof, line 2: (maybe worth mentioning in the commit message that these pins are gone)

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/LoggerClass.scala
```diff
@@ -26,6 +27,7 @@ import pekko.util.OptionVal
 private[pekko] object LoggerClass {
 
   // just to get access to the class context
+  @nowarn("msg=deprecated")
```
raboof, line 30: what is deprecated here? can we replace it?

project/JdkOptions.scala
```diff
@@ -19,89 +19,50 @@ import sbt.librarymanagement.VersionNumber
 
 object JdkOptions extends AutoPlugin {
   object autoImport {
-    lazy val jdk8home = settingKey[String]("JDK 8 home. Only needs to be set when it cannot be auto-detected by sbt")
     lazy val targetSystemJdk = settingKey[Boolean](
```
raboof: this whole concept might not be needed anymore either - it was introduced to make development easier and not requiring 2 separate JDKs to be installed, but that is not the case anymore.

project/JdkOptions.scala
```diff
@@ -19,89 +19,50 @@ import sbt.librarymanagement.VersionNumber
...
-      // FIXME replace with Runtime.version() when we no longer support Java 8
-      // See Oracle section 1.5.3 at:
-      // https://docs.oracle.com/javase/8/docs/technotes/guides/versioning/spec/versioning2.html
-      val version = specificationVersion.split('.')
-      val majorString =
-        if (version(0) == "1") version(1) // Java 8 will be 1.8
-        else version(0) // later will be 9, 10, 11 etc
-      majorString.toInt
-    }
+    val majorVersion: Int = java.lang.Runtime.version().feature()
```
raboof, line 25: do we still need this int at all?

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/LoggerClass.scala
pjfanning: The whole SecurityManager class is deprecated (since Java 19).
We might need to change this to get the context classes from a stack trace instead. If you see the StackBench change in this PR, it seems like there is a perf benefit (or maybe used to be a benefit) of doing it with the SecurityManager.

I logged #1970

project/JdkOptions.scala
pjfanning: it's still used but could possibly be removed in a later change

.scala-steward.conf
pjfanning: updated PR title - which becomes the commit message on a squash merge

project/JdkOptions.scala
pjfanning: I removed this in a subsequent commit

project/Jdk9.scala
```diff
@@ -15,14 +15,11 @@ import sbt.Keys._
...
 
-  // The version 9 is special for any Java versions >= 9
-  // and the version 11 is special for any Java versions >= 11
-  // and the version 17 is special for any Java versions >= 17
+  // The version 17 is special for any Java versions >= 17
   // and the version 21 is special for any Java versions >= 21
-  private val supportedJavaLTSVersions = List("9", "11", "17", "21")
+  private val supportedJavaLTSVersions = List("17", "21")
 
   lazy val CompileJdk9 = config("CompileJdk9").extend(Compile)
```
raboof, line 24: I suppose we can remove this whole concept - but perhaps that's for a subsequent PR

pjfanning: I intend to follow up and have noted in the PR description that this work is still needed. I had a quick look and the build plumbing for this Jdk9 option is surprisingly large so I was hoping to tackle it separately.

pjfanning:
@He-Pin @raboof the upgrade to the paradox plugins in this PR is needed to fix the build. I'm not sure why the Java 17 CI changes went ok in the PR phase but now that they are merged, the builds are failing with pegdown issues - the old version of pegdown does not seem to work in Java 17.

Would it be possible to merge this an I'll follow up with further build simplifications in further PRs?

raboof (approved):
I suppose we can add comments to the 'nowarn' occurances later

He-Pin (approved):
lgtm

PR #1945, opened 2025-07-12
Remove some deprecated code in pekko-actor

pjfanning:
* Targeted to Pekko 2.0
* A lot more could be removed - I can do more PRs so that no PR gets too big
* Some deprecated code is used under the hood
* I think it is better to record the MIMA issues for reference as opposed to turning off MIMA checking
* the tests that I removed were directly testing the deprecated methods that I am removing

PR #1969, opened 2025-07-29
remove some more deprecated code from pekko-actor

pjfanning:
* follow up to #1945 
* TypedActor has been replaced by the more fully formed actor-typed lib


He-Pin (approved):
lgtm

PR #1975, opened 2025-07-30
simplify forkjoin code

### LINKED ISSUE #1729: Build Bug: org.apache.pekko.dispatch.PekkoJdk9ForkJoinPool is not on the classpath
Even the PekkoJdk9ForkJoinPool is on the final result jar, but it does not on the classpath.

How to reproduce:
1. using Java 21
2. sbt console
3.
```scala
Welcome to Scala 2.13.16 (Java HotSpot(TM) 64-Bit Server VM, Java 21.0.5).

scala> classOf[org.apache.pekko.dispatch.PekkoJdk9ForkJoinPool]
                                         ^
       error: type PekkoJdk9ForkJoinPool is not a member of package org.apache.pekko.dispatch

scala> 
```

I think that's maybe caused by of the `Jdk9.scala`

pjfanning:
with Java 8 support gone in Pekko 2.x, I can remove the reflection based code for fork join dispatchers

fixes #1729

### DIFF
--- a/actor/src/main/mima-filters/2.0.x.backwards.excludes/simplify-forkjoin.excludes
+++ b/actor/src/main/mima-filters/2.0.x.backwards.excludes/simplify-forkjoin.excludes
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala
```diff
@@ -30,15 +28,12 @@ object ForkJoinExecutorConfigurator {
...
       asyncMode: Boolean)
-      extends ForkJoinPool(parallelism, threadFactory, unhandledExceptionHandler, asyncMode)
+      extends ForkJoinPool(parallelism, threadFactory, unhandledExceptionHandler, asyncMode,
+        0, maximumPoolSize, 1, null, ForkJoinPoolConstants.DefaultKeepAliveMillis, TimeUnit.MILLISECONDS)
       with LoadMetrics {
-    def this(
-        parallelism: Int,
-        threadFactory: ForkJoinPool.ForkJoinWorkerThreadFactory,
-        unhandledExceptionHandler: Thread.UncaughtExceptionHandler) =
-      this(parallelism, threadFactory, unhandledExceptionHandler, asyncMode = true)
```
raboof, line 41: I suppose this was here for binary compatibility, but indeed we don't need to preserve that

pjfanning: it's an internal class so I think it is safe to remove the unused constructors in a major release

PR #1976, opened 2025-07-30
move jdk9 source code

pjfanning:
move source code out of jdk9 dirs

### DIFF
--- a/stream/src/main/java/org/apache/pekko/stream/javadsl/JavaFlowSupport.java
+++ b/stream/src/main/java/org/apache/pekko/stream/javadsl/JavaFlowSupport.java
@@ -19,8 +19,8 @@ import org.apache.pekko.japi.Creator;
 import org.apache.pekko.stream.impl.JavaFlowAndRsConverters;
 
 /**
- * Provides support for `java.util.concurrent.Flow.*` interfaces which mirror the Reactive
- * Streams interfaces from `org.reactivestreams`. See <a
+ * Provides support for `java.util.concurrent.Flow.*` interfaces which mirror the Reactive Streams
+ * interfaces from `org.reactivestreams`. See <a
  * href="http//www.reactive-streams.org/">reactive-streams.org</a>.
  */
 public final class JavaFlowSupport {


actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/jfr/Events.scala
```diff
@@ -96,7 +96,7 @@ final class DeliveryProducerReceived(val producerId: String, val currentSeqNr: L
 @StackTrace(false)
 @Category(Array("Pekko", "Delivery", "ProducerController")) @Label(
   "Delivery ProducerController received demand request")
-final class DeliveryProducerReceivedRequest(val producerId: String, val requestedSeqNr: Long, confirmedSeqNr: Long)
+final class DeliveryProducerReceivedRequest(val producerId: String, val requestedSeqNr: Long, val confirmedSeqNr: Long)
```
pjfanning, line 99: * now that the source has moved to main dir - more checks are done - and this field is treated as unused
* seems best to make it a val like all the other fields - exposing it to users

cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/internal/jfr/Events.scala
```diff
@@ -12,8 +12,9 @@
  */
 
 package org.apache.pekko.cluster.sharding.internal.jfr
+
 import org.apache.pekko.annotation.InternalApi
-import jdk.jfr.{ Category, Enabled, Event, Label, StackTrace, Timespan }
+import jdk.jfr.{ Category, Event, Label, StackTrace, Timespan }
```
pjfanning, line 17: now source has moved, the build fails when imports are unused

remote/src/main/scala/org/apache/pekko/remote/artery/jfr/JFRRemotingFlightRecorder.scala
```diff
@@ -17,7 +17,6 @@ import java.net.InetSocketAddress
 
 import org.apache.pekko
 import pekko.actor.Address
-import pekko.actor.ExtendedActorSystem
```
pjfanning, line 20: now source has moved, the build fails when imports are unused

stream-tests/src/test/java/org/apache/pekko/stream/javadsl/FlowUnfoldAsyncTest.java
```diff
@@ -30,28 +30,29 @@
 import java.util.concurrent.TimeUnit;
 
 public class FlowUnfoldAsyncTest extends StreamTest {
-    @ClassRule
```
pjfanning, line 33: now source has moved, the build fails when wrong java format is used

stream/src/main/java/org/apache/pekko/stream/javadsl/JavaFlowSupport.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.pekko.stream.impl.JavaFlowAndRsConverters;
+
+/**
+ * For use only with `JDK 9+`.
+ *
+ * <p>Provides support for `java.util.concurrent.Flow.*` interfaces which mirror the Reactive
+ * Streams interfaces from `org.reactivestreams`. See <a
+ * href="http//www.reactive-streams.org/">reactive-streams.org</a>.
+ */
+public final class JavaFlowSupport {
```
He-Pin, line 26: Should we remove this `For use only with `JDK 9+`.`

pjfanning: done

PR #1979, opened 2025-07-30
remove use of org.agrona.hints.ThreadHints

pjfanning:
* relates to #1973 
* the jar upgrades in 1973 won't work for v1.2 because they require Java 17
* ThreadHints just delegates to java.lang.Thread anyway
* removing the use of org.agrona.hints.ThreadHints allows Pekko users who use Java 17+ to try running with newer versions of Aeron/Agrona

https://github.com/aeron-io/agrona/blob/master/agrona/src/main/java/org/agrona/hints/ThreadHints.java#L56


pjfanning:
build failed

PR #1980, opened 2025-07-31
Add back Scala 3 inline

mdedetrich:
In https://github.com/apache/pekko/pull/1079 we had to remove the Scala 3 inline keyword because it technically broke backwards compatibility (in short, the scala 3 inline keyword prevents that method from being called in Java, see https://github.com/lampepfl/dotty/issues/19346 and https://contributors.scala-lang.org/t/can-we-still-generate-bytecode-when-inline-def-method-been-called-from-java/6500).

This PR adds back the Scala 3 inline keyword for the Pekko 2.0.x series since breaking changes are allowed. Note that the Scala 3 inline keyword is _**ONLY**_ being added for methods which are to be called within Scala 3 and _**NOT**_ Java (the conversion util methods have already been split up between methods called from Scala and methods called from Java due to https://github.com/apache/pekko/pull/894 and as a result there isn't a single section in the entire pekko codebase where Java source calls these Scala methods)


actor/src/main/mima-filters/2.0.x.backwards.excludes/future-converters.excludes
```diff
@@ -0,0 +1,6 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#CompletionStageOps.asScala$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#CompletionStageOps.asScala")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#CompletionStageOps.asScala$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#FutureOps.asJava$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#FutureOps.asJava")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#FutureOps.asJava$extension")
```
pjfanning: line 6 is a duplicate of line 4

also need a source header and we also use add a one line comment describing the reason for the excludes

actor/src/main/mima-filters/2.0.x.backwards.excludes/option-converters.excludes
```diff
@@ -0,0 +1,30 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.OptionConverters#RichOption.toJava$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.OptionConverters#RichOption.toJavaPrimitive$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.OptionConverters#RichOption.toJava")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.OptionConverters#RichOption.toJavaPrimitive")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.OptionConverters#RichOption.toJava$extension")
```
pjfanning: line 5 duplicates line 1 and there might be more

raboof (commented):
Should this also include a note in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html that, for the purpose of binary compatibility, we should consider all methods that are called by inline functions that are part of the public API as 'public API' as well, even if MiMa won't warn about them?

mdedetrich:
@pjfanning Just resolved your comments and force pushed

mdedetrich:
> Should this also include a note in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html that, for the purpose of binary compatibility, we should consider all methods that are called by inline functions that are part of the public API as 'public API' as well, even if MiMa won't warn about them?

These are all internal API functions, none of them are meant to be called outside of pekko and this part is remarked in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#the-internal-api-and-marker.

Or am I missing something?

raboof:
> > Should this also include a note in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html that, for the purpose of binary compatibility, we should consider all methods that are called by inline functions that are part of the public API as 'public API' as well, even if MiMa won't warn about them?
> 
> These are all internal API functions, none of them are meant to be called outside of pekko and this part is remarked in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#the-internal-api-and-marker.
> 
> Or am I missing something?

Ah, I didn't realize, no problem then :+1:

raboof:
> > > Should this also include a note in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html that, for the purpose of binary compatibility, we should consider all methods that are called by inline functions that are part of the public API as 'public API' as well, even if MiMa won't warn about them?
> > 
> > 
> > These are all internal API functions, none of them are meant to be called outside of pekko and this part is remarked in https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#the-internal-api-and-marker.
> > Or am I missing something?
> 
> Ah, I didn't realize, no problem then 👍

Oh, some are `@InternalStableApi`, meaning they may get called by satellite projects, meaning they need to be de-facto binary compatible. We should probably document that _somewhere_ to remind our future selves I think.

mdedetrich:
> Oh, some are `@InternalStableApi`, meaning they may get called by satellite projects, meaning they need to be de-facto binary compatible. We should probably document that _somewhere_ to remind our future selves I think.

Oh I see, not sure if this should be in official documentation as that may confuse users as `@InternalStableApi` is still internal, there shouldn't be any project outside of the official pekko ones that use these functions but if by sattelite projects you mean what we have as pekko core then indeed this is true. Also upgrading the epic version (from 1 to 2) trumps anything including binary compatibility, so this change is all allowed.

This is why we should be quite careful/cautious whenever we bump the epoch version as we force entire ecosystem to also maintain branches (and as a result we should lump as many breaking changes as we can in 2.0.x)

`@InternalStableApi` is already documented at https://github.com/apache/pekko/blob/b0fdac259bd57fdd481483f3fe9a7aec6e1ff38a/actor/src/main/java/org/apache/pekko/annotation/InternalStableApi.java#L22-L32

raboof:
> > Oh, some are `@InternalStableApi`, meaning they may get called by satellite projects, meaning they need to be de-facto binary compatible. We should probably document that _somewhere_ to remind our future selves I think.
> 
> Oh I see, not sure if this should be in official documentation as that may confuse users as `@InternalStableApi` is still internal

I agree. Maybe in `CONTRIBUTING.md`?

> Also upgrading the epic version (from 1 to 2) trumps anything including binary compatibility, so this is all allowed.

Definitely, I'm just saying we should document that the functions called from inline `@InternalStableApi` functions should be kept binary compatible within 2.x, as that's not entirely obvious and MiMa will not help us here.
 
> This is why we should be quite careful/cautious whenever we bump the epoch version as we force entire ecosystem to also maintain branches (and as a result we should lump as many breaking changes as we can in 2.0.x)

I agree (though on the other hand we should not take that as a reason to delay 2.x too long - it's a balancing act ;) )

mdedetrich:
> I agree. Maybe in `CONTRIBUTING.md`?

Check previous commented (sorry I edited it). Although don't have an issue also adding it in `CONTRIBUTING.md` (should be a separate PR).

raboof:
> if by sattelite projects you mean what we have as pekko core then indeed this is true.

I mean 'sattelite projects' as defined in https://pekko.apache.org/docs/pekko/current/project/downstream-upgrade-strategy.html: Apache Pekko projects that are outside of 'core', such as Pekko HTTP.

> > I agree. Maybe in `CONTRIBUTING.md`?
> 
> Although don't have an issue also adding it in `CONTRIBUTING.md` (should be a separate PR).

This PR introduces the complexity that, if a function is marked both `inline` and `@InternalStableAPI`, all methods that are called by this function should be considered `@InternalStableAPI` as well (because they will be called 'directly' from other artifacts due to the inlining). This is not documented anywhere yet (as it's new), and I think this PR should document it. I don't have a strong opinion if this documentation goes into `CONTRIBUTING.md` or in the `InternalStableAPI` scaladocs.

mdedetrich:
> This PR introduces the complexity that, if a function is marked both `inline` and `@InternalStableAPI`, all methods that are called by this function should be considered `@InternalStableAPI` as well (because they will be called 'directly' from other artifacts due to the inlining). This is not documented anywhere yet (as it's new), and I think this PR should document it. I don't have a strong opinion if this documentation goes into `CONTRIBUTING.md` or in the `InternalStableAPI` scaladocs.

I don't believe this is the case, at least with Scala 3 inlining since Scala 3 inlining is true inlining on the AST level (it works differently than Scala 2 inline). 

What I mean is that if we have in project A (for Scala 3)

```scala
inline def toScala: Option[Int] = scala.jdk.OptionConverters.RichOptionalInt(o).toScala
```

and in a satellite project B we have
```scala
def someMethod = Option(3).toScala
```

Then the result will be

```scala
def someMethod = scala.jdk.OptionConverters.RichOptionalInt(Option(3)).toScala
```

But this will only happen at compile time for project B, not at method definition when the artifact is being compiled  in project A (in this case the pekko-actor artifact). This is also why Scala 3 inline prevents the method being called entirely by ulterior means, i.e. it won't appear in the JVM bytecode as a method as you can get wild results being that Scala 3 inline enforces that inlining **_must_** occur (or it will fail to compile).

Given this I don't see why caller methods that happen to call Scala 3 inline methods marked as `@InternalStableApi` need to also be marked as `@InternalStableApi`. At least as I understand it,  the only difference between `@InternalStableApi`and `@InternalApi` is that the former gives guarantees that the method will exist for the lifetime of the epoch version and its method signature won't change (which is what determines and allows it to only be used by pekko satellite projects). The java specific comments in the current `@InternalStableApi` documentation don't apply here since these methods are not meant to be called in Java source code from satellite projects (and after this change its impossible for them to be called).

Or are you stating that since the inlining is an actual literal inline we should mark the callers of the Scala 3 inline methods in some fashion because its possible that if we happen to change the implementation of the Scala 3 inline methods that it may not be reflected depending on what artifact gets resolved in the satellite project? If so I can see the concern here, however in the specific case of this PR those scala 3 inline method implementations are pretty much guaranteed to never change. They have never changed in the pekko 1.0.x series given that they just happen to delegate to methods from Scala SDK (or Scala SDK equivalent).

mdedetrich:
PR is now ready for review (all tests passing), I wont merge until current discussion with @raboof is cleared.

raboof:
> > if a function is marked both `inline` and `@InternalStableAPI`, all methods that are called by this function should be considered `@InternalStableAPI` as well
> 
> I don't see why caller methods that happen to call Scala 3 inline methods marked as `@InternalStableApi` need to also be marked as `@InternalStableApi`.

Indeed caller methods don't need to be stable. However, all methods that _are called by_  the inlined method need to be stable. Adapting your example, if we have in project A

```scala
inline def someCoreMethod: SomeTrait = InternalImplementation(42)
```
 
and in a satellite project B we have

```scala
def someSatelliteMethod = someCoreMethod()
```

Then the result will be

```scala
def someSatelliteMethod = InternalImplementation(42)
```

... meaning for example now we cannot "just" add a second parameter to the `InternalImplementation` apply method, as artifacts compiled against the previous method would still have inlined references to the old apply method.

Of course the inlined methods in this PR are completely fine, I just would sleep better if we have this documented to help avoid falling foul of this in the future.

mdedetrich:
> ... meaning for example now we cannot "just" add a second parameter to the `InternalImplementation` apply method, as artifacts compiled against the previous method would still have inlined references to the old apply method.

So indeed this is a case of what I said earlier, specifically

> Or are you stating that since the inlining is an actual literal inline we should mark the callers of the Scala 3 inline methods in some fashion because its possible that if we happen to change the implementation of the Scala 3 inline methods that it may not be reflected depending on what artifact gets resolved in the satellite project? If so I can see the concern here, however in the specific case of this PR those scala 3 inline method implementations are pretty much guaranteed to never change. They have never changed in the pekko 1.0.x series given that they just happen to delegate to methods from Scala SDK (or Scala SDK equivalent).
 
> Of course the inlined methods in this PR are completely fine, I just would sleep better if we have this documented to help avoid falling foul of this in the future.

Agreed, with this PR its entirely non issue since the implementation of the Scala 3 inline functions is never going to change for the lifecycle of 2.0.x but indeed newer contributors can run afoul of this if they don't understand the context.

Maybe its best that we add a statement in `CONTRIBUTING.md` stating that any functions that use inline (whether its Scala 2 or Scala 3) must have its implementations remain static if they use `@InternalStableApi`? I can make a separate PR for this.

pjfanning:
I'm ok with this change but I just want to mention that if, as seems likely, we drop Scala 2.12 support - we could start using the scala converters directly.
We'd should keep these util classes anyway for compatibility reasons but inlining them becomes less important if we stop using them in most places.

mdedetrich:
> I'm ok with this change but I just want to mention that if, as seems likely, we drop Scala 2.12 support - we could start using the scala converters directly. We'd should keep these util classes anyway for compatibility reasons but inlining them becomes less important if we stop using them in most places.

Thats indeed true, although the discussion to drop Scala 2.12 is a separate one (there are also other converters not related to Scala 2.12)

pjfanning (dismissed):
lgtm

actor/src/main/mima-filters/2.0.x.backwards.excludes/future-converters.excludes
```diff
@@ -0,0 +1,22 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Result of adding the Scala 3 inline keyword, these methods are no longer visible as ordinary callable methods
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#CompletionStageOps.asScala$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#CompletionStageOps.asScala")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#FutureOps.asJava$extension")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.util.FutureConverters#FutureOps.asJava")
```
pjfanning, line 22: Does this mean that code compiled with Pekko 1 that relies on this asJava method will not work with Pekko 2 jars?
This would mean that only code compiled with Pekko 2 would work with Pekko 2.

pjfanning: This change only affects Scala 3 users and it could be some shortcoming of the MIMA lib that it thinks this method (and others) are removed. If it is a genuine backward compatibility issue then I think we should look at alternatives. These methods are regarded as internal but are widely used in other Pekko modules.
One left field option is to create a new pekko.util.inline package and put the inlineable methods in there. The existing util methods could be changed to use the inlined methods but their method signatures would not be changed. We could change all the code in this repo to use the pekko.util.inline methods. We could also backport a fix like that to Pekko 1.2.0.

mdedetrich: > Does this mean that code compiled with Pekko 1 that relies on this asJava method will not work with Pekko 2 jars?
This would mean that only code compiled with Pekko 2 would work with Pekko 2.

Yes, and that is the intention for Pekko 2.0.x. Its really not possible and/or sane to claim otherwise. Even at the most basic level we will have satellite pekko projects with libraries that have minimum JVM 11/17 and its going to cause chaos/mess to our users to try and claim that x and y artifacts can be hot swappable but z is not.

Furthermore we may actually need to adjust/create some API's for some functionality, i.e. adding hooks for tracing may require this.

Pekko has never claimed or cared about source compatbility, its always binary compatibility and since we follow SemVer, bumping the epoch version from 1 to 2 indicates a clean break (including binary compatibility). Having a compromising attitude on what is binary compatibility is a recipe for disaster, we are either fully binary compatible or not, there isn't an inbetween that won't cause headaches for users.

mdedetrich:
> I've been a bit perturbed by the mima excludes file. I'm worried that the bin compat change is too much.

2.0.x is not binary compatible with 1.0.x at all, there has never been such a claim and its misleading/harmful to say otherwise.

He-Pin:
I think this is fine, because it's 2.0.x

He-Pin (approved):
lgtm for 2.0.x, does the jmh show real boost?

mdedetrich:
> lgtm for 2.0.x, does the jmh show real boost?

JVM will normally inline such calls, however the JVM only inlines up until a limit (which can be overridden but a user manually has to pass a flag into the JVM instance), see https://stackoverflow.com/questions/32503669/why-does-the-jvm-have-a-maximum-inline-depth. I have actually hit this in production, including in contexts outside of akka/pekko

So in summary, if we can inline calls (especially trivial ones like utility dispatch methods) then we should as it reduces the chance that users will hit that `MaxInlineLevel`. If we are dealing with deeply nested calls that contain a lot of these utility calls (not **that** unlikely given they are used everywhere when converting from Java to Scala methods and vice versa) then yes it will have an impact.

He-Pin:
@mdedetrich yes, but jvm inline will need some warmup and may not be steady sometimes.

mdedetrich:
> @mdedetrich yes, but jvm inline will need some warmup and may not be steady sometimes.

Even if the JVM is entirely warmed up, if you hit that `MaxInlineLevel` it will prevent any kind of inlining. If you google `XX:MaxInlineLevel` you will see plenty of cases of people needing to manually pass the arg, i.e. https://scalacenter.github.io/bloop/docs/performance-guide#tweak-the-java-jit.

In any case I will merge the PR, as its best to see if there are any potential issues right now rather than later in the release process when we will be a bit stressed.

PR #1983, opened 2025-07-31
remove deprecated code from pekko-remote

pjfanning:
* needs mima excludes
* avoids removing any of the classic remoting impl even though it is officially deprecated - but it is used by people


He-Pin (approved):
lgtm

Issue #39, opened 2022-11-10
Drop 2.12 support


He-Pin:
Scala 2.12 support (e.g. ByteString has forked sources for 2.12 and 2.13 that lead to various bugs in the past), see also https://github.com/mdedetrich/akka-apache-project/discussions/27

_Originally posted by @jrudolph in https://github.com/mdedetrich/akka-apache-project/discussions/8#discussioncomment-3904101_
   
And now JDK 21 is released too.

nvollmar:
I agree, it should be good enough to support the last two major versions

He-Pin:
Yes, Scala 3.3 lts is near the corner too.

spangaer:
Still on 2.12 ✅ 

Admittedly, we're not on the latest Akka either and we sure had the intention to move Scala 2.13 before doing so.

We're blocked from doing that before, but metals/VSCode is our route forward now.

Anyhow, am I wrong to state it's an Apache "thing" to support fairly "mature" runtimes?

Is it odd to suggest that dropping support should align with Scala center end-of-life of a version?

SethTisue:
> Is it odd to suggest that dropping support should align with Scala center end-of-life of a version?

Scala 2.12.x itself would quite conceivably have been EOLed by now if it weren't for commercial sponsorship. So you can argue that for open source Pekko to drop 2.12 is actually consistent with the approach that Lightbend and the Center have taken with our stewardship of Scala 2.12.x — namely yes if commercially funded, otherwise perhaps no.

spangaer:
I did do some research before writing that response, but it seems like these things aren't communicated very openly. (Till what you just wrote).
A published timeline for such things would really be helpful to base a decision on.
If not EOL, call it deep maintenance.

If it where Java, I'd think you'd target lts's and could start considering dropping 8.

yishenggudou:
I feel that many people are still using version 2.12, including me

[Still on 2.12](https://trends.google.com/trends/explore?q=scala%202.12,scala%202.13&hl=en-US)

nvollmar:
To get more facts, what is preventing you from upgrading to Scala 2.13?

jrudolph:
I'd like to invite everyone to view issues like this from a community perspective. The whole first goal of this project is to preserve the existing status of the open source project and set up administrative structures to enable broader participation. So, there's no attempt to drop 2.12 from the initial release.

[...]

PR #1986, opened 2025-07-31
remove scala 2.12 support

pjfanning:
We are removing Scala 2.12 support for the 2.0.0 release
while maintaining it in all our future 1.x releases.


mdedetrich:
> still under discussion for 2.0.x

This should definitely be discussed, afaik the Scala 2.12 version of Apache Pekko is still the most downloaded one so the argument for dropping it is even weaker than JDK 1.8

He-Pin:
I think we should do this, I'm using Scala 3.3.x at work, and maintaining 2.12 is a pain, the next Scala 3.8 LTS is comming soon too.

He-Pin:
@mdedetrich User, which is using 2.12 can use pekko 1.2.x too.

raboof:
Thanks for illustrating what dropping 2.12 support would look like. For future reference, the discussion on when to drop 2.12 is at #39

He-Pin:
refs: https://github.com/apache/pekko/pull/878
Once we drop 2.12.x we can try to cross compile pekko with latest scala version.

He-Pin:
All the ccompat module can be removed and migrating to `scala.jdk.*`

pjfanning:
> All the ccomp module can be removed and migrating to `scala.jdk.*`

The plan is to leave that for another PR. Most of this util and compat methods are inlined so they are not urgent to remove.

He-Pin:
But as the vote is still in progress,  what if the vote doesn't pass.

He-Pin:
@pjfanning Did the vote pass?

He-Pin:
close and open to trigger a build.

pjfanning:
Vote Thread has passed
https://lists.apache.org/thread/p9d5h6ycny3r435f187gr2ok3vg7hy8m

PR #1993, opened 2025-08-01
remove Unsafe.isIsJavaVersion9Plus

pjfanning:
* pekko.util.Unsafe is an internal class
* we only support Java 17+ in Pekko 2


He-Pin (approved):
lgtm

PR #1994, opened 2025-08-02
Make use of eq and ne.

He-Pin:
Motivation:
eq and ne are more performance

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/SpawnProtocol.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/SpawnProtocol.scala
@@ -67,7 +67,7 @@ object SpawnProtocol {
       msg match {
         case Spawn(bhvr: Behavior[t], name, props, replyTo) =>
           val ref =
-            if (name == null || name.equals(""))
+            if ((name eq null) || name.equals(""))
               ctx.spawnAnonymous(bhvr, props)
             else {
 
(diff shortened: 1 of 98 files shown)


actor-typed/src/main/scala/org/apache/pekko/actor/typed/SpawnProtocol.scala
```diff
@@ -67,7 +67,7 @@ object SpawnProtocol {
       msg match {
         case Spawn(bhvr: Behavior[t], name, props, replyTo) =>
           val ref =
-            if (name == null || name.equals(""))
+            if ((name eq null) || name.equals(""))
```
pjfanning, line 70: are the extra parentheses needed?

pjfanning: likewise for all the extra parentheses?

He-Pin: Yes, otherwise compiling error

mdedetrich:
@He-Pin Can you make backported/cherry picked commits for 1.2.x and 1.1.x?

He-Pin:
yes,  but not sure @pjfanning can agree

pjfanning:
* 1.2.x is fine for backports
* 1.1.x - I'm not quite so sure - it doesn't feel like a change that you would put in a patch release

PR #1995, opened 2025-08-02
remove final use of sun.misc.Unsafe

pjfanning:
* String value field is a byte array in Java 17+ - no need to support it being a char array like it was in older JDKs
* the VarHandle still needs `--add-opens=java.base/java.lang=ALL-UNNAMED` or equivalent so I think is ok to keep the Pekko class name as pekko.util.Unsafe - and this is better for backward compatibility too because the public methods are used in other Pekko modules

### DIFF
--- a/actor/src/main/java/org/apache/pekko/util/Unsafe.java
+++ b/actor/src/main/java/org/apache/pekko/util/Unsafe.java
@@ -15,47 +15,36 @@ package org.apache.pekko.util;
 
 import org.apache.pekko.annotation.InternalApi;
 
-import java.lang.reflect.Field;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
 import java.nio.charset.StandardCharsets;
 
 /** INTERNAL API */
 @InternalApi
 public final class Unsafe {
-  private static final sun.misc.Unsafe instance;
-
-  private static final long stringValueFieldOffset;
+  private static final VarHandle stringValueFieldHandle;
   private static final int copyUSAsciiStrToBytesAlgorithm;
 
   static {
     try {
-      sun.misc.Unsafe found = null;
-      for (Field field : sun.misc.Unsafe.class.getDeclaredFields()) {
-        if (field.getType() == sun.misc.Unsafe.class) {
-          field.setAccessible(true);
-          found = (sun.misc.Unsafe) field.get(null);
-          break;
-        }
-      }
-      if (found == null) throw new IllegalStateException("Can't find instance of sun.misc.Unsafe");
-      else instance = found;
-
-      long fo;
+      VarHandle handle;
       try {
-        fo = instance.objectFieldOffset(String.class.getDeclaredField("value"));
+        MethodHandles.Lookup lookup =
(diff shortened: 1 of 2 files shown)


mdedetrich:
> * the VarHandle still needs `--add-opens=java.base/java.lang=ALL-UNNAMED` or equivalent so I think is ok to keep the Pekko class name as pekko.util.Unsafe - and this is better for backward compatibility too because the public methods are used in other Pekko modules

Hmm, will have to think about this because iirc one algorithm is quite a bit faster than the other and the one that is faster is the one that uses `.value` @raboof @He-Pin any comments?

pjfanning:
> > * the VarHandle still needs `--add-opens=java.base/java.lang=ALL-UNNAMED` or equivalent so I think is ok to keep the Pekko class name as pekko.util.Unsafe - and this is better for backward compatibility too because the public methods are used in other Pekko modules
> 
> Hmm, will have to think about this because iirc one algorithm is quite a bit faster than the other and the one that is faster is the one that uses `.value` @raboof @He-Pin any comments?

@mdedetrich  The change keeps the code that gets the `value` byte array from the string and uses it. The only change is that I use a VarHandle to get it.

actor/src/main/java/org/apache/pekko/util/Unsafe.java
```diff
@@ -15,47 +15,37 @@
...
-        // algorithm 0
-        fo = -1;
+        MethodHandles.Lookup lookup =
+            MethodHandles.privateLookupIn(String.class, MethodHandles.lookup());
+        handle = lookup.findVarHandle(String.class, "value", byte[].class);
+      } catch (NoSuchFieldException | IllegalAccessException e) {
+        // The platform's implementation of String doesn't have a 'value' field
+        // or the field is inaccessible, so we have to use algorithm 0.
+        // You need `--add-opens=java.base/java.lang=ALL-UNNAMED` or similar to access it.
+        handle = null;
```
He-Pin, line 39: should we print this to error message like kyo?

pjfanning: * this exact same issue exists in the old code
* sun.misc.Unsafe requires exactly the same level of access granted by --add-opens
* I'm not against logging something though
* the code does not fail here though if insufficient access, it just works with standard algorithm instead (ie String getBytes public API)

pjfanning: Would a system.err.println be ok? I think you need an ActorSystem to use Pekko's logging framework and this pekko.util.Unsafe class does not have access to an ActorSystem.

pjfanning: @He-Pin is system.err.println ok for the new logging?

mdedetrich: I would leave this as is and not worry about adding a logging statement.

He-Pin: I'm fine, but in kyo, they log it to logger, or just to the exception error msg

He-Pin (approved):
lgtm

PR #2001, opened 2025-08-03
chore: Migrate to pekko.japi.function

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/267

Make use of `pekko.japi.function`

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/pattern/Patterns.scala
+++ b/actor/src/main/scala/org/apache/pekko/pattern/Patterns.scala
@@ -126,7 +126,7 @@ object Patterns {
    *   timeout);
    * }}}
    */
-  def askWithReplyTo(actor: ActorRef, messageFactory: japi.Function[ActorRef, Any], timeout: Timeout): Future[AnyRef] =
+  def askWithReplyTo(actor: ActorRef, messageFactory: japi.function.Function[ActorRef, Any], timeout: Timeout): Future[AnyRef] =
     extended.ask(actor, messageFactory.apply _)(timeout).asInstanceOf[Future[AnyRef]]
 
   /**
@@ -197,7 +197,7 @@ object Patterns {
    */
   def askWithReplyTo(
       actor: ActorRef,
-      messageFactory: japi.Function[ActorRef, Any],
+      messageFactory: japi.function.Function[ActorRef, Any],
       timeoutMillis: Long): Future[AnyRef] =
     extended.ask(actor, messageFactory.apply _)(Timeout(timeoutMillis.millis)).asInstanceOf[Future[AnyRef]]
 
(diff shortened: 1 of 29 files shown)


actor/src/main/scala/org/apache/pekko/pattern/Patterns.scala
```diff
@@ -126,7 +126,8 @@ object Patterns {
    *   timeout);
    * }}}
    */
-  def askWithReplyTo(actor: ActorRef, messageFactory: japi.Function[ActorRef, Any], timeout: Timeout): Future[AnyRef] =
+  def askWithReplyTo(actor: ActorRef, messageFactory: japi.function.Function[ActorRef, Any], timeout: Timeout)
```
pjfanning, line 129: I am not aboard with this. This is a breaking change where we did not deprecate the old code first.
I don't know what is the best way to approach this but I think this change needs serious debate.

Would it be feasible to have 2 copies of this method? This version and also the old version but that is marked as deprecated.

He-Pin: That will not work when using java lambda

pjfanning: what would the migration instructions be for affected users?

pjfanning: the binary compat issues are large:
https://github.com/apache/pekko/actions/runs/16704071539/job/47279560921?pr=2001

pjfanning: I'll go with the consensus on this but it would be useful if we knew if external users will need to rewrite their code that uses our Java stream API due to this change or if a recompile is enough. I'm making assumptions but if Java use lambdas already when passing functions to our APIs then a recompile should be enough. But if they have code that uses japi.Function directly (or any of the other function classes that are being removed) then the user needs to do a code rewrite.

He-Pin: Yes, this pain is one time; otherwise, we will have to live with two functions forever ...

mdedetrich: > the binary compat issues are large: https://github.com/apache/pekko/actions/runs/16704071539/job/47279560921?pr=2001

We do not care about binary compatibility for 2.0.x, so this is a non issue.

There is a fair point about these functions not being deprecated, I am on board with marking the `org.apache.pekko.japi.*` functions as deprecated for the 1.2.x series of Pekko (which is still yet to have its final release) and then this PR to me is approvable.

The whole point of 2.0.x is to remove cruft like this, we are allowed to break binary compatibility. Semver allows this in an epoch version change

He-Pin: @mdedetrich So you mean we depracate these classes in 1.2.x right? thanks.

mdedetrich: Yes since 1.2.0 isn't fully released yet

pjfanning:
@He-Pin The Pekko JavaAPI.Option is still not removed but I will assume that it will be in #2004

He-Pin:
@pjfanning I will do that in a separated pr.

pjfanning (approved):
lgtm

pjfanning:
see https://github.com/apache/pekko/issues/1670

Issue #2000, opened 2025-08-02
investigate stream-tests failures when run with Java 24


pjfanning:
A large number of stream-tests fail when Java 24 is used
* most relate to messages getting lost and ending up on the dead-letter-queue
* also TLSActor issues related to cipher suites not being supported or Content Renegotiation not working

https://github.com/apache/pekko/actions/runs/16697884430/job/47264617078

pjfanning:
One example of a failure with Java 24 (`sbt "stream-tests/testOnly *HubSpec*"`)
```
[info] HubSpec:
Coroner Thread Count starts at 17 in org.apache.pekko.stream.scaladsl.HubSpec
[info] MergeHub
[info] - must work in the happy case (270 milliseconds)
[info] - must notify new producers if consumer cancels before first producer (173 milliseconds)
[info] - must do not throw exceptions when upstream completes normally (55 milliseconds)
[info] - must notify existing producers if consumer cancels after a few elements (58 milliseconds)
[info] - must notify new producers if consumer cancels after a few elements (86 milliseconds)
[info] - must respect buffer size (666 milliseconds)
[WARN] [08/03/2025 18:59:17.276] [HubSpec-pekko.actor.default-dispatcher-6] [pekko://HubSpec/user/StreamSupervisor*] received dead letter from Actor[pekko://HubSpec/system/testProbe-24#1843412432]: GetChildren
--- Stream actors debug dump ---
Stream is completed. No debug information is available
[info] - must work with long streams *** FAILED *** (5 seconds, 43 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Waited 3000000000 nanoseconds. (HubSpec.scala:150)
[info]   org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2:
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:339)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:281)
[info]   at org.apache.pekko.stream.scaladsl.HubSpec.$anonfun$new$12(HubSpec.scala:150)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[...]

PR #2003, opened 2025-08-03
don't test SessionRenegotiationFirstOne on java 24

pjfanning:
* Part of #2000 but there are many other broken tests still
* SessionRenegotiationFirstTwo is still tested and that uses a stronger cipher suite
* SessionRenegotiationFirstOne uses a weak cipher suite, TLS_RSA_WITH_AES_128_CBC_SHA, that no longer seems to work with Java 24

```
javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate)
[info]   at java.base/sun.security.ssl.HandshakeContext.<init>(HandshakeContext.java:162)
[info]   at java.base/sun.security.ssl.ClientHandshakeContext.<init>(ClientHandshakeContext.java:103)
```


He-Pin (approved):
lgtm

PR #2009, opened 2025-08-05
chore: Use Java's CompletionStage in Futures.

He-Pin:
Motivation:
It should returns Java's CompletionStage for Java users.

based on https://github.com/apache/pekko/pull/2007


actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -70,7 +69,7 @@ trait FutureTimeoutSupport {
       implicit ec: ExecutionContext): CompletionStage[T] =
     if (duration.isFinite && duration.length < 1) {
```
pjfanning: `def afterCompletionStage[T](duration: FiniteDuration`

should this become?

`def afterCompletionStage[T](duration: java.time.Duration`

He-Pin: Yes, but that can be done in a separate pr.

actor/src/main/scala/org/apache/pekko/dispatch/Future.scala
```diff
@@ -137,6 +137,7 @@ object Futures {
   /**
    * Creates an already completed CompletionStage with the specified exception
    */
+  @deprecated("Use `CompletableFuture#failedStage` instead.", since = "2.0.0")
   def failedCompletionStage[T](ex: Throwable): CompletionStage[T] = {
```
pjfanning: *if this is not needed any more - could we do a 1.2.x change to deprecate it and use CompletableFuture#failedStage instead?
* the changes that affect binary compatibility would have to be for 2.x only

He-Pin: No, because the method is only added in Java 9...

mdedetrich (changes requested):
So I am blocking this for now as it seems that the entire point of the `Future`'s object at `org.apache.pekko.dispatch.Futures` is that its meant to handle Scala's futures within Java i.e. https://github.com/apache/pekko/blob/836ae6025e1c61a88c4673130ce98d2723cf6f60/actor/src/main/scala/org/apache/pekko/dispatch/Future.scala#L103-L105 so converting them to use `CompletionStage` is kind of besides the point.

To me it seems that instead we should be making sure that every single Java API only exposes `CompletionStage`, users of the Java API shouldn't be exposed to Scala Future at all.

pjfanning:
> So I am blocking this for now as it seems that the entire point of the `Future`'s object at `org.apache.pekko.dispatch.Futures` is that its meant to handle Scala's futures within Java i.e.
> 
> https://github.com/apache/pekko/blob/836ae6025e1c61a88c4673130ce98d2723cf6f60/actor/src/main/scala/org/apache/pekko/dispatch/Future.scala#L103-L105
> 
> so converting them to use `CompletionStage` is kind of besides the point.
> To me it seems that instead we should be making sure that every single Java API only exposes `CompletionStage`, users of the Java API shouldn't be exposed to Scala Future at all and then we can actually deprecate/remove this entire object/helper methods?

We could still do the part of the PR to deprecate `def failedCompletionStage` where we switch to CompletableFuture#failedStage instead.

mdedetrich:
> We could still do the part of the PR to deprecate `def failedCompletionStage` where we switch to CompletableFuture#failedStage instead.

Even this method sounds a bit funny, you are right that `def failedCompletionStage` should work with `CompletionStage` and not `Future` but then it shouldn't even sit in this `Future`'s object, it should be somewhere else.

From what I can tell, this `Future`'s object should **_only_** work with Scala `Future`'s (specifically in the context of the Java code) in which case if anything the method should be renamed to `def failedFuture` and keep its current parameter args.

I think we need a stronger look at what the purpose of these utility methods are. Definitely don't have an issue with utility helpers for Java users with `CompletionStage`, just that it shouldn't be here because that doesn't appear to be the intent of the `Future` object.

He-Pin:
@mdedetrich I am a daily Java developer, this api is in a bad design for Java developers, because Java developers work with Java's CompletionStage, not Scala's Future, here, so the result type should be a Java type。The only Scala type should be `Future` in the `input` side, and the `output/result` should always be a `CompletionStage`.

And we only need to keep the `success` and `failed` for creating Scala's Future.

He-Pin:
I think we may need another `CompletionStages` helper for Java developers and keep this `Futures` returns `CompletionStage` too. We have this `CompletableFuture` with `sequence/reduce...` etc methods at work.

mdedetrich:
> @mdedetrich I am a daily Java developer, this api is in a bad design for Java developers, because Java developers work with Java's CompletionStage, not Scala's Future, here, so the result type should be a Java type。The only Scala type should be `Future` in the `input` side, and the `output/result` should always be a `CompletionStage`.

Oh I agree with you, I also find the current design questionable. My main point is that the premise of the current util helpers that you are working on **is that they are for using Future within Java**.

In summary, we should go back to the drawing board (i.e. 从头再来/重新开始) and question the original premise of these helpers in the first place. Like why do we even have util helpers in Java for working with Scala future? As you rightfully pointed out Java users shouldn't even be using Scala future in the first place. Maybe this code is ancient and was created at a time before Java's `CompletableFuture`?

In my view we should maybe even deprecate this entire object and create a new API for Java users that just works with `CompletableFuture`. iirc The Scala SDK already has helper methods to convert from Scala `Future` to Java `CompletionStage` and vice versa. Question is whether we should deprecate this object in 1.2.x or 2.0.x (I am personally thinking 2.0.x because currently in 1.2.x there isn't any alternative to some of these methods, so if we deprecate them in 1.2.x users don't have a non deprecated version they can use, something that we can only fix in 2.0.x as some changes may be breaking)

He-Pin:
Another option for this is to delete these helper all. I think a `CompletableFutures` with timeout, sequence  and tranverse maybe help,better than the current one.

He-Pin:
@mdedetrich I think you are right, we can deprecate the current helpers and add a `CompletableFutures` helper class

He-Pin:
@pjfanning I think we should deprecate these methods in 1.2.x and then remove it in 2.0.0

He-Pin:
Use https://github.com/apache/pekko/pull/2060 now.

PR #2011, opened 2025-08-05
try to remove some deprecated materializer methods

pjfanning:
* lots of the materializer deprecated code is still actually used
* removing some parts is good but it could take a while to remove the majority of the deprecated code

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/ActorMaterializer.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/ActorMaterializer.scala
@@ -522,22 +522,10 @@ object IOSettings {
   @deprecated(
     "Use setting 'pekko.stream.materializer.io.tcp.write-buffer-size' or attribute TcpAttributes.writeBufferSize instead",
     "Akka 2.6.0")
-  def apply(system: ActorSystem): IOSettings =
-    apply(system.settings.config.getConfig("pekko.stream.materializer.io"))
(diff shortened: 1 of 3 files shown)


stream/src/main/scala/org/apache/pekko/stream/ActorMaterializer.scala
```diff
@@ -764,45 +522,15 @@ final class ActorMaterializerSettings @InternalApi private (
...
-
+@InternalApi
+private[pekko] object IOSettings {
   @deprecated(
     "Use setting 'pekko.stream.materializer.io.tcp.write-buffer-size' or attribute TcpAttributes.writeBufferSize instead",
     "Akka 2.6.0")
   def apply(config: Config): IOSettings =
     new IOSettings(
       tcpWriteBufferSize = math.min(Int.MaxValue, config.getBytes("tcp.write-buffer-size")).toInt,
       coalesceWrites = config.getInt("tcp.coalesce-writes"))
```
He-Pin, line 533: Should this be removed too?

pjfanning: This method is still used. My idea is to try to get rid of the unused methods first and then to come back later and try to remove the code that still uses this method so that it can be removed too.

He-Pin (approved):
lgtm

PR #2012, opened 2025-08-05
remove deprecated stream future operators

pjfanning:
removes some deprecated stream operators and the related docs


He-Pin (approved):
lgtm

PR #2013, opened 2025-08-05
use CompletableFuture#failedStage

pjfanning:
* part of #2009 but that other PR is blocked
* this part should be non-controversial - CompletableFuture#failedStage is a Java 9 change but Pekko 2 is Java 17 min


He-Pin (approved):
lgtm

PR #2015, opened 2025-08-07
remove more deprecated operators from pekko-stream

pjfanning:


pjfanning:
abandoned because Flow has an undeprecated recoverWith which causes DSL consistency check exceptions if I remove the Source deprecated methods

Issue #2019, opened 2025-08-07
investigate use of sun.misc.Cleaner


pjfanning:
* this looks inefficient - the Java reflection code is rerun on every call to CleanDirectBuffer https://github.com/apache/pekko/blob/c11434abb42eb7273593bb307e7654e1c8fc8532/actor/src/main/scala/org/apache/pekko/io/DirectByteBufferPool.scala#L90 
* sun.misc.Cleaner may only exist in Java 8 - see the way that this code uses sun.misc.Cleaner in Java 8 but uses sun.misc.Unsafe for Java 9+
    * https://github.com/apache/pinot/blob/584119fce06d47387e8b68a9eceb8493eebe7b02/pinot-segment-spi/src/main/java/org/apache/pinot/segment/spi/memory/CleanerUtil.java#L89
    * also https://github.com/apache/commons-io/blob/441115a4b5cd63ae808dd4c40fc238cb52c8048f/src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java#L35
* sun.misc.Unsafe has its own problems and we've only just managed to remove it from Pekko 2.0.0 code base
* the sun.misc.Unsafe cleaner code was deprecated in Java 23.
    * https://github.com/openjdk/jdk/blob/c71be802b530034169d17325478dba6e2f1c3238/src/jdk.unsupported/share/classes/sun/misc/Unsafe.java#L1739

pjfanning:
#2020 merged

PR #2020, opened 2025-08-07
use better byte buffer cleaner

pjfanning:
* see #2019 
* the existing cleaner code will not work in Java 9+
* this new code copied from commons-io will mean that the cleaner works in all Java versions
* the support it needs in sun.misc.Unsafe is deprecated in Java 23 but still exists and should work
* worst case scenario is that Java 23+ users who don't allow sun.misc.Unsafe to work will not get the Cleaner to proactively clean the byte buffers and will need to rely on the Garbage Collector
* I would like to backport this to Pekko 1.2.0

### DIFF
--- a/actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
+++ b/actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
@@ -51,7 +51,7 @@ final class ByteBufferCleaner {
     private final Method cleanerMethod;
     private final Method cleanMethod;
 
-    private Java8Cleaner() throws ReflectiveOperationException, SecurityException {
+    private Java8Cleaner() throws ReflectiveOperationException {
       cleanMethod = Class.forName("sun.misc.Cleaner").getMethod("clean");
       cleanerMethod = Class.forName("sun.nio.ch.DirectBuffer").getMethod("cleaner");
     }
@@ -69,7 +69,7 @@ final class ByteBufferCleaner {
 
     private final MethodHandle invokeCleaner;
 
-    private Java9Cleaner() throws ReflectiveOperationException, SecurityException {
+    private Java9Cleaner() throws ReflectiveOperationException {
       final Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
       final Field field = unsafeClass.getDeclaredField("theUnsafe");
       field.setAccessible(true);
--- a/actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
+++ b/actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
@@ -17,9 +17,13 @@
 
 package org.apache.pekko.io;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
+
+import static java.lang.invoke.MethodType.methodType;
 
 /**
  * Cleans a direct {@link ByteBuffer}. Without manual intervention, direct ByteBuffers will be
@@ -39,7 +43,7 @@ final class ByteBufferCleaner {
   // https://github.com/apache/commons-io/blob/441115a4b5cd63ae808dd4c40fc238cb52c8048f/src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java
 
   private interface Cleaner {
-    void clean(ByteBuffer buffer) throws ReflectiveOperationException;
+    void clean(ByteBuffer buffer) throws Throwable;
   }
 
   private static final class Java8Cleaner implements Cleaner {
@@ -53,7 +57,7 @@ final class ByteBufferCleaner {
     }
 
     @Override
-    public void clean(final ByteBuffer buffer) throws ReflectiveOperationException {
+    public void clean(final ByteBuffer buffer) throws Throwable {
       final Object cleaner = cleanerMethod.invoke(buffer);
       if (cleaner != null) {
         cleanMethod.invoke(cleaner);
@@ -63,20 +67,24 @@ final class ByteBufferCleaner {
 
   private static final class Java9Cleaner implements Cleaner {
 
-    private final Object theUnsafe;
-    private final Method invokeCleaner;
+    private final MethodHandle invokeCleaner;
 
     private Java9Cleaner() throws ReflectiveOperationException, SecurityException {
       final Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
       final Field field = unsafeClass.getDeclaredField("theUnsafe");
       field.setAccessible(true);
-      theUnsafe = field.get(null);
-      invokeCleaner = unsafeClass.getMethod("invokeCleaner", ByteBuffer.class);
+      final Object theUnsafe = field.get(null);
+      MethodHandles.Lookup lookup = MethodHandles.lookup();
+      MethodHandle invokeCleaner =
+          lookup.findVirtual(
+              unsafeClass, "invokeCleaner", methodType(void.class, ByteBuffer.class));
+      invokeCleaner = invokeCleaner.bindTo(theUnsafe);
+      this.invokeCleaner = invokeCleaner;
     }
 
     @Override
-    public void clean(final ByteBuffer buffer) throws ReflectiveOperationException {
-      invokeCleaner.invoke(theUnsafe, buffer);
+    public void clean(final ByteBuffer buffer) throws Throwable {
+      invokeCleaner.invoke(buffer);
     }
   }
 
@@ -91,8 +99,8 @@ final class ByteBufferCleaner {
   static void clean(final ByteBuffer buffer) {
     try {
       INSTANCE.clean(buffer);
-    } catch (final Exception e) {
-      throw new IllegalStateException("Failed to clean direct buffer.", e);
+    } catch (final Throwable t) {
+      throw new IllegalStateException("Failed to clean direct buffer.", t);
     }
   }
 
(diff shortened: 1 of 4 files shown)


actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private static Cleaner getCleaner() {
+    try {
+      return new Java8Cleaner();
+    } catch (final Exception e) {
+      try {
+        return new Java9Cleaner();
+      } catch (final Exception e1) {
+        throw new IllegalStateException("Failed to initialize a Cleaner.", e);
```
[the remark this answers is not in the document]
pjfanning: this is from the commons-io code and in the end, we ignore the exception anyway

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private final Object theUnsafe;
+    private final Method invokeCleaner;
+
+    private Java9Cleaner() throws ReflectiveOperationException, SecurityException {
+      final Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
+      final Field field = unsafeClass.getDeclaredField("theUnsafe");
+      field.setAccessible(true);
+      theUnsafe = field.get(null);
+      invokeCleaner = unsafeClass.getMethod("invokeCleaner", ByteBuffer.class);
```
He-Pin: use method handle

He-Pin:
@pjfanning The ForkjoinPool related thing should be updated too.

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
pjfanning: * I didn't write this code - I copied it from commons-io.
* the old reflection methods are not deprecated
* this code is called once in lifetime of the JVM

pjfanning:
> @pjfanning The ForkjoinPool related thing should be updated too.

the fokjoinpool code was already rewritten and no longer uses reflection or method handles

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
pjfanning: importantly, I intend to backport this code to 1.2.0 because it is a bug fix - so I need the code to work with Java 8

mdedetrich:
> * the support it needs in sun.misc.Unsafe is deprecated in Java 23 but still exists and should work

To clarify, this is also the case for the current state of affairs right?

pjfanning:
> > * the support it needs in sun.misc.Unsafe is deprecated in Java 23 but still exists and should work
> 
> To clarify, this is also the case for the current state of affairs right?

* the existing code for cleaning byte buffers only supports Java 8 and doesn't use sun.misc.Unsafe - sun.misc.Unsafe only added this support in Java 9 (the support that was deprecated in Java 23)
* the other Pekko code that uses sun.misc.Unsafe has all been removed in main branch
    * some of those usages were mandatory (ie the Pekko code used sun.misc.Unsafe and if it failed then Pekko failed
    * this usage and some of the old usages are optional - if it fails, Pekko will carry on with a less optimal solution

mdedetrich:
> the existing code for cleaning byte buffers only supports Java 8 and doesn't use sun.misc.Unsafe - sun.misc.Unsafe only added this support in Java 9 (the support that was deprecated in Java 23)

Right so this is what I am wary of, because if I understand correctly we may have a case where this code fails to work in Java 23+ (due to `sun.misc.unsafe` being deprecated) unless the user supplies a flag to JVM instance in order to allow calling of deprecated methods.

pjfanning:
> > the existing code for cleaning byte buffers only supports Java 8 and doesn't use sun.misc.Unsafe - sun.misc.Unsafe only added this support in Java 9 (the support that was deprecated in Java 23)
> 
> Right so this is what I am wary of, because if I understand correctly we may have a case where this code fails to work in Java 23+ (due to `sun.misc.unsafe` being deprecated) unless the user supplies a flag to JVM instance in order to allow calling of deprecated methods.

The code ignores exceptions - old version and new version. If you don't allow access to sun.misc.Unsafe, code will still work.
The new code even remembers that it failed so it won't keep trying sun.misc.Unsafe.
We can add documentation to warn users that they may need to set `--sun-misc-unsafe-memory-access=allow`. I'm not sure if that java flag is needed for the Cleaner support in sun.misc.Unsafe or if that flag is more about the memory access functions like the field offset ones.

In a later issue or PR, I can look at seeing if I can write a different DirectByteBuffer implementation that Java 23+ users will benefit from. That would involve creating Java MemorySegments and converting those to ByteBuffers and those MemorySegments can be cleaned using the appropriate APIs. Based on https://github.com/openjdk/jdk/blob/c71be802b530034169d17325478dba6e2f1c3238/src/jdk.unsupported/share/classes/sun/misc/Unsafe.java#L1731

pjfanning:
I've added the documentation bit to https://github.com/apache/pekko/issues/1840

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
He-Pin: It was using reflection too in Netty's CleanerJava9, but in 4.2, it's now using method handle, method handle is a litter faster.

He-Pin:
```java
MethodHandles.Lookup lookup = MethodHandles.lookup();
MethodHandle invokeCleaner = lookup.findVirtual(
unsafeClass, "invokeCleaner", methodType(void.class, ByteBuffer.class));
invokeCleaner = invokeCleaner.bindTo(PlatformDependent0.UNSAFE);
```

He-Pin (approved):
lgtm

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -0,0 +1,128 @@
...
+        cleanMethod.invoke(cleaner);
+      }
+    }
+  }
+
+  private static final class Java9Cleaner implements Cleaner {
+
+    private final MethodHandle invokeCleaner;
+
+    private Java9Cleaner() throws ReflectiveOperationException, SecurityException {
```
He-Pin: SecurityException seems will be removed , how about just an Exception

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -0,0 +1,128 @@
...
+      final Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
+      final Field field = unsafeClass.getDeclaredField("theUnsafe");
+      field.setAccessible(true);
+      final Object theUnsafe = field.get(null);
+      MethodHandles.Lookup lookup = MethodHandles.lookup();
+      MethodHandle invokeCleaner =
+          lookup.findVirtual(
+              unsafeClass, "invokeCleaner", methodType(void.class, ByteBuffer.class));
+      this.invokeCleaner = invokeCleaner.bindTo(theUnsafe);
```
He-Pin, line 81: In Netty, there is an buffer(1) and try to clean it ,but seems we don;t have it here.

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -0,0 +1,128 @@
...
+   * Releases memory held by the given {@link ByteBuffer}.
+   *
+   * @param buffer to release.
+   * @throws IllegalStateException on internal failure.
+   */
+  static void clean(final ByteBuffer buffer) {
+    try {
+      INSTANCE.clean(buffer);
+    } catch (final Throwable t) {
+      throw new IllegalStateException("Failed to clean direct buffer.", t);
```
He-Pin, line 102: UnsupportedOperationException

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
pjfanning: the code is not a copy of netty - it is copied from the Apache project, commons-io

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
pjfanning: It isn't an unsupported exception if it gets here - INSTANCE is null if it is unsupported. I check for this at runtime.
This is not a public API. It is not even a public class. It only needs to suit the way that I use it internally.

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
He-Pin: so we can't adjust it?

pjfanning: I've added the initialisation time check that you requested

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -104,15 +104,33 @@ static void clean(final ByteBuffer buffer) {
...
+        System.err.println(
+            "WARNING: Failed to initialize a ByteBuffer Cleaner. This means "
+                + "direct ByteBuffers will only be cleaned upon garbage collection. Reason: "
+                + e1);
       }
     }
+    if (cleaner != null) {
+      try {
+        ByteBuffer testByteBuffer = ByteBuffer.allocateDirect(1);
+        cleaner.clean(testByteBuffer);
```
He-Pin, line 123: nice!

Issue #2026, opened 2025-08-12
Allow disabling AsyncWriteJournal.Resequencer to improve latency


migesok:
Previous discussion regarding `AsyncWriteJournal.Resequencer`: https://github.com/apache/pekko/discussions/1881

Why I think that `AsyncWriteJournal.Resequencer` should be safe to disable: https://github.com/apache/pekko/discussions/1881#discussioncomment-13367927

### More context:

In my company we are running a lot of code based on classic persistent actors with our custom journal plugin: https://github.com/evolution-gaming/kafka-journal.

Its write path goes through Kafka. We noticed that actor persist latency stats do not match the stats we obtain from the plugin's Kafka producer. Moreover, latency issues in one partition somehow disproportionally affected latency for all the write requests, not only to that particular partition. After a lengthy investigation it was found that the culprit is `AsyncWriteJournal.Resequencer`, which made all write requests wait if one request takes too long.

We disabled it in our internal Akka fork and we run it in production for some time with good results: in some cases latency was down 70%. No issues observed so far.

Since we are in the process of migrating to Pekko, we would like to contribute the fix there.

The PR will be provided shortly.

pjfanning:
I've marked #2027 as milestone 2.0.x but I see no reason not to merge it to 1.2.x if and when 2027 is merged.

migesok:
> I've marked [#2027](https://github.com/apache/pekko/pull/2027) as milestone 2.0.x but I see no reason not to merge it to 1.2.x if and when 2027 is merged.

Thank you!

pjfanning:
@migesok #2434 seems to possibly make the AsyncWriteJournalResponseOrderSpec brittle - or even breaks it altogether

pjfanning:
With #2434, a copy of an Akka change, this change now is starting to seem potentially dangerous.
We definitely would never want to consider making `write-response-global-order = off` the default.

migesok:
> With [#2434](https://github.com/apache/pekko/pull/2434), a copy of an Akka change, this change now is starting to seem potentially dangerous. We definitely would never want to consider making `write-response-global-order = off` the default.

@pjfanning could you elaborate why is it related and how it could be dangerous with the default dispatcher?

Internally, we have been running on the default dispatcher for a long time already.

pjfanning:
[...]

PR #2027, opened 2025-08-12
Add AsyncWriteJournal option for disabling Resequencer (#2026)

migesok:
https://github.com/apache/pekko/issues/2026

I left the previous behaviour to be the default one, just in case, but I think that eventually AsyncWriteJournal.Resequencer should be removed.

Please, double check the wording on the reference.conf config option. Also I'm not sure whether I need to update the documentation and where should I update it.

### DIFF
--- a/persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala
+++ b/persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala
@@ -21,8 +21,8 @@ import scala.util.{ Failure, Success, Try }
 import scala.util.control.NonFatal
 
 import org.apache.pekko
-import org.apache.pekko.annotation.InternalApi
 import pekko.actor._
+import pekko.annotation.InternalApi
 import pekko.pattern.CircuitBreaker
 import pekko.pattern.pipe
 import pekko.persistence._
--- a/persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala
+++ b/persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala
@@ -21,6 +21,7 @@ import scala.util.{ Failure, Success, Try }
 import scala.util.control.NonFatal
 
 import org.apache.pekko
+import org.apache.pekko.annotation.InternalApi
 import pekko.actor._
 import pekko.pattern.CircuitBreaker
 import pekko.pattern.pipe
@@ -74,6 +75,7 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
 
     // should be a private method in the trait, but it needs the enableGlobalWriteResponseOrder field which can't be
     // moved to the trait level because adding any fields there breaks bincompat
+    @InternalApi
     def sendWriteResponse(msg: Any, snr: Long, target: ActorRef, sender: ActorRef): Unit = {
       if (enableGlobalWriteResponseOrder) {
         resequencer ! Desequenced(msg, snr, target, sender)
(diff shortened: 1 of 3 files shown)


persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -67,9 +67,21 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
...
     // cannot be a val in the trait due to binary compatibility
     val replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")
+    val enableGlobalWriteResponseOrder: Boolean = config.getBoolean("write-response-global-order")
+
     val eventStream = context.system.eventStream // used from Future callbacks
     implicit val ec: ExecutionContext = context.dispatcher
 
+    // should be a private method in the trait, but it needs the enableGlobalWriteResponseOrder field which can't be
+    // moved to the trait level because adding any fields there breaks bincompat
+    def sendWriteResponse(msg: Any, snr: Long, target: ActorRef, sender: ActorRef): Unit = {
```
pjfanning, line 79: can this be package private?
if not can you annotate it with `@InternalApi`?

migesok: Unfortunately, no amount of annotations would save me from runtime class errors here.

This method needs access to the "memoized" config option, which has to be a val/var/lazy val - enableGlobalWriteResponseOrder. Which I currently put under ` final val receiveWriteJournal: Actor.Receive = {...}` scope. My first version had the field on the trait level as a private val - right under here:
https://github.com/apache/pekko/blob/e7e71a3adffdd8b19d800b22843780b54f56d4b8/persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala#L60

Then it lead to runtime issues:
- you have a concrete plugin class extending AsyncWriteJournal built against the old version, without the new private val
- you try to run it against the new AsyncWriteJournal version
- adding a new private/packag-private/private[this] val to trait results in new abstract methods in the generated interface
- classloading of the plugin class fails in runtime because an interface it implements has new abstract methods which the concrete class does not implement

It seems I was not the first one to hit this issue in this trait:
```
  final val receiveWriteJournal: Actor.Receive = {
    // cannot be a val in the trait due to binary compatibility
    val replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")
```

TBH before this ordeal I wasn't aware that adding private fields to a trait breaks bincompat but I was able to confirm it on small examples:
https://gist.github.com/migesok/71c5e5feb1ba4fff8c6515f42922df11

Please let me know if there is a better way to structure it without breaking bincompat.

pjfanning: add the `@InternalApi` annotation to signal that it is not a public API and we can refactor it later as part of 2.0.0 dev (where we are allow to break bin compat)

migesok: @pjfanning could you elaborate please? If I move the field and the method to the trait level, it will break all the 3rd party journal plugins built against the previous version. In this case we would have to mark the whole `AsyncWriteJournal` as an internal API, leaving no proper persistence plugin API for the users.

Ideally the plugin API should be a pure trait, without fields, so Pekko-Persistence internals can evolve freely but this is what we have right now.

pjfanning: This thread is attached to line 77 in AsyncWriteJournal.scala (of your PR).
I am asking you to add an annotation to `def sendWriteResponse`.
The annotation is `@org.apache.pekko.annotation.InternalApi` - although would be nicer to have an import for this class and to then just use `@InternalApi`.

migesok: @pjfanning added. I hope I got it right this time.

pjfanning (approved):
lgtm

migesok:
@pjfanning thank you! Can this be merged? What's the procedure?

pjfanning:
I left it unmerged in the hope that someone else will look at it. I will merge over the next few days if there are no comments.

pjfanning:
merged - thanks

PR #2037, opened 2025-08-17
Scheduler: remove more deprecated methods

pjfanning:
continuing effort to remove deprecated methods for 2.0.0


actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala
```diff
@@ -303,7 +299,16 @@ trait Scheduler {
...
       sender: ActorRef = Actor.noSender): Cancellable =
-    schedule(initialDelay, interval, receiver, message)
+    schedule(
+      initialDelay,
+      interval,
+      new Runnable {
+        def run(): Unit = {
+          receiver ! message
+          if (receiver.isTerminated)
+            throw SchedulerException("timer active for terminated actor")
```
Roiocam, line 309: does scheduler catch SchedulerException propertly?

actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala
```diff
@@ -340,37 +345,7 @@ trait Scheduler {
...
-      executor: ExecutionContext,
-      sender: ActorRef = Actor.noSender): Cancellable =
-    schedule(
-      initialDelay,
-      interval,
-      new Runnable {
-        def run(): Unit = {
-          receiver ! message
-          if (receiver.isTerminated)
-            throw SchedulerException("timer active for terminated actor")
```
He-Pin, line 362: The code was ported from here.

actor/src/main/scala/org/apache/pekko/actor/Scheduler.scala
He-Pin: Do you mean the scheduler maybe die because of this?

pjfanning: https://github.com/apache/pekko/pull/2037#discussion_r2287091109

As He Pin indicates, this is not new code. This method used to call a deprecated method. I removed that method and moved its implementation to here.

PR #2046, opened 2025-08-22
chore: disable batch if isVirtualized

He-Pin:
Motivation:
There is a threadlocal in AbstractDispatcher, which will cause large memory usage, so when using virtual thread, we should disable that batching.

And this should be backport to 1.2.x too

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/Dispatcher.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/Dispatcher.scala
@@ -64,6 +64,7 @@ class Dispatcher(
     new LazyExecutorServiceDelegate(executorServiceFactoryProvider.createExecutorServiceFactory(id, threadFactory))
 
   protected final def executorService: ExecutorServiceDelegate = executorServiceDelegate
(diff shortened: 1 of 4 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/Dispatcher.scala
```diff
@@ -64,16 +64,20 @@ class Dispatcher(
...
    */
   protected[pekko] def dispatch(receiver: ActorCell, invocation: Envelope): Unit = {
     val mbox = receiver.mailbox
     mbox.enqueue(receiver.self, invocation)
-    registerForExecution(mbox, true, false)
+    registerForExecution(mbox, hasMessageHint = true, hasSystemMessageHint = false)
   }
 
+  final override def batchable(runnable: Runnable): Boolean =
+    if (isVirtualized) true else super.batchable(runnable)
```
pjfanning: the PR title says 'disable batch if isVirtualized' but this seems to be the opposite

He-Pin: Opps...

PR #2051, opened 2025-08-22
chore: Fix some sbt deprecations

He-Pin:
Motivation:
Fix some deprecates when do sbt reload


He-Pin:
@pjfanning This should be backported to 1.2.x too

PR #2052, opened 2025-08-22
chore: Remove two deprecated methods in AbstractScheduler

He-Pin:
Motivation:

Remove the depcated methods from Java side


pjfanning:
Changed milestone to 2.0.x. This is not something that we should aim to backport to 1.2.

PR #2056, opened 2025-08-23
remove deprecated monitor operation

pjfanning:
another method that can be removed in Pekko 2.0.0


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Flow.scala
```diff
@@ -4425,17 +4425,6 @@ trait FlowOpsMat[+Out, +Mat] extends FlowOps[Out, Mat] {
...
-  /**
-   * Materializes to `FlowMonitor[Out]` that allows monitoring of the current flow. All events are propagated
-   * by the monitor unchanged. Note that the monitor inserts a memory barrier every time it processes an
-   * event, and may therefor affect performance.
-   *
-   * The `combine` function is used to combine the `FlowMonitor` with this flow's materialized value.
-   */
-  @deprecated("Use monitor() or monitorMat(combine) instead", "Akka 2.5.17")
-  def monitor[Mat2]()(combine: (Mat, FlowMonitor[Out]) => Mat2): ReprMat[Out, Mat2] =
-    viaMat(GraphStages.monitor)(combine)
```
He-Pin, line 4437: mima is pushed to another pr :)

pjfanning: thanks - I've fixed this now

PR #2057, opened 2025-08-23
feat: Add CompletionStages util.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/2008

Add a helper for CompletionStages and then we can deprecate the methods in Futures.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
@@ -1,0 +1,219 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
```diff
@@ -0,0 +1,217 @@
...
+   * @param stages   the stages to aggregate
+   * @param function the associative function to use for aggregation
+   * @return a `CompletionStage` that, when completed, will contain the aggregated result
+   * @since 1.2.0
+   */
+  def fold[T <: AnyRef, R](
+      zero: R,
+      stages: java.lang.Iterable[_ <: CompletionStage[_ <: T]],
+      function: pekko.japi.function.Function2[R, T, R]): CompletionStage[R] = {
+    foldWithNext[T, R](stages.iterator(), zero, function)
```
He-Pin, line 131: Scala 2.12 needs these `[T,R]` :(

PR #2060, opened 2025-08-23
feat: Add CompletionStages util.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/2008

Add a helper for CompletionStages and then we can deprecate the methods in Futures.

This should be backported to 1.2.x

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/dispatch/JavaFutureTests.java
+++ b/actor-tests/src/test/java/org/apache/pekko/dispatch/JavaFutureTests.java
@@ -33,6 +33,7 @@ import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.LinkedList;
 import java.lang.Iterable;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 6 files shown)


He-Pin:
Once this is merged, I will backport these methods to 1.2.x and then add deprecations for `Futures#find` and etcs, after that, we can remove these methods from the main branch.

actor-tests/src/test/java/org/apache/pekko/dispatch/JavaFutureTests.java
```diff
@@ -48,7 +49,14 @@ public class JavaFutureTests extends JUnitSuite {
   private final ActorSystem system = actorSystemResource.getSystem();
   private final Duration timeout = Duration.create(5, TimeUnit.SECONDS);
 
-  @Test
+    @Test
```
pjfanning: the indenting seems out of line with the rest of the file

He-Pin: yes, it was like that , but anyway, these will be removed soon.

PR #2063, opened 2025-08-23
chore: Make timeoutCompletionStage accept java Duration.

He-Pin:
Motivation:

This method should accept Java's duration

Should be backported to 1.2.0 too

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/pattern/PatternsTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/pattern/PatternsTest.java
@@ -536,7 +536,7 @@ public class PatternsTest extends JUnitSuite {
       delayedStage.toCompletableFuture().get(3, SECONDS);
     } catch (ExecutionException e) {
       assertTrue(e.getCause() instanceof TimeoutException);
-      assertEquals("Timeout of 200 milliseconds expired", e.getCause().getMessage());
+      assertEquals("Timeout of PT0.2S expired", e.getCause().getMessage());
     }
   }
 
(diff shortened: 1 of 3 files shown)


actor-tests/src/test/java/org/apache/pekko/pattern/PatternsTest.java
```diff
@@ -531,7 +531,7 @@ public void testCompletedWithTimeout() throws Exception {
       delayedStage.toCompletableFuture().get(3, SECONDS);
     } catch (ExecutionException e) {
       assertTrue(e.getCause() instanceof TimeoutException);
-      assertEquals("Timeout of 200 milliseconds expired", e.getCause().getMessage());
+      assertEquals("Timeout of PT0.2S expired", e.getCause().getMessage());
```
pjfanning, line 539: Is there any way to preserve the more readable '200 milliseconds'? Not everyone is familar with ISO-8601 duration notation.

actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -124,7 +124,7 @@ trait FutureTimeoutSupport {
    * @since 1.2.0
    */
   @deprecated("Use `CompletableFuture#orTimeout instead.", "Pekko 2.0.0")
```
pjfanning: we generally don't put "Pekko" in the version, we just use "Akka" in the version if the version is an Akka version.

actor-tests/src/test/java/org/apache/pekko/pattern/PatternsTest.java
pjfanning: I had a quick look for ways to do this and I found pages like https://www.baeldung.com/java-ms-to-hhmmss
It looks like there isn't a tidy solution because we don't want to use a 3rd party lib - so I'm ok with keeping the change as is.

He-Pin: @pjfanning No, we lost the origin unit when using Java's Duration,

PR #2064, opened 2025-08-23
chore: Remove JAPI

He-Pin:
Motivation:
Remove JAPI


actor/src/main/java/org/apache/pekko/japi/JAPI.java
```diff
@@ -1,32 +0,0 @@
...
-
-import scala.collection.Seq;
-
-@Deprecated
-public class JAPI {
-
-  /**
-   * Create a Scala seq from a Java array.
-   *
-   * @deprecated since 1.1.0. Use <code>
```
pjfanning, line 24: I don't really mind if this method is unused - let's then get rid of it. So far, I've avoided removing methods that were deprecated by the Pekko team. I have concentrated on the older deprecations done by the Akka team before we forked off Pekko.

He-Pin:
```scala
  @SafeVarargs
  @varargs
  def createImmutableSeq[T](elements: T*): immutable.Seq[T] =
    if (elements != null && elements.nonEmpty) elements.toIndexedSeq else Nil
```
Maybe we can add something like this to Util

pjfanning: only if it is needed

He-Pin: The only reason I think it was there is be cause if `seq(1,2,3)` to create an immutableSeq, otherwise, that will cause some overloading issue if it's the same name in `Util`

PR #2070, opened 2025-08-23
chore: Remove deprecated methods in Futures

He-Pin:
Motivation:
Remove the deprecated methods and classes in Futures, we should use CompletionStage now.


actor/src/main/scala/org/apache/pekko/dispatch/Future.scala
```diff
@@ -153,266 +146,4 @@ object Futures {
...
- * if the Future that this callback is registered on becomes completed with a success.
- * This callback is the equivalent of an org.apache.pekko.japi.Function
- *
- * Override "apply" normally, or "checkedApply" if you need to throw checked exceptions.
- *
- * SAM (Single Abstract Method) class
- *
- * Java API
- */
-abstract class Mapper[-T, +R] extends scala.runtime.AbstractFunction1[T, R] {
```
He-Pin, line 402: already deprecated in 1.2.x

PR #2075, opened 2025-08-23
chore: Deprecate FI

He-Pin:
Motivation:
chore: Depreacte FI

refs: https://github.com/apache/pekko/issues/948


actor/src/main/java/org/apache/pekko/japi/pf/FI.java
```diff
@@ -30,14 +30,17 @@ private FI() {}
    *
    * @param <I> the input type, that this Apply will be applied to
    * @param <R> the return type, that the results of the application will have
+   * @deprecated Will be removed in 2,0.0
```
pjfanning: 2.0.0

PR #2082, opened 2025-08-24
release notes for 1.2.0

pjfanning:
release vote has passed

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-1.2.md
+++ b/docs/src/main/paradox/release-notes/releases-1.2.md
@@ -70,7 +70,7 @@ Most of the changes appeared in the milestone releases (1.2.0-M1 and 1.2.0-M2) b
 * Change aggregateWithBoundary operator in javadsl to use Optional ([PR1876](https://github.com/apache/pekko/pull/1876))
 * TLS v1.3 is now the default ([PR1901](https://github.com/apache/pekko/pull/1901))
 * Set vector builder to null after stage completed to avoid leak ([PR1917](https://github.com/apache/pekko/pull/1917))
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/release-notes/releases-1.2.md
```diff
@@ -2,24 +2,34 @@
...
 Apache Pekko 1.2.x releases support Java 8 and above.
 
-# 1.2.0-M2
+# 1.2.0
 
-Pekko 1.2.0-M2 has some new features, performance updates and dependency upgrades. See the [GitHub Milestone](https://github.com/apache/pekko/milestone/15?closed=1) for a fuller list of changes.
+Pekko 1.2.0 has some new features, performance updates and dependency upgrades. See [GitHub Milestone for 1.2.0-M1](https://github.com/apache/pekko/milestone/6?closed=1), [GitHub Milestone for 1.2.0-M2](https://github.com/apache/pekko/milestone/15?closed=1) and [GitHub Milestone for 1.2.0](https://github.com/apache/pekko/milestone/16?closed=1) for a fuller list of changes.
 
-This is milestone release and is aimed at testing some new changes. This release should ideally not be used in production.
+Most of the changes appeared in the milestone releases (1.2.0-M1 and 1.2.0-M2) but some additional changes were made for the 1.2.0 release. These extra changes include deprecating some methods that will be removed in a future release.
```
He-Pin, line 9: I think we should contain that：
The next release will be 2.0.0, which will remove deprecated methods and require Java 17, but 1.2.0 will be maintained too.

pjfanning: I don't think that we need to be specific here. Release plans might change and I'd prefer not to have to retroactively update the release notes. For release information, users are better off subscribing the Pekko dev list.

PR #2087, opened 2025-08-25
Java Docs: use instanceof with name binding

pjfanning:
* Available since Java 16
* pekko 2.0.0 supports min Java 17


He-Pin (approved):
lgtm

He-Pin:
We may need to rewrite some code with pattern matching too.

PR #2092, opened 2025-08-25
use Java CompletionStage instead of Scala Future in pekko-persistence Java APIs

pjfanning:
* based on #1419
* breaking change so will target Pekko 2.0.0
* it might be useful to do some follow up changes but I think this is enough to start with

### DIFF
--- a/docs/src/test/java/jdocs/persistence/LambdaPersistencePluginDocTest.java
+++ b/docs/src/test/java/jdocs/persistence/LambdaPersistencePluginDocTest.java
@@ -30,7 +30,6 @@ import org.apache.pekko.persistence.japi.snapshot.JavaSnapshotStoreSpec;
 import org.apache.pekko.persistence.journal.japi.*;
 import org.apache.pekko.persistence.journal.leveldb.SharedLeveldbJournal;
 import org.apache.pekko.persistence.journal.leveldb.SharedLeveldbStore;
-import org.apache.pekko.persistence.serialization.MessageFormats.AtomicWrite;
 import org.apache.pekko.persistence.snapshot.japi.*;
 // #plugin-imports
 import org.iq80.leveldb.util.FileUtils;


persistence/src/main/scala/org/apache/pekko/persistence/snapshot/japi/SnapshotStore.scala
```diff
@@ -29,15 +30,15 @@ abstract class SnapshotStore extends SSnapshotStore with SnapshotStorePlugin {
   override final def loadAsync(
       persistenceId: String,
       criteria: SnapshotSelectionCriteria): Future[Option[SelectedSnapshot]] =
-    doLoadAsync(persistenceId, criteria).map(option)
+    doLoadAsync(persistenceId, criteria).asScala.map(option)
```
He-Pin, line 33: sould pass a `parasitic`

persistence/src/main/scala/org/apache/pekko/persistence/snapshot/japi/SnapshotStore.scala
```diff
@@ -29,15 +30,15 @@ abstract class SnapshotStore extends SSnapshotStore with SnapshotStorePlugin {
...
 
   override final def saveAsync(metadata: SnapshotMetadata, snapshot: Any): Future[Unit] =
-    doSaveAsync(metadata, snapshot).map(_ => ())
+    doSaveAsync(metadata, snapshot).asScala.map(_ => ())
```
He-Pin: parasitic

He-Pin: `ConstantFun.anyToUnit`

pjfanning: there is no ConstantFun.anyToUnit

He-Pin: I mean `  def scalaAnyToUnit[A]: A => Unit = unit`,
Sorry not check the code.

pjfanning: I found `ConstantFun.unit` - I thought this was closer

pjfanning: I see that is private

persistence/src/main/scala/org/apache/pekko/persistence/snapshot/japi/SnapshotStore.scala
```diff
@@ -33,12 +34,12 @@ abstract class SnapshotStore extends SSnapshotStore with SnapshotStorePlugin {
     doLoadAsync(persistenceId, criteria).asScala.map(option)
 
   override final def saveAsync(metadata: SnapshotMetadata, snapshot: Any): Future[Unit] =
-    doSaveAsync(metadata, snapshot).asScala.map(_ => ())
+    doSaveAsync(metadata, snapshot).asScala.map(ConstantFun.unit)
```
He-Pin: map(ConstantFun.unit)(org.apache.pekko.dispatch.ExecutionContexts.parasitic)

persistence/src/main/scala/org/apache/pekko/persistence/snapshot/japi/SnapshotStore.scala
He-Pin: `map(option)(org.apache.pekko.dispatch.ExecutionContexts.parasitic)`

persistence/src/main/scala/org/apache/pekko/persistence/snapshot/japi/SnapshotStore.scala
pjfanning: `import context.dispatcher` - shouldn't this be enough?

Or is there a good reason to prefer `parasitic`?

He-Pin: Yes, `parasitic` is the same thread, with which we can avoid a context switching.

persistence/src/main/scala/org/apache/pekko/persistence/snapshot/japi/SnapshotStore.scala
```diff
@@ -26,7 +26,7 @@ import pekko.util.FutureConverters._
  * Java API: abstract snapshot store.
  */
 abstract class SnapshotStore extends SSnapshotStore with SnapshotStorePlugin {
-  import context.dispatcher
+  import pekko.dispatch.ExecutionContexts.parasitic
```
He-Pin: I think we should only use it in the last step `map(...)(parasitic)`

persistence/src/main/scala/org/apache/pekko/persistence/journal/japi/AsyncWriteJournal.scala
```diff
@@ -42,6 +43,7 @@ abstract class AsyncWriteJournal extends AsyncRecovery with SAsyncWriteJournal w
         .to(immutable.IndexedSeq)
     }
 
-  final def asyncDeleteMessagesTo(persistenceId: String, toSequenceNr: Long) =
-    doAsyncDeleteMessagesTo(persistenceId, toSequenceNr).map(_ => ())
+  final def asyncDeleteMessagesTo(persistenceId: String, toSequenceNr: Long): Future[Unit] = {
+    doAsyncDeleteMessagesTo(persistenceId, toSequenceNr).asScala.map(_ => ())
```
He-Pin: unit

He-Pin (approved):

persistence/src/main/scala/org/apache/pekko/persistence/journal/japi/AsyncWriteJournal.scala
```diff
@@ -33,7 +34,7 @@ abstract class AsyncWriteJournal extends AsyncRecovery with SAsyncWriteJournal w
   import context.dispatcher
 
   final def asyncWriteMessages(messages: immutable.Seq[AtomicWrite]): Future[immutable.Seq[Try[Unit]]] =
-    doAsyncWriteMessages(messages.asJava).map { results =>
+    doAsyncWriteMessages(messages.asJava).asScala.map { results =>
```
He-Pin, line 38: parasites

pjfanning: changed

mdedetrich:
@pjfanning Regarding `parasitic`, for future reference whenever you do a `map` operation on Scala's `Future`, always pass in the explicit `ExecutionContexts.parasitic`. If you don't do this, the block inside of the `map` has to go into the implicit `ExecutionContext` (usually some kind of fork join pool), and that context switch is unnecessary overhead for what typically happens inside of a map.

Context switch is fine (and needed) for long running async operations, but those are typically composed with `flatMap` (since they would also be running in some kind of `Future`). When it comes to long running blocking that aren't inside of a `Future`, you would use the `blocking`/specific thread based dispatcher instead

pjfanning:
@He-Pin I made a couple of changes after your review was made including fixing an issue. Do you still think it is ok to merge as is?

persistence/src/main/java/org/apache/pekko/persistence/journal/japi/AsyncWritePlugin.java
```diff
@@ -70,16 +70,22 @@ interface AsyncWritePlugin {
    * and the max batch size is reached.
    *
    * <p>This call is protected with a circuit-breaker.
+   *
+   * @return a CompletionStage that will be completed when the write is done (in Pekko 1.x, this was
+   *     a Scala Future)
    */
-  Future<Iterable<Optional<Exception>>> doAsyncWriteMessages(Iterable<AtomicWrite> messages);
+  CompletionStage<Iterable<Optional<Exception>>> doAsyncWriteMessages(
+      Iterable<AtomicWrite> messages);
```
He-Pin, line 78: Should this be a Java Iterable？

PR #2096, opened 2025-08-26
feat: Add terminateAndAwait to ActorSystem.

He-Pin:
refs: https://github.com/apache/pekko/issues/2093

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
@@ -678,6 +678,32 @@ abstract class ActorSystem extends ActorRefFactory with ClassicActorSystemProvid
   def terminate(): Future[Terminated]
 
   /**
+   * Terminates this actor system and blocks the current thread until it is terminated,
+   * waiting at most the given duration.
+   *
+   * @param atMost the maximum duration to wait
+   * @throws java.util.concurrent.TimeoutException if the wait timed out
+   * @throws InterruptedException if the current thread is interrupted while waiting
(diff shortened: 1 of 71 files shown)


actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
```diff
@@ -675,7 +675,36 @@ abstract class ActorSystem extends ActorRefFactory with ClassicActorSystemProvid
...
+   *
+   * @param atMost the maximum duration to wait
+   * @throws java.util.concurrent.TimeoutException if the wait timed out
+   * @throws InterruptedException if the current thread is interrupted while waiting
+   */
+  def terminateAndAwait(atMost: Duration): Unit = {
+    import scala.concurrent.Await
+    terminate()
+    Await.result(whenTerminated, atMost)
+  }
```
He-Pin, line 691: Added two methods for both Java and Scala

He-Pin:
A close is a `terminateAndWait(Duration.Inf)`

He-Pin:
@pjfanning @raboof wdyt？

pjfanning:
Let's see what alternative proposals are made before choosing one. In this PR, I would prefer if the classic ActorSystem terminate was not changed so that existing users won't have a migration problem.

He-Pin:
OK, I was alias it with the typed one.

He-Pin:
@pjfanning I changed back to the old `terminate`

He-Pin:
@mdedetrich Wdyt, if I change the `terminateAndAwait` to `close()`?

PR #2097, opened 2025-08-26
feat: Add Untyped actor with stash for Java 21

He-Pin:
Motivation:
With will works better than the AbstractActor with Java 21's pattern matching

### DIFF
--- a/actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/StashJavaAPITestUntypedActors.java
+++ b/actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/StashJavaAPITestUntypedActors.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+/*
+ * Copyright (C) 2018-2022 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package org.apache.pekko.actor;
+
+import static org.junit.Assert.assertEquals;
+
+public class StashJavaAPITestUntypedActors {
+    private static int testReceive(Object msg, int count, ActorRef sender, ActorRef self, UnrestrictedStash stash) {
+        switch (msg) {
+            case String s when count < 0 -> {
+                sender.tell(s.length(), self);
+                return count;
+            }
+            case String ignored when count == 2 -> {
+                stash.unstashAll();
+                return -1;
+            }
+            case String ignored -> {
+                stash.stash();
+                return count + 1;
+            }
+            case Integer value -> {
+                assertEquals(5, value.intValue());
+                return count;
+            }
+            default -> {
+                return count;
+            }
+        }
+    }
+
+    public static class WithStash extends UntypedAbstractActorWithStash {
+        int count = 0;
+
+        @Override
+        public void onReceive(final Object message) throws Throwable {
+            count = testReceive(message, count, getSender(), getSelf(), this);
+        }
+    }
+
+    public static class WithUnboundedStash extends UntypedAbstractActorWithUnboundedStash {
+        int count = 0;
+
+        @Override
+        public void onReceive(final Object message) throws Throwable {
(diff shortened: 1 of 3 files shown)


pjfanning:
I'm not sure that we need this. Can't we just document that Java 21+ users might want to mix in the Stash? I'd really prefer god documentation over adding more base classes. It is not hard for someone with good docs to learn to mix in the Stash themselves.

He-Pin:
But you can not mixing it from Java, Pure Java.

actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/StashJavaAPITestUntypedActors.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
```
pjfanning, line 3: should this not have the normal Apache header - was it really copied from Akka like this header claims?

actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/UntypedStashJavaAPI.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
```
pjfanning, line 4: should this not have the normal Apache header - was it really copied from Akka like this header claims?

actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/StashJavaAPITestUntypedActors.java
He-Pin: I adjusted some code from it, maybe that's fine?

actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/UntypedStashJavaAPI.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.apache.pekko.actor;
+
+import org.apache.pekko.testkit.PekkoJUnitActorSystemResource;
+import org.apache.pekko.testkit.TestProbe;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.scalatestplus.junit.JUnitSuite;
+
+public class UntypedStashJavaAPI extends JUnitSuite {
```
pjfanning, line 22: can this test really need Java 21+? Can we add some tests that don't need Java 21 anyway? I can understand that the tests that use the new `switch` support need Java 21+ but Stash works with all Java versions.

actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/StashJavaAPITestUntypedActors.java
pjfanning: if this copies important code from an existing Akka origin class then the header is best left as is

actor-tests/src/test/java-jdk21-only/org/apache/pekko/actor/UntypedStashJavaAPI.java
He-Pin: there are tests there for receiveBuilder.

In Java 17, we need a preview flag to enable it, that's why's it's Java 21.

actor/src/main/scala/org/apache/pekko/actor/AbstractActor.scala
```diff
@@ -363,6 +363,13 @@ abstract class UntypedAbstractActor extends AbstractActor {
  */
 abstract class AbstractLoggingActor extends AbstractActor with ActorLogging
 
+/**
+ * Java API: compatible with lambda expressions
+ *
+ * Actor base class that mixes in logging into the Actor.
+ */
```
pjfanning, line 372: Can you put `@since 2.0.0` in Javadoc of the new classes?

He-Pin: Thanks, seems like we now need both new methods and classes annotated.

pjfanning (approved):
lgtm

PR #2104, opened 2025-08-30
chore: Test 3.3.7-RC1

He-Pin:


He-Pin:
https://github.com/scala/scala3/issues/23842
https://github.com/scala/scala3/issues/23843

stream/src/main/scala/org/apache/pekko/stream/impl/ActorRefBackpressureSource.scala
```diff
@@ -80,6 +80,7 @@ private object ActorRefBackpressureSource {
             }
             tryPush()
           }
+        case _ => throw new RuntimeException() // compiler exhaustiveness check pleaser
```
pjfanning, line 83: could you put messages in the exceptions just in case? - I know the exceptions are unlikely but it would help if they ever did happen to be able track down why they were thrown

He-Pin: I can ,but maybe do it in another PR, there are many of them.

pjfanning: sure - there are quite a few like this already, so I guess this is consistent with existing code

He-Pin:
@pjfanning I think this is ready.

pjfanning:
@raboof does merging this complicate your RC? I am ok with it. I'll leave it to you if you want to merge it or delay it till after the RC.

PR #2105, opened 2025-08-30
chore: Remove left org.apache.pekko.japi.Function in doc

He-Pin:
Motivation:
It's missing when remove `org.apache.pekko.japi.Function`


pjfanning (approved):
lgtm

PR #2106, opened 2025-08-30
chore: Remove SecurityManager usage.

He-Pin:
Motivation:
Refs: https://github.com/apache/pekko/issues/1970


actor/src/main/java/org/apache/pekko/util/ClassContext.java
```diff
@@ -0,0 +1,18 @@
...
+    throw new UnsupportedOperationException("Cannot instantiate utility class");
+  }
+
+  private static final Function<Stream<StackWalker.StackFrame>, Class<?>[]> CLASS_STACK_WALKER =
+      frames -> frames.map(StackWalker.StackFrame::getDeclaringClass).toArray(Class<?>[]::new);
+
+  public static Class<?>[] getClassStack() {
+    return StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE)
+        .walk(CLASS_STACK_WALKER);
+  }
```
He-Pin, line 40: much easier to write it in Java, I have a problem when supporting Scala 2.12

He-Pin:
@raboof Do you know why it passed on my local env with 3.3.6 and Java 17

actor/src/main/java/org/apache/pekko/util/ClassContext.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+public final class ClassContext {
+  private ClassContext() {
+    throw new UnsupportedOperationException("Cannot instantiate utility class");
+  }
+
+  private static final Set<StackWalker.Option> OPTIONS =
+      Set.of(StackWalker.Option.RETAIN_CLASS_REFERENCE, StackWalker.Option.SHOW_HIDDEN_FRAMES);
```
He-Pin, line 33: `SHOW_HIDDEN_FRAMES` is needed, otherwise , zhe `$$Lambda` is missing.

He-Pin:
Should this be backport to 1.2.0 @pjfanning

pjfanning:
I'd prefer not to backport this. SecurityManager still works. It's just good to future proof 2.0.x by removing the use there.

He-Pin:
thanks, I just checked, it just a noop on jdk 25.

pjfanning:
> thanks, I just checked, it just a noop on jdk 25.

If we find issues where users can't use Pekko with LTS versions of Java, then we might need to make releases with changes like this. If users can readily configure Java to avoid the issue then that becomes less of issue. For me, this is a wait and see. My gut feeling is that Pekko 1.x as is will work ok with Java 25.
We have unit test issues with Java 24 and 25 but I think those issues are caused by compiling with Java 24 and 25. So far, it appears that you can use Pekko with Java 24 and 25 if Pekko is compiled with an older Java compiler. We have issues labeled with `java24` for these compile/test issues.

He-Pin:
Netty just bumped their ci to Java 24 too https://github.com/netty/netty/pull/15102

PR #2111, opened 2025-08-30
Fix pekkoImport() range

rovarga:
The range does not reflect current version, update it to [2,2.1).

Signed-off-by: Robert Varga <robert.varga@pantheon.tech>

### DIFF
--- a/project/OSGi.scala
+++ b/project/OSGi.scala
@@ -143,7 +143,7 @@ object OSGi {
       "!scala.util.parsing.*",
       scalaImport(scalaVersion),
       "*")
-  def pekkoImport(packageName: String = "org.apache.pekko.*") = versionedImport(packageName, "1.1", "1.2")
+  def pekkoImport(packageName: String = "org.apache.pekko.*") = versionedImport(packageName, "2", "2.1")
   def configImport(packageName: String = "com.typesafe.config.*") = versionedImport(packageName, "1.4.0", "1.5.0")
   def scalaImport(version: String) = {
     val packageName = "scala.*"


He-Pin (approved):
I see, it should be 2.0 at least now.

project/OSGi.scala
```diff
@@ -143,7 +143,7 @@ object OSGi {
       "!scala.util.parsing.*",
       scalaImport(scalaVersion),
       "*")
-  def pekkoImport(packageName: String = "org.apache.pekko.*") = versionedImport(packageName, "1.1", "1.2")
+  def pekkoImport(packageName: String = "org.apache.pekko.*") = versionedImport(packageName, "2", "2.1")
```
pjfanning, line 146: 2.0, 2.1 might be better

He-Pin: btw, I'm not sure if we should still keep osgi in 2.0.0, at least Akka dropped it and seems no much complain.

pjfanning: https://github.com/apache/pekko/pull/705 - but let's assume that that is not agreed yet

pjfanning:
@rovarga it would be good to keep the version change in its own PR instead of merging other changes here - we may end up cherry picking some changes to other version branches but we can't do that if you change version.
My suggestion is to leave this as version change only and to look at the comments that we have raised about the version strings. Reopen #2112 and use that for all the import fixes.

pjfanning (changes requested):
2.0, 2.1 might be better

pjfanning:
@rovarga I'm going to close this due to https://github.com/apache/pekko/pull/2313

Feel free to reopen this or raise an issue if that is insufficient

PR #2117, opened 2025-08-31
chore: Test if materializerPromise is already completed first.

He-Pin:
Extract from https://github.com/apache/pekko/pull/1666


stream/src/main/scala/org/apache/pekko/stream/SystemMaterializer.scala
```diff
@@ -96,7 +96,11 @@ final class SystemMaterializer(system: ExtendedActorSystem) extends Extension {
 
   val materializer: Materializer = {
     // block on async creation to make it effectively final
-    Await.result(systemMaterializerPromise.future, materializerTimeout.duration)
+    val future = systemMaterializerPromise.future
+    future.value match {
+      case Some(value) => value.get
+      case _           => Await.result(future, materializerTimeout.duration)
+    }
```
He-Pin, line 103: I think this can improve a little performance

PR #2119, opened 2025-08-31
chore: Remove the deprecated annotation from recoverWith

He-Pin:
Motivation:
These methods are actually ok, the removal of deprecated is incomplete.


He-Pin:
@pjfanning I want to backport this to 1.2.0 too

pjfanning:
I am not completely against undeprecating this. This has not been removed in main branch because something must still use it - but would it be feasible to check if the existing uses could with be changed to use recoverWithRetries?

He-Pin:
The method is actually fine, you see I removed the depreciation in pekko and akka once, but I forgot some place

PR #2120, opened 2025-08-31
feat: Add Flow/Source#onErrorResume for javadsl.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/2116

The current `recoverWith` is using a Scala partial function, which is hard to use from the Java side, with this, we can achieve the same as `reactor-core/rxjava` friendly to Java developers.

The `onErrorComplete` and `onErrorResume` are widely used in MCP Java SDK.

Modification:
1. Add `onErrorResume` operator to both Source and Flow, javadsl only
2. Adjust the current FlowTest, which was testing the Source dsl, but now Flow DSL.
3. Add both Flow and Source testing for `onErrorResume` and `onErrorComplete`

Result:
Easy to use for Java developers.

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/onErrorResume.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/onErrorResume.md
@@ -1,0 +1,34 @@
+# onErrorResume
+
+Allows transforming a failure signal into a stream of elements provided by a factory function.
+
+@ref[Error handling](../index.md#error-handling)
+
+This method is Java API only, use @ref[recoverWith](recoverWith.md) in Scala.
+
+## Signature
+
+@apidoc[Source.onErrorResume](Source) { java="#onErrorResume(org.apache.pekko.japi.function.Function)" }<br>
+@apidoc[Source.onErrorResume](Source) { java="#onErrorResume(java.lang.Class,org.apache.pekko.japi.function.Function)" }<br>
+@apidoc[Source.onErrorResume](Source) { java="#onErrorResume(org.apache.pekko.japi.function.Predicate,org.apache.pekko.japi.function.Function)" }<br>
+@apidoc[Flow.onErrorResume](Flow) { java="#onErrorResume(org.apache.pekko.japi.function.Function)" }<br>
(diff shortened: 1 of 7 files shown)


stream-tests/src/test/java/org/apache/pekko/stream/javadsl/FlowTest.java
```diff
@@ -1338,13 +1338,32 @@ public void mustBeAbleToOnErrorComplete() {
                 return elem;
               }
             })
-        .onErrorComplete()
+        .via(Flow.of(Integer.class).onErrorComplete())
```
pjfanning, line 1341: why does this need to change?

docs/src/main/paradox/stream/operators/Source-or-Flow/onErrorResume.md
```diff
@@ -0,0 +1,32 @@
+# onErrorResume
+
+Allows transforming a failure signal into a stream of elements provided by a factory function.
```
pjfanning, line 3: Should you mention that this is Java Only, that Scala users should use `recoverWith`?

stream-tests/src/test/java/org/apache/pekko/stream/javadsl/FlowTest.java
He-Pin: Before, the tests is testing the method on Source, But the File is FlowTest

docs/src/main/paradox/stream/operators/Source-or-Flow/onErrorResume.md
```diff
@@ -0,0 +1,34 @@
...
+
+@ref[Error handling](../index.md#error-handling)
+
+This method is Java API only, use @ref[recoverWith](recoverWith.md) in Scala.
```
He-Pin, line 7: @pjfanning Done with this line.

pjfanning (approved):
lgtm

He-Pin:
@pjfanning Can we also backport this to 1.2.1?

pjfanning:
Not usual to add new features in a patch release. I'm not 100% against but we have other PMC members who I would suspect would have more stringent views.

He-Pin:
As this will help Java developers' experience, if they are working with MCP or etc, but anyway, they can use the `Match` to build a PartialFunction too.

Issue #2125, opened 2025-09-02
Drop ssl-config


He-Pin:
Drop https://github.com/lightbend/ssl-config

pjfanning:
@He-Pin can you provide some background as to why we should drop this?

He-Pin:
@pjfanning It's been marked deprecated, and Akka dropped that too , years ago.

PR #2127, opened 2025-09-02
remove ssl-config dependency

pjfanning:
* see #2125 
* the code that uses ssl-confif is deprecated - this PR removes that deprecated code


stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsModule.scala
```diff
@@ -37,8 +36,8 @@ import pekko.util.ByteString
     cipherOut: Outlet[ByteString],
     shape: BidiShape[SslTlsOutbound, ByteString, ByteString, SslTlsInbound],
     attributes: Attributes,
-    createSSLEngine: ActorSystem => SSLEngine, // ActorSystem is only needed to support the PekkoSSLConfig legacy, see #21753
-    verifySession: (ActorSystem, SSLSession) => Try[Unit], // ActorSystem is only needed to support the PekkoSSLConfig legacy, see #21753
+    createSSLEngine: () => SSLEngine,
+    verifySession: SSLSession => Try[Unit],
```
He-Pin, line 40: I think this can be a Try[Done] too.

He-Pin (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsModule.scala
pjfanning: I'd prefer to treat that as a separate change. `Try[Unit]` is the existing return type

He-Pin: Yes, but that's an internal class, doesn't matter anyway.

PR #2128, opened 2025-09-02
Annotate PekkoLoggerBridge as InternalApi

pjfanning:
Added internal API annotations and documentation for PekkoLoggerBridge.

Will be removed as part of #2127 (in Pekko 2.0.0). So let's tidy up Pekko 1.2.1 to ensure that users know the class was never part of public API.

### DIFF
--- a/stream/src/main/scala/com/typesafe/sslconfig/pekko/util/PekkoLoggerBridge.scala
+++ b/stream/src/main/scala/com/typesafe/sslconfig/pekko/util/PekkoLoggerBridge.scala
@@ -24,6 +24,8 @@ import com.typesafe.sslconfig.util.{ LoggerFactory, NoDepsLogger }
 /**
  * Internal Usage only. Will be removed in Pekko 2.0.0.
  */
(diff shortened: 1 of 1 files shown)


stream/src/main/scala/com/typesafe/sslconfig/pekko/util/PekkoLoggerBridge.scala
```diff
@@ -15,18 +15,27 @@ package com.typesafe.sslconfig.pekko.util
...
 
   override def apply(name: String): NoDepsLogger =
     new PekkoLoggerBridge(system.eventStream, name, classOf[DummyClassForStringSources])
 }
 
+/**
+ * Internal Usage only. Will be removed in Pekko 2.0.0.
+ */
+@InternalApi
 class PekkoLoggerBridge(bus: EventStream, logSource: String, logClass: Class[_]) extends NoDepsLogger {
```
He-Pin, line 43: Should we add deprecated annotation here too?

pjfanning: done

He-Pin (approved):
lgtm

PR #2129, opened 2025-09-02
chore: Remove the deprecated onDownstreamFinish method in GraphStage

He-Pin:
Motivation:

This method should and can be removed completely, it was deprecated.

Modification:

Remvoe the private onDownstreamFinish method

Result:
Clean code and better performance.


stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
```diff
@@ -699,10 +698,6 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
 
   // Variable used from `OutHandler.onDownstreamFinish` to carry over cancellation cause in cases where
```
pjfanning: can you remove this comment and next line too?

pjfanning (approved):
lgtm

Issue #2080, opened 2025-08-24
Remove Cluster client


He-Pin:
Cluster client can be replaced by grpc/http

pjfanning:
I'm not sure about this. The alternative for user's is not a strightforward switch. I would argue that it might be worth keeping the ClusterClient for its ease of use.

PR #2131, opened 2025-09-03
chore: Drop the deprecated Cluster Client

He-Pin:
Motivation:
The Cluster client is deprecated since Akka 2.6.0 and suggests using Pekko Grpc instead.

Modification:

Delete Cluster client code and doc.

Result:
Clean code.


pjfanning:
I'm not 100% in favour of this. pekko-grpc is pretty complicated to use - this class is somewhat easier to use.

He-Pin:
But seems this will not provide more function than it is now, but with grpc, a user can easy to write an Cluster endpoint.

pjfanning:
Unless we can document how users are meant to use pekko-grpc in detail, I would favour not deleting this code - even possibly undeprecating it.

He-Pin:
Just a gRPC server which receive the message and then forwarding to the cluster:)

pjfanning:
> Just a gRPC server which receive the message and then forwarding to the cluster:)

That is much more complicated than us providing a ClusterClient.

He-Pin:
That's true, so it seems like we can cut a 2.0.0-M1 now? I mean for early adoption.

pjfanning (changes requested):
I would like to get the Pekko team's opinion on #2080 and whether this is better to keep for ease of use.

He-Pin (commented):
Removing deprecated Cluster Client code is a reasonable cleanup. A few points:

1. **Backwards compatibility**: As @pjfanning noted in #2080, there is a question about whether keeping it for ease of use outweighs the maintenance burden. Since it was deprecated since Akka 2.6.0, users have had ample time to migrate.

2. **Clean removal**: The diff looks thorough -- removing all implementation, tests, protobuf definitions, docs, and MiMa filters. 

3. **Changes requested**: @pjfanning has requested changes pending the team's opinion on #2080. It would be good to get a consensus from the PMC/committers on whether to proceed with removal or keep it longer.

4. **Staleness**: Open since September 2025. The team should decide on the #2080 question so this can either move forward or be closed.

He-Pin:
Closing per @pjfanning's feedback. The team's position is that Cluster Client should be retained (and potentially undeprecated) because pekko-grpc is not a straightforward replacement for users. The removal should not proceed without a documented migration path. See also #2080.

PR #2136, opened 2025-09-03
remove more deprecation annotations by making fields less accessible

pjfanning:
these values are used internally but we can stop exposing them publicly


He-Pin:
This ActorMaterializerSettings should be removed.

pjfanning:
> This ActorMaterializerSettings should be removed.

The code is used still internally. I would remove ActorMaterializerSettings if it was easy but it is heavily wired into the code.

pjfanning:
@He-Pin is it ok to merge this? - it is an improvment - I can follow up with future PRs to try to further remove deprecated Materializer code.

PR #2137, opened 2025-09-03
chore: Drop ActorMaterializer class.

He-Pin:
Motivation:

Drop ActorMaterializer , because Materializer has all its methods.


pjfanning (approved):
lgtm

I've been experimenting with pekko-http and seeing if I can compile it with Pekko 2 snapshot jars. This change is going to make that harder because pekko-http has a few usages of ActorMaterializer but in the end, we need to try to rewrite these anyway.

PR #2143, opened 2025-09-04
chore: Make use of japi.function in stream api.

He-Pin:
Motivation:
1. Make it more consistent in the stream java api
2. Make it throw, which will help UX in Java; otherwise, you have to do `try (...) catch (...）{ //ignore}` in Java

refs: https://github.com/apache/pekko/issues/2118


He-Pin:
I will finish that after done this, and I think we can add the `andThen` to `Function2` in scala/scala too

He-Pin:
Thanks @pjfanning for the review, this will help java developers.

PR #2146, opened 2025-09-05
chore: Remove the deprecated overloaded aggregateWithBoundary  method

He-Pin:
Motivation：
Remove the deprecated one for javadsl.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
@@ -4305,54 +4305,18 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
    * @param harvest     this is invoked before emit within the current stage/operator
    * @param emitOnTimer decide whether the current aggregated elements can be emitted, the custom function is invoked on every interval
    */
-  @deprecated("Use the overloaded one which accepts an Optional instead.", since = "1.2.0")
-  def aggregateWithBoundary[Agg, Emit](allocate: function.Creator[Agg],
-      aggregate: function.Function2[Agg, Out, Pair[Agg, Boolean]],
-      harvest: function.Function[Agg, Emit],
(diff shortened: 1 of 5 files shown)


pjfanning:
I'd prefer not to remove recently deprecated methods unless absolutely necessary. I've been concentrating on removing methods deprecated years ago.

He-Pin:
The deprecated method force user to start a timer, which is not right. and gives a very bad practice for java api.

pjfanning:
We've deprecated it. If users continue to use - then that's their issue. We've signaled to them to stop using it.
When you deprecate a method, it is usually regarded as good practice to keep it for a few releases.

He-Pin:
I think we already broke things, so why not do it as much as possible and get a fresh start then?

He-Pin:
1. These methods only affect the Java API, and Pekko doesn't have many Java users.
2. To expand our user base, we need a clean, tidy, and best-practice codebase.
3. The old method forced users to enter a Function, but the new method is much simpler. If users don't care about the time, they can simply enter Optional.empty， and match the scaladsl.
4. This time is a good time, as it would be awkward to clean up later.

mdedetrich:
> We've deprecated it. If users continue to use - then that's their issue. We've signaled to them to stop using it. When you deprecate a method, it is usually regarded as good practice to keep it for a few releases.

We haven't been following this guideline, there have been plenty of recent deprecations in 1.x series (such as 1.2.x) which are being removed in 2.0.0 so at least in this regard the PR is totally fine.

The number of releases isn't the metric that we should be looking at, rather the time period between those releases. A full non milestone 2.0.0 release is still some way out, so users will have plenty of time to handle deprecations.

He-Pin:
The 1.2.x is still maintained, I think 1.2.x will be maintained quite a long time, because its the last version which support Java 8.

He-Pin:
@pjfanning As there will be an 1.3 release, so can we remove it in 2.0.0 now?

pjfanning:
sure, I guess the consensus is to remove these methods in 2.0.0

He-Pin:
@pjfanning I have updated the PR

PR #2147, opened 2025-09-05
feat: Add andThen and compose to japi function

He-Pin:
After we change some methods to Scala, we should add these methods for a better migration path.


pjfanning:
ideally, we should have some basic tests

He-Pin:
But these methods are actually logical, I can have ai do check a round?

He-Pin:
@Copilot Review this pr

pjfanning:
I don't believe that getting an AI agent to check the code today is enough. All code should have tests. They are very important for regression purposes.

He-Pin:
I will add the test in java

Issue #2150, opened 2025-09-05
Optimise ByteString lastIndexOf


pjfanning:
We don't explicitly implement this but we pick up an unoptimised one from scala.Seq.
See #2148

I checked our main Pekko repos and found no uses of ByteString.lastIndexOf

pjfanning:
@He-Pin I can't find anywhere where lastIndexOf is used for ByteStrings. An optimisation might be nice but it is unlikely to help anyone.

He-Pin:
I checked both pekko and pekko-http, seems true, we can defer that optimization for now.

pjfanning:
#2838 merged

PR #2153, opened 2025-09-05
improve ByteString lastIndexOf

pjfanning:
* see #2150 
* WIP
* doesn't yet use SWAR but having the new methods that take Byte inputs instead of the `[B >: Byte]` versions is the biggest win
* needs a jmh benchmark


He-Pin (commented):
Good performance improvement for `ByteString.lastIndexOf`. A few observations:

1. **Specialized `Byte` overload**: Adding the `Byte`-typed overload (avoiding the `[B >: Byte]` generic version) is indeed the biggest win since it avoids boxing/unboxing.

2. **SWAR optimization**: The PR mentions it doesn't yet use SWAR (SIMD Within A Register). That would be a nice follow-up for even better performance on longer ByteStrings.

3. **Missing JMH benchmark**: As noted in the PR body, a JMH benchmark is needed to quantify the improvement. This should be added before merge.

4. **Staleness**: Open since September 2025 with no reviews yet. The implementation looks solid -- the main gap is the benchmark. @pjfanning would you like to finish this up?

pjfanning:
going with #2838

PR #2155, opened 2025-09-06
chore: Use while loop

He-Pin:
Motivation:
Make it performant better in Scala 3

https://github.com/scala/scala3/issues/8880


pjfanning (approved):
lgtm

PR #2157, opened 2025-09-06
chore: make toAttributes a lazy val

He-Pin:
Motivation:
Make it a lazy val to cache the value.


pjfanning (approved):
lgtm

there is a small overhead to `lazy val`s but if it is possible that this code is called more than once - then it is better to use `lazy val`

PR #2158, opened 2025-09-06
add tests for ByteStrings with non-zero startIndex

pjfanning:
* the indexOf code is  affected by the startIndex (where the array backing the ByteString has initial bytes that we are meant to ignore)
* seems best to test that indexOf behaves ok with startIndex > 0


actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
```diff
@@ -701,6 +701,16 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       byteString1.indexOf('a', 0) should ===(0)
       byteString1.indexOf('a', 1) should ===(-1)
 
+      val array = Array[Byte]('x', 'y', 'z', 'a', 'b', 'c')
+      val byteString2 = ByteString1(array, 3, 3)
+      byteString2.indexOf('x', -1) should ===(-1)
```
He-Pin, line 706:
```java
        fromIndex = Math.max(fromIndex, 0);
        if (fromIndex >= toIndex || buffer.capacity() == 0) {
            return -1;
        }
```
In Netty's Bytebuf, the fromIndex is always >=0

pjfanning: our code allows negative values - I'm not going to change that in this in PR - which is about missing test coverage

He-Pin: learn something everyday:)

Issue #2160, opened 2025-09-06
Remove ReentrantGuard


He-Pin:
Motivation:
In Scala 2.x it will create a closure every time, that code is simple, if we can use it with Scala 3 's inline, then that's fine, I think we should just remove the code.

Should deprecate it first in 1.2.x

pjfanning:
I am against this. No justification has been provided. The class is used in only a few places and and it will be hard to test if removing it causes concurrency issues.

https://github.com/search?q=repo%3Aapache%2Fpekko%20ReentrantGuard&type=code

PR #2161, opened 2025-09-06
chore: Remove ReentrantGuard

He-Pin:
It creates a closure every time, making it use less, to write high-performance code.

If you think this can be removed, we can start the deprecation in 1.2.x


pjfanning:
I suppose this is ok. #2160 makes no mention that you intended to just use ReentrantLock directly.
ReentrantGuard seems/seemed useful as it was hard to forget to unlock the lock in every case.
The new code looks correct but every time someone uses ReentrantGuard directly, we need to very closely examine the code to ensure the try/finally code is implemented correctly.

He-Pin:
The old code will just reduce performance with Scala3 inline

PR #2164, opened 2025-09-06
chore: deprecate ReentrantGuard

He-Pin:
Motivation:
This class will be removed in 2.0.0

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/LockUtil.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/LockUtil.scala
@@ -16,7 +16,7 @@ package org.apache.pekko.util
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.locks.ReentrantLock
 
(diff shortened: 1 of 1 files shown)


pjfanning:
@He-Pin why wouldn't we merge this? If the class is removed in 2.0.0, it makes sense to deprecate it in 1.2.1 or 1.3.0.

He-Pin:
I forgot this

He-Pin:
@pjfanning Would like to take a look at this, I want to reduce the pending pr queue.

actor/src/main/scala/org/apache/pekko/util/LockUtil.scala
```diff
@@ -16,6 +16,7 @@ package org.apache.pekko.util
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.locks.ReentrantLock
 
+@deprecated("Use java.util.concurrent.locks.ReentrantLock directly, will be removed in Pekko 2.0.0", "1.2.0")
```
pjfanning:
```suggestion
@deprecated("Use java.util.concurrent.locks.ReentrantLock directly, will be removed in Pekko 2.0.0", "1.2.1")
```

PR #2169, opened 2025-09-07
feat: Add virtualize support for thread-pool-executor

He-Pin:
Motivation:
Add virtualize support for thread-pool-executor

refs: https://github.com/apache/pekko/issues/2163

will rebase on https://github.com/apache/pekko/issues/2171


actor/src/main/resources/reference.conf
```diff
@@ -543,6 +543,18 @@ pekko {
 
         # Allow core threads to time out
         allow-core-timeout = on
+
+        # This config is new in Pekko v1.2.0 and only has an effect if you are running with JDK 21 and above,
+        # When set to `on` but underlying runtime does not support virtual threads, an Exception will throw.
```
pjfanning:
```suggestion
        # When set to `on` but underlying runtime does not support virtual threads, an Exception will be thrown.
```

actor/src/main/resources/reference.conf
```diff
@@ -600,6 +612,17 @@ pekko {
 
       thread-pool-executor {
         fixed-pool-size = 16
+        # This config is new in Pekko v1.2.0 and only has an effect if you are running with JDK 21 and above,
+        # When set to `on` but underlying runtime does not support virtual threads, an Exception will throw.
```
pjfanning:
```suggestion
        # When set to `on` but underlying runtime does not support virtual threads, an Exception will be thrown.
```

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
```diff
@@ -93,8 +93,8 @@ private[dispatch] object VirtualThreadSupport {
       }
       val nameMethod = ofVirtualClass.getDeclaredMethod("name", classOf[String], classOf[Long])
       val factoryMethod = builderClass.getDeclaredMethod("factory")
-      val zero = java.lang.Long.valueOf(0L)
```
pjfanning, line 96: How about declaring a constant in an `object VirtualThreadSupport` with `val zero = java.lang.Long.valueOf(0L)`? Then you avoid boxing in the new code. 0L has to be boxed to a java.util.Long on every call.

He-Pin: @pjfanning https://github.com/apache/pekko/issues/2170

I think we should just drop the counter, wdyt?

actor/src/main/resources/reference.conf
```diff
@@ -543,6 +543,18 @@ pekko {
 
         # Allow core threads to time out
         allow-core-timeout = on
+
+        # This config is new in Pekko v1.2.0 and only has an effect if you are running with JDK 21 and above,
+        # When set to `on` but underlying runtime does not support virtual threads, an Exception will thrown.
```
pjfanning:
```suggestion
        # When set to `on` but the underlying runtime does not support virtual threads, an Exception will be thrown.
```

actor/src/main/resources/reference.conf
```diff
@@ -482,7 +482,7 @@ pekko {
         maximum-pool-size = 32767
 
         # This config is new in Pekko v1.2.0 and only has an effect if you are running with JDK 21 and above,
-        # When set to `on` but underlying runtime does not support virtual threads, an Exception will throw.
+        # When set to `on` but underlying runtime does not support virtual threads, an Exception will thrown.
```
pjfanning:
```suggestion
        # When set to `on` but the underlying runtime does not support virtual threads, an Exception will be thrown.
```

actor/src/main/resources/reference.conf
```diff
@@ -600,6 +612,17 @@ pekko {
 
       thread-pool-executor {
         fixed-pool-size = 16
+        # This config is new in Pekko v1.2.0 and only has an effect if you are running with JDK 21 and above,
+        # When set to `on` but underlying runtime does not support virtual threads, an Exception will thrown.
```
pjfanning:
```suggestion
        # When set to `on` but the underlying runtime does not support virtual threads, an Exception will be thrown.
```

PR #2174, opened 2025-09-07
chore: Make ThreadPoolConfig an internal api

He-Pin:
Motivation:
This should be an internal api
refs: https://github.com/apache/pekko/issues/2171


pjfanning (approved):
seems ok - marking it as InternalApi may not be seen by users who think that this class is  for public use but the chances of people using this class directly seems unlikely

PR #2176, opened 2025-09-07
chore: avoid boxing in virtual thread support

He-Pin:
Motivation:

Avoid boxing in virtual thread support

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
@@ -27,6 +27,7 @@ import scala.util.control.NonFatal
 
 @InternalApi
 private[dispatch] object VirtualThreadSupport {
+  val zero = java.lang.Long.valueOf(0L)
   private val lookup = MethodHandles.publicLookup()
 
   /**
@@ -67,7 +68,7 @@ private[dispatch] object VirtualThreadSupport {
         MethodType.methodType(ofVirtualClass, classOf[String], classOf[Long]))
       // TODO support replace scheduler when we drop Java 8 support
       val factoryMethod = lookup.findVirtual(builderClass, "factory", MethodType.methodType(classOf[ThreadFactory]))
-      builder = nameMethod.invoke(builder, prefix + "-virtual-thread-", 0L)
+      builder = nameMethod.invoke(builder, prefix + "-virtual-thread-", zero)
       factoryMethod.invoke(builder).asInstanceOf[ThreadFactory]
     } catch {
       case NonFatal(e) =>
@@ -85,7 +86,7 @@ private[dispatch] object VirtualThreadSupport {
       val ofVirtualClass = ClassLoader.getSystemClassLoader.loadClass("java.lang.Thread$Builder$OfVirtual")
       val ofVirtualMethod = classOf[Thread].getDeclaredMethod("ofVirtual")
       var builder = ofVirtualMethod.invoke(null)
-      if (executor != null) {
+      if (executor ne null) {
         val clazz = builder.getClass
         val field = clazz.getDeclaredField("scheduler")
         field.setAccessible(true)
@@ -93,7 +94,6 @@ private[dispatch] object VirtualThreadSupport {
       }
       val nameMethod = ofVirtualClass.getDeclaredMethod("name", classOf[String], classOf[Long])
       val factoryMethod = builderClass.getDeclaredMethod("factory")
-      val zero = java.lang.Long.valueOf(0L)
       builder = nameMethod.invoke(builder, prefix + "-virtual-thread-", zero)
       factoryMethod.invoke(builder).asInstanceOf[ThreadFactory]
     } catch {


actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
```diff
@@ -67,7 +68,7 @@ private[dispatch] object VirtualThreadSupport {
         MethodType.methodType(ofVirtualClass, classOf[String], classOf[Long]))
       // TODO support replace scheduler when we drop Java 8 support
       val factoryMethod = lookup.findVirtual(builderClass, "factory", MethodType.methodType(classOf[ThreadFactory]))
-      builder = nameMethod.invoke(builder, prefix + "-virtual-thread-", 0L)
+      builder = nameMethod.invoke(builder, prefix + "-virtual-thread-", zero)
```
raboof, line 71: I find it hard to believe the boxing would be a bottleneck/hotspot in the creation of thread factories, but OK I suppose.

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
```diff
@@ -85,15 +86,14 @@ private[dispatch] object VirtualThreadSupport {
       val ofVirtualClass = ClassLoader.getSystemClassLoader.loadClass("java.lang.Thread$Builder$OfVirtual")
       val ofVirtualMethod = classOf[Thread].getDeclaredMethod("ofVirtual")
       var builder = ofVirtualMethod.invoke(null)
-      if (executor != null) {
+      if (executor ne null) {
```
raboof, line 89: Does this really make any difference?

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
He-Pin: Actually,  I was wondering should we have a thread I'd for virtual thread....

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
He-Pin: Yes，it was a function call.but now just a bytecode

He-Pin:
OK, seems the backport can only be bug fix :(

actor/src/main/scala/org/apache/pekko/dispatch/VirtualThreadSupport.scala
raboof: Hmm, that's interesting.

Indeed the spec at https://scala-lang.org/files/archive/spec/3.4/12-the-scala-standard-library.html seems to say if the left-hand size is not null it should call `equals` on the left-hand side, but if I actually try it I do get a direct `null` check. I wonder where/how that's optimized out.

mdedetrich: Indeed I thought it was the same but `ne` is faster as it doesn't have to check as many things.

Not sure if it's a bottleneck and/or if JVM optimises it out but if @He-Pin finds something it's usually because it's being hit somewhere

He-Pin: I was find that with IDEA 's show bytecode

mdedetrich:
> I see this is a backport of changes already on `main`. I'm not sure we should be backporting such subtle performance tweaks tbh.

Uhh, the main branch is currently at 2.00 which is a long time away from being released, we are actually backporting all the time now from 2.0.0 to 1.2.x

mdedetrich:
@pjfanning @raboof @He-Pin Now that `main` is at 2.0.0 I think we need to setup a process given that some PRs are creating confusion.

If there is a change that's only specific to 2.0.0 such as removing deprecated methods, dropping Scala 2.12 etc etc than those PRs should target `main`.

On the other hand, general changed not specific to 2.0.0 should target the 1.2.x/1.3.x branches and be forward ported to `main`.

wdyt?

raboof:
> @pjfanning @raboof @He-Pin Now that `main` is at 2.0.0 I think we need to setup a process given that some PRs are creating confusion.
> 
> If there is a change that's only specific to 2.0.0 such as removing deprecated methods, dropping Scala 2.12 etc etc than those PRs should target `main`.
> 
> On the other hand, general changed not specific to 2.0.0 should target the 1.2.x/1.3.x branches and be forward ported to `main`.
> 
> wdyt?

Should we open a separate thread for this on https://github.com/apache/pekko/discussions or the list, instead of discussion it on an only semi-related issue?

The risk with 'forward-porting' is that we might forget, causing a 'regression' in 2.x . Forgetting a backport is less problematic. For that reason I'd prefer first targeting main and then backporting, even for 'general' changes.

mdedetrich:
> > @pjfanning @raboof @He-Pin Now that `main` is at 2.0.0 I think we need to setup a process given that some PRs are creating confusion.
> 
> > 
> 
> > If there is a change that's only specific to 2.0.0 such as removing deprecated methods, dropping Scala 2.12 etc etc than those PRs should target `main`.
> 
> > 
> 
> > On the other hand, general changed not specific to 2.0.0 should target the 1.2.x/1.3.x branches and be forward ported to `main`.
> 
> > 
> 
> > wdyt?
> 
> 
> 
> Should we open a separate thread for this on https://github.com/apache/pekko/discussions or the list, instead of discussion it on an only semi-related issue?

Yes, will do!

mdedetrich:
@He-Pin If this is a yet to be unreleased change it's fine to backport it

He-Pin:
@mdedetrich @raboof I think we can learn the Playframework, where @mkurz has a bot, and he just need the comment, and then the bot will do the backport.

mkurz:
> @mdedetrich @raboof I think we can learn the Playframework, where @mkurz has a bot, and he just need the comment, and then the bot will do the backport.

We "installed" Mergify (https://mergify.com/) as GitHub app. (In Pekko you should see it later here: https://github.com/apache/pekko/settings/installations).
We set up a global mergify config in our `.github` repo: https://github.com/playframework/.github/blob/main/.github/mergify.yml

The global config can then be extended for each repo, for example our main repo: https://github.com/playframework/playframework/blob/main/.github/mergify.yml

The way we make use of mergify is making comments like `@Mergifyio backport 3.0.x` or `@Mergifyio rebase`.

raboof:
I see @He-Pin opened https://github.com/apache/pekko/discussions/2180, let's move the discussion there.

pjfanning:
The ASF does not let us use GitHub Settings directly. They do not allow us to add GitHub apps.
These are for Security reasons.
We have a yml script where we can control some of the settings but we can only enable what the ASF Infra team supports.
https://github.com/apache/infrastructure-asfyaml/blob/main/README.md

pjfanning:
@He-Pin Could this be rebased? I think some of these changes already appeared in a recent PR that was merged in last few days.

raboof:
> The ASF does not let us use GitHub Settings directly. They do not allow us to add GitHub apps. These are for Security reasons. We have a yml script where we can control some of the settings but we can only enable what the ASF Infra team supports. https://github.com/apache/infrastructure-asfyaml/blob/main/README.md

This is true, but if there's any functionality that would be particularly helpful for us then we could definitely look into how we could make that happen. While it's possible mergify specifically might require too many privileges, there's little a backport plugin needs to do that requires privileges, so I'm sure we could find some alternative. So far creating the backports manually hasn't felt like too much of a burden AFAIK, though.

PR #2194, opened 2025-09-11
chore: Deprecated japi.option

He-Pin:
Deprecate it in 1.2.x


actor/src/main/scala/org/apache/pekko/japi/JavaAPI.scala
```diff
@@ -305,6 +305,7 @@ object Util {
 
   // TODO in case we decide to pull in scala-java8-compat methods below could be removed - https://github.com/akka/akka/issues/16247
 
+  @deprecated("This method will be removed in Pekko 2.0.0", "1.2.0")
```
pjfanning:
```suggestion
  @deprecated("This method will be removed in Pekko 2.0.0", "1.2.1")
```

PR #2195, opened 2025-09-12
add back BackoffSupervisor.calculateDelay due to bin compat issues

pjfanning:
See https://github.com/apache/pekko-persistence-cassandra/issues/305 and #1940 

Aim for main branch is to forward fit the release note - and to move the mima filter to 2.0.x with a note about the bug in Pekko 1.2.0.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/pattern/BackoffSupervisor.scala
+++ b/actor/src/main/scala/org/apache/pekko/pattern/BackoffSupervisor.scala
@@ -312,7 +312,7 @@ object BackoffSupervisor {
   @InternalApi
   private[pekko] case class ResetRestartCount(current: Int) extends DeadLetterSuppression
 
- /**
+  /**
    * INTERNAL API
    *
    * Calculates an exponential back off delay.


actor/src/main/scala/org/apache/pekko/pattern/BackoffSupervisor.scala
```diff
@@ -311,6 +311,22 @@ object BackoffSupervisor {
    */
   @InternalApi
   private[pekko] case class ResetRestartCount(current: Int) extends DeadLetterSuppression
+
+  /**
+   * INTERNAL API
+   *
+   * Calculates an exponential back off delay.
+   *
+   * Was removed in 1.2.0 but added back in 1.2.1 for binary compatibility reasons.
```
He-Pin, line 320: We should point to the new methods in 2.0.0

He-Pin:
Seems like some internal api is been used across akka projects :(

pjfanning:
> Seems like some internal api is been used across akka projects :(

we can't assume package private code at [pekko] level is only used in this repo - it can be used in any of our repos

actor/src/main/scala/org/apache/pekko/pattern/BackoffSupervisor.scala
pjfanning: I do - in the deprecation message below

He-Pin:
seems like we will do a release soon

Issue #2197, opened 2025-09-12
are we going to replace java.util.function params with pekko.util.function in Pekko 2.0.0?


pjfanning:
I don't code much in Java. Pekko 2.0.0 dev already switched a lot of Java functions to pekko.util.function Functions.

We still have quite a lot of java.util.function Functions

https://github.com/search?q=repo%3Aapache%2Fpekko+java.util.function&type=code&p=1

It would be good to work out sooner rather than later, which, if any, of these usages are going to be rewritten for Pekko 2.0.0. If we do a 2.0.0-M0 release soon and then start doing even more breaking changes, this will mean that we end up with a 2.0.0-M1 which is still unstable - dragging out an ultimate 2.0.0 release.

So, I'm saying that we need to finalise all the changes now but it would be good to have some inkling as to what to expect later in the 2.0.0 dev cycle.

pjfanning:
@He-Pin could you have a look at the remaining Java functions in Pekko main branch and work out if any of them need to change to Pekko functions?

He-Pin:
I will take care of them tomorrow

He-Pin:
@pjfanning I think I can continue this once https://github.com/apache/pekko/pull/1986 got merged.

mdedetrich:
So there are a couple of things flying around here. One seems to be a general cleanup that should be done as a result of Pekko 2.0.0 allowing breaking changes, this is absolutely fine.

The more critical change that has been silently been pushed into PR's without proper discussion is how we treat lambda functions that can throw exceptions, this is actually quite important because its part of API design and there is historical precedent to how it has worked in both Akka and in Pekko. There are PR's that have been merged which changes this behaviour and it to be blunt, it should have been discussed before changing the behaviour.

Both Akka and Pekko have historically and consistently designed their API in such a way that have you 2 types of functions. "pure" functions which cannot throw and "non pure" functions which can throw. Operators which do resource acquisition (such as `mapWithResource`) have used functions which allow throwing because you can hit cases where if you try to create a file (as part of a resource) it can throw an exception because the file failed to create.

[...]

PR #2199, opened 2025-09-13
chore: Add boolean supplier that can throws.

He-Pin:
Motivation:
Better Java dsl support.

on top of https://github.com/apache/pekko/pull/2198

refs: https://github.com/apache/pekko/issues/2197

### DIFF
--- a/actor/src/main/java/org/apache/pekko/japi/pf/FSMStateFunctionBuilder.java
+++ b/actor/src/main/java/org/apache/pekko/japi/pf/FSMStateFunctionBuilder.java
@@ -54,7 +54,7 @@ public class FSMStateFunctionBuilder<S, D> {
         FSM.Event.class,
         new Predicate<FSM.Event>() {
           @Override
-          public boolean test(FSM.Event e) {
+          public boolean test(FSM.Event e) throws Exception {
             boolean res = true;
             if (eventOrType != null) {
               if (eventOrType instanceof Class eventType) {
--- a/actor-tests/src/test/java/org/apache/pekko/japi/FunctionLawTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/japi/FunctionLawTest.java
@@ -53,7 +53,7 @@ public class FunctionLawTest {
   }
 
   @Test
-  public void testPredicateNegate() {
+  public void testPredicateNegate() throws Exception {
     final Predicate<Object> alwaysTrue = o -> true;
     final Predicate<Object> alwaysFalse = alwaysTrue.negate();
     Assert.assertTrue(alwaysTrue.test(new Object()));
@@ -62,7 +62,7 @@ public class FunctionLawTest {
   }
 
   @Test
-  public void testPredicate2Negate() {
+  public void testPredicate2Negate() throws Exception {
     final Predicate2<Object, Object> alwaysTrue = (o1, o2) -> true;
     final Predicate2<Object, Object> alwaysFalse = alwaysTrue.negate();
     Assert.assertTrue(alwaysTrue.test(new Object(), new Object()));
--- a/actor/src/main/java/org/apache/pekko/japi/function/BooleanSupplier.java
+++ b/actor/src/main/java/org/apache/pekko/japi/function/BooleanSupplier.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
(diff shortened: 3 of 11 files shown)


actor/src/main/java/org/apache/pekko/japi/pf/FSMStateFunctionBuilder.java
```diff
@@ -54,7 +54,7 @@ private FSMStateFunctionBuilder<S, D> erasedEvent(
         FSM.Event.class,
         new Predicate<FSM.Event>() {
           @Override
-          public boolean test(FSM.Event e) {
+          public boolean test(FSM.Event e) throws Exception {
```
He-Pin, line 57: I will rewrite these kinds to lambda once https://github.com/apache/pekko/pull/1986 gets merged

mdedetrich (changes requested):
I am not sure what the intention of this PR is, in these cases you should actually **NOT** allow throwing exceptions because we are just supplying a predicate.

To be clear, the only cases where we should have functions that allow throwing exceptions is in cases where we can have resource acquasition, i.e. `mapWithResource` where you need to supply a function creates a resource (lets say creating a file) and that creation of a file can fail with an exception.

He-Pin:
@mdedetrich Fine,take the https://github.com/apache/pekko/issues/2197 from here.

1. The current implementation does not match the Scala doc of both `Predicate` and `Predicate2·
2. The Java BooleanSupplier can not throw, but you can actually inside a Scala's PartialFunction.

image

He-Pin:
By the way, you can never limit a user; do not throw an exception, as a user can always do something like sneakyThrow, with this, we can use some function which can throw inside a lambda

mdedetrich:
> By the way, you can never limit a user; do not throw an exception, as a user can always do something like sneakyThrow, with this, we can use some function which can throw inside a lambda

No we should definitely limit what users can do, its part of good API design. Throwing exceptions in predicates make zero sense and leads to undefined behaviour in a lot of cases

He-Pin:
I think you need to refresh your knowledge about lambda, please check some other project first, eg: https://github.com/smallrye/smallrye-mutiny

He-Pin:
@mdedetrich  There are some Java api should be changed. In the persistence plugin, I will leave that to you.

mdedetrich:
> I think you need to refresh your knowledge about lambda, please check some other project first, eg: https://github.com/smallrye/smallrye-mutiny

Linking me to some other projects without an explanation isn't very helpful. There is a reason why the pekko API originally created their own versions of functional interfaces like `Creator`, its because `Creator` could throw exceptions and in some cases supplying functions that throw exceptions makes sense (again resource acquisition).

But if you have functions that are basic predicates, throwing functions especially in the context of pekko-streams can produce weird results. It also doesn't make any sense in terms of API design, what does an exception mean when thrown during a predicate? Should it be false? Should it be true? Should we cancel altogether? If the user is dealing with exceptions they should explicitly catch and map the exception to a `true`/`false`.

He-Pin:
I don't want to explain any more. You understand RPC, right? You understand rate limiting, right? Can I use distributed rate limiting? Distributed rate limiting can use a Predicate abstraction, but will the distributed rate limiter throw an exception? I've decided to move on to other things and don't want to be blocked by you. This exception won't show up in a lambda scenario. You've been writing in Kotlin so much that you've forgotten how to write in Java.

mdedetrich:
> I don't want to explain any more. You understand RPC, right? You understand rate limiting, right? Can I use distributed rate limiting? Distributed rate limiting can use a Predicate abstraction, but will the distributed rate limiter throw an exception? I've decided to move on to other things and don't want to be blocked by you. This exception won't show up in a lambda scenario. You've been writing in Kotlin so much that you've forgotten how to write in Java.

Again, this is the reason

> what does an exception mean when thrown during a predicate? Should it be false? Should it be true? Should we cancel altogether? If the user is dealing with exceptions they should explicitly catch and map the exception to a `true`/`false`.

Also this is how akka/pekko have historically always worked, only functions used in resource acquisitions (like `mapWithResource`) are the ones that were allowed to throw exceptions

pjfanning:
Let's delay 2.0.0-M1 to remove any time pressure so that we can get a few people to give their ideas in #2197. 
I think everyone has valid points on this and I'm sure that we can find a consensus.

He-Pin:
Other reactive libraries are also acceptable. This makes it easier for Java developers to interpret and call methods that throw exceptions, allowing them to be called normally in lambda. This is a very common problem that does not exist in Kotlin and Scala.

Issue #2185, opened 2025-09-09
Rewrite org.apache.pekko.japi.function.* in Java?


He-Pin:
<img width="722" height="179" alt="Image" src="https://github.com/user-attachments/assets/7a7f745d-6592-4242-b584-b107917f9ab3" />

It renders badly in Java.

He-Pin:
I will go to do this, because we need some boxing free functions too.

mdedetrich:
If we are going to do big decisions like move everything to Java or vice versa to Scala can we please discuss this first before doing it i.e. sneakily putting it in PRs like https://github.com/apache/pekko/pull/2198? Just because you personally think this is a good idea doesn't mean the community agrees. This is a community run project, not a dictatorship by a company like Akka.

Whenever we make a big decision like this, it should be discussed first. For example my suggestion of using sconfig, a discussion was created first at https://github.com/apache/pekko/discussions/2186 and even though I opened a PR it wasn't intended to be merged until a consensus is made (the PR was for evaluation reasons).

He-Pin:
@mdedetrich That's new, not rewriting, ok?

He-Pin:
And in case of exceptions, do you know:
```java
public interface UncheckedBooleanSupplier extends BooleanSupplier {

    @Override
    boolean get();
```
An unchecked one can override the one that throws

mdedetrich:
[...]

PR #2200, opened 2025-09-13
Move IntFunction to scala and don't allow throwing

mdedetrich:
In https://github.com/apache/pekko/pull/2198 we defined an `IntFunction` but there were some issues with the definition

- There is no reason for `IntFunction` to allow throwing, this is used in retry functions for the `delayFunction` and supplying a function that can throw makes no sense and can lead to undefined behaviour
- Moves implementation to scala since thats where all of the other functions sit, there is no reason to pointlessly split our implementations in java and scala, and having it sit in Scala is beneficial if we want to explore native/js. Java users can still use these functions without any problems (it in fact creates the exact same bytecode)


actor/src/main/scala/org/apache/pekko/japi/function/Function.scala
```diff
@@ -152,3 +152,10 @@ trait Creator[+T] extends Serializable {
   @throws(classOf[Exception])
   def create(): T
 }
+
+@nowarn("msg=@SerialVersionUID has no effect")
+@SerialVersionUID(1L)
+@FunctionalInterface
+trait IntFunction[T] extends Serializable {
+  def apply(value: Int): T
```
He-Pin: ? what's the poin of this?

He-Pin:
Check https://github.com/apache/pekko/issues/2185 first

And then why can't the `IntFunction` throw? Are you kidding me?

Then tell why an `Array(index: Int) can throw `IOBE`?

actor/src/main/scala/org/apache/pekko/japi/function/Function.scala
mdedetrich: We already have a place where we define all of our functional interfaces, its in `actor/src/main/scala/org/apache/pekko/japi/function/Function.scala`. In your PR you created another implementation in Java which means we have functional interfaces defined in both Scala and Java for no good reason.

Lets keep all of the functional interfaces defined where they already are in `actor/src/main/scala/org/apache/pekko/japi/function/Function.scala` so its consistent.

He-Pin: The reason is in https://github.com/apache/pekko/issues/2185, ok?
Even Akka-sdk is now Java-first.

mdedetrich:
> Check #2185 first
> 
> And then why can't the `IntFunction` throw? Are you kidding me?
> 
> Then tell why an `Array(index: Int) can throw `IOBE`?

Its not that Int functions can never throw, its that throwing an exception as a supplier function in a pekko stream doesn't make any sense.

actor/src/main/scala/org/apache/pekko/japi/function/Function.scala
He-Pin: These are Java APIs, so shouldn't they be written in Java? Just because they were done in the past doesn't mean they should be done now.

He-Pin:
These are common functions. Thanks. Do you actually write Java on a daily basis? How much Java do you write on a daily basis?

actor/src/main/scala/org/apache/pekko/japi/function/Function.scala
mdedetrich: > The reason is in #2185, ok? Even Akka-sdk is now Java-first.

If you are going to make a decision like move everything to Java (or to Scala) for that matter, you need to discuss it with community first. You can't just sneak it into PR's because you think its a good idea.

Please discuss this with community first, also whether to throw an exception or not is a separate matter from whether functions can throw exceptions.

He-Pin:
And you should never define such a method in Scala , there is already one, In java's stdlib.

mdedetrich:
> And you should never define such a method in Scala , there is already one, In java's stdlib.

Then we should use that one, I will change the PR to do so.

> These are common functions. Thanks. Do you actually write Java on a daily basis? How much Java do you write on a daily basis?

I write Java/Kotlin frequently, and there is a reason why the Java functional interfaces do not allow exception throwing (albeit they should have also provided alternatives that do allow throwing)

He-Pin (changes requested):
There is already a Java's IntFunction in stdlib

mdedetrich:
> There is already a Java's IntFunction in stdlib

Fixed and pushed

pjfanning (approved):
lgtm - I'm ok with this if @He-Pin is happy with this change

He-Pin:
actually we should just do a revert instead of this.

He-Pin (approved):
lgtm

He-Pin:
Although this is not the result I want, I can accept it, reduce the change set help anyway.

PR #2203, opened 2025-09-13
use Scala Option Converter

pjfanning:
* still WIP, I need to reorder many of the imports
* the advantage to using the Scala OptionConverter directly is that the Pekko one does not inline the toScala method
* I intend to leave the Pekko OptionConverter in the code at least for now, maybe I can deprecate it


pjfanning:
I'll come back this later in the weekend.

He-Pin:
@pjfanning I'm working on the Duration one

He-Pin:
@pjfanning We should remove those two classes.

pjfanning:
I would prefer to deprecate in 2.0.0-M1 and maybe remove before full release.

He-Pin:
I think we should delete it, otherwise it will cause trouble when users import or when IDE automatically completes.

PR #2204, opened 2025-09-13
chore: Remove serialversion-remover-plugin

He-Pin:
refs: https://github.com/apache/pekko/issues/2202


mdedetrich:
@raboof Can you look into this? I think you have a better history as to why this plugin exists, I am afraid that by removing it we may break something

He-Pin:
It's not an error now in Scala 3 now.

mdedetrich (approved):
Oh I see, lgtm to me then

PR #2206, opened 2025-09-13
Remove ccompat and organize imports

mdedetrich:
Now that Scala 2.12 is dropped we no longer need an inlined `ccompat`. This PR removes that in place of just using `scala.jdk.CollectionConverters._` directly.

Since I had to change a lot of imports, this PR goes ahead and cleans up a lot of inconsistency when it comes to imports.


He-Pin:
@mdedetrich I think this can be merged after https://github.com/apache/pekko/pull/2209

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/LoggingEvent.scala
```diff
@@ -15,11 +15,12 @@ package org.apache.pekko.actor.testkit.typed
 
 import java.util.Optional
 
+import scala.jdk.CollectionConverters._
```
He-Pin, line 18: Yes, now `scala.jdk.CollectionConverters._`

He-Pin:
@mdedetrich There are some compiling issue.

He-Pin (commented):
Need fix the mima

mdedetrich:
> Need fix the mima

Yes I will get back to it tonight, have to do multiple passes

mdedetrich:
Got through a lot done tonight, still have a bit more to go through (should be done tomorrow)

mdedetrich:
@pjfanning @He-Pin This is taking a bit longer (there are a LOT of source files in pekko) but I should be done tonight. I will split out the removal of ccompat into another PR due to the massive size of this one, I also want to add this commit into `git-blame-ignore-revs` so people don't have a polluted git blame

He-Pin:
@mdedetrich I think that's fine, or you may want to split into two commits

He-Pin:
@mdedetrich what's the status of this:)

pjfanning:
If the build is passing, can we just merge what we have to unblock other work?

He-Pin:
I think so，otherwise too mach lag

mdedetrich:
It a bit more annoying as you have to cherry pick and merge multiple commits (I wanted to cherry pic this to 1.2.x series so that future cherry pick commits are easier to do), Ill see how much I can do tonight and if I don't finish it in time then will have to split it out

pjfanning:
> It a bit more annoying as you have to cherry pick and merge multiple commits (I wanted to cherry pic this to 1.2.x series so that future cherry pick commits are easier to do), Ill see how much I can do tonight and if I don't finish it in time then will have to split it out

We can't switch away from ccompat in Pekko 1.x due to Scala 2.12 needing support

He-Pin:
I suggest we do this automatically,  with short imports instead of ✋️ rolling

pjfanning:
Let's use https://github.com/apache/pekko/discussions/2228 to discuss the automated sorting of imports

pjfanning:
@mdedetrich we have `sbt sortImports` set up for this repo. Its output seems good. It is set up using `.scalafix.conf` and the docs and jdocs packages are excluded. Could we not just start using that command?

mdedetrich:
> @mdedetrich we have `sbt sortImports` set up for this repo. Its output seems good. It is set up using `.scalafix.conf` and the docs and jdocs packages are excluded. Could we not just start using that command?

It still won't work with `import org.apache.pekko` and then future `import pekko` imports, see https://github.com/scalacenter/scalafix/issues/2100

He-Pin:
I think so, I hate ✋️ tweaking result, just don't scale

mdedetrich:
> @mdedetrich we have `sbt sortImports` set up for this repo. Its output seems good. It is set up using `.scalafix.conf` and the docs and jdocs packages are excluded. Could we not just start using that command?

Also not sure why `jdocs` are excluded, scalafix only works on Scala sources?

He-Pin:
There are rules there, you can tweak the rule, in the scalafix.conf

He-Pin:
Iirc, I excludes that because of we already have java formatter

mdedetrich:
> Iirc, I excludes that because of we already have java formatter 

Fair enough, it is pointless to format that folder so best to exclude it

He-Pin:
`  "re:^(org\\.apache\\.pekko|pekko\\.)\\w*"` is there , and the https://github.com/liancheng/scalafix-organize-imports has some problems too.

I still think we should just work with the current setup and avoid handing it over manually by tweaking it, because style is always some kind of tasty.

He-Pin:
If it breaks docs, we can go back to it later, actually, imports can always be auto-imported by IDEA, so it doesn't matter too much.

mdedetrich:
@pjfanning @He-Pin Please don't let this PR block you anymore, it evidently took a lot longer than expected and I don't want to cause that amount of disruption. If you want to standard pull requests against Pekko, please do so (and also you don't need to take account any kind of formatting/import ordering stuff, just make a standard PR as you would).

This PR still has merit if we can't get scalafix to work with imports and I can handle rebasing this PR if we decide to merge it, but in any case this is a decision we need to make together and for the time being no formatting/import related PR's should be merged as it needs to be setup in a very specific way (I mentioned this in other threads).

He-Pin:
You said you don't want it to be blocked, so please kindly send me the PR CRs that you think are OK, thank you, otherwise it will be stuck all the time.

He-Pin:
Mostly done in other pr.

He-Pin:
Feel free to reopen a new PR

PR #2218, opened 2025-09-14
feat: Add sneakyThrow for Java

He-Pin:
Motivation:
When lambda doesn't support throwing exceptions, but the method we want to use declares that it throws exceptions, we can use this method to support it, similar to lombok's @sneakyThrows.

Usage:

```java

() -> {
 try {
     .....// can throw
   } catch(Throwable e) {
      Throwables.sneakThrow(e)
   }
}
```

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/japi/ThrowablesTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/japi/ThrowablesTest.java
@@ -29,4 +29,13 @@ public class ThrowablesTest {
     Assert.assertTrue(Throwables.isFatal(new InterruptedException("fatal")));
     Assert.assertTrue(Throwables.isFatal(new LinkageError("fatal")));
   }
+
+  private void doSneakyThrow() {
+    Throwables.sneakyThrow(new Exception("sneaky"));
+  }
+
+  @Test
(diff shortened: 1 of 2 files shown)


mdedetrich:
Thanks, this is a better solution for providing throwable function interfaces in pekko API that doesn't allow it as its opt in by the user

He-Pin:
That's true, this does not require a large refactoring.

mdedetrich (approved):
lgtm, although not entirely sure about the name. @pjfanning any thoughts on this?

He-Pin:
@mdedetrich The lombok and vavr both use this name, so it should be ok.

mdedetrich:
@He-Pin 

Can you add docs for this new function? Also you can cherry pick both this and the docs to Pekko 1.x series since it's an addition

He-Pin:
Don't find a good place to add

mdedetrich:
> Don't find a good place to add

Okay I can do it, don't forget to backport this

pjfanning:
> > Don't find a good place to add
> 
> Okay I can do it, don't forget to backport this

Can we make the backport be in 1.3.0? I haven't yet create a 1.3.x branch and it makes life easier for me to wait until after the 1.2.1 release to do this.

PR #2234, opened 2025-09-19
chore: Remove ccompat.JavaConverters

He-Pin:
The ccompat is not needed now


He-Pin:
@mdedetrich I would like to have this get merged first and then you can work on the sorting on you dedaicated pr.

PR #2244, opened 2025-09-20
feat: Add Sink#count operator.

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/2182

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Sink.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Sink.scala
@@ -314,6 +314,27 @@ object Sink {
   }
 
   /**
+   * A `Sink` that counts all incoming elements until upstream terminates.
+   *
+   * Since upstream may be unbounded, consider using `Flow[T].take` or the stricter `Flow[T].limit`
+   * (and their variants) to ensure boundedness. The sink materializes into a `CompletionStage` of `Long`
+   * containing the total count of elements that passed through.
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Backpressures when''' never (counting is a lightweight operation)
+   *
+   * '''Cancels when''' never
+   *
+   * @return a `Sink` that materializes to a `CompletionStage[Long]` with the element count
+   * @since 2.0.0
+   *
+   * See also [[Flow.limit]], [[Flow.limitWeighted]], [[Flow.take]], [[Flow.takeWithin]], [[Flow.takeWhile]]
+   */
+  def count[In]: Sink[In, CompletionStage[java.lang.Long]] = new Sink(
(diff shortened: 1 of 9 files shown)


stream/src/main/scala/org/apache/pekko/stream/javadsl/Sink.scala
```diff
@@ -313,6 +313,27 @@ object Sink {
...
+   *
+   * '''Cancels when''' never
+   *
+   * @return a `Sink` that materializes to a `CompletionStage[Long]` with the element count
+   * @since 2.0.0
+   *
+   * See also [[Flow.limit]], [[Flow.limitWeighted]], [[Flow.take]], [[Flow.takeWithin]], [[Flow.takeWhile]]
+   */
+  def count[In]: Sink[In, CompletionStage[java.lang.Long]] = new Sink(
+    scaladsl.Sink.count[In].mapMaterializedValue(_.mapTo[java.lang.Long].asJava))
```
He-Pin: Same as the reactor core, but this conversion looks bad. Is there a better way?

mdedetrich: This is what looks natural to me

He-Pin: Why scala compiler can't do this for me？

mdedetrich: Not on laptop, I'll check when I get t back home

He-Pin: Not sure why A Scala.Long can not be automaticly convert to a Java java.lang.Long, if this is the case, then there may be some `Int` `Long` leak to java api, I have not grep that, but I think so.

mdedetrich: iirc A Scala long can be used in Java fine (and Java sees a Scala long as a Java long) but it doesn't work the other way around.

In other words just return Scala long in both Java and Scala dsl, you can also look at other code for reference

He-Pin: no, the `CompletableFuture[Long]` seems not a `CompletableFuture[java.lang.Long]`

mdedetrich:
@He-Pin I added an extra commit for a more performant solution to the `java.lang.Long`/`scala.lang.Long` issue. While it is true that `scala.long.Long` is a supertype of `java.lang.Long`, because `CompletionStage` is invariant it doesn't compile as is so you have to use `asInstanceOf`.

The Java `SinkTest` is passing locally without any issues, but ill wait till tomorrow morning (my time) for all the tests to pass so I can approve the PR.

PR #2246, opened 2025-09-20
remove unnecessary items from license

pjfanning:
* the 3rd party code that we copied from scala-collection-compat has been removed
* likewise the Scala library borrowings only now apply to Helpers.scala


He-Pin (approved):
lgtm

PR #2247, opened 2025-09-20
remove unused script

pjfanning:
* the fact that it doesn't have a license header makes the file even worse


He-Pin (approved):

He-Pin:
will this be backprted？

pjfanning:
> will this be backprted？

not urgent - I might backport it later

Issue #2214, opened 2025-09-13
Drop vavr


He-Pin:
Now java 21:)

pjfanning:
We have a couple of doc references. Is there anything else?

He-Pin:
Iirc, there is an example.somewhere

PR #2248, opened 2025-09-20
Ronantakizawa/dropvavr

ronantakizawa:
Fixes issue #2214


He-Pin:
The java 21 switch expression for actor typed is current missing :(

PR #2250, opened 2025-09-20
feat: Add Sink#source

He-Pin:
Refs: https://github.com/apache/pekko/issues/2222

I think the optimization can be done in the materializer, but that's another story.

### DIFF
--- a/docs/src/main/paradox/stream/operators/Sink/source.md
+++ b/docs/src/main/paradox/stream/operators/Sink/source.md
@@ -1,0 +1,27 @@
+# Sink.source
+
+Always backpressure never cancel and never consume any elements from the stream.
+
+@ref[Sink operators](../index.md#sink-operators)
+
+## Signature
+
+@apidoc[Sink.source](Sink$) { java="#source()" }
+@apidoc[Sink.source](Sink$) { scala="#source()" }
+
+
+## Description
+
+A `Sink` that materialize this `Sink` itself as a `Source`.
+
+## Reactive Streams semantics
+
+@@@div { .callout }
+
+**cancels** When the materialized `Source` is cancelled or timeout with subscription.
+
+**backpressures** When the materialized `Source` backpressures or not ready to receive elements.
+
+@@@
+
+
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Sink.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Sink.scala
@@ -27,7 +27,7 @@ import pekko.annotation.InternalApi
 import pekko.stream._
 import pekko.stream.impl._
 import pekko.stream.impl.Stages.DefaultAttributes
-import pekko.stream.impl.fusing.GraphStages
+import pekko.stream.impl.fusing.{ GraphStages, SourceSink }
 import pekko.stream.stage._
 
 import org.reactivestreams.{ Publisher, Subscriber }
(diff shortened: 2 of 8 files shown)


He-Pin:
@mkurz ping you on this.

docs/src/main/paradox/stream/operators/Sink/source.md
```diff
@@ -0,0 +1,27 @@
+# Sink.source
+
+Always backpressure never cancel and never consume any elements from the stream.
```
raboof: This doesn't seem accurate, and doesn't seem like an accurate description of what this sink does.

From the javadoc, "A `Sink` that materializes this `Sink` itself as a `Source`." seems like a better description?

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Sink.scala
```diff
@@ -312,6 +312,19 @@ object Sink {
       if (fanout) new FanoutPublisherSink[T](DefaultAttributes.fanoutPublisherSink, shape("FanoutPublisherSink"))
       else new PublisherSink[T](DefaultAttributes.publisherSink, shape("PublisherSink")))
 
+  /**
+   * A `Sink` that materializes this `Sink` itself as a `Source`.
+   * The returned `Source` is a "live view" onto the `Sink` and only supports a single `Subscriber`.
```
raboof, line 317: Do I understand correctly that this means if you materialize a stream with this sink, you'll get a `Source` as a materialized value, but this `Source` will be 'special' in that it can only be materialized once?

While I agree having 'object creation' and 'materialization' as separate phases sometimes makes it challenging to design good API's, I'm concerned that adding APIs that violate these concepts will make things harder and more confusing, instead of simpler.

Before adding such methods I think we should identify a real-world example which such a sink would be valuable, and there's no better way to achieve the desired behavior?

docs/src/main/paradox/stream/operators/Sink/source.md
He-Pin: I copyied from never...., not changed it in time.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Sink.scala
He-Pin: Yes, that's true; otherwise, I have to with the help of the broadcast hub, should I?
The real-world example is 
https://github.com/playframework/playframework/blob/main/core/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala#L279

He-Pin: You can see, in the current usage of Play, the `fanout = false`, which will be covered by this PR, the broadcast one would be more complicated, so I would like to have that in a seperate pr, if that's needed.

To support the `fanout` I need to make use of `BroadcastHub.sink`.

He-Pin: The original requirement is similar to implementing a proxy, where one API requires a Sink for consumption, and another API requires a Source for data provision, connecting them together. However, the current API is just an API, and because it hasn't been optimized for the materializer, it introduces asynchronous boundaries, but it should be better than before.

He-Pin:
Yes, this is mainly useful in scenarios such as transparent proxy, or file upload, etc. In other scenarios such as chat rooms, you can use broadcastHub. In this case, later subscribers may lose some information.

@raboof

mdedetrich:
@raboof @He-Pin I believe this is also related to https://github.com/apache/pekko/issues/865 in which case there are definitely useful reasons for having this (and also that issue should be closed if this is solved).

i.e. think of a common flow, where you have a http request and in response to that request you want to send a topic to Kafka and then you want to deliver a http response. If you want the entire thing to be a flow you need this functionality as sending a message to a Kafka is a sink where as providing a http response (using pekko http) is a source.

These kind of cases are quite common and I actually bumped into the "I need to create a source from a sink" problem fairly frequently

raboof:
It would be great if you could take such an example and see if this implementation indeed meets that use case in practice!

mdedetrich:
> It would be great if you could take such an example and see if this implementation indeed meets that use case in practice!

Definitely, unfortunately this usecase I described was in the previous company where I worked (and the one before that), currently not using Pekko where I am working but if I have more time at some point I should be able to make such a working example.

He-Pin:
Nice to know, @mdedetrich are you writing Java or Kotlin nowadays, I'm a dayliy java developer

mdedetrich:
> Nice to know, @mdedetrich are you writing Java or Kotlin nowadays, I'm a dayliy java developer

Current day job is pure kotlin

PR #2251, opened 2025-09-20
chore: Make val debug final

He-Pin:


stream/src/main/scala/org/apache/pekko/stream/impl/PhasedFusingActorMaterializer.scala
```diff
@@ -60,7 +60,7 @@ import org.reactivestreams.Subscriber
  */
 @InternalApi private[pekko] object PhasedFusingActorMaterializer {
 
-  val Debug = false
+  final val Debug = false
```
He-Pin, line 63: Should be `final val`

pjfanning: Aren't Scala objects essentially final classes - in that they can't be extended ?
If a class is final then its fields are basically final too.
Is there really a benefit to marking a val in an object as final?

He-Pin: by making it a val , then it's a compiler constant, then the if branch will be purged by the comipler。

mdedetrich: > by making it a val , then it's a compiler constant, then the if branch will be purged by the comipler。
>  

Indeed this is correct, it will be constant folded because compiler knows it can't be overridden.

https://stackoverflow.com/a/24911844

PR #2253, opened 2025-09-21
refactor compiler options

pjfanning:
* I was experimenting with building with Java 25. Not working at all with Scala 2.13.17 but I can get Scala 3.3.6 to build jars.
* I tidied up the Java version settings to try to help


He-Pin:
We can rename the jdk9 plugin name too

pjfanning:
> We can rename the jdk9 plugin name too

I raised https://github.com/apache/pekko/issues/2256 - the change is probably not hard but it isn't trivial.

PR #2262, opened 2025-09-22
chore: Use array list for better performance in BroadcastHub

He-Pin:
Motivation:
The current implementation uses Lis,t which can be slow when there are many subscribers.
`jmh:run -i 10 -wi 5 -f2 -t1 org.apache.pekko.stream.BroadcastHubBenchmark`
```

[info] BroadcastHubBenchmark.broadcast             64  thrpt   10  207115.274 ± 17719.445  ops/s
[info] BroadcastHubBenchmark.broadcast            256  thrpt   10  158131.716 ±  5081.131  ops/s

////// array list

[info] BroadcastHubBenchmark.broadcast             64  thrpt   10  223218.562 ± 23896.486  ops/s
[info] BroadcastHubBenchmark.broadcast            256  thrpt   10  169280.588 ± 16349.613  ops/s

```

### DIFF
--- a/bench-jmh/src/main/scala/org/apache/pekko/stream/BroadcastHubBenchmark.scala
+++ b/bench-jmh/src/main/scala/org/apache/pekko/stream/BroadcastHubBenchmark.scala
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream
+
+import com.typesafe.config.ConfigFactory
+import org.apache.pekko.NotUsed
+import org.apache.pekko.actor.ActorSystem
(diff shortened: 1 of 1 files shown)


pjfanning:
Why not use a Scala Vector?

He-Pin:
Not try that yet, maybe array buffer？

pjfanning:
> Not try that yet, maybe array buffer？

An ArrayBuffer might be a good choice

He-Pin:
@pjfanning The arraybuffer's `foreach` is not overridden to work with array, maybe ArrayList is quite good for now.

He-Pin:
@mdedetrich wdyt about  this?

mdedetrich:
> Why not use a Scala Vector?

Scala's vector is extremely slow for small collection sizes (see https://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#conclusion). The only areas where Vectors are worth it is in multi-threaded/concurrent scenarios and you are dealing with large collection sizes (unlike other immutable collections such as java's `String`, Vector is triemap based so it doesn't need to copy the entire collection and it being immutable makes concurrency trivial as you are always making a new reference). Even though other collections require full copying (either because they are non tree based immutable collections or you are dealing with Arrays), CPU caches are so large that this type of copying in a hot loop is very fast.

This is where `ByteString` shines however, its backed by a Scala vector but due to the fact that we are dealing with large sizes (http responses/requests can be quite big) and its used in situations where concurrency is needed (we don't want a purely single threaded http client/server).

tl;dr is, especially if you don't need to care about concurrency just use standard array/arraybuffer etc etc depending on your requirements, its going to be much faster than Vector. Even if you do care about concurrency, often copying an entire collection is still faster than ones like `Vector` as long as the collection size isn't **that** big.

> @mdedetrich wdyt about this?

Ill look into it in a bit but at first glance this is a non controversial performance improvement

mdedetrich (approved):
lgtm

brendo-m:
Is it possible to backport this fix to a 1.x release please?

mdedetrich:
> Is it possible to backport this fix to a 1.x release please?

I made a cherry pick against 1.5.x here https://github.com/apache/pekko/pull/2596

PR #2269, opened 2025-09-25
no need for separate Scala 2 and 3 ByteStrings

pjfanning:
I diffed the Scala 2.13 and Scala 3 versions of ByteString and other than a few comment diffs and some other minor diffs they seemed the same.

#2270 shows the diff when overwriting the Scala 3 version of ByteString with the Scala 2.13 code

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -874,6 +874,9 @@ sealed abstract class ByteString
 
   override def indexWhere(p: Byte => Boolean, from: Int): Int = iterator.indexWhere(p, from)
 
+  // optimized in subclasses
+  override def indexOf[B >: Byte](elem: B, from: Int): Int = super.indexOf(elem, from)
+
   // optimized version of indexOf for bytes, optimized in subclasses
   /**
    * Finds index of first occurrence of some byte in this ByteString after or at some start index.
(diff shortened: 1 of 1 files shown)


pjfanning:
Mima tooling shows a binary compatibility issue in indexOf and this is worth analysing. It could be a relatively unimportant issue that we could choose to accept.

pjfanning:
In d65d8166b29dec5b0cc5302b999347800968fefc, I copied over some changes from the Scala 3 ByteString. This fixed the mima issue.

He-Pin:
Iirc , was planning to add some inline

pjfanning:
@He-Pin It's a lot of code to duplicate. At the moment, there doesn't appear to be much reason to support it.
If you want to keep the code duplication (over 1500 lines of duplication), can you provide a summary of what changes you want to make?

I am far from convinced about the benefits of `inline def` and `@inline def`. The Hotspot compiler already does a good job at doing this for you. It would be good to do jmh benchmarks to show that any `inline def` or `@inline def` changes actually really help.

pjfanning:
If we do want to do more inlining, could we consider isolating just the defs that we want to be inlinable into dedicated classes instead of having to maintain separate ByteString code altogether?

He-Pin:
@mdedetrich Wdyt?

pjfanning:
One option is to merge the code for now and if we want to make the Scala 2.13 and Scala 3 code differ in future, we could consider using metaprogramming to allow small parts of the class to differ instead of having to duplicate the full 1500 lines.
pekko-http has code that I removed in main branch recently because it really about Scala 2.12 support but a similar approach might be feasible to support Scala 2.13 vs Scala 3 diffs. https://github.com/apache/pekko-http/pull/785

He-Pin:
Jit actually helps , so maybe just merge the code for now to avoid duplication

PR #2271, opened 2025-09-25
ByteString Indexof with from and to

pjfanning:
* only specialised to take Byte input for search
* relates to https://github.com/apache/pekko/pull/2148#issuecomment-3333052773
* if this is ok, I think it can be backfit to 1.3.0

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -234,26 +234,24 @@ object ByteString {
     override def indexOf[B >: Byte](elem: B, from: Int): Int = {
       if (from >= length) -1
       else {
-        var found = -1
         var i = math.max(from, 0)
-        while (i < length && found == -1) {
-          if (bytes(i) == elem) found = i
+        while (i < length) {
+          if (bytes(i) == elem) return i
           i += 1
         }
-        found
+        -1
       }
     }
 
     override def indexOf(elem: Byte, from: Int): Int = {
       if (from >= length) -1
       else {
-        var found = -1
         var i = math.max(from, 0)
-        while (i < length && found == -1) {
-          if (bytes(i) == elem) found = i
+        while (i < length) {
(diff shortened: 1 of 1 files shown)


actor/src/main/scala-2.13/org/apache/pekko/util/ByteString.scala
```diff
@@ -257,6 +257,20 @@ object ByteString {
...
+      if (from >= length || to <= from) -1
+      else {
+        val upto = math.min(to, length)
+        var found = -1
+        var i = math.max(from, 0)
+        while (i < upto && found == -1) {
+          if (bytes(i) == elem) found = i
+          i += 1
+        }
+        found
```
He-Pin: Use return found here to avoid the &&

pjfanning: * this is a copy/paste of the existing `indexOf(Byte, Int)`
* I am not against changing this but I should then also change the existing `indexOf` methods too

Is the idea to change
```
      var found = -1
      while (i < upto && found == -1) {
        if (bytes(startIndex + i) == elem) found = i
          i += 1
        }
        found
      }
```
to this?
```
      while (i < upto && found == -1) {
        if (bytes(startIndex + i) == elem) return i
          i += 1
        }
        -1
      }
```
This saves having to do the while check one final time at the end of the loop. Not a huge win but should make a minor improvement in performance. Purists will hate the explicit returns.

He-Pin: I'm on a phone, the we should remove the `＆＆  found== -1` too

He-Pin: while (i < upto) {
        if (bytes(startIndex + i) == elem) 
           return i
        i+=1
     }
  -1

He-Pin:
I think we should remove the duplicated Bytestring before merge this

PR #2274, opened 2025-09-26
use Scala FunctionConverter

pjfanning:
* with Scala 2.12 support removed for 2.0.0, we can use the Scala built-in converter
* also tidy up some other scala 2.12 leftovers and remove a few unnecessary binary excludes


stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/StreamConvertersSpec.scala
```diff
@@ -45,7 +45,7 @@ class StreamConvertersSpec extends StreamSpec with DefaultTimeout {
     import java.util.stream.IntStream
     import java.util.stream.Stream
 
-    import pekko.util.FunctionConverters._
+    import scala.jdk.FunctionConverters._
```
He-Pin, line 48: we should remove FunctionConverters class too

pjfanning: pekko.util.FunctionConverters is deleted in this PR

He-Pin (approved):

CONTRIBUTING.md
```diff
@@ -533,7 +533,7 @@ Scala has proven the most viable way to do it, as long as you keep the following
 
 1. Use `scala.jdk.javaapi.OptionConverters` to translate `Option`s to Java `Optional`s (or vice versa).
 
-1. Use `org.apache.pekko.util.FunctionConverters` to translate Scala Functions to Java Functions (or vice versa).
+1. Use `scala.jdk.javapi.FunctionConverters` to translate Scala Functions to Java Functions (or vice versa).
```
ihostage, line 536: `scala.jdk.FunctionConverters`? 🤔

pjfanning: thanks - I'll correct this

pjfanning: actually https://www.scala-lang.org/api/2.13.16/scala/jdk/javaapi/FunctionConverters$.html does exist - this section is about people writing Java APIs and some of the Java APIs are written in Java (but some are in Scala) - so it sort of depends, Java users should use https://www.scala-lang.org/api/2.13.16/scala/jdk/javaapi/FunctionConverters$.html while Scala code should use scala.jdk.FunctionConverters implicits.

ihostage: In this case, we need to change the package name from `scala.jdk.javapi` to `scala.jdk.javaapi` 😉

PR #2286, opened 2025-09-30
chore: Set reuseAddr/port in temporaryServerAddresses

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/2268

Try to fix this by set reuse address to true.

### DIFF
--- a/testkit/src/main/scala/org/apache/pekko/testkit/SocketUtil.scala
+++ b/testkit/src/main/scala/org/apache/pekko/testkit/SocketUtil.scala
@@ -13,10 +13,9 @@
 
 package org.apache.pekko.testkit
 
-import java.net.{ DatagramSocket, InetSocketAddress, NetworkInterface, StandardProtocolFamily }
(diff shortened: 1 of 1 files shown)


raboof:
I'm not sure this is the right fix. While it will avoid the 'bind' error, presumably the test performs the bind because it actually wants to connect to the port. With this change, the bind may succeed, but the connect may connect to the 'wrong' test. In that case the test will still be flaky, but the flakiness will be even harder to understand. I'm not sure that's a net improvement...

He-Pin (commented):
Simple and correct fix for the `temporaryServerAddress` utility. Setting `reuseAddr = true` prevents port conflicts in tests that bind to ephemeral ports.

One note: `reusePort` is also set, which is Linux-specific (not supported on macOS/Windows). On platforms that don't support it, this will throw `UnsupportedOperationException`. The code should catch and ignore this exception for cross-platform compatibility.

Otherwise looks good.

PR #2300, opened 2025-10-06
use UnsynchronizedByteArrayInputStream

pjfanning:
Main advantage is that it doesn't copy the input array. This saves on time and memory.


He-Pin (approved):

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -590,7 +592,7 @@ object ByteString {
     }
 
     override def asInputStream: InputStream =
```
He-Pin, line 594: Should we update the scaladoc here？

pjfanning: This inherits this javadoc:
```
  /**
   * Return the bytes in this ByteString as an InputStream.
   *
   * @return the bytes in this ByteString accessible as an InputStream
   * @see [[asByteBuffer]]
   * @since 1.1.0
   */
```

I think this is ok as is. It doesn't mention anything about the InputStream implementation which is good because it gives us flexibility.

PR #2305, opened 2025-10-07
ci: stage release candidate source archive

raboof:
Needs Infra to configure these secrets before merging, but I welcome a review before we request those.

### DIFF
--- a/.github/workflows/stage-release-candidate.yml
+++ b/.github/workflows/stage-release-candidate.yml
@@ -31,13 +31,18 @@ jobs:
     steps:
       - name: Check version parameter
         run: |-
+          if [[ "$REF" != "v"* ]]; then
+            echo "Trigger this workflow on a version tag"
+            exit 1
+          fi
           if [[ "$REF" != *"-RC"* ]]; then
             echo "Trigger this workflow on an RC tag"
             exit 1
           fi
-          export VERSION=$(echo $REF | sed -e "s/\(.*\)-.*/\\1/")
+          export VERSION=$(echo $REF | sed -e "s/v\(.*\)-.*/\\1/")
+          export RC_VERSION=$(echo $REF | tail -c +2)
           echo "Version: $VERSION"
-          echo "RC: $REF"
+          echo "RC Version: $RC_VERSION"
         env:
           REF: ${{ github.ref_name }}
 
@@ -50,7 +55,7 @@ jobs:
 
       - name: Generate source archive
         run: |-
-          VERSION=$(echo $REF | sed -e "s/\(.*\)-.*/\\1/")
+          VERSION=$(echo $REF | sed -e "s/v\(.*\)-.*/\\1/")
           PREFIX=apache-pekko-$VERSION
           DATE=$(git log -n1 --format=%cs | tr -d -)
           TARBALL=$PREFIX-src-$DATE.tgz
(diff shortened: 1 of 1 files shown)


.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,95 @@
...
+      # We intentionally do not use the Coursier cache for release candidates,
+      # to reduce attack surface
+
+      # We intentionally postpone the signing to a separate step, to reduce
+      # attack surface
+
+      - name: Generate source dist
+        run: |-
+          export VERSION=$(echo $REF | sed -e "s/refs\/tags\/\(.*\)-.*/\\1/")
+          sbt "clean; set ThisBuild / version := \"$VERSION\"; sourceDistGenerate"
```
pjfanning: What is the plan for tagging? I presume we will tag with something like `2.0.0-M1-RC1` but we will want the versioned artifacts to have `2.0.0-M1`.
Ideally, we wouldn't want to add the `2.0.0-M1` tag until the release vote passes.
Our git setup makes it hard to move tags. We would need to remove the protections to move the tag or get Infra team to do it for us.

.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Sign source dist
+        run: |-
+          echo $PEKKO_GPG_SECRET_KEY | base64 -d | gpg --batch --import --import-options import-show
+          gpg -ab target/dist/*
+        env:
+          PEKKO_GPG_SECRET_KEY: ${{ secrets.PEKKO_GPG_SECRET_KEY }}
+
+      - name: Upload source dist
+        run: |-
+          export RC_VERSION=$(echo $REF | sed -e "s/refs\/tags\/\(.*\)/\\1/")
```
pjfanning: the RC_VERSION should have the `-RC1` suffix because we'll want that in the directory name on dist.apache.org while the version used to generate the jars and the source tar should leave the `-RC1` suffix off

.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,95 @@
...
+
+          svn checkout https://dist.apache.org/repos/dist/dev/pekko dist
+          cd dist
+          mkdir $RC_VERSION
+          cp ../target/dist/* $RC_VERSION
+          svn add $RC_VERSION $RC_VERSION/*
+          svn commit --username $PEKKO_SVN_DEV_USERNAME --password $PEKKO_SVN_DEV_PASSWORD --message "Stage Pekko $RC_VERSION" $RC_VERSION
+        env:
+          PEKKO_SVN_DEV_USERNAME: ${{ secrets.PEKKO_SVN_DEV_USERNAME }}
+          PEKKO_SVN_DEV_PASSWORD: ${{ secrets.PEKKO_SVN_DEV_PASSWORD }}
```
pjfanning, line 90: Do we intend to also stage the jars in repository.apache.org?

.github/workflows/stage-release-candidate.yml
raboof: The plan is not to change that :). the sed substitution chops off that `-RC1` in this case (and the earlier step checks it's there)

.github/workflows/stage-release-candidate.yml
raboof: Correct

.github/workflows/stage-release-candidate.yml
raboof: Yes, I think we should (but perhaps not in this PR :) )

.github/workflows/stage-release-candidate.yml
pjfanning: this command looks the same as the one for `$VERSION`

raboof: the one for `$VERSION` has an extra `-.*` :)

raboof: though I agree it's not obvious. It looks like we could use `github.ref_name` which should make this more readable. Let me try that.

raboof:
I moved to `git archive` prompted by the discussion in https://github.com/pjfanning/sbt-source-dist/issues/44 - seems nice to reduce the dependencies/complexity/attacksurface for this part of the process.

pjfanning:
lgtm - if we can get all the secrets set up, I don't see any issues with getting this merged and trying it out for our next RC

raboof:
Requested in https://issues.apache.org/jira/browse/INFRA-27312

raboof:
closing in favour of #2314

PR #2306, opened 2025-10-07
specialized version of indexOfSlice

pjfanning:
* We use indexOfSlice in pekko-http
* This specialized version should in theory be faster than the version that ByteString inherits from scala.Seq
* the existing code in this PR would probably not be great for ByteStrings implementation and it would be good to override it in that class - I suspect using indexOf for the first byte in the slice is still good but comparing the bytes after that would benefit from creating a new ByteString that drops the bytes before the startPos and that continues the search on the new ByteString with the dropped bytes
* I intend to some jmh benchmarking before looking to merge this
* if the benchmarks look good, I may also try to override `def indexOfSlice(slice: ByteString, from: Int)`

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -1076,7 +1076,10 @@ sealed abstract class ByteString
       else if (check(startPos)) startPos
       else rec(startPos + 1)
     }
-    if (slice.isEmpty) 0 else rec(math.max(0, from))
+    val sliceLength = slice.length
+    if (sliceLength == 0) 0
+    else if (sliceLength == 1) indexOf(headByte, from)
+    else rec(math.max(0, from))
   }
 
   /**
@@ -1108,7 +1111,10 @@ sealed abstract class ByteString
       else if (check(startPos)) startPos
       else rec(startPos + 1)
     }
-    if (slice.isEmpty) 0 else rec(math.max(0, from))
+    val sliceLength = slice.length
+    if (sliceLength == 0) 0
+    else if (sliceLength == 1) indexOf(slice.head, from)
+    else rec(math.max(0, from))
   }
 
   /**


pjfanning:
Benchmark testing with https://github.com/pjfanning/pekko-bench/tree/indexOf shows that the new code has better performance.

```
[info] Benchmark                                          Mode  Cnt          Score           Error  Units
[info] IndexOfSliceBench.indexOfSliceByteString0Compact  thrpt    3  170972026.909 ± 950921237.250  ops/s
[info] IndexOfSliceBench.indexOfSliceByteString0Concat   thrpt    3   62923083.158 ±  80172894.188  ops/s
[info] IndexOfSliceBench.indexOfSliceByteStringCompact   thrpt    3   52554838.440 ±   8384708.327  ops/s
[info] IndexOfSliceBench.indexOfSliceByteStringConcat    thrpt    3   22211242.061 ±  38497866.157  ops/s
[info] IndexOfSliceBench.indexOfSliceBytes0Compact       thrpt    3  142160935.848 ± 842800972.850  ops/s
[info] IndexOfSliceBench.indexOfSliceBytes0Concat        thrpt    3   68818647.909 ±  43520695.803  ops/s
[info] IndexOfSliceBench.indexOfSliceBytes1Compact       thrpt    3  106181468.854 ±  72102056.049  ops/s
[info] IndexOfSliceBench.indexOfSliceBytes1Concat        thrpt    3   31358546.706 ±  63361128.277  ops/s
```

* indexOfSliceByteStringCompact is the legacy
* indexOfSliceByteString0Compact is a new override
* indexOfSliceBytes0Compact searches for `Array[Byte]` instead of `Seq[B :> Byte]`
* indexOfSliceBytes1Compact uses a different algorithm that is theoretically better for concatenated ByteStrings but in practice is worse than the 0 algorithm at least in the use cases tested

This PR drops use of the second algorithm.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1056,6 +1056,72 @@ sealed abstract class ByteString
    */
   def indexOf(elem: Byte): Int = indexOf(elem, 0)
 
+  override def indexOfSlice[B >: Byte](slice: scala.collection.Seq[B], from: Int): Int = {
+    // this is only called if the first byte matches, so we can skip that check
```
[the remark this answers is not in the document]
pjfanning: the existing comment is more correct

pjfanning:
@He-Pin would you have time to re-review this?

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1056,6 +1056,72 @@ sealed abstract class ByteString
...
+    def check(startPos: Int): Boolean = {
```
He-Pin, line 1061: Those two methods `check` can create closures; we can make it a private method.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1056,6 +1056,72 @@ sealed abstract class ByteString
...
+        j += 1
+      }
+      true
+    }
+    val headByte = slice.head.asInstanceOf[Byte]
+    @tailrec def rec(from: Int): Int = {
+      val startPos = indexOf(headByte, from, length - slice.length + 1)
+      if (startPos == -1) -1
+      else if (check(startPos)) startPos
+      else rec(startPos + 1)
```
He-Pin, line 1077: This can be optimized, because they will know how many index already been checked, so we should be able to skip that length instead of just `1` step here.

pjfanning: * it has to be +1 because there is an edge case where the the slice has repeated characters - I have carefully tested this change
* this PR implementation as is is already much faster than the existing implementation

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: the 2 check methods are actually different and trying to force the 2 check methods to be the same will involve casting and that will damage performance - the type of the slice parameter is different

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
He-Pin: see io.netty.buffer.ByteBufUtil#indexOf(io.netty.buffer.ByteBuf, io.netty.buffer.ByteBuf)

He-Pin (approved):
lgtm, but performance can be better with https://en.wikipedia.org/wiki/Two-way_string-matching_algorithm , as it's been used in Netty's `ByteBufUtil.indexOf`

pjfanning:
@He-Pin I logged https://github.com/apache/pekko/issues/2321 and would prefer to treat that as a follow up. This PR already makes big gains and porting the Netty code can be complicated. I will look at 2321 but it could take me a week or 2 to get back to it.

PR #2310, opened 2025-10-08
chore: fix up some GitHub Actions pins

raboof:
Notably don't rely on the scala-steward tag

done with octopin (but skipped commits for `actions/` actions)


pjfanning (approved):
lgtm

PR #2314, opened 2025-10-09
ci: stage release candidate sources and jars

raboof:
on top of #2305

### DIFF
--- a/.github/workflows/stage-release-candidate.yml
+++ b/.github/workflows/stage-release-candidate.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Sign source archive
         run: |-
-          echo $PEKKO_GPG_SECRET_KEY | base64 -d | gpg --batch --import --import-options import-show
+          echo $PEKKO_GPG_SECRET_KEY | gpg --batch --import --import-options import-show
           gpg -ab archive/*.tgz
         env:
           PEKKO_GPG_SECRET_KEY: ${{ secrets.PEKKO_GPG_SECRET_KEY }}


.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,143 @@
...
+      # It would be better to split this into 3 steps, where only the first
+      # uses sbt and the signing/staging are done with well-known tools
+      # reducing attack surface, but this seems to be the state of the art:
+      - name: Build, sign and stage artifacts
+        run: |-
+          VERSION=$(echo $REF | sed -e "s/\(.*\)-.*/\\1/")
+          PGP_PASSPHRASE=
+
+          sbt "set ThisBuild / version := \"$VERSION\"; +publishSigned"
+        env:
```
pjfanning, line 155: with the sbt setup that we have, `sbt publishSigned` when a version tag is found on most recent commit, the jars are just staged in the target directory locally - you need to run an sbt-sonatype command to push to nexus.

`sbt "sonatypePrepare; sonatypeBundleUpload; sonatypeClose"`

https://github.com/apache/pekko-site/wiki/Pekko-Release-Process

I have previously found that setting the version in the sbt command can be problematic with this command - the version seems not to propagate in all stages. It is worth playing with the commands locally.

You could drop an RC tag locally and run `sbt "set ThisBuild / version := \"$VERSION\"; publishSigned` (maybe just one scala version to cut down on build time in a dry run) and then the `sbt "set ThisBuild / version := \"$VERSION\"; sonatypePrepare; sonatypeBundleUpload; sonatypeClose"`.

A bit of trial and error, you might find something that works.
It is easy to go into repository.apache.org and drop the staged repos.

raboof: Good pointer - indeed we need to set the version again before `sonatypeBundleUpload`. It doesn't seem to need to happen before `sonatypeClose` again.

(actually it seems `sonatypeBundleUpload` will `sonatypePrepare` implicitly so `sonatypePrepare` might not be strictly needed - but let's keep it to stay close to https://github.com/xerial/sbt-sonatype?tab=readme-ov-file#publishing-your-artifact)

raboof:
waiting for https://issues.apache.org/jira/browse/INFRA-27312

pjfanning:
In theory, this job could be used to publish 1.x releases too.
We would want to use Java 11 instead of Java 17 for a 1.x release.

.github/workflows/stage-release-candidate.yml
```diff
@@ -0,0 +1,159 @@
...
+          echo "RC Version: $RC_VERSION"
+        env:
+          REF: ${{ github.ref_name }}
+
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          persist-credentials: false
```
pjfanning, line 65: do we need to derive and apply a `ref`, basically the tag identifier so the checkout would be at a tag as opposed to HEAD?

raboof: we don't, because `workflow_dispatch` already allows us to choose the tag to run against.

PR #2319, opened 2025-10-12
chore: bump Scala version to  3.3.7 (#2104)

He-Pin:
(cherry picked from commit 71ff3ff9b567b21cdd699d38ca1a411b42e2ddfd)


pjfanning:
One issue is that our CI builds in other Pekko repos will need to be updated too because sbt enforces that any transitive dependency on a new Scala lib breaks the build until you upgrade your own build.sbt.

He-Pin:
@pjfanning Would you like to merge this?

pjfanning:
it breaks a lot of our CI builds in other modules when we upgrade Scala in pekko core - it forces us to go to lots of branches in other repos and upgrade scala there too - specifically to check for broken builds and fix them

I guess we can merge this but would appreciate some help fixing any builds in pekko repos that occur after this.

Issue #2321, opened 2025-10-18
Revisit indexOfSlice performance improvements


pjfanning:
https://github.com/apache/pekko/pull/2306#pullrequestreview-3352698604

PR2306 already improves performance quite a bit.

There is a suggestion that we could eke out some more perf gains by using the https://en.wikipedia.org/wiki/Two-way_string-matching_algorithm - see the link above - basically trying to port Netty's code

pjfanning:
I ran the new benchmarks from #2325 against the main branch, PR2324 and PR2323 code.
So far, PR2323 helps a little but PR2324 has poor performance - possibly a bug in my code that damages the performance.


```
indexOfSlice

Scala 2.13 main branch (Java 17.0.16)
[info] Benchmark                                            Mode  Cnt         Score         Error  Units
[info] ByteString_indexOfSlice_Benchmark.bs_indexOfSlice   thrpt    3  32742448.312 ± 7405884.505  ops/s
[info] ByteString_indexOfSlice_Benchmark.bss_indexOfSlice  thrpt    3   3944861.418 ± 1012500.206  ops/s


Scala 2.13 full 2 way impl (Java 17.0.16)
[info] Benchmark                                            Mode  Cnt        Score         Error  Units
[info] ByteString_indexOfSlice_Benchmark.bs_indexOfSlice   thrpt    3  3243337.351 ± 1006365.481  ops/s
[info] ByteString_indexOfSlice_Benchmark.bss_indexOfSlice  thrpt    3  3045235.141 ± 5055886.316  ops/s


Scala 2.13 with just getLong changes (Java 17.0.16)
[info] Benchmark                                            Mode  Cnt         Score         Error  Units
[info] ByteString_indexOfSlice_Benchmark.bs_indexOfSlice   thrpt    3  35207523.319 ± 9632183.164  ops/s
[info] ByteString_indexOfSlice_Benchmark.bss_indexOfSlice  thrpt    3   3583970.019 ±   34365.425  ops/s
```

PR #2323, opened 2025-10-18
[WIP] use getLong when matching byte sequences in ByteString

pjfanning:
* see #2321
* there are more changes that could be attempted based on what is in Netty ByteBufUtil but we've also seen getLong helping quite a bit with indexOf
* aim is to benchmark this and if it helps, get it merged before possibly trying further Netty ByteBufUtil approaches 
https://github.com/netty/netty/blob/a5343227b10456ec889a3fdc5fa4246f036a216d/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L242
* the CI issue is due to a package private method causing a mima issue - it can be easily filtered later

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -222,6 +222,7 @@ Copyright EPFL and Lightbend, Inc.
 pekko-actor contains code from Netty which was released under an Apache 2.0 license.
 Copyright 2014 The Netty Project
 - actor/src/main/scala/org/apache/pekko/io/dns/DnsSettings.scala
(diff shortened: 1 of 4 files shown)


He-Pin (commented):
[WIP] The simpler approach compared to #2324. Using `getLong` for byte sequence matching is straightforward and likely provides good performance for most cases.

**Recommendation:** Same as #2324 -- benchmark this against the Two-Way approach and merge the winner. The MiMa issue mentioned in the description ('package private method causing a mima issue') can be handled with exclusion filters.

PR #2324, opened 2025-10-18
[WIP] ByteString Two-Way string matching in indexOfSlice

pjfanning:
* See #2323 and #2321
* the aim is to benchmark both solutions - this one is more complicated so it would need to be much better to justify being used

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -385,7 +385,7 @@ object ByteString {
 
     // Derived from code in Netty
     // https://github.com/netty/netty/blob/d28a0fc6598b50fbe8f296831777cf4b653a475f/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L366-L408
-    override private[util] def bytesMatch(fromIndex: Int, checkBytes: Array[Byte], bytesFromIndex: Int,
+    override final private[util] def bytesMatch(fromIndex: Int, checkBytes: Array[Byte], bytesFromIndex: Int,
         checkLength: Int): Boolean = {
       var aIndex = fromIndex
       var bIndex = bytesFromIndex
@@ -725,7 +725,7 @@ object ByteString {
 
     // Derived from code in Netty
     // https://github.com/netty/netty/blob/d28a0fc6598b50fbe8f296831777cf4b653a475f/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L366-L408
-    override private[util] def bytesMatch(fromIndex: Int,
+    override final private[util] def bytesMatch(fromIndex: Int,
         checkBytes: Array[Byte],
         bytesFromIndex: Int,
         checkLength: Int): Boolean = {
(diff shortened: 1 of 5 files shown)


actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -575,6 +668,86 @@ object ByteString {
...
+      val m = slice.length
+      if (m == 0) return 0
+      // When the needle has only one byte that can be read,
+      // the indexOf() can be used
+      if (m == 1) return indexOf(slice.head, from)
+      var i = 0
+      var j = 0
+      val aStartIndex = 0
+      val bStartIndex = from + startIndex
+      val suffixes = SWARUtil.maxSuf(slice, m, aStartIndex, true)
```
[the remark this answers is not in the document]
pjfanning: fixed

He-Pin (commented):
[WIP] The Two-Way string matching approach from Netty is a good candidate for optimizing `indexOfSlice` on large ByteStrings. However:

1. This is marked WIP and the PR description says 'the aim is to benchmark both solutions'. Without benchmark results showing a clear win, this is hard to evaluate.
2. The licensing update (adding Netty-derived code to LICENSE) is correctly handled.
3. This relates to #2323 which takes a simpler approach (just using `getLong` for byte matching). Consider benchmarking both approaches and merging whichever wins.

**Recommendation:** Complete the benchmarks and either merge or close based on results. Don't leave as WIP indefinitely.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -314,6 +329,85 @@ object ByteString {
       else -1
     }
 
+    // Derived from code in Netty
+    // https://github.com/netty/netty/blob/d28a0fc6598b50fbe8f296831777cf4b653a475f/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L242-L325
+    override def indexOfSlice(slice: Array[Byte], from: Int): Int = {
+      val fromIndex = math.max(0, from)
+      val n = length - fromIndex
+      val m = slice.length
+      if (m == 0) return 0
```
He-Pin, line 338: Bug: `if (m == 0) return 0` is a behavioral regression. The current `main` contract for `indexOfSlice` returns `if (from > length) -1 else math.max(from, 0)` when the slice is empty (see the base implementation around line 2069 on main). This hard-coded `return 0` breaks that contract:

- `ByteString("xxyz").indexOfSlice(Array.empty[Byte], 1)` would return `0` instead of `1`
- `ByteString("xxyz").indexOfSlice(Array.empty[Byte], 10)` would return `0` instead of `-1`

The existing spec (`ByteStringSpec.scala:1274-1281`) documents the intended behavior.

Suggested fix:
```scala
if (m == 0) return if (fromIndex > length) -1 else fromIndex
```

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -575,6 +669,87 @@ object ByteString {
...
```
He-Pin, line 678: Same issue as ByteString1 above: `if (m == 0) return 0` should be `if (m == 0) return if (fromIndex > length) -1 else fromIndex` to match the base `indexOfSlice` contract.

He-Pin (changes requested):
The Two-Way algorithm itself is correctly ported from Netty. However, there is a behavioral regression for the empty-slice case that needs fixing before this can merge.

pjfanning:
I don't have much interest in going back to this. My recollection is that the performance was much worse with this change.

Issue #2329, opened 2025-10-19
Aeron tests failing every night: ArteryFailedToBindSpec - must not start if port is taken


pjfanning:
eg https://github.com/apache/pekko/actions/runs/18266203685/job/52000863954

I think the only recent change was an upgrade in aeron jar - #2295

Seems to be mainly this test:
ArteryFailedToBindSpec: must not start if port is taken

```
[info] - must not start if port is taken *** FAILED *** (340 milliseconds)
[info]   Expected exception org.apache.pekko.remote.RemoteTransportException to be thrown, but io.aeron.exceptions.RegistrationException was thrown (ArteryFailedToBindSpec.scala:56)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
[info]   at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)
[info]   at org.scalatest.wordspec.AnyWordSpec.newAssertionFailedException(AnyWordSpec.scala:1880)
[info]   at org.scalatest.Assertions.intercept(Assertions.scala:756)
[info]   at org.scalatest.Assertions.intercept$(Assertions.scala:746)
[info]   at org.scalatest.wordspec.AnyWordSpec.intercept(AnyWordSpec.scala:1880)
[info]   at org.apache.pekko.remote.artery.ArteryFailedToBindSpec.$anonfun$new$2(ArteryFailedToBindSpec.scala:56)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1240)
[info]   at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)
[info]   at org.scalatest.TestSuite.withFixture$(TestSuite.scala:195)
[info]   at org.scalatest.wordspec.AnyWordSpec.withFixture(AnyWordSpec.scala:1880)
[...]

PR #2331, opened 2025-10-19
revert recent aeron upgrade while we investigate test issue

pjfanning:
see #2329


He-Pin:
IIRC, @jrudolph suggested we drop the aeron .
If it really matters (cross geo region performance , we can use quic/http3)

PR #2336, opened 2025-10-19
feat: Add more recover operators  for java dsl.

He-Pin:
I was plan to add `onErrorReturn` but `recover` alias with the current.


He-Pin:
Once gets merged , need backport to 1.3.0

stream/src/main/scala/org/apache/pekko/stream/javadsl/Flow.scala
```diff
@@ -1894,6 +1894,72 @@ final class Flow[In, Out, Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Gr
       case elem if clazz.isInstance(elem) => creator.create()
     }
 
+  /**
+   * Recover allows to send last element on failure and gracefully complete the stream
```
pjfanning: Could you add a full stop at the end of this line? And also in other places where this same comment is added in this PR.

He-Pin: I see

PR #2337, opened 2025-10-19
feat: Add missing onErrorResume to SubFlow and SubSource

He-Pin:
Motivation:
1. Add the missing onErrorResume method to SubFlow and SubSource
2. Change to only resume the stream when non fatal error


He-Pin:
Need backport to 1.3.x

He-Pin:
The error is unrelated.

PR #2344, opened 2025-10-19
Update sbt-header to 5.11.0 (#2217)

He-Pin:
* Update sbt-header to 5.11.0

* chore: fix imports

---------

(cherry picked from commit 94ae86fa6f1af0c7177e2665167b39fb54571696)


pjfanning:
This really doesn't feel like a priority to me. I don't see why upgrading an sbt plugin that appears to have no effect is of use.
Fine to have it on the main branch - I just question the value of backporting it.

He-Pin:
@pjfanning because we need to do some backporting, so I think this will reduce the in case difference.

PR #2348, opened 2025-10-20
Support Jackson 3

pjfanning:
Idea is to have a new pekko-serialization-jackson3 lib and to keep pekko-serialization-jackson as is with its Jackson 2 support.

Jackson 3 is a big change with significant modifications to the default set up.

I already have my own lib - https://github.com/pjfanning/pekko-serialization-jackson3 - but this PR is intended to bring the code into the Pekko project.


nvollmar (approved):
lgtm

PR #2358, opened 2025-10-23
pekko-stream: change default TLS protocol from TLSv1.2 to TLSv1.3

pjfanning:
#1901 only really covers pekko-remote use cases - pekko-stream has its own setting that presumably also affects pekko-http

### DIFF
--- a/stream/src/main/resources/reference.conf
+++ b/stream/src/main/resources/reference.conf
@@ -182,7 +182,7 @@ pekko {
 
   # configure overrides to ssl-configuration here (to be used by pekko-streams, and pekko-http – i.e. when serving https connections)
   ssl-config {
-    protocol = "TLSv1.2"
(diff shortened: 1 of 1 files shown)


mkurz:
IMHO good candidate for 1.3

He-Pin (approved):
lgtm

pjfanning:
@mkurz @He-Pin this setting no longer appears to be used in pekko main branch - the code that uses it was removed

I will look at 1.3.x branch separately but the code that uses it is marked as deprecated. If users don't use the deprecated code then they will be unaffected. It is possible that pekko-http 1.x uses the deprecated code though (internally).

pjfanning:
closing in favour of #2359 - we may end up with a branch 1.3.x variant of this PR but it only affects the long deprecated class PekkoSSLConfig

PR #2360, opened 2025-10-23
change pekko.ssl-config.protocol default to TLSv1.3

pjfanning:
* affects deprecated code (PekkoSSLConfig) but code that is used in deprecated pekko-http 1.x code - so users may still be using it
* this code is removed from main branch
* see #2358 for discussion


pjfanning:
https://github.com/apache/pekko-http/pull/855 passed so it looks like
* either we don't test the deprecated code that uses this config well
* or if we do test the deprecated code that the tests work ok with this change

He-Pin (approved):

PR #2365, opened 2025-10-25
fix close in cancel for statefulMap

pjfanning:
* based on https://github.com/akka/akka-core/pull/31630
* Akka 2.7.0 changes are now available under Apache license
* this test fail in Pekko so it looks like we need changes to get them to work

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -2274,7 +2274,7 @@ private[pekko] final class StatefulMap[S, In, Out](create: () => S, f: (S, In) =
           push(out, newElem)
         } catch {
           case ex: NullStateException => throw ex // don't cover with supervision
-          case NonFatal(ex) =>
+          case NonFatal(ex)           =>
             decider(ex) match {
               case Supervision.Stop    => closeStateAndFail(ex)
               case Supervision.Resume  => pull(in)
@@ -2315,7 +2315,7 @@ private[pekko] final class StatefulMap[S, In, Out](create: () => S, f: (S, In) =
       }
 
       private def closeStateAndFail(ex: Throwable): Unit =
-       completeStateIfNeeded() match {
+        completeStateIfNeeded() match {
           case Some(elem) => emit(out, elem, () => failStage(ex))
           case None       => failStage(ex)
         }


He-Pin (approved):
lgtm

He-Pin:
@pjfanning A different is , IIRC, statefulmap in pekko support `null` as state, but not in akka

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -2250,69 +2260,86 @@ private[pekko] final class StatefulMap[S, In, Out](create: () => S, f: (S, In) =
...
       override def onPull(): Unit = pull(in)
 
       override def postStop(): Unit = {
-        if (needInvokeOnCompleteCallback) {
-          onComplete(state)
-        }
+        completeStateIfNeeded()
+      }
+
+      private def throwIfNoState(): Unit = {
```
He-Pin, line 2338: This seems to change the behavior. @pjfanning I have not tested it ,but , if that's true, we should revert some changes.

stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowStatefulMapSpec.scala
```diff
@@ -455,4 +454,86 @@ class FlowStatefulMapSpec extends StreamSpec {
...
+      Source
+        .single("one")
+        .statefulMap(() => "state")((_, t) => (null, t), _ => None)
+        .runWith(Sink.seq)
+        .failed
+        .futureValue shouldBe a[NullPointerException]
+    }
+  }
+
+  "not allow null state on restart" in {
```
He-Pin, line 524: behavior change.

pjfanning: in theory, we can change behaviour in 1.3.0 - but if we truly want to keep null support, we could try to modify this test and the other new null tests and then update the code to make the modified tests work

He-Pin: I still like to keep the old behavior, we can improve both document and test.

pjfanning:
I had a look at modifying the new code to not throw the NullStateException when the state evaluated as null but the new tests end up with NoSuchElementExceptions instead.

I think the options are:
* revert this and just keep the existing code and agree that this change is not important
* keep this change and mandate the users who rely on 'null' state can modify their code to use something like scala.None or java.lang.Optional.empty() instead.
* a significant rewrite of the new code that somehow allows nulls but that allows the new tests to work - this code is not an area of expertise for me so if someone wants to go this route, can they provide a PR?

wdyt? @He-Pin @mdedetrich @raboof @jrudolph

He-Pin:
I will take care of this @pjfanning

PR #2372, opened 2025-10-25
Replaying a replicated event sourced actor from snapshot caused bug

pjfanning:
* copied from https://github.com/akka/akka-core/pull/31645
* Akka 2.7.0 changes are now available under Apache 2.0 license

### DIFF
--- a/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
+++ b/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
@@ -417,7 +417,7 @@ private[pekko] object Running {
           replication.allReplicas.mkString(", "))
         this
       } else {
-        val expectedSequenceNumber = state.seenPerReplica(originReplicaId) + 1
+        val expectedSequenceNumber = state.seenPerReplica.getOrElse(originReplicaId, 0L) + 1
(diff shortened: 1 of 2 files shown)


He-Pin (approved):
lgtm

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
```diff
@@ -751,7 +751,7 @@ private[pekko] object Running {
     }
 
     def onGetSeenSequenceNr(get: GetSeenSequenceNr): PersistingEvents = {
-      get.replyTo ! state.seenPerReplica(get.replica)
+      get.replyTo ! state.seenPerReplica.getOrElse(get.replica, 0L)
```
He-Pin, line 754: I think we should extract the 0L to a constant if possible.

He-Pin:
Need backport to 1.3.0 too

persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/Running.scala
pjfanning: the compiler will handle this nicely - I don't think we need a constant for 0

PR #2373, opened 2025-10-25
fix close in cancel for statefulMap (#2365)

pjfanning:
cherry pick dbdbe03a13eb491dd474400ad644f7d6bf8ee67a #2365


He-Pin:
https://github.com/thatdot/quine/blob/main/quine-core/src/main/scala/com/thatdot/quine/util/PekkoStreams.scala#L38

@pjfanning IIRC, the statefulMap supports `null` as the first `state` in pekko, but not in akka, so we should be careful about this .

pjfanning:
let's not merge this then while we investigate the implications

He-Pin:
Yes, we should partial revert changes in main, and add the behavior of support null as state.

PR #2378, opened 2025-10-25
chore: Remove unnecessary watchTermination parentheses

He-Pin:
refs: https://github.com/apache/pekko/issues/2376


pjfanning (approved):

He-Pin:
Can we break api in 1.3.x ? @pjfanning

pjfanning:
I'd like some more opinions on whether this is safe for 1.3.0. I'm wary of it.

He-Pin:
This will breaking user's code, so I think it's better to be 2.0.0 only.

PR #2385, opened 2025-10-26
fix: Update statefulMap scaladoc and tests

He-Pin:
Motivation:

Update the Scala doc about pekko's statefulmap support `null`, and add a fix with the updated test.


He-Pin:
This need backport to 1.3.x and 1.2.x

PR #2389, opened 2025-10-26
chore: Update stew on 1.3.x too

He-Pin:
This will make it work on 1.3.x too


pjfanning:
I would be against this. I think we need to be more careful about dependency upgrades in 1.x branches. I see no harm in manually creating PRs fo upgrades there.

He-Pin:
But what about the scalafmt ? It can update automatically; you can handle all the dependencies of 1.1.x if you like:)

PR #2400, opened 2025-10-26
chore: Migrate to Strings.CI.contains

He-Pin:


project/CopyrightHeader.scala
```diff
@@ -12,8 +12,8 @@
  */
 
 import sbtheader.{ CommentCreator, HeaderPlugin, NewLine }
-import org.apache.commons.lang3.StringUtils
-import sbt._
+import org.apache.commons.lang3.Strings
+import sbt.*
```
pjfanning: can we keep the _ syntax?

He-Pin: Yes, but we should migrate to this sooner or later.

PR #2405, opened 2025-10-27
Rename gunzip to gzipDecompress

mdedetrich:
While working on adding zstd compression I noticed that the vernacular in the method names that we use for gzip decompression is quite atypical. For starers the g in `gunzip` could refer to anything (not just gzip) and the `unzip` part only makes sense in the context of unzipping a file where as `gunzip` is a generic flow that can work on anything (i.e. just a stream of bytes).

This PR renames `gunzip` to `gzipDecompress` to make it ultra clear, also deprecates the old method name so it can be removed in 2.0.0 (I will make a follow up PR in 2.0.0 that adds this method without the deprecation).


stream/src/main/scala/org/apache/pekko/stream/javadsl/Compression.scala
```diff
@@ -25,8 +25,17 @@ object Compression {
    *
    * @param maxBytesPerChunk Maximum length of the output [[pekko.util.ByteString]] chunk.
    */
+  def gzipDecompress(maxBytesPerChunk: Int): Flow[ByteString, ByteString, NotUsed] =
```
He-Pin, line 29: @since 1.3.0

mdedetrich: Done and fixed

pjfanning:
Should `gzip` become `gzipCompress` then?

mdedetrich:
> Should `gzip` become `gzipCompress` then?

I suppose if we want to be completely consistent although one can argue its implied due to the `Compress` object that its contained within.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Compression.scala
```diff
@@ -24,9 +24,19 @@ object Compression {
...
+    scaladsl.Compression.gzipDecompress(maxBytesPerChunk).asJava
+
+  /**
+   * Creates a Flow that decompresses gzip-compressed stream of data.
+   *
+   * @param maxBytesPerChunk Maximum length of the output [[pekko.util.ByteString]] chunk.
+   */
+  @deprecated("Use gzipDecompress instead", "Pekko 1.3.0")
   def gunzip(maxBytesPerChunk: Int): Flow[ByteString, ByteString, NotUsed] =
```
He-Pin, line 38: @mdedetrich Will you forward port to main and then delete these methods?

He-Pin (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/javadsl/Compression.scala
mdedetrich: Yes after its merged

He-Pin:
@mdedetrich need rename and update `gunzip.md` too

mdedetrich:
> @mdedetrich need rename and update `gunzip.md` too

Yes I did (check the PR), note sure where this error is coming from, looking into it

gmethvin:
Are we going to rename `deflate`/`inflate` as well then?

mdedetrich:
> Are we going to rename `deflate`/`inflate` as well then?

Yeah I have been thinking about this, its a bit of a weird case since deflate is the name of an algorithm but its also the action of compressing (and inflate is the action of decompression) so renaming `inflate` to `inflateDecompress` can sound a bit weird.

Gotta love english, in any case we should maybe look into this in a separate PR?

mdedetrich:
> @mdedetrich need rename and update `gunzip.md` too

Figured out the issue, we still need the old `gunzip.md` as the function is still there and visible, should work now (in Pekko 2.0.0 we can delete this file as the functions will also not exist).

pjfanning:
This is just my view. Naming is hard but renaming methods once released presents its own problems (and the community is using it). I see little reason not to have 2 methods but working towards the fast removal of the pre-existing method is something that I would question.
Deprecate the existing one by all means but I would question the idea of removing the existing ione n 2.0.0. We can remove it in 3.0.0.

mdedetrich:
> This is just my view. Naming is hard but renaming methods once released presents its own problems (and the community is using it). I see little reason not to have 2 methods but working towards the fast removal of the pre-existing method is something that I would question. Deprecate the existing one by all means but I would question the idea of removing the existing ione n 2.0.0. We can remove it in 3.0.0.

I am not sure what the issue here is, we have removed many (if not all?) deprecated methods in Pekko 2.0.0 a lot of which have more consequences than this PR as those removed methods weren't just aliases. Ontop of that, due to how the documentation is generated if we leave both `gunzip` and `gzipDecompress` then we have duplicate documentation entries.

I can see what you mean by fast removal, but practically speaking the kinds of people that will move to Pekko 2.0.x will be able to easily handle the dropped duplicated of method (along with everything else) where as the other set of people who can't move that fast will continue to use Pekko 1.x which will still be supported and have features backported.

pjfanning:
For me, there is a big difference between removing a method that was deprecated a few years ago and one that was deprecated a few days ago.

mdedetrich:
> For me, there is a big difference between removing a method that was deprecated a few years ago and one that was deprecated a few days ago.

I understand, but the 1.3.0 release will likely happen by next week where as a Pekko 2.0.x full (non milestone) release will be many months and by the time the the kinds of people that will migrate will get around to it, it will be even longer. So sure it won't be many years, but it also also won't be days.

Pekko 2.0.x already has an implicit disclaimer as part of semver that deprecated methods will be removed, this is understood by everyone. We never made a contract for users regarding how long methods need to be in existing versions of Pekko before they are removed. We can of course discuss this for the future but I would posit that if we don't get complaints from users then we shouldn't put the cart before the horse.

He-Pin:
I think we should remove it in 2.0.0, we ahave lready broken many things in 2.0.0

mdedetrich:
Okay I am going to go ahead and merge, if we wan't to rename other functions they should be discussed in separate PR's

PR #2406, opened 2025-10-27
Add gzip decompress

mdedetrich:
Forward ports https://github.com/apache/pekko/pull/2405 as well as removing the old `gunzip` functions as they are deprecated and can be removed on a major version bump.

If users complain about this when the 2.0.0 milestone is released, we can always revert the `Remove deprecated gUnzip function` commit


He-Pin:
@mdedetrich Need update the mima filter.

mdedetrich:
> @mdedetrich Need update the mima filter.

Thanks, pushed again

PR #2409, opened 2025-10-28
Add zstd compression and decompression for pekko streams

### LINKED ISSUE #2404: Add zstd compression/decompression IO Stream
[zstd decompression/compression](https://en.wikipedia.org/wiki/Zstd) should be added a standard pekko IO stream as its far superior to gzip/lzma in every way. The API design should be the same as it is for gzip.

Doing this requires using https://github.com/luben/zstd-jni, which uses the standard reference C library via JNI (unfortunately there is no such implementation in native Java and even if there would be I am not sure it would be as fast as the C one).

mdedetrich:
Resolves: https://github.com/apache/pekko/issues/2404

This PR adds a zstd compression/decompression stream/flow using [zstd-jni](https://github.com/luben/zstd-jni), this project was chosen because its the only one that has high performance (it uses the [reference zstd implementation](https://github.com/facebook/zstd) via JNI) and also supports at least JDK 17 (its published with JDK 11).

The PR still needs to be completed (documentation needs to be added along with MiMa exclusions) but I am creating a PR now with the necessary barebones so that people can comment on whether the PR is on the right track, tests have been added (there is already a base testing framework for pekko-streams compression flows `CoderSpec`).

The implementation of `ZstdCompressor`/`ZstdDecompressor` uses `ZstdDirectBufferCompressingStreamNoFinalizer`/`ZstdDirectBufferDecompressingStreamNoFinalizer` as these are the abstractions provided by zstd-jni to do streaming compression, note that the versions with `NoFinalizer` just mean that you need to explicitly shutdown the resource (which is what we want since Pekko Streams handles resource cleaning). These compression abstractions need to use a direct `ByteBuffer` in order to handle the shuffling of data between the JNI boundary so that the C implementation can do its work directly in memory,  the [zstd-jni tests](https://github.com/luben/zstd-jni/blob/9c3386d306086078155f58116a4d905e07239db4/src/test/scala/Zstd.scala) was the basis used to write the implementation.

Some extra notes

- `CoderSpec` had to be modified as the test which catches the exception to be thrown on corrupt input was hardcoded to `DataFormatException` where as zstd throws its own bespoke exception on corrupt input
- The `ZstdCompressor` implements the `Compressor` abstraction which does a lot of heavy lifting (especially when it comes to tests) however the `ZstdDecompressor` intentionally does not implement `DeflateDecompressorBase` as the design is heavily tied to Java's deflate/compression API's, instead we use `SimpleLinearGraphStage[ByteString]` backed by `ZstdDirectBufferDecompressingStreamNoFinalizer`
- The current API also allows you to specify a [dictionary](https://github.com/facebook/zstd?tab=readme-ov-file#dictionary-compression-how-to) when doing compression. Note that to do this, you need to pass a `com.github.luben.zstd.ZstdDictCompress` datastructure which is tied to the implementation of zstd-jni. There is an argument to create our own pekko equivalent of `ZstdDictCompress` which will internally map to a `com.github.luben.zstd.ZstdDictCompress`, doing so would allow us to swap to a different implementation of zstd in the future without breaking the API.
  - This is the only part of the API (aside from the `com.github.luben.zstd.ZstdIOException` exception that is thrown on corrupt input) that is tied to the zstd-jni implementation

### DIFF
--- a/actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
+++ b/actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
@@ -23,6 +23,7 @@ import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
+import org.apache.pekko.annotation.InternalStableApi;
 
 /**
  * Cleans a direct {@link ByteBuffer}. Without manual intervention, direct ByteBuffers will be
@@ -36,7 +37,8 @@ import java.nio.ByteBuffer;
  *
  * <p>See <a href=https://bugs.openjdk.java.net/browse/JDK-4724038>JDK-4724038</a>
  */
-final class ByteBufferCleaner {
+@InternalStableApi
+public final class ByteBufferCleaner {
 
   // adapted from
   // https://github.com/apache/commons-io/blob/441115a4b5cd63ae808dd4c40fc238cb52c8048f/src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java
@@ -75,7 +77,7 @@ final class ByteBufferCleaner {
    * @param buffer to release.
    * @throws IllegalStateException on internal failure.
    */
-  static void clean(final ByteBuffer buffer) {
+  public static void clean(final ByteBuffer buffer) {
     try {
       INSTANCE.clean(buffer);
     } catch (final Throwable t) {
@@ -116,7 +118,7 @@ final class ByteBufferCleaner {
    *
    * @return {@code true} if cleaning is supported, {@code false} otherwise.
    */
-  static boolean isSupported() {
+  public static boolean isSupported() {
     return INSTANCE != null;
   }
 }
--- a/docs/src/main/paradox/stream/operators/Compression/zstd.md
+++ b/docs/src/main/paradox/stream/operators/Compression/zstd.md
@@ -1,0 +1,30 @@
+# Compression.zstd
+
+Creates a flow that zstd-compresses a stream of ByteStrings.
+
+@ref[Compression operators](../index.md#compression-operators)
+
+## Signature
+
+@apidoc[Compression.zstd](stream.*.Compression$) { scala="#zstd:org.apache.pekko.stream.scaladsl.Flow[org.apache.pekko.util.ByteString,org.apache.pekko.util.ByteString,org.apache.pekko.NotUsed]" java="#zstd()" }
+
+## Description
+
+Creates a flow that zstd-compresses a stream of ByteStrings. Note that the compressor
+will SYNC_FLUSH after every @apidoc[util.ByteString] so that it is guaranteed that every @apidoc[util.ByteString]
+coming out of the flow can be fully decompressed without waiting for additional data. This may
+come at a compression performance cost for very small chunks.
+
+Use the overload method to control the compression level.
+
+## Reactive Streams semantics
+
+@@@div { .callout }
+
+**emits** when the compression algorithm produces output for the received `ByteString`
+
+**backpressures** when downstream backpressures
+
+**completes** when upstream completes
+
+@@@
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -352,7 +352,8 @@ object Dependencies {
 
   // pekko stream
 
-  lazy val stream = l ++= Seq[sbt.ModuleID](reactiveStreams, TestDependencies.scalatest)
+  lazy val stream =
+    l ++= Seq[sbt.ModuleID](reactiveStreams, "com.github.luben" % "zstd-jni" % "1.5.7-6", TestDependencies.scalatest)
 
   lazy val streamTestkit = l ++= Seq(
     TestDependencies.scalatest,
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.impl.io.compression
+
+import java.nio.ByteBuffer
+import com.github.luben.zstd.{ ZstdDictCompress, ZstdDirectBufferCompressingStreamNoFinalizer }
+import org.apache.pekko
+import org.apache.pekko.stream.scaladsl.Compression
+import pekko.annotation.InternalApi
+import pekko.util.ByteString
+
+/** INTERNAL API */
+@InternalApi private[pekko] class ZstdCompressor(
+    compressionLevel: Int =
+      Compression.ZstdDefaultCompressionLevel, dictionary: Option[ZstdDictionaryImpl] = None) extends Compressor {
+
+  private val targetBuffer = ByteBuffer.allocateDirect(65536)
+  private val compressingStream = new ZstdDirectBufferCompressingStreamNoFinalizer(targetBuffer, compressionLevel)
+
+  dictionary.foreach { dict =>
+    (dict.level, dict.length, dict.offset) match {
+      case (None, None, None) =>
+        compressingStream.setDict(dict.dictionary)
+      case (Some(dictLevel), None, None) =>
+        compressingStream.setDict(new ZstdDictCompress(dict.dictionary, dictLevel))
+      case (Some(dictLevel), Some(dictLength), Some(dictOffset)) =>
+        compressingStream.setDict(new ZstdDictCompress(dict.dictionary, dictLevel, dictLength, dictOffset))
+      case _ =>
+        throw new IllegalArgumentException("Invalid combination of ZstdDictionary parameters")
+    }
+  }
+
+  override def compress(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
+    inputBB.put(input.toArrayUnsafe())
+    inputBB.flip()
+    compressingStream.compress(inputBB)
+    val result = ByteString.fromByteBuffer(targetBuffer)
+    targetBuffer.flip()
+    result
+  }
+
+  override def flush(): ByteString = {
+    targetBuffer.flip()
+    val result = ByteString.fromByteBuffer(targetBuffer)
+    targetBuffer.clear()
+    compressingStream.flush()
+    result
+  }
+
+  override def finish(): ByteString = {
+    compressingStream.close()
+    targetBuffer.flip()
+    val arr = Array.ofDim[Byte](targetBuffer.limit())
+    targetBuffer.get(arr)
+    val result = ByteString.fromArrayUnsafe(arr)
+    result
+  }
+
+  override def compressAndFlush(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
+    inputBB.put(input.toArrayUnsafe())
+    inputBB.flip()
+    compressingStream.compress(inputBB)
+    compressingStream.flush()
+    targetBuffer.flip()
+
+    val arr = new Array[Byte](targetBuffer.limit())
+    targetBuffer.get(arr)
+    targetBuffer.clear()
+    ByteString.fromArrayUnsafe(arr)
+  }
+
+  override def compressAndFinish(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
+    inputBB.put(input.toArrayUnsafe())
+    inputBB.flip()
+    compressingStream.compress(inputBB)
+    compressingStream.close()
+    targetBuffer.flip()
+
+    val arr = new Array[Byte](targetBuffer.limit())
+    targetBuffer.get(arr)
+    ByteString.fromArrayUnsafe(arr)
+  }
+
+  override def close(): Unit = compressingStream.close()
+}
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdDecompressor.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdDecompressor.scala
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.impl.io.compression
+
+import java.nio.ByteBuffer
+
+import scala.util.control.NonFatal
+
+import com.github.luben.zstd.{ Zstd, ZstdDirectBufferDecompressingStreamNoFinalizer }
+
+import org.apache.pekko
+import pekko.annotation.InternalApi
+import pekko.io.ByteBufferCleaner
+import pekko.stream.Attributes
+import pekko.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import pekko.stream.stage.{ GraphStageLogic, InHandler, OutHandler }
+import pekko.util.ByteString
+
+/** INTERNAL API */
+@InternalApi private[pekko] class ZstdDecompressor(maxBytesPerChunk: Int =
+      Zstd.blockSizeMax()) extends SimpleLinearGraphStage[ByteString] {
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    new GraphStageLogic(shape) with InHandler with OutHandler {
+
+      private val sourceBuffer = ByteBuffer.allocateDirect(maxBytesPerChunk)
+
+      // This is initialized here to avoid an allocation per onPush, remember to clear before using
+      private val outputBuffer = ByteBuffer.allocateDirect(maxBytesPerChunk)
+      private val decompressingStream = new ZstdDirectBufferDecompressingStreamNoFinalizer(sourceBuffer)
+
+      override def onPush(): Unit = {
+        sourceBuffer.clear()
+        val inputArray = grab(in).toArrayUnsafe()
+        sourceBuffer.put(inputArray)
+        sourceBuffer.flip()
+        if (sourceBuffer.hasRemaining) {
+          var result = ByteString.empty
+          while (sourceBuffer.hasRemaining) {
+            outputBuffer.clear()
+            decompressingStream.read(outputBuffer)
+            outputBuffer.flip()
+            val outputArray = new Array[Byte](outputBuffer.limit())
+            outputBuffer.get(outputArray)
+            result = result.concat(ByteString.fromArrayUnsafe(outputArray))
+          }
+
+          // Fencepost case, it's possible to still have sourceBuffer.hasRemaining and yet decompression result
+          // not output anything
+          if (result.nonEmpty)
+            push(out, result)
+          else
+            pull(in)
+        } else pull(in)
+        sourceBuffer.flip()
+      }
+
+      override def onPull(): Unit = pull(in)
+
+      override def onUpstreamFinish(): Unit = {
+        sourceBuffer.flip()
+        if (sourceBuffer.hasRemaining) {
+          var result = ByteString.empty
+          while (sourceBuffer.hasRemaining) {
+            outputBuffer.clear()
+            decompressingStream.read(outputBuffer)
+            outputBuffer.flip()
+            val outputArray = new Array[Byte](outputBuffer.limit())
+            outputBuffer.get(outputArray)
+            result = result.concat(ByteString.fromArrayUnsafe(outputArray))
+          }
+          decompressingStream.close()
+
+          // Fencepost case, it's possible to still have sourceBuffer.hasRemaining and yet decompression result
+          // not output anything
+          if (result.nonEmpty)
+            emit(out, result)
+        }
+        completeStage()
+      }
+
+      override def postStop(): Unit = {
+        decompressingStream.close()
+
+        if (ByteBufferCleaner.isSupported)
+          try {
+            ByteBufferCleaner.clean(sourceBuffer)
+            ByteBufferCleaner.clean(outputBuffer)
+          } catch {
+            case NonFatal(_) => /* ok, best effort attempt to cleanup failed */
+          }
+      }
+
+      setHandlers(in, out, this)
+    }
+  }
+
+}
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
@@ -1,0 +1,210 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.javadsl.compression;
+
+import org.apache.pekko.annotation.InternalApi;
+import org.apache.pekko.stream.impl.io.compression.ZstdDictionaryImpl;
+import org.apache.pekko.util.ByteString;
+import org.apache.pekko.util.OptionalUtil;
+import scala.Option;
+
+import java.util.OptionalInt;
+
+/**
+ * Configuration class based on the official C zstd library to specify dictionary options for
+ * compression.
+ *
+ * @see <a href="https://github.com/facebook/zstd?tab=readme-ov-file#dictionary-compression-how-to">Zstd dictionary readme</a>
+ * @see <a href="https://facebook.github.io/zstd/zstd_manual.html#Chapter10">Zstd C manual</a>
+ * @since 2.0.0
+ */
+public class ZstdDictionary {
+    byte[] dictionary;
+    private OptionalInt level = OptionalInt.empty();
+    private OptionalInt offset = OptionalInt.empty();
+    private OptionalInt length = OptionalInt.empty();
+
+    public ZstdDictionary(byte[] dictionary) {
+        this.dictionary = dictionary;
+    }
+
+    public ZstdDictionary(byte[] dictionary, int level) {
+        this.dictionary = dictionary;
+        this.level = OptionalInt.of(level);
+    }
+
+    public ZstdDictionary(byte[] dictionary, Integer level) {
+        this.dictionary = dictionary;
+        if (level != null) {
+            this.level = OptionalInt.of(level);
+        }
+    }
+
+    public ZstdDictionary(byte[] dictionary, int level, int offset, int length) {
+        this.dictionary = dictionary;
+        this.level = OptionalInt.of(level);
+        this.offset = OptionalInt.of(offset);
+        this.length = OptionalInt.of(length);
+    }
+
+    public ZstdDictionary(byte[] dictionary, Integer level, Integer offset, Integer length) {
+        this.dictionary = dictionary;
+        if (level != null) {
+            this.level = OptionalInt.of(level);
+        }
+        if (offset != null) {
+            this.offset = OptionalInt.of(offset);
+        }
+        if (length != null) {
+            this.length = OptionalInt.of(length);
+        }
+    }
+
+    public ZstdDictionary(ByteString dictionary) {
+        this(dictionary.toArray());
+    }
+
+    public ZstdDictionary(ByteString dictionary, int level) {
+        this(dictionary.toArray(), level);
+    }
+
+    public ZstdDictionary(ByteString dictionary, Integer level) {
+        this(dictionary.toArray(), level);
+    }
+
+    public ZstdDictionary(ByteString dictionary, int level, int offset, int length) {
+        this(dictionary.toArray(), level, offset, length);
+    }
+
+    public ZstdDictionary(ByteString dictionary, Integer level, Integer offset, Integer length) {
+        this(dictionary.toArray(), level, offset, length);
+    }
+
+    public ZstdDictionary(String dictionary) {
+        this(dictionary.getBytes());
+    }
+
+    public ZstdDictionary(String dictionary, int level) {
+        this(dictionary.getBytes(), level);
+    }
+
+    public ZstdDictionary(String dictionary, Integer level) {
+        this(dictionary.getBytes(), level);
+    }
+
+    public ZstdDictionary(String dictionary, int level, int offset, int length) {
+        this(dictionary.getBytes(), level, offset, length);
+    }
+
+    public ZstdDictionary(String dictionary, Integer level, Integer offset, Integer length) {
+        this(dictionary.getBytes(), level, offset, length);
+    }
+
+    public byte[] getDictionary() {
+        return dictionary;
+    }
+
+    public OptionalInt getLevel() {
+        return level;
+    }
+
+    public OptionalInt getOffset() {
+        return offset;
+    }
+
+    public OptionalInt getLength() {
+        return length;
+    }
+
+    public void setDictionary(byte[] dictionary) {
+        this.dictionary = dictionary;
+    }
+
+    public void setLevel(int level) {
+        this.level = OptionalInt.of(level);
+    }
+
+    public void setLevel(OptionalInt level) {
+        this.level = level;
+    }
+
+    public void setLevel(Integer level) {
+        if (level == null) {
+            this.level = OptionalInt.empty();
+        } else {
+            this.level = OptionalInt.of(level);
+        }
+    }
+
+    public void setOffset(int offset) {
+        this.offset = OptionalInt.of(offset);
+    }
+
+    public void setOffset(Integer offset) {
+        if (offset == null) {
+            this.offset = OptionalInt.empty();
+        } else {
+            this.offset = OptionalInt.of(offset);
+        }
+    }
+
+    public void setOffset(OptionalInt offset) {
+        this.offset = offset;
+    }
+
+    public void setLength(int length) {
+        this.length = OptionalInt.of(length);
+    }
+
+    public void setLength(Integer length) {
+        if (length == null) {
+            this.length = OptionalInt.empty();
+        } else {
+            this.length = OptionalInt.of(length);
+        }
+    }
+
+    public void setLength(OptionalInt length) {
+        this.length = length;
+    }
+
+    @InternalApi
+    public ZstdDictionaryImpl toImpl() {
+        return new ZstdDictionaryImpl() {
+            @Override
+            public byte[] dictionary() {
+                return dictionary;
+            }
+
+            @Override
+            public Option<Object> level() {
+                return OptionalUtil.convertOptionalToScala(level);
+            }
+
+            @Override
+            public Option<Object> offset() {
+                return OptionalUtil.convertOptionalToScala(offset);
+            }
+
+            @Override
+            public Option<Object> length() {
+                return OptionalUtil.convertOptionalToScala(length);
+            }
+        };
+    }
+}
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala
@@ -15,9 +15,12 @@ package org.apache.pekko.stream.scaladsl
 
 import java.util.zip.Deflater
 
+import com.github.luben.zstd._
+
 import org.apache.pekko
 import pekko.NotUsed
 import pekko.stream.impl.io.compression._
+import pekko.stream.scaladsl.compression.ZstdDictionary
 import pekko.util.ByteString
 
 object Compression {
@@ -104,4 +107,68 @@ object Compression {
    */
   def inflate(maxBytesPerChunk: Int, nowrap: Boolean): Flow[ByteString, ByteString, NotUsed] =
     Flow[ByteString].via(new DeflateDecompressor(maxBytesPerChunk, nowrap)).named("inflate")
+
+  /**
+   * The minimum compression level supported by zstd
+   * @since 2.0.0
+   */
+  final val ZstdMinCompressionLevel: Int = Zstd.minCompressionLevel()
+
+  /**
+   * The maximum compression level supported by zstd
+   * @since 2.0.0
+   */
+  final val ZstdMaxCompressionLevel: Int = Zstd.maxCompressionLevel()
+
+  /**
+   * The zstd default compression level
+   * @since 2.0.0
+   */
+  final val ZstdDefaultCompressionLevel: Int = Zstd.defaultCompressionLevel()
+
+  /**
+   * @since 2.0.0
+   */
+  def zstd: Flow[ByteString, ByteString, NotUsed] = zstd(ZstdDefaultCompressionLevel, None)
+
+  /**
+   * Same as [[zstd]] with a custom level and an optional dictionary.
+   *
+   * @param level      The compression level, must be greater or equal to [[ZstdMinCompressionLevel]] and less than or equal
+   *                   to [[ZstdMaxCompressionLevel]]
+   * @param dictionary An optional dictionary that can be used for compression
+   * @since 2.0.0
+   */
+  def zstd(level: Int, dictionary: Option[ZstdDictionary]): Flow[ByteString, ByteString, NotUsed] = {
+    require(level <= ZstdMaxCompressionLevel && level >= ZstdMinCompressionLevel)
+    CompressionUtils.compressorFlow(() => new ZstdCompressor(level, dictionary.map(_.toImpl)))
+  }
+
+  /**
+   * Same as [[zstd]] with a custom level, optional dictionary and configurable flush mode.
+   *
+   * @param level      The compression level, must be greater or equal to [[ZstdMinCompressionLevel]] and less than or equal
+   *                   to [[ZstdMaxCompressionLevel]]
+   * @param dictionary An optional dictionary that can be used for compression
+   * @param autoFlush  If true will automatically flush after every single element in the stream.
+   * @since 2.0.0
+   */
+  def zstd(level: Int, dictionary: Option[ZstdDictionary], autoFlush: Boolean)
+      : Flow[ByteString, ByteString, NotUsed] = {
+    require(level <= ZstdMaxCompressionLevel && level >= ZstdMinCompressionLevel)
+    CompressionUtils.compressorFlow(() => new ZstdCompressor(level, dictionary.map(_.toImpl)), autoFlush)
+  }
+
+  /**
+   * The maximum block size used by zstd decompression
+   * @since 2.0.0
+   */
+  final val ZstdDecompressMaxBlockSize: Int = Zstd.blockSizeMax()
+
+  /**
+   * @since 2.0.0
+   */
+  def zstdDecompress(maxBytesPerChunk: Int = ZstdDecompressMaxBlockSize): Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString].via(new ZstdDecompressor(maxBytesPerChunk)).named("zstdDecompress")
+
 }
--- a/docs/src/main/paradox/stream/operators/Compression/zstdDecompress.md
+++ b/docs/src/main/paradox/stream/operators/Compression/zstdDecompress.md
@@ -1,0 +1,26 @@
+# Compression.zstdDecompress
+
+Creates a flow that zstd-decompresses a stream of ByteStrings.
+
+@ref[Compression operators](../index.md#compression-operators)
+
+## Signature
+
+@apidoc[Compression.zstdDecompress](stream.*.Compression$) { scala="#zstdDecompress(maxBytesPerChunk:Int):org.apache.pekko.stream.scaladsl.Flow[org.apache.pekko.util.ByteString,org.apache.pekko.util.ByteString,org.apache.pekko.NotUsed]" java="#zstdDecompress(int)" }
+
+## Description
+
+Creates a flow that zstd-decompresses a stream of ByteStrings. If the input is truncated, uses invalid
+compression method or is invalid (failed CRC checks) this operator fails with a `com.github.luben.zstd.ZstdIOException`.
+
+## Reactive Streams semantics
+
+@@@div { .callout }
+
+**emits** when the decompression algorithm produces output for the received `ByteString` (the emitted `ByteString` is of `maxBytesPerChunk` maximum length)
+
+**backpressures** when downstream backpressures
+
+**completes** when upstream completes
+
+@@@
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/CoderSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/CoderSpec.scala
@@ -15,11 +15,11 @@ package org.apache.pekko.stream.io.compression
 
 import java.io.{ ByteArrayOutputStream, InputStream, OutputStream }
 import java.util.concurrent.ThreadLocalRandom
-import java.util.zip.DataFormatException
 
 import scala.annotation.tailrec
 import scala.concurrent.Await
 import scala.concurrent.duration._
+import scala.reflect.ClassTag
 import scala.util.control.NoStackTrace
 
 import org.apache.pekko
@@ -31,7 +31,8 @@ import pekko.util.ByteString
 import org.scalatest.Inspectors
 import org.scalatest.wordspec.AnyWordSpec
 
-abstract class CoderSpec(codecName: String) extends AnyWordSpec with CodecSpecSupport with Inspectors {
+abstract class CoderSpec[CorruptInputException: ClassTag](codecName: String) extends AnyWordSpec with CodecSpecSupport
+    with Inspectors {
   import CompressionTestingTools._
 
   protected def newCompressor(): Compressor
@@ -85,7 +86,7 @@ abstract class CoderSpec(codecName: String) extends AnyWordSpec with CodecSpecSu
       "throw an error on corrupt input" in {
         (the[RuntimeException] thrownBy {
           ourDecode(corruptContent)
-        }).ultimateCause should be(a[DataFormatException])
+        }).ultimateCause should be(a[CorruptInputException])
       }
     }
 
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/DeflateSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/DeflateSpec.scala
@@ -21,7 +21,7 @@ import pekko.stream.impl.io.compression.{ Compressor, DeflateCompressor }
 import pekko.stream.scaladsl.{ Compression, Flow }
 import pekko.util.ByteString
 
-class DeflateSpec extends CoderSpec("deflate") {
+class DeflateSpec extends CoderSpec[DataFormatException]("deflate") {
   import CompressionTestingTools._
 
   protected def newCompressor(): Compressor = new DeflateCompressor
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/GzipSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/GzipSpec.scala
@@ -15,14 +15,14 @@ package org.apache.pekko.stream.io.compression
 
 import java.io.{ InputStream, OutputStream }
 import java.nio.charset.StandardCharsets
-import java.util.zip.{ GZIPInputStream, GZIPOutputStream, ZipException }
+import java.util.zip.{ DataFormatException, GZIPInputStream, GZIPOutputStream, ZipException }
 
 import org.apache.pekko
 import pekko.stream.impl.io.compression.{ Compressor, GzipCompressor }
 import pekko.stream.scaladsl.{ Compression, Flow }
 import pekko.util.ByteString
 
-class GzipSpec extends CoderSpec("gzip") {
+class GzipSpec extends CoderSpec[DataFormatException]("gzip") {
   import CompressionTestingTools._
 
   protected def newCompressor(): Compressor = new GzipCompressor
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/ZstdAutoFlushSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/ZstdAutoFlushSpec.scala
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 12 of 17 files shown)


project/Dependencies.scala
```diff
@@ -351,7 +351,8 @@ object Dependencies {
 
   // pekko stream
 
-  lazy val stream = l ++= Seq[sbt.ModuleID](reactiveStreams, TestDependencies.scalatest)
+  lazy val stream =
+    l ++= Seq[sbt.ModuleID](reactiveStreams, "com.github.luben" % "zstd-jni" % "1.5.7-6", TestDependencies.scalatest)
```
pjfanning, line 356: can we make the zstd dependency optional or provided?

mdedetrich: Making it optional/provided is not going to give as any benefits as we would need to force users to redundantly provide a zstd implementation (which is going to be `"com.github.luben" % "zstd-jni" % "1.5.7-6"` anyways).

To do this properly, we would need to make a zstd agnostic API and then provided concrete implementations in a different artifact and considering that there is only a single candidate for a proper implementation (as discussed at https://github.com/apache/pekko/issues/2404#issuecomment-3450106736) we wouldn't gain any benefit out of it.

Hence zstd is being treated the same way as lz4Java is (for the same reasons), where we have a [simple direct dependency](https://github.com/apache/pekko/blob/1b54c97914e91af287af8681875bee317e968c58/project/Dependencies.scala#L106)

pjfanning: How about having a separate pekko-streams-ztsd or pekko-streams-compression so that pekko-streams core lib doesn't need this dependency? ZIO have done something similar where a small number of compression operators are in zio-streams but others have their own libs, eg zio-streams-compress-zstd.

mdedetrich: On another note there is a silver lining here, since zstd-jni uses JNI, the C shared objects which are platform dependent are contained within the jar artifact. The `"com.github.luben" % "zstd-jni" % "1.5.7-6"` artifact contains the C shared objects for **all** of the platforms (see https://github.com/luben/zstd-jni?tab=readme-ov-file#binary-releases for which ones are supported) but there are platform dependent jars if you only care about a single platform.

This does give some credence to using provided, as a user can provide their own zstd-jni artifact however there are still some things to note here. If we end up using `provided` then we force all pekko end users to add a zstd-jni artifact to their dependencies (with `"com.github.luben" % "zstd-jni" % "1.5.7-6"` being the documented default) as its not going to be including by in the dependency resolution.

mdedetrich: > How about having a separate pekko-streams-ztsd or pekko-streams-compression so that pekko-streams core lib doesn't need this dependency? ZIO have done something similar where a small number of compression operators are in zio-streams but others have their own libs, eg zio-streams-compress-zstd.

You could make the same argument as to why lz4Java isn't its own artifact, or any of the other compression algorithms. Also, due to the way the current API is designed, all of the compression algorithms are contained within the pekko-streams `Compression` object (which we could monkey patch to add more methods as a separate artifact, but thats starting to look ugly already). Ontop of that, when adding zstd compression to pekko http request/response headers (see https://github.com/apache/pekko-http/issues/860), we would likewise have to create yet another artifact for pekko-http.

`zio-streams-compress` is also not the best to compare to, since if you look at https://zio.dev/zio-streams-compress/ you will see they also have separate artifacts for `gzip` and `lz4` something is contained within pekko-streams core artifact for us. zio appears to have a different design ethos here, where they err on the side of hyper modularity and put everything in its own artifact where as pekko is on the other side.

I understand the sentiment, but honestly this is creating a huge amount of ceremony for no real benefit and the only reason we are having this discussion is because of how behind Java is in adding what are de facto standard compression algorithms to the stdlib (of which zstd has been for at least half a decade now). The only reason I can see to use separate artifacts would be to support Pekko 1.x series, but zstd-jni only supports JDK 11 or higher and this would then need to be advertised to end users.

`zio-streams-compress` is also not the best comparison, since if you look at https://zio.dev/zio-streams-compress/ you will see they also have separate artifacts for `gzip` and `lz4`, where as for us these compression algorithms are contained within the core pekko-streams artifact. zio appears to have a different design ethos here, where they err on the side of hyper modularity and put everything in its own artifact where as pekko is on the other side.

Put differently, the entire point of `Compression` object for scaladsl/javadsl is to contain the standard compression/decompression algorithms, and zstd is one of those. While having to use zstd-jni is not ideal, to the end user there is no difference as it already supports almost every architecture that would be used in the wild.

mdedetrich:
Since the core tests have passed (the github action says failed but thats due to a missing paradox documentation) I have added other reviewers

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala
```diff
@@ -85,4 +87,28 @@ object Compression {
...
+   * @since 2.0.0
+   */
+  def zstd: Flow[ByteString, ByteString, NotUsed] = zstd(Zstd.defaultCompressionLevel())
+
+  /**
+   * Same as [[zstd]] with a custom level and an optional dictionary.
+   * @param level The compression level, must be greater or equal to [[Zstd.minCompressionLevel]] and less than or equal
+   *              to [[Zstd.maxCompressionLevel]]
+   * @param dictionary An optional dictionary that can be used for compression
+   * @since 2.0.0
```
He-Pin, line 140: 1.3.0 ?

He-Pin, line 140: 1.3.0 ?

mdedetrich:
@He-Pin This cannot be backported as the zstd-jni library is built against JDK 11 and Pekko 1.x series requires every dependency to be built against JDK 1.8 or lower

He-Pin:
@mdedetrich Not sure if it's better to be a dedicated module, I will review it this weekend.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdDecompressor.scala
```diff
@@ -0,0 +1,99 @@
...
+          }
+
+          // Fencepost case, it's possible to still have sourceBuffer.hasRemaining and yet compression result
+          // not output anything
+          if (result.nonEmpty)
+            push(out, result)
+          else
+            pull(in)
+        } else pull(in)
+        sourceBuffer.flip()
```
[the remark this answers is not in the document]
mdedetrich: This just plain wrong

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdDecompressor.scala
```diff
@@ -0,0 +1,99 @@
...
+            outputBuffer.clear()
+            decompressingStream.read(outputBuffer)
+            outputBuffer.flip()
+            val outputArray = new Array[Byte](outputBuffer.limit())
+            outputBuffer.get(outputArray)
+            result = result.concat(ByteString.fromArrayUnsafe(outputArray))
+          }
+          decompressingStream.close()
+
+          // Fencepost case, it's possible to still have sourceBuffer.hasRemaining and yet compression result
```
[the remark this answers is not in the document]
mdedetrich: Done

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdDecompressor.scala
```diff
@@ -0,0 +1,99 @@
...
+          while (sourceBuffer.hasRemaining) {
+            outputBuffer.clear()
+            decompressingStream.read(outputBuffer)
+            outputBuffer.flip()
+            val outputArray = new Array[Byte](outputBuffer.limit())
+            outputBuffer.get(outputArray)
+            result = result.concat(ByteString.fromArrayUnsafe(outputArray))
+          }
+
+          // Fencepost case, it's possible to still have sourceBuffer.hasRemaining and yet compression result
```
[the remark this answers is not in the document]
mdedetrich: Done

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  private val targetBuffer = ByteBuffer.allocateDirect(65536)
+  private val compressingStream = new ZstdDirectBufferCompressingStreamNoFinalizer(targetBuffer, compressionLevel)
+
+  dictionary.foreach(compressingStream.setDict)
+
+  override def compress(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
+    inputBB.put(input.toArrayUnsafe())
+    inputBB.flip()
```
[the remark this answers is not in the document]
mdedetrich: Fixed and pushed.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,93 @@
...
+    compressingStream.compress(input.asByteBuffer)
+    val result = ByteString.fromByteBuffer(targetBuffer)
+    targetBuffer.flip()
```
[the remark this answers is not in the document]
mdedetrich: The ordering here doesn't really matter, as all that matters is the `targetBuffer.flip()` happens after the read at `compressingStream.compress(inputBB)`.

mdedetrich:
I have finished the docs for paradox (hence completing the PR) and also fixed some minor things, all resolved comments from the copilot review (including ones which were outright wrong).

There is still an open question about whether we should make our own `com.github.luben.zstd.ZstdDictCompress` datastructure so that the API is abstracted away from the implementation.

mdedetrich:
> @mdedetrich Not sure if it's better to be a dedicated module, I will review it this weekend.

There is a discussion on this topic at https://github.com/apache/pekko/pull/2409#discussion_r2469468261, I personally think making this an extra module adds a lot of ceremony (as we would also have to add it another module in pekko-http) for no real benefit to end user but if a lot of people feel strongly about it I can reconsider.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala
```diff
@@ -85,4 +87,28 @@ object Compression {
...
+  def zstd(level: Int, dictionary: Option[ZstdDictCompress] = None): Flow[ByteString, ByteString, NotUsed] = {
+    require(level <= Zstd.maxCompressionLevel() && level >= Zstd.minCompressionLevel())
+    CompressionUtils.compressorFlow(() => new ZstdCompressor(level, dictionary))
+  }
+
+  /**
+   * @since 2.0.0
+   */
+  def zstdDecompress(maxBytesPerChunk: Int = MaxBytesPerChunkDefault): Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString].via(new ZstdDecompressor(maxBytesPerChunk)).named("zstdDecompress")
```
He-Pin, line 172: Can we make this live inside `ZstdCompression`?

He-Pin: I mean, what if the user wants to use it in Android?

mdedetrich: Why, its sitting in the same spot that all other compression flows like deflate or gzip do? If we decide to put this in another artifact then sure, but that is another debate.

Either we treat zstd as a standard widely adopted industry compression algorithm (which it is) in which this PR has it in the appropriate place or we put it in another artifact and if so we should have the discussion then.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,93 @@
...
+    compressionLevel: Int =
+      Zstd.defaultCompressionLevel(), dictionary: Option[ZstdDictCompress] = None) extends Compressor {
+
+  private val targetBuffer = ByteBuffer.allocateDirect(65536)
+  private val compressingStream = new ZstdDirectBufferCompressingStreamNoFinalizer(targetBuffer, compressionLevel)
+
+  dictionary.foreach(compressingStream.setDict)
+
+  override def compress(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
```
He-Pin, line 49: There is a ByteBufferPool, can we use that here?

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala
jrudolph: One difference is that zlib is distributed differently and already part of the JDK while zstd-jni might bring extra weight that not everyone is interested in. The full jar including binaries for all platforms is >7MB. Whether that is significant or not is a different question but it goes somewhat against the usual idea to keep pekko-stream reasonably minimal and add stuff like this to the connectors (which probably does not work if we want to depend on it from http). So, I'd also slightly favor having an extra module at least in pekko core.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,93 @@
...
+import java.nio.ByteBuffer
+
+import com.github.luben.zstd.{ Zstd, ZstdDictCompress, ZstdDirectBufferCompressingStreamNoFinalizer }
+
+import org.apache.pekko
+import pekko.annotation.InternalApi
+import pekko.util.ByteString
+
+/** INTERNAL API */
+@InternalApi private[pekko] class ZstdCompressor(
```
He-Pin, line 28: mark it as not threadsafe

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
mdedetrich: Possibly? I just read the implementation now and it appears that the main use case of `ByteBufferPool ` is if you need a pool of buffers where as in our case we only have a static set amount of direct byte buffers and they have different sizes.

However from reading the `ByteBufferPool` code I did found another improvement, we can use `ByteBufferCleaner.clean(byteBuffer)` to clean up the byte buffers in a timely manner (will implement this).

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
mdedetrich: You mean with a comment or is there some annotation I am not aware of?

mdedetrich:
I have just added another commit where we use the `ByteBufferCleaner` to clean up the direct byte buffers rather than resorting to GC (see my [comment](https://github.com/apache/pekko/pull/2409#discussion_r2472754494)). There is however an open question about whether we should do

```scala
if (ByteBufferCleaner.isSupported)
  ByteBufferCleaner.clean(inputBB)
```

on the `compress` method in `ZstdCompressor` as this will trigger a cleanup on every `onPush` of the stream. I don't know how expensive this cleanup is, there can be an argument that for this specific case we should defer it for GC as GC cleaning up many ByteBuffers as a batch would be faster, @jrudolph can you comment?

jrudolph:
> on the `compress` method in `ZstdCompressor` as this will trigger a cleanup on every `onPush` of the stream. I don't know how expensive this cleanup is, there can be an argument that for this specific case we should defer it for GC as GC cleaning up many ByteBuffers as a batch would be faster, @jrudolph can you comment?

There used to be a reason to use buffer pools especially for direct buffers, probably not least because you might fragment the native heap. Would probably make sense to not churn through them in quick succession. Why not use the buffer pool from the zstd-jni jar?

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala
mdedetrich: > One difference is that zlib is distributed differently and already part of the JDK while zstd-jni might bring extra weight that not everyone is interested in. The full jar including binaries for all platforms is >7MB. Whether that is significant or not is a different question but it goes somewhat against the usual idea to keep pekko-stream reasonably minimal and add stuff like this to the connectors (which probably does not work if we want to depend on it from http). So, I'd also slightly favor having an extra module at least in pekko core.

What would you recommend here, because there are many ways to go about this? For example, it would be possible to have the `zstd`/`zstdDecompress` method in the `Compression` object so that the API is clean but then the user would have to provide their own artifact (of which we can provide a default that uses `zstd-jni`).

Or should the artifact just contain everything related to zstd and should the artifact be a `pekko-streams-compress-extra` (which initially contains a zstd)  or `pekko-streams-zstd`?

One thing to note is that when doing this, I also have https://github.com/apache/pekko-http/issues/860 in mind, and I wanted to add zstd to the standard spot where all of our content encoder's are i.e. https://github.com/apache/pekko-http/blob/4833a8e42f946682a72a72a0f3bee4c4d662b9a6/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/headers/HttpEncoding.scala#L63-L76

mdedetrich:
> > on the `compress` method in `ZstdCompressor` as this will trigger a cleanup on every `onPush` of the stream. I don't know how expensive this cleanup is, there can be an argument that for this specific case we should defer it for GC as GC cleaning up many ByteBuffers as a batch would be faster, @jrudolph can you comment?
> 
> There used to be a reason to use buffer pools especially for direct buffers, probably not least because you might fragment the native heap. Would probably make sense to not churn through them in quick succession. Why not use the buffer pool from the zstd-jni jar?

Ah I just noticed the buffer pool in zstd-jni, it wasn't being used in the sample code with `ZstdDirectBufferCompressingStream`/`ZstdDirectBufferDecompressingStream`, ill look into it, thanks!

mdedetrich:
> > on the `compress` method in `ZstdCompressor` as this will trigger a cleanup on every `onPush` of the stream. I don't know how expensive this cleanup is, there can be an argument that for this specific case we should defer it for GC as GC cleaning up many ByteBuffers as a batch would be faster, @jrudolph can you comment?
> 
> There used to be a reason to use buffer pools especially for direct buffers, probably not least because you might fragment the native heap. Would probably make sense to not churn through them in quick succession. Why not use the buffer pool from the zstd-jni jar?

So I just tried doing this and it doesn't work with `ZstdDirectBufferCompressingStreamNoFinalizer`/`ZstdDirectBufferDecompressingStreamNoFinalizer` as they expect direct byte buffers. zstd-jni's `RecyclingBufferPool` doesn't return direct byte buffers, instead its used internally with `ZstdInputStreamNoFinalizer`/`ZstOutputStreamNoFinalizer` which leaves me with 2 options, either

1. Use Pekko's `ByteBufferPool` and rewrite the `ZstdCompressor` logic to deal with fixed size direct `ByteBuffer`'s (shouldn't be too bad)

or

2. Use zstd-jni's `ZstdInputStreamNoFinalizer`/`ZstOutputStreamNoFinalizer`, which would likely be simpler but also would be less performant

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,104 @@
...
+
+import org.apache.pekko
+import pekko.annotation.InternalApi
+import pekko.io.ByteBufferCleaner
+import pekko.util.ByteString
+
+/** INTERNAL API */
+@InternalApi private[pekko] class ZstdCompressor(
+    compressionLevel: Int =
+      Zstd.defaultCompressionLevel(), dictionary: Option[ZstdDictCompress] = None) extends Compressor {
```
He-Pin: dictionary can be a `OptionVal`

He-Pin: And should we provide a helper utility to train the dictionary, too?

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,104 @@
...
+  private val compressingStream = new ZstdDirectBufferCompressingStreamNoFinalizer(targetBuffer, compressionLevel)
+
+  dictionary.foreach(compressingStream.setDict)
+
+  override def compress(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
+    inputBB.put(input.toArrayUnsafe())
+    inputBB.flip()
+    compressingStream.compress(inputBB)
+    val result = ByteString.fromByteBuffer(targetBuffer)
```
He-Pin, line 53: I missed Netty's bytebuf here, when we release the bytebuf, it will returns to the pool, which avloid the fragment.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
He-Pin: what about the `compressDirectByteBuffer` method?

mdedetrich: This won't give as anything, as we would have to redundantly convert from an `Optional`/`Optional` to an `OptionVal` since the public scaladsl/javadsl uses `Option`/`Optional`

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
mdedetrich: I will rewrite the implementation to use the pekko `DirectByteBufferPool`

mdedetrich:
So I have been thinking about this, and I am coming to the conclusion that I should just reimplement the zstd compressor from scratch using the low level zstd-jni API (i.e. `public int compress(byte[] dst, byte[] src)` within the `ZstdCompressCtx` class. The reasoning is as follows

- `ZstdDirectBufferCompressingStreamNoFinalizer` and related classes were designed to work with sources that provide direct `ByteBuffer`'s i.e. Netty/Java NIO however in our case we are dealing with `ByteString` where we don't have direct `ByteBuffer`'s (`ByteString.asByteBuffer` is not a direct byte buffer) so we always need up having to copy into an array that then gets pushed into a direct `ByteBuffer` before going into the compress function. Using `ZstdCompressCtx` we can just directly feed/get the byte arrays.
- Now lets get onto buffering
  -  When it comes to compression, whether we want to buffer or not actually depends on the size of the `ByteString` element thats incoming onto the stream.
    - If the element happens to be really "big", its actually ideal to compress that element as is without feeding it into any buffer as the larger the element, the better the compression ratio is. In this case we can just use `input.toByteArrayUnsafe` and feed it directly into `ZstdCompressCtx.compress`, even better we don't need to worry about any copying in happy path case for small `ByteString`.
    - If the input `ByteStrings` is "small", then we can keep on asking for elements to buffer and once it hits a constant size limit we can then feed it into `ZstdCompressCtx.compress`
- When it comes to decompression, the logic can reman as is. We don't need to use a `DirectByteBufferPool` because we only create 2 buffers per stream instantiation (**not** per element of stream) and those `ByteBuffer`'s are reused for each incoming element, albeit it should be updated so that those `ByteBuffer`'s get cleaned up with `if (ByteBufferCleaner.isSupported) ByteBufferCleaner.clean(byteBuffer)`. Using `ZstdDecompressCtx` isn't going to gain us much net benefit as we need to use the native streaming capabilities of zstd-jni to accept incoming data until we can get to a point where we can decompress.

What are peoples thoughts here, is my thinking of the right track here?

pjfanning:
* avoiding unnecessary intermediate transformation to InputStreams/ByteBuffers seems good
* worth prototyping anyway

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,104 @@
...
+    targetBuffer.get(arr)
+    val result = ByteString.fromArrayUnsafe(arr)
+    if (ByteBufferCleaner.isSupported)
+      ByteBufferCleaner.clean(targetBuffer)
+    result
+  }
+
+  override def compressAndFlush(input: ByteString): ByteString = {
+    val inputBB = ByteBuffer.allocateDirect(input.size)
+    inputBB.put(input.toArray)
```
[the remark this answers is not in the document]
pjfanning: input.toArrayUnsafe seems ok and better here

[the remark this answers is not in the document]
mdedetrich: Ah right that was an accident, in any case as per https://github.com/apache/pekko/pull/2409#issuecomment-3462846239 this is going to be recoded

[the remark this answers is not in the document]
He-Pin: Should we mark it `@ApiMayChange` too?

raboof:
> So I have been thinking about this, and I am coming to the conclusion that I should just reimplement the zstd compressor from scratch using the low level zstd-jni API (i.e. `public int compress(byte[] dst, byte[] src)` within the `ZstdCompressCtx` class. The reasoning is as follows
> 
> * `ZstdDirectBufferCompressingStreamNoFinalizer` and related classes were designed to work with sources that provide direct `ByteBuffer`'s i.e. Netty/Java NIO however in our case we are dealing with `ByteString` where we don't have direct `ByteBuffer`'s (`ByteString.asByteBuffer` is not a direct byte buffer) so we always need up having to copy into an array that then gets pushed into a direct `ByteBuffer` before going into the compress function. Using `ZstdCompressCtx` we can just directly feed/get the byte arrays.

This seems like a convincing argument

> * Now lets get onto buffering
>       
>   * When it comes to compression, whether we want to buffer or not actually depends on the size of the `ByteString` element thats incoming onto the stream.
>   * If the element happens to be really "big", its actually ideal to compress that element as is without feeding it into any buffer as the larger the element, the better the compression ratio is. In this case we can just use `input.toByteArrayUnsafe` and feed it directly into `ZstdCompressCtx.compress`, even better we don't need to worry about any copying in happy path case for small `ByteString`.

This sounds like the resulting bytes will be different depending on how the input was split, which could be nondeterministic (e.g. depending on network conditions). I guess it's fine since this will most likely be used for 'ephemeral' streams, we can 'blame' the nondeterminism on the component that does the splitting, and someone who cares about determinism could insert a component to convert the input into fixed-size chunks.

>   * If the input `ByteStrings` is "small", then we can keep on asking for elements to buffer and once it hits a constant size limit we can then feed it into `ZstdCompressCtx.compress`

This is a trade-off between latency and compression performance: in a naive implementation, if we for instance 'keep asking elements to buffer' on the boundary of 2 elements of a gRPC stream, it would delay sending+processing the first element while waiting for the next element to show up - which could take a long time on 'event streams', for example. It's possible to avoid this problem: we could 'keep asking for elements to buffer' only when there's currently no demand from downstream, but compress and send 'everything we have' as soon as demand is signaled. That might be significantly more complex, though.

> * When it comes to decompression, the logic can remain as is. We don't need to use a `DirectByteBufferPool` because we only create 2 buffers per stream instantiation (**not** per element of stream) and those `ByteBuffer`'s are reused for each incoming element, albeit it should be updated so that those `ByteBuffer`'s get cleaned up with `if (ByteBufferCleaner.isSupported) ByteBufferCleaner.clean(byteBuffer)`. Using `ZstdDecompressCtx` isn't going to gain us much net benefit as we need to use the native streaming capabilities of zstd-jni to accept incoming data until we can get to a point where we can decompress.

Sounds good to me.

mdedetrich:
> This sounds like the resulting bytes will be different depending on how the input was split, which could be nondeterministic (e.g. depending on network conditions). I guess it's fine since this will most likely be used for 'ephemeral' streams, we can 'blame' the nondeterminism on the component that does the splitting, and someone who cares about determinism could insert a component to convert the input into fixed-size chunks.

Yes this is precisely my thinking as well. Hypothetically speaking someone could just feed 1gbyte sized chunks as individual stream elements but if you are doing that then your already in a troublesome spot, regardless of compression or not.

The more I think about it, this limit should be configurable (and maybe even optional) with sensible defaults, wdyt?

> This is a trade-off between latency and compression performance: in a naive implementation, if we for instance 'keep asking elements to buffer' on the boundary of 2 elements of a gRPC stream, it would delay sending+processing the first element while waiting for the next element to show up - which could take a long time on 'event streams', for example. It's possible to avoid this problem: we could 'keep asking for elements to buffer' only when there's currently no demand from downstream, but compress and send 'everything we have' as soon as demand is signaled. That might be significantly more complex, though.

Good point, I did miss that. I am currently doing an initial implementation that follows the logic I described, but it can be improved on later to bypass the accumulation of data if downstream is demanding data right now. I assume that you would use `hasBeenPulled(port)` to figure out if there is downstream demand?

raboof:
> > This is a trade-off between latency and compression performance: in a naive implementation, if we for instance 'keep asking elements to buffer' on the boundary of 2 elements of a gRPC stream, it would delay sending+processing the first element while waiting for the next element to show up - which could take a long time on 'event streams', for example. It's possible to avoid this problem: we could 'keep asking for elements to buffer' only when there's currently no demand from downstream, but compress and send 'everything we have' as soon as demand is signaled. That might be significantly more complex, though.
> 
> Good point, I did miss that. I am currently doing an initial implementation that follows the logic I described, but it can be improved on later to bypass the accumulation of data if downstream is demanding data right now. I assume that you would use `hasBeenPulled(port)` to figure out if there is downstream demand?

From https://pekko.apache.org/docs/pekko/current/stream/stream-customize.html it looks like `hasBeenPulled` is for input ports, and `isAvailable` would be the thing to check. You'd also need to make sure that your flow signals demand even when downstream doesn't (until the buffer is 'full').

jrudolph:
> `ZstdDirectBufferCompressingStreamNoFinalizer` and related classes were designed to work with sources that provide direct `ByteBuffer`'s i.e. Netty/Java NIO however in our case we are dealing with `ByteString` where we don't have direct `ByteBuffer`'s (`ByteString.asByteBuffer` is not a direct byte buffer) so we always need up having to copy into an array that then gets pushed into a direct `ByteBuffer` before going into the compress function. Using `ZstdCompressCtx` we can just directly feed/get the byte arrays.

You can ignore the copy cost, compression is not cheap anyway. Also, the copying is likely happening anyways as you otherwise have to globally lock the heap array and interfere with GC...

> Now lets get onto buffering

It's a complex topic how and when a (streaming) compressor should outputs compressed data. Roughly speaking, for best compression, you leave the decision completely to the compressor. For more interactive usage you want to send any data as early as possible (kind of a nagle situation but for compression). The compressor cannot know what you want and we cannot know what the user wants. Also note, that streaming does not equal interactive usage, you might want to stream data to run with less memory, so for the (native) compressor API you don't want to make the size of the chunks part of the decision of when to flush.

For our streaming API, the problem is that we cannot easily provide the signal when to flush using a raw ByteString based stream. For that reason, in the deflate compressor we allow to set a flush mode and flush after each ByteString. That way the user still has full control over when the flushing does happen (by buffering data in the stream if required). We should not add any inner logic on top of that. Let's not overthink the interface between our compression component and the native component. In general, if we can provide all of incoming data in one go to the compressor that's ok because it's already on the heap anyway and owned by us. But we should never hold anything back.

> When it comes to decompression, the logic can reman as is. We don't need to use a `DirectByteBufferPool` because we only create 2 buffers per stream instantiation (**not** per element of stream) and those `ByteBuffer`'s are reused for each incoming element, albeit it should be updated so that those `ByteBuffer`'s get cleaned up with `if (ByteBufferCleaner.isSupported) ByteBufferCleaner.clean(byteBuffer)`. Using `ZstdDecompressCtx` isn't going to gain us much net benefit as we need to use the native streaming capabilities of zstd-jni to accept incoming data until we can get to a point where we can decompress.

In any case, we should avoid cycling through direct buffers. E.g. using one for the lifetime of the whole stream could be acceptable.

Btw. zstd is somewhat different from deflate in that regard, you can only ever decompress full blocks while deflate can at least in theory start streaming while still receiving a block (see https://github.com/facebook/zstd/issues/1359 and https://fastcompression.blogspot.com/2016/04/working-with-streaming.html, the tldr; is that a compressed block is max 128kb anyway and that usually dwarfs any LZ77 context buffers you might have to keep around anyways).

mdedetrich:
> You can ignore the copy cost, compression is not cheap anyway. Also, the copying is likely happening anyways as you otherwise have to globally lock the heap array and interfere with GC...

Yes definitely true, my main point here is that even though the copying is already happening behind the scenes and in context of compression the copying is cheap, doing this

```scala
ByteString.fromArrayUnsafe(zstdCompressCtx.compress(input.toArrayUnsafe()))
```

is far simpler and easier to maintain than

```scala
val inputBB = ByteBuffer.allocateDirect(input.size)
inputBB.put(input.toArray)
inputBB.flip()
compressingStream.compress(inputBB)
compressingStream.flush()
targetBuffer.flip()

val arr = new Array[Byte](targetBuffer.limit())
targetBuffer.get(arr)
targetBuffer.clear()
ByteString.fromArrayUnsafe(arr)
```

And that version isn't even the ideal one since we are allocating a direct `ByteBuffer` on every input element (which we shouldn't be doing since direct `ByteBuffer` allocation is extremely expensive). Rather if we want to work with direct `ByteBuffer`, we should get a `ByteBuffer` from a `DirectByteBufferPool` which will complicate the logic even further since `DirectByteBufferPool` only supports direct `ByteBuffer`'s of a static size (and in our case we have a dynamic `input.size` length)

> It's a complex topic how and when a (streaming) compressor should outputs compressed data. Roughly speaking, for best compression, you leave the decision completely to the compressor. For more interactive usage you want to send any data as early as possible (kind of a nagle situation but for compression). The compressor cannot know what you want and we cannot know what the user wants. Also note, that streaming does not equal interactive usage, you might want to stream data to run with less memory, so for the (native) compressor API you don't want to make the size of the chunks part of the decision of when to flush.

So my mentality behind is https://github.com/apache/pekko/pull/2409#issuecomment-3462846239 is that with all else being equal, its always better to compress larger chunks of data than small. Now the critical part here is all else being equal, what @raboof said earlier is entirely correct in that even if we have stream elements as small as 1 byte (or even less), for some kind of applications you want to compress and send as fast as possible due to latency reasons.

> For our streaming API, the problem is that we cannot easily provide the signal when to flush using a raw ByteString based stream. For that reason, in the deflate compressor we allow to set a flush mode and flush after each ByteString.

Unless I am missing something I don't think the flush mode is configurable as of now, in fact you even left a comment and an issue saying it should be configurable (see https://github.com/apache/pekko/blob/085bb4275a70d8db4c2c53946712ee6b54e74688/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Compression.scala#L29-L34)

Also on the same point, are you saying that @raboof 's suggestion i.e.

> It's possible to avoid this problem: we could 'keep asking for elements to buffer' only when there's currently no demand from downstream, but compress and send 'everything we have' as soon as demand is signaled. That might be significantly more complex, though.

Is not possible or is it just overcomplicated/not necessary? To me, with the assumption that we can reliably figure out if downstream is demanding data this should be able to get the best of both worlds, no? i.e. if downstream is asking for data, we just compress the input element as is and send it downstream  but if downstream is not asking for data then we can buffer elements (up to a sane limit) and we send the compressed data as soon as that limit is hit and/or downstream asks for demand again.

Also if this suggested logic ends up being implemented, it would be configurable so users can disable it which would just compress each element as it is (and this wouldn't just be for latency reasons, as you pointed out if a user wants to reduce memory/CPU pressure as well).

> That way the user still has full control over when the flushing does happen (by buffering data in the stream if required). We should not add any inner logic on top of that. Let's not overthink the interface between our compression component and the native component. In general, if we can provide all of incoming data in one go to the compressor that's ok because it's already on the heap anyway and owned by us. But we should never hold anything back.

If I am understanding what you are saying, then essentially you are suggesting that the implementation of the compressor would be a trivial `SimpleLinearGraphStage[ByteString]` that on each incoming processing element just does `ByteString.fromArrayUnsafe(zstdCompressCtx.compress(input.toArrayUnsafe()))` and we then call it a day?

Don't have to worry about flushing at all, or buffering or anything like that and if a user wants buffering then they can just use the standard streaming operators to implement it on their own.

> Btw. zstd is somewhat different from deflate in that regard, you can only ever decompress full blocks while deflate can at least in theory start streaming while still receiving a block (see [facebook/zstd#1359](https://github.com/facebook/zstd/issues/1359) and https://fastcompression.blogspot.com/2016/04/working-with-streaming.html, the tldr; is that a compressed block is max 128kb anyway and that usually dwarfs any LZ77 context buffers you might have to keep around anyways).

Thanks for reminding, Zstd actually provides a constant called `Zstd.blockSizeMax` which I believe points to the `128kb` that you are talking about and that should be the default value for `maxBytesPerChunk`.

mdedetrich:
I have just updated the PR with two small minor changes, one is setting the default of `maxBytesPerChunk` for the zstd decompressor to be `Zstd.blockSizeMax()` which happens to be 128kb (thanks @jrudolph !) and the other is cleaning up the direct `ByteBuffer` in the `postStop` hook.

I am holding on my current work of buffering data up to a limit until I get feedback from my latest comment at https://github.com/apache/pekko/pull/2409#issuecomment-3469014680 as it sounds like I might be over-engineering the problem.

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
mdedetrich, line 1: I had to make both this class and specific methods public because currently its package private to `org.apache.pekko.io` and the pekko stream implementations are not in that package.

I marked the class to be `InternalStableApi`, let me know if should be something else.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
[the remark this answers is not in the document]
mdedetrich: > Should we mark it `@ApiMayChange` too?

This is internal api, `@ApiMayChange` is only for public API's

He-Pin:
@mdedetrich, in Netty, we have `ZstdOptions` too. I think we can take a look at Netty's one for some inspiration before we merge this.

I think we should have this in 1.3.0 release.

mdedetrich:
> @mdedetrich, in Netty, we have `ZstdOptions` too. I think we can take a look at Netty's one for some inspiration before we merge this.

So I have been waiting for a response to https://github.com/apache/pekko/pull/2409#issuecomment-3469014680 but haven't gotten one yet.

In thinking about it, I am increasingly getting to the view of implementing the more complicated but smart solution, i.e. buffering data up until upper limit or if downstream signals demand. Reason being I think this is a fairly novel approach that takes advantage of pekko-streams bi-directional nature and it can even makes good advertisement for pekko streams (can make a blog post about it)

> I think we should have this in 1.3.0 release.

This would have to be a separate artifact and that would open a whole can of worms, not sure I could even make it for 1.3.0

He-Pin:
zstd is indeed quite complex. In our work, it actually includes a complete platform and toolchain for dictionary training, end-to-end deployment, version synchronization, etc.

Therefore, could we consider releasing a separate zstd toolkit, such as pekko-zstd or pekko-http-zstd, etc.? This would allow for a more complete design and implementation, and also enable independent evolution.

Without a dictionary, zstd is simply a slightly faster gzip implementation.

mdedetrich:
> zstd is indeed quite complex. In our work, it actually includes a complete platform and toolchain for dictionary training, end-to-end deployment, version synchronization, etc.

Actually the zstd-api is incredibly simple, especially if we decide to the simple solution of doing `ByteString.fromArrayUnsafe(zstdCompressCtx.compress(input.toArrayUnsafe()))`. In this case the compressor would just be a few lines of code and its far simpler than what is done in [gzip/deflate](https://github.com/apache/pekko/blob/b0fdac259bd57fdd481483f3fe9a7aec6e1ff38a/stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/DeflateCompressor.scala) which has to deal with arbitrary behavior in the JVM implementation.

The complexity that I am describing comes from doing something very smart that is only possible with pekko-streams, i.e. taking into account downstream demand in a stream when it comes to buffering, afaik no other implementation does this.

On the other hand the complexity you describe is from your specific usecase where you work. Training dictionaries would be entirely out of scope for this, but of course you can specify a dictionary in the same way you can specify a compression level and this is also a single of code.

> Therefore, could we consider releasing a separate zstd toolkit, such as pekko-zstd or pekko-http-zstd, etc.? This would allow for a more complete design and implementation, and also enable independent evolution.

Assuming the initial implementation is correct I don't see how this could evolve aside from version bumps in zstd-jni (at least with assumption that dictionary training is out of scope)

> Without a dictionary, zstd is simply a slightly faster gzip implementation.

Its not just slightly faster, it has significantly better compression ratios. And slightly is downplaying it a bit, it can be many factors faster (ofc it depends on the level and where the bottleneck is).

He-Pin:
Our API gateway currently uses zstd, so I know the benefits are significant, both in terms of performance and compression ratio. Although Netty's codec directly depends on zstd-jni, I don't know if Pekko-stream users will find this troublesome; personally, I think it's fine.

By standard standards, I think this PR is already good enough, but I'm unsure how much complex effort you're willing to invest. For example, in our work, some of our APIs have separate configurations, dictionary names, versions, etc., and they negotiate with the gateway during application startup. But here, I think we're probably good enough if you're not considering further complexity and extreme optimization.

mdedetrich:
Currently doing work on making our own separate `ZstdDictionary` class so we don't depend on the zstd-jni one but its blocked by https://github.com/apache/pekko/pull/2464

mdedetrich:
I have just updated the PR by adding our own `ZstdDictionary` datastructure so that the public API of the zstd compress/decompress doesn't depend on any internals from zstd-jni.

Aside from making it possible to swap out zstd-jni in the future for any alternative implementations it also makes the API cleaner for both scala and java users.

@He-Pin @pjfanning @raboof Let me know if the design `ZstdDictionary` works (I made it idiomatic for both scala and javadsl) and also that its location (i.e. `org.apache.pekko.stream.javadsl.compression.ZstdDictionary`/`org.apache.pekko.stream.scaladsl.compression.ZstdDictionary`) is appropriate

mdedetrich:
> Although Netty's codec directly depends on zstd-jni, I don't know if Pekko-stream users will find this troublesome; personally, I think it's fine.

Its not just netty, even Apache Commons Compress uses zstd-jni and includes it directly as an artifact (see https://commons.apache.org/proper/commons-compress/ and https://mvnrepository.com/artifact/org.apache.commons/commons-compress/1.28.0).

The only difference is that apache-commons has it as an `optional` maven dependency. We can do the same here, we would just have to document to end users that they have to add the zstd-jni artifact dependency themselves (this line is also just a single change in code).

He-Pin (dismissed):
lgtm

stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.pekko.annotation.InternalApi;
+import org.apache.pekko.stream.impl.io.compression.ZstdDictionaryImpl;
+import org.apache.pekko.util.ByteString;
+import org.apache.pekko.util.OptionalUtil;
+import scala.Option;
+
+import java.util.OptionalInt;
+
+/**
+ * @since 2.0.0
```
He-Pin, line 34: Add java/scala doc

mdedetrich:
I have updated the PR, I have made the various zstd constants (i.e. min level, max level, default block size etc etc) be part of the `Compression` object which makes the Pekko API fully separate from the underlying zstd-jni implementation. 

I have also rebased against main to use the latest configurable `autoFlush` changes add as part of https://github.com/apache/pekko/pull/2467

docs/src/main/paradox/stream/operators/Compression/zstd.md
```diff
@@ -0,0 +1,30 @@
...
+@ref[Compression operators](../index.md#compression-operators)
+
+## Signature
+
+@apidoc[Compression.zstd](stream.*.Compression$) { scala="#zstd:org.apache.pekko.stream.scaladsl.Flow[org.apache.pekko.util.ByteString,org.apache.pekko.util.ByteString,org.apache.pekko.NotUsed]" java="#zstd()" }
+
+## Description
+
+Creates a flow that zstd-compresses a stream of ByteStrings. Note that the compressor
+will SYNC_FLUSH after every @apidoc[util.ByteString] so that it is guaranteed that every @apidoc[util.ByteString]
```
raboof, line 14: This also means that if the way the incoming data is cut up into `ByteString`s is not deterministic, the compressed stream will not be deterministic, right? I think that is acceptable, but would be worth calling out.

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
```diff
@@ -36,7 +37,8 @@
  *
  * <p>See <a href=https://bugs.openjdk.java.net/browse/JDK-4724038>JDK-4724038</a>
  */
-final class ByteBufferCleaner {
+@InternalStableApi
```
raboof, line 40: This annotation is for APIs that need to remain binary compatible because other, independently-released Pekko components depend on them.

As you're using this API within Pekko Core, and we expect all Pekko Core artifact versions to be consistent, the annotation is not needed for this use.

Are you planning to use this API in other Pekko modules? If not then we can remove this annotation here. If so then we should add a note to the scaladoc pointing to where it's (going to be) used.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
raboof: Are you still planning on this?

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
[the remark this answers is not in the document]
raboof: the recoding is now complete, right? This line still seems to be here, should it still be changed to `Unsafe`?

.scala-steward.conf
```diff
@@ -5,6 +5,8 @@ updates.pin = [
   { groupId = "org.scala-lang", artifactId = "scala3-library", version = "3.3." }
   # sbt-assembly 2.3 causes build issues (https://github.com/apache/pekko/pull/1744)
   { groupId = "com.eed3si9n", artifactId = "sbt-assembly", version = "2.2." }
+  # zstd-jni upgrades are fine as long as they are backwards compatible
+  { groupId = "com.github.luben", artifactId = "zstd-jni", version = "1."}
```
raboof: Shouldn't we even allow major version updates? Then we can decide on a case-by-case basis whether the incompatibility that caused them to bump the major version is a reason for us to avoid the update.

raboof (commented):
I have not fully digested the PR but on first read it looks reasonable. Noted a couple of small things that are potential leftovers/nits.

raboof:
Needs a rebase for the merge conflict and the fix for the #2476 failure

stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
```diff
@@ -0,0 +1,205 @@
...
+ */
+public class ZstdDictionary {
+    byte[] dictionary;
```
raboof, line 37: Is this somehow a standardized "zstd thing" or something still tied to this particular implementation? I think that needs to be documented. We could consider keeping the dictionary-related things `@ApiMayChange` to expedite the mergability of this PR while still giving us room to iterate later.

stream/src/main/scala/org/apache/pekko/stream/impl/io/compression/ZstdCompressor.scala
```diff
@@ -0,0 +1,103 @@
...
+
+/** INTERNAL API */
+@InternalApi private[pekko] class ZstdCompressor(
+    compressionLevel: Int =
+      Compression.ZstdDefaultCompressionLevel, dictionary: Option[ZstdDictionaryImpl] = None) extends Compressor {
+
+  private val targetBuffer = ByteBuffer.allocateDirect(65536)
+  private val compressingStream = new ZstdDirectBufferCompressingStreamNoFinalizer(targetBuffer, compressionLevel)
+
+  dictionary.foreach { dict =>
```
raboof, line 35: (nit: I find `match` almost always easier to read compared to `foreach` on `Option`)

stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
mdedetrich: > Is this somehow a standardized "zstd thing" or something still tied to this particular implementation

Yes its entirely standerdized as part of the official C API (see https://github.com/facebook/zstd?tab=readme-ov-file#dictionary-compression-how-to and https://facebook.github.io/zstd/zstd_manual.html#Chapter10.

>  think that needs to be documented. We could consider keeping the dictionary-related things `@ApiMayChange` to expedite the mergability of this PR while still giving us room to iterate later.

So the core premise of `ZstdDictionary` is for it to be a stable API that is decoupled from zstd-jni (incase other future better implementations come along). I can still put in `@ApiMayChange` but it is in contention with that core point.

.scala-steward.conf
mdedetrich: > Shouldn't we even allow major version updates? Then we can decide on a case-by-case basis whether the incompatibility that caused them to bump the major version is a reason for us to avoid the update.

Well on the premise of it following semver doing major version updates could be problematic because it can imply behaviour and/or API changes?

mdedetrich:
> I have not fully digested the PR but on first read it looks reasonable. Noted a couple of small things that are potential leftovers/nits.

So I still want to get this into 2.0.0-M1 but I haven't had time with the buffer implementation I was planning to do. For now I am thinking of doing the simple dumb solution so that we at least have a implementation with a stable API which unlocks zstd-compression for http responses in pekko-http and I can always do a better version for a future milestone release of 2.0.0.

stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
raboof: Cool, would be good to add that reference.

.scala-steward.conf
raboof: It _could_, but it's not _necessarily_ problematic. We specifically reserve the right to do major version updates of dependencies when we deem them unproblematic (https://pekko.apache.org/docs/pekko/current/common/binary-compatibility-rules.html#binary-compatibility-of-dependencies). We let scala-steward open PRs for such upgrades, and then when that happens we can decide whether to follow upstream or restrict ourselves to the previous major version.

He-Pin:
@mdedetrich I think you can mark it as experimental and then polish it in a later release.

stream/src/main/scala/org/apache/pekko/stream/javadsl/compression/ZstdDictionary.java
mdedetrich: Done

.scala-steward.conf
mdedetrich: Okay Ill remove the entry then

raboof:
@mdedetrich are you still planning further changes? I see you pushed some changes yesterday but AFAICT there's still some open review comments

mdedetrich:
> @mdedetrich are you still planning further changes? I see you pushed some changes yesterday but AFAICT there's still some open review comments

Yes, I have a couple of tests to fix to pass and then I will push changes

docs/src/main/paradox/stream/operators/Compression/zstd.md
mdedetrich: Actually its the opposite, since a sync flush happens on every stream element it means that there is a direct one to one relationship of every element having the exact same data compressed (there is actually a test that validates this)

raboof: But that is not the case "if the way the incoming data is cut up into `ByteString`s is not deterministic", right? E.g when the data is coming from the network?

(this is not a fault of the stream compression, but a truth that would be good to highlight)

mdedetrich: > But that is not the case "if the way the incoming data is cut up into ByteStrings is not deterministic", right? E.g when the data is coming from the network?

Yes true, I believe that comment is strictly talking about the source happens right before going into compression and not the entire "stack". All the comment is trying to convey is that if `autoFlush` is true, any buffers will get processed and cleared for each incoming element, so if you have "abc" as an incoming element, you will always get the compressed "abc" as a result, and not lets say "a" and then "bc" as a further element or some variation.

The network can always split up data as it wants, but thats an orthogonal concern

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
mdedetrich: > Are you planning to use this API in other Pekko modules? If not then we can remove this annotation here. If so then we should add a note to the scaladoc pointing to where it's (going to be) used.

As you allude to, the reason why I added `@InternalStableApi` is that I can imagine cases where other Pekko libraries (i.e. pekko-http or even pekko-connectors) might want to use `ByteBufferCleaner` and obviously this would need to be available in the Pekko 2.0.0 release to fullfil this.

Also this API is quite stable, hence why I added `@InternalStableApi`, but yes in general it was a forward thinking rationality.

docs/src/main/paradox/stream/operators/Compression/zstd.md
raboof: That is correct. I wasn't trying to say the original comment is wrong (it isn't), I meant to say that it would be valuable to highlight that the way the incoming data is split up influences the result of the compression.

mdedetrich:
@raboof So I tried to get my fancy buffering implementation done, but I am coming across a very weird issue/problem where even if I change the current implementation to something like this

```scala
override def compressAndFlush(input: ByteString): ByteString =
  ByteString.fromArrayUnsafe(Zstd.compress(input.toArrayUnsafe))
```

The tests start failing because `onUpstreamFinish` never ends up being executed. 

So I think we should opt for the simple option for now, which is going to be the current implementation but avoid creating a direct byte buffer on each incoming element (instead using a pool), doing the changes now.

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
raboof: OK. I'm not super fond of committing ourselves to `@InternalStableApi` "just in case" without a specific plan on where to use it. Also we should seriously consider using the implementation from `commons-io` - definitely in satellite projects, but also here. The use of `sun.misc.Unsafe` is icky as well.

I'm OK with this not blocking the merge though.

docs/src/main/paradox/stream/operators/Compression/zstd.md
mdedetrich: Got it, ill better clarify this!

actor/src/main/java/org/apache/pekko/io/ByteBufferCleaner.java
pjfanning: * this code doesn't fail if Unsafe is not accessible - it accepts that GC will eventually clean but it is better for users if this Unsafe code cleans the instance earlier
* it's one of those areas where the Java team keep telling us to stop using Unsafe but in this case, they have not yet given an alternative

mdedetrich:
@raboof @pjfanning @He-Pin  So here is a status update, I actually found a bug in the `ZstdDecompressor` which I am now trying to fix. The bug is quite simple, it happens when [here](https://github.com/apache/pekko/pull/2409/files#diff-36b15bf36eaf7b9bd2c64a650c0d1f602a475383340766c45c0907aedf4f9b6cR50) when the source buffer is larger than the bytebuffer causing a basic overflow. This bug didn't show in tests because individual `ByteString` elements in the tests happen always be smaller than the `maxBytesPerChunk`, but if you set `maxBytesPerChunk` to small value such as 8 then it immediately shows.

To solve this, I wanted to create an ultra simple implementation which would essentially just be `ByteString.fromArrayUnsafe(Zstd.decompress(input.toArrayUnsafe))` but this is not possible because `Zstd.decompress` actually needs to know the original size of the decompressed element (and we have no context of this in a stream).

Due to this I have to shuffle data from the input `ByteString` to that direct bytebuffer while taking into account the fact that the input `ByteString` can be of any size. There is a naive implementation that transfers one byte at a time (which is really slow), so instead the implementation copies as much as it can depending on the free amount of space in the `outputBuffer` but understandably this is a more complicated solution. Note this solution is similar to how zstd-jni test also [does](https://github.com/luben/zstd-jni/blob/9c3386d306086078155f58116a4d905e07239db4/src/test/scala/Zstd.scala#L516-L551). 

The advantage here is that this is the exact same logic that is needed to avoid the `ByteBuffer.allocateDirect(input.size)` in `ZstdCompressor` in the current PR implementation which we need to remove anyways because of how slow `ByteBuffer.allocateDirect(input.size)` is (which is the only change that is needed on the compressor side).

tl;dr Almost at the finish line!

He-Pin:
https://github.com/netty/netty/pull/15861

@mdedetrich There is a related pr in netty too.

mdedetrich:
@raboof @pjfanning Apologies for missing this for the Pekko 2.0.0-M1 release train, I am still working on this but its taking longer as expected. I am actually currently implementing this work as a test in zstd-jni to make my life easier and also with the hope of getting feedback from the author that it is on the right track.

Essentially the problem is that there isn't any real prior art for this, all of the current zstd streaming tests in zstd-jni either involve creating a direct `ByteBuffer` with the same size of the input data (which we obviously don't want to do in pekko-streams as these allocations are extremely expensive), the input source is already a direct `ByteBuffer` (i.e. java nio) or in the case of decompression they use methods which require the size of the original uncompressed input (which we cannot do in pekko-streams) i.e. the tests are still streaming decompression a file but since they are dealing with a file they easily have access to the original file size.

I am slightly annoyed that I didn't make the 2.0.0-M1 release, but as long as we are happy with using a nightly in pekko-http to implement the http request/response zstd work in the end it shouldn't be that problematic. You could also argue its better this way, as pekko 2.0.0 milestone was primarily about updating all of the core libraries/jvm versions + removing deprecated code and bookkeeping and we can have 2.0.0-M2 be about major features.

mdedetrich:
Relevant issue https://github.com/luben/zstd-jni/issues/324 and made a comment there.

mdedetrich:
So upstream has commented at https://github.com/luben/zstd-jni/issues/324#issuecomment-3614203657, it may make more sense to add this functionality (specifically `ZstdCompressCtx.compressByteBufferStream()` and `ZstdDecompressCtx.decompressByteBufferStream()`) into zstd-jni proper as it would help others and in turn make the pekko-streams implementation far more trivial.

He-Pin:
@mdedetrich what's the status of this pr?

He-Pin (commented):
Valuable addition -- zstd is increasingly the compression algorithm of choice for performance-sensitive workloads.

**Concerns:**
1. PR is marked as incomplete ('documentation needs to be added along with MiMa exclusions'). The docs appear to be in place now but MiMa exclusions need verification.
2. The dictionary API ties directly to `com.github.luben.zstd.ZstdDictCompress` -- as noted in the PR description, this creates a hard dependency on zstd-jni. Consider the abstraction layer suggestion (Pekko wrapper around the zstd-jni dict classes) before merging.
3. The PR has been open since October 2025 with extensive discussion. It needs a final decision on the API design and completion of the remaining items.

Otherwise the implementation looks solid. The test coverage is good and the choice of zstd-jni is appropriate.

mdedetrich:
> @mdedetrich what's the status of this pr?

It requires upstream changes to zstd in order for the PR to make sense

He-Pin:
@mdedetrich Thanks for the update,i think you can prepare a pr with ai help to zstd repo.

PR #2414, opened 2025-11-01
feat: Add mapOption operator

He-Pin:
Motivation:
Add an `mapOptional` operator as RxJava for java/scala.

refs: https://github.com/apache/pekko/issues/2403

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/mapOption.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/mapOption.md
@@ -1,0 +1,31 @@
+# mapOption
+
+Transform each element in the stream by calling a mapping function with it and emits the contained item if present.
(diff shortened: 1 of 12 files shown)


docs/src/main/paradox/stream/operators/Source-or-Flow/mapOption.md
```diff
@@ -0,0 +1,31 @@
+# map
```
pjfanning: map option ?

He-Pin: will update ,thanks

He-Pin:
I have updated this pr, thanks @pjfanning I think it will works well for Java developer

pjfanning (approved):
lgtm

PR #2422, opened 2025-11-02
feat: Add Source#elements

He-Pin:
Motivation:
refs: https://github.com/apache/pekko/issues/2416

It's named `just` in Flux and `ambArray` in Flowable.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
@@ -248,6 +248,23 @@ object Source {
     new Source(scaladsl.Source.single(element))
 
   /**
+   * Create a `Source` from the given elements.
+   *
+   * @since 1.3.0
+   */
+  @varargs
+  @SafeVarargs
+  def elements[T](elements: T*): javadsl.Source[T, NotUsed] = {
+    if (elements.isEmpty) {
+      empty()
+    } else if (elements.length == 1) {
+      single(elements.head)
+    } else {
+      new Source(scaladsl.Source(elements))
+    }
+  }
+
+  /**
    * Create a `Source` that will continually emit the given element.
    */
   def repeat[T](element: T): Source[T, NotUsed] =
--- a/docs/src/main/paradox/stream/operators/Source/elements.md
+++ b/docs/src/main/paradox/stream/operators/Source/elements.md
@@ -1,0 +1,26 @@
+# Source.elements
+
+Create a `Source` from the given elements.
+
+@ref[Source operators](../index.md#source-operators)
+
+## Signature
+
+@apidoc[Source.elements](Source$) { scala="#elements[T](elements:T*):org.apache.pekko.stream.scaladsl.Source[T,org.apache.pekko.NotUsed]" java="#elements[T](T)" }
+
+
+## Description
+
+Create a `Source` from the given elements.
+
+## Examples
+
+## Reactive Streams semantics
+
+@@@div { .callout }
+
+**emits** the elements one by one
+
+**completes** when the last element has been emitted
+
+@@@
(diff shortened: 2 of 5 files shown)


He-Pin:
IIRC, it was there, but later been removed because of the bug in Scala 2.10.

stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
```diff
@@ -247,6 +247,23 @@ object Source {
...
+   */
+  @varargs
+  @SafeVarargs
+  def elements[T](elements: T*): javadsl.Source[T, NotUsed] = {
+    if (elements.isEmpty) {
+      empty()
+    } else if (elements.length == 1) {
+      single(elements.head)
+    } else {
+      new Source(scaladsl.Source(elements))
```
He-Pin, line 263: This line needs to be adjusted when merging into 1.3.x

mdedetrich:
> IIRC, it was there, but later been removed because of the bug in Scala 2.10.

Can't we just overload `Source.apply` instead of creating yet another factory method named `elements`?

He-Pin:
That will not work, because `Source(Array(1))` will not compile, because two methods match.

mdedetrich:
> That will not work, because `Source(Array(1))` will not compile, because two methods match.

I just added

```scala
@varargs
@SafeVarargs
def apply[T](elements: T*): Source[T, NotUsed] = {
  if (elements.isEmpty) {
    empty[T]
  } else if (elements.length == 1) {
    single(elements.head)
  } else {
    Source(elements)
  }
}
```

Locally to my `Source` in `scaladsl` and wrote this test
  
```scala
"product elements with varargs" in {
  val p = Source(1, 2, 3).runWith(Sink.asPublisher(false))
  val c = TestSubscriber.manualProbe[Int]()
  p.subscribe(c)
  val sub = c.expectSubscription()
  sub.request(1)
  c.expectNext(1)
  sub.request(2)
  c.expectNext(2)
  sub.request(3)
  c.expectNext(3)
  c.expectComplete()
}
```

And it compiled fine and I verified that `Source(1, 2, 3)` is picking the `def apply[T](elements: T*)` method

He-Pin:
Try Source(Array(1))

mdedetrich:
> Try Source(Array(1))

```scala
"product elements with Array" in {
  val p = Source(Array(1, 2, 3)).runWith(Sink.asPublisher(false))
  val c = TestSubscriber.manualProbe[Int]()
  p.subscribe(c)
  val sub = c.expectSubscription()
  sub.request(3)
  c.expectNext(1)
  c.expectNext(2)
  c.expectNext(3)
  c.expectComplete()
}

"product elements with single Array element" in {
  val p = Source(Array(1)).runWith(Sink.asPublisher(false))
  val c = TestSubscriber.manualProbe[Int]()
  p.subscribe(c)
  val sub = c.expectSubscription()
  sub.request(1)
  c.expectNext(1)
  c.expectComplete()
}

"product elements with varargs" in {
  val p = Source(1, 2, 3).runWith(Sink.asPublisher(false))
  val c = TestSubscriber.manualProbe[Int]()
  p.subscribe(c)
  val sub = c.expectSubscription()
  sub.request(3)
  c.expectNext(1)
  c.expectNext(2)
  c.expectNext(3)
  c.expectComplete()
}
```

Works fine, just tried it locally

He-Pin:
Then you can not let the source emit Array(1) with the apply method

He-Pin:
@mdedetrich The problem is that when there is only one element, `Source.apply(Array(1))` and `Source.apply(1)` have the same meaning. However, `Source.apply(Array(1), Array(1))` means two `Array(1)` objects, while `Source.apply(Array(1))` does not represent a single `Array`, which is simply incomprehensible. I think we should follow the original design.

He-Pin:
https://github.com/akka/akka-core/issues/17923

@mdedetrich You can see, it was this name too

mdedetrich:
> @mdedetrich The problem is that when there is only one element, `Source.apply(Array(1))` and `Source.apply(1)` have the same meaning. However, `Source.apply(Array(1), Array(1))` means two `Array(1)` objects, while `Source.apply(Array(1))` does not represent a single `Array`, which is simply incomprehensible. I think we should follow the original design.

For starters, doing `Source(Array(1),Array(1))` seems a bit off because wanting to send entire mutable arrays is almost never what the user wants, rather they almost always want the elements in those arrays. I agree it's not consistent but its only for one definition of consistent. We can care about consistency of the precise args being sent in (in which case you are right) but there is also consistency regarding what you want in an `Source`, and 99% of the time if someone is constructing a Source from a collection what they want to do is to have all of the elements in that collection to be individual emitted elements in that `Source`

The real question then though, is what are we solving if we have a `Source.elements` method then? The only reason to have these kinds of methods is to save some keystrokes, and if you have `Source.elements(1,2,3)` you can just do `Source(List(1,2,3))` instead.

He-Pin:
Let's just skip it for now.

PR #2425, opened 2025-11-02
chore: sortImports

He-Pin:
Motivation:

Sort some imports

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -19,14 +19,14 @@ import java.nio.{ ByteBuffer, ByteOrder }
 import java.nio.charset.{ Charset, StandardCharsets }
 import java.util.Base64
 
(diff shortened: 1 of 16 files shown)


samueleresca (approved):

mdedetrich:
Did we make a decision on this? I don't think it makes sense to sort imports like this manually until we get the tooling to work properly. Also commits need to be setup in a certain way so that we can add them to gitignore otherwise we will pollute the git history.

He-Pin:
will submit to ignore the commit in blame log

samueleresca:
> Did we make a decision on this? I don't think it makes sense to sort imports like this manually until we get the tooling to work properly.

This is the result of running `sbt sortImports` with the rules defined in the repo

mdedetrich:
> > Did we make a decision on this? I don't think it makes sense to sort imports like this manually until we get the tooling to work properly.
> 
> This is the result of running `sbt sortImports` with the rules defined in the repo

I know, but there is already a discussion on this and the rules have some limitations (see https://github.com/apache/pekko/discussions/2228)

PR #2427, opened 2025-11-02
feat: Add fromArray to Source scaladsl (#2424)

He-Pin:
(cherry picked from commit 2c2daa88a10a4858a77311928cd405690a3fd50b)

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source/fromArray.md
+++ b/docs/src/main/paradox/stream/operators/Source/fromArray.md
@@ -6,7 +6,7 @@ Stream the values of an `array`.
 
 ## Signature
 
-@apidoc[Source.from](Source$) { java="#fromArray(java.lang.Object[])" }
+@apidoc[Source.fromArray](Source$) { scala="#apply[T](array:scala.Array[T]):org.apache.pekko.stream.scaladsl.Source[T,org.apache.pekko.NotUsed]" java="#fromArray(java.lang.Object[])" }
 
 ## Description
 
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Source.scala
@@ -36,13 +36,14 @@ import pekko.japi.function.Creator
 import pekko.stream._
 import pekko.stream.impl.{ LinearTraversalBuilder, UnfoldAsyncJava, UnfoldJava }
 import pekko.stream.impl.Stages.DefaultAttributes
-import pekko.stream.impl.fusing.{ ArraySource, StatefulMapConcat, ZipWithIndexJava }
+import pekko.stream.impl.fusing.{ StatefulMapConcat, ZipWithIndexJava }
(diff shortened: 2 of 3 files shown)


mdedetrich (changes requested):
I am quickly blocking this for now, but only because I think we can just use an overloaded  `Source.apply` instead of having to create yet another method name, prototyping this in PR

He-Pin:
@mdedetrich It's not that explicit with `apply`, you're not sure if the result is a `Source.single(array)`

mdedetrich:
> @mdedetrich It's not that explicit with `apply`, you're not sure if the result is a `Source.single(array)`

Technically true, but at the same time you never want a source of containing array as that doesn't make sense (in fact you can argue if you want a source that contains a mutable array you should be explicit about it and do `Source.single(array)`).

99% perfect on the time, if a user is making a source from an array that want to make a source from all of the elements in the array

He-Pin:
I think you are right but I actually followed the `Flowable` /`Flux`, where both are named `fromArray`, or we can ignore this, because we can treat .

We can already do that without this method but with an error:
```
 method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated (since 2.13.0): implicit conversions from Array to immutable.IndexedSeq are implemented by copying; use `toIndexedSeq` explicitly if you want to copy, or use the more efficient non-copying ArraySeq.unsafeWrapArray
[error] Applicable -Wconf / @nowarn filters for this fatal warning: msg=<part of the message>, cat=deprecation, site=org.apache.pekko.stream.scaladsl.FlowOnErrorCompleteSpec, origin=scala.LowPriorityImplicits2.copyArrayToImmutableIndexedSeq, version=2.13.0
```

mdedetrich:
> I think you are right but I actually followed the `Flowable` /`Flux`, where both are named `fromArray`, or we can ignore this, because we can treat .
> 
> We can already do that without this method but with an error:
> 
> ```
>  method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated (since 2.13.0): implicit conversions from Array to immutable.IndexedSeq are implemented by copying; use `toIndexedSeq` explicitly if you want to copy, or use the more efficient non-copying ArraySeq.unsafeWrapArray
> [error] Applicable -Wconf / @nowarn filters for this fatal warning: msg=<part of the message>, cat=deprecation, site=org.apache.pekko.stream.scaladsl.FlowOnErrorCompleteSpec, origin=scala.LowPriorityImplicits2.copyArrayToImmutableIndexedSeq, version=2.13.0
> ```

That might be true for `javadsl` but in Scala its idiomatic to have a single `def apply` method with overloaded alternatives. Made a PR, see https://github.com/apache/pekko/pull/2428

mdedetrich:
@He-Pin Can you copy over test from https://github.com/apache/pekko/pull/2428 so that we verify that it compiles fine for all scala targets? Then I can approve (and I will close that PR).

He-Pin:
@mdedetrich feel free to update this pr:) it's 6am here.

PR #2428, opened 2025-11-02
Change Source.fromArray to Source.apply

mdedetrich:
Its not idiomatic in Scala to have multiple different named methods from constructors of a data type, instead we should use `def apply` with overloaded alternatives.

I also wrote a test to make sure that everything compiles fine and also behaves as expected.

@He-Pin This is what I mean


He-Pin (approved):
lgtm, I just prepared one

mdedetrich:
Alrighty, I will merge this PR when all tests pass and then you can cherry pick both to the 1.3.x series

He-Pin:
@mdedetrich This is need for 1.3.x too

pjfanning:
@He-Pin if there is a delay, you can create the 1.3.x PR and I will approve it.

PR #2430, opened 2025-11-05
change propagate-harmless-quarantine-events default to off

### LINKED ISSUE #2141: change pekko.remote.artery.propagate-harmless-quarantine-events config default to off
* see #1555
* hopefully, people will set it to off explicity in 1.2.0+ and will get back to us if there are issues

pjfanning:
* see #1555
* fixes #2141


He-Pin:
we use it in production now

PR #2434, opened 2025-11-06
Change default persistence plugin-dispatcher

pjfanning:
https://github.com/akka/akka-core/pull/31598 is now available under Apache License, Version 2.0 because Akka 2.7.0 was released over 3 years ago.

breaks AsyncWriteJournal tests


He-Pin:
We will need to release a newer version of the persistent plugins, too, after this update.

He-Pin (approved):
lgtm, we need add the migration guide too.

pjfanning:
blocking this from being merged - we need to work what to do about #2026 first

pjfanning:
I intend to come back to this but I want to get more info and ensure that people know about the side effects before attempting to get this merged.

PR #2439, opened 2025-11-07
chore: Deprecate stream testkit's probe methods

He-Pin:
Motivation:
Port https://github.com/akka/akka-core/pull/31596


He-Pin:
Need backport to 1.3.0

stream-testkit/src/main/scala/org/apache/pekko/stream/testkit/javadsl/TestSource.scala
```diff
@@ -23,15 +23,15 @@ import pekko.stream.testkit._
 object TestSource {
 
   /**
-   * A Source that materializes to a [[pekko.stream.testkit.TestPublisher.Probe]].
+   * A Source that materializes to a [[pekko.stream.testkit.TestPublisher]].
    */
-  def probe[T](system: ActorSystem): Source[T, TestPublisher.Probe[T]] =
-    new Source(scaladsl.TestSource.probe[T](system))
+  @deprecated("Use `TestSource.create` with ClassicActorSystemProvider instead of ActorSystem", "1.3.0")
```
pjfanning, line 28: may be better to also add Java `@Deprecated` annotation and add `@deprecated Use `TestSource.create` with ClassicActorSystemProvider instead of ActorSystem (since 1.3.0)` to the Javadoc

He-Pin: It's give me an error when using Java's `@ Deprecated`, and that's my origin code, it doesn't compile with our current setup.

He-Pin:
Another `AsyncDnsResolverIntegrationSpec` failed

pjfanning (approved):
lgtm

PR #2460, opened 2025-11-08
move Java 9 Flow support into main Source and Sink classes 

pjfanning:
* Currently users have to use JavaFlowSupport.Source and JavaFlowSupport.Sink instead
* Still some doc changes to make 
* In particular, I need to create asJavaSubscriber.md and asJavaPublisher.md

### DIFF
--- a/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/IterablePublisherViaJavaFlowPublisherTest.scala
+++ b/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/IterablePublisherViaJavaFlowPublisherTest.scala
@@ -17,7 +17,7 @@ import java.util.concurrent.{ Flow => JavaFlow }
 
 import org.apache.pekko
 import pekko.NotUsed
-import pekko.stream.scaladsl.{ JavaFlowSupport, Sink, Source }
+import pekko.stream.scaladsl.{ Sink, Source }
 
 import org.reactivestreams._
 
@@ -25,10 +25,10 @@ class IterablePublisherViaJavaFlowPublisherTest extends PekkoPublisherVerificati
 
   override def createPublisher(elements: Long): Publisher[Int] = {
     val sourceViaJavaFlowPublisher: JavaFlow.Publisher[Int] = Source(iterable(elements))
-      .runWith(JavaFlowSupport.Sink.asPublisher(fanout = false))
+      .runWith(Sink.asJavaPublisher(fanout = false))
 
     val javaFlowPublisherIntoPekkoSource: Source[Int, NotUsed] =
-      JavaFlowSupport.Source.fromPublisher(sourceViaJavaFlowPublisher)
+      Source.fromPublisher(sourceViaJavaFlowPublisher)
 
     javaFlowPublisherIntoPekkoSource
       .runWith(Sink.asPublisher(false)) // back as RS Publisher
--- a/docs/src/main/paradox/stream/operators/Sink/asJavaPublisher.md
+++ b/docs/src/main/paradox/stream/operators/Sink/asJavaPublisher.md
@@ -1,0 +1,39 @@
+# Sink.asJavaPublisher
+
+Integration with Java Flow API, materializes into a `java.util.concurrent.Flow.Publisher`.
+
+@ref[Sink operators](../index.md#sink-operators)
+
+## Signature
+
+@apidoc[Sink.asJavaPublisher](Sink$) { scala="#asJavaPublisher[T](fanout:Boolean):org.apache.pekko.stream.scaladsl.Sink[T,java.util.concurrent.Flow.Publisher[T]]" java="#asJavaPublisher(org.apache.pekko.stream.javadsl.asJavaPublisher)" }
+
+
+
+## Description
+
+This method gives you the capability to publish the data from the `Sink` through a Java Flow [Publisher](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Flow.Publisher.html).
+Generally, in Pekko Streams a `Sink` is considered a subscriber, which consumes the data from source. To integrate with other Reactive Stream implementations `Sink.asJavaPublisher` provides a `Publisher` materialized value when run.
+Now, the data from this publisher can be consumed by subscribing to it. We can control if we allow more than one downstream subscriber from the single running Pekko stream through the `fanout` parameter.
+If you want to support a ReactiveStreams Publisher, there is [Sink.asPublisher](asPublisher.md).
+
+## Example
+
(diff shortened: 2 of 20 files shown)


stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/IterablePublisherViaJavaFlowPublisherTest.scala
```diff
@@ -17,18 +17,18 @@ import java.util.concurrent.{ Flow => JavaFlow }
...
+import pekko.stream.scaladsl.{ Sink, Source }
 
 import org.reactivestreams._
 
 class IterablePublisherViaJavaFlowPublisherTest extends PekkoPublisherVerification[Int] {
 
   override def createPublisher(elements: Long): Publisher[Int] = {
     val sourceViaJavaFlowPublisher: JavaFlow.Publisher[Int] = Source(iterable(elements))
-      .runWith(JavaFlowSupport.Sink.asPublisher(fanout = false))
+      .runWith(Sink.asJavaPublisher(fanout = false))
```
He-Pin, line 28: nice, much clear

raboof:
I wonder if we shouldn't go a step further, and instead of adding a `asJavaPublisher`, actually changing the existing `asPublisher` to use the Java classes instead of the old `org.reactivestreams` ones

pjfanning:
> I wonder if we shouldn't go a step further, and instead of adding a `asJavaPublisher`, actually changing the existing `asPublisher` to use the Java classes instead of the old `org.reactivestreams` ones

There is https://github.com/apache/pekko/issues/1977. I would prefer to treat that as a separate issue. In that issue, @He-Pin points out that some libs still prefer reactivestreams lib (due to inertia or continuing desire to support Java 8). I have little issue with supporting both.

It just seems like we shouldn't treat the Java Flow API as a 2nd class citizen.

raboof:
> It just seems like we shouldn't treat the Java Flow API as a 2nd class citizen.

Much agreed. Perhaps we could make `asPublisher` the Java Flow one and rename the Reactive Streams one to something like `asReactiveStreamsPublisher`?

mdedetrich:
> > It just seems like we shouldn't treat the Java Flow API as a 2nd class citizen.
> 
> Much agreed. Perhaps we could make `asPublisher` the Java Flow one and rename the Reactive Streams one to something like `asReactiveStreamsPublisher`?

Another alternative would be to have static methods i.e. `ReactiveStreams.asPublisher(source)` vs `JavaFlow.asPublisher(source)` where `ReactiveStreams` would be a Scala object and `JavaFlow` would be a Java class with `asPublisher` being static methods.

That way the method name is the same, but you import a different object depending on if you want Java's flow or Reactive Streams

pjfanning:
@mdedetrich @raboof Is it just the asJavaSubscriber and asJavaProducer that is causing concerns?

If I separate out those changes but keep the moves of the fromPublisher and fromSubscriber methods, is that ok? Those methods can be overloaded because they don't hit issues with generics and type erasure.

mdedetrich:
> @mdedetrich @raboof Is it just the asJavaSubscriber and asJavaProducer that is causing concerns?
> 
> If I separate out those changes but keep the moves of the fromPublisher and fromSubscriber methods, is that ok? Those methods can be overloaded because they don't hit issues with generics and type erasure.

For me the design is cleaner to have `asSubscriber`/`asPublisher` and `fromSubscriber`/`fromPublisher` method names reused for both reactive streams and Java flow and just have different FQN (including the object), i.e. `pekko.stream.JavaFlow.asSubscriber` vs `pekko.stream.ReactiveStreams.asSubscriber`.

If the design is breaking with Pekko 1.3.x you can create alias methods in Pekko 1.3.x that point to the new design in 2.0.0 and deprecate the older ones (Pekko 1.3.x hasn't been released yet).

pjfanning:
I am focusing on 2.x changes. I'm not going to rush a change to suit the 1.3.0 timeline. Pekko 2.x is built with Java 17 making it easier to support and test Java Flow API integration.

pjfanning:
We already have JavaFlowSupport so I guess the feedback is that some people prefer that to us moving the code to Source and Sink directly. The question then arises about us deprecating the org.reactivestreams code in Source and Sink and creating a ReactiveStreamsSupport class that mirrors the JavaFlowSupport.

A more radical reworking sees us deprecating JavaFlowSupport and creating a new class that has similar utility functions but possibly with less verbose naming and then to mirror this new class to create an org.reactivestreams equivalent.

mdedetrich:
> I am focusing on 2.x changes. I'm not going to rush a change to suit the 1.3.0 timeline. Pekko 2.x is built with Java 17 making it easier to support and test Java Flow API integration.

I am aware, the point is not about rushing changes to suite a 1.3.x changeline but rather creating an ideal API design given that we have both reactive streams and java 9 flow

If that API design requires changing the current reactive-streams interopt then its possible to create the necessary changes (i.e. aliases/deprecations) in 1.3.0 branch, thats all

mdedetrich:
> We already have JavaFlowSupport so I guess the feedback is that some people prefer that to us moving the code to Source and Sink directly.

If the classes/datastructure didn't have the exact same name for both reactive streams and java 9 flow I would have also preferred this option, but in both cases they are called `Publishers`/`Subscribers` which means that you would have `source.asJava9Subscriber`/`source.asReactiveStreamsSubscriber` which is a bit of a mouthful.

You could avoid this on the scaladsl side since you can use implicit classes/extension methods and you only import the support that you can are about (i.e. import `java9FlowSupport` vs import `reactivestreams` support) but that doesn't work for javadsl.

Another disadvantage of inlining the source is that its not modular, given that already have java 9 flow/reactive streams flow we could have yet another streaming interface in the future which we might want to implement. Hence another argument for having separate objects/classes with static methods that provide the conversion methods.

He-Pin (commented):
Good direction -- promoting Java Flow API to first-class citizen in Source/Sink is the right move for Pekko 2.x (Java 17+).

**Design discussion points from the thread:**
1. The debate about whether to rename `asPublisher` to the Java Flow version and create `asReactiveStreamsPublisher` is valid. Given Pekko 2.x can break compatibility, making Java Flow the default makes sense.
2. The `ReactiveStreams.asPublisher(source)` vs `JavaFlow.asPublisher(source)` approach proposed by mdedetrich is elegant and would solve the naming conflict.

**Concerns:**
1. PR description says 'Still some doc changes to make' and mentions needing to create `asJavaSubscriber.md` and `asJavaPublisher.md` -- those files are now in the diff, so looks like docs are in progress.
2. The PR seems to have stalled since November 2025. The design discussion is valuable but a decision needs to be made on the API approach before proceeding.

**Recommendation:** Decide on the API naming strategy (inline methods vs separate support objects) and rebase. The current diff looks correct for the chosen approach.

PR #2464, opened 2025-11-09
Add OptionalUtil to Java util

mdedetrich:
When working on https://github.com/apache/pekko/pull/2409 I needed to convert from Java `OptionalInt` to Scala `Option[Int]` and I realized that due to Java Optional being invariant and Scala's `Int` being a subclass of Java's `Integer` you need some boilerplate in order convert from one to the other.

This boilerplate already lives in pekko-http since pekko-http already has this issue (see https://github.com/apache/pekko-http/blob/4833a8e42f946682a72a72a0f3bee4c4d662b9a6/http-core/src/main/java/org/apache/pekko/http/impl/util/Util.java) so this PR moves that code from pekko-http to pekko (since this will be the first time we need to use this code in pekko core).

Once 2.0.0-M1 will be published then we can change pekko-http 2.0.0 to use this code.

### DIFF
--- a/actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
+++ b/actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+/*
+ * Copyright (C) 2009-2022 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package org.apache.pekko.util;
+
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
+import org.apache.pekko.annotation.InternalStableApi;
+import scala.None$;
+import scala.jdk.javaapi.OptionConverters;
+
+/** INTERNAL API */
+@InternalStableApi
+public final class OptionalUtil {
+  private static final scala.Option<?> noneValue = None$.MODULE$;
+
+  public static <T> scala.Option<T> scalaNone() {
+    return (scala.Option<T>) noneValue;
+  }
+
+  @SuppressWarnings("unchecked") // no support for covariance of option in Java
+  // needed to provide covariant conversions that the Java interfaces don't provide automatically.
+  // The alternative would be having to cast around everywhere instead of doing it here in a central
+  // place.
+  public static <U, T extends U> Optional<U> convertOption(scala.Option<T> o) {
+    return (Optional<U>) (Object) OptionConverters.toJava(o);
+  }
+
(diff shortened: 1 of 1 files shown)


actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
```diff
@@ -0,0 +1,64 @@
+/*
```
mdedetrich, line 1: Note that even though this is technically a new file, the code is copied from https://github.com/apache/pekko-http/blob/4833a8e42f946682a72a72a0f3bee4c4d662b9a6/http-core/src/main/java/org/apache/pekko/http/impl/util/Util.java so we still need to use the lightbend header

actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
+import org.apache.pekko.annotation.InternalStableApi;
+import scala.None$;
+import scala.jdk.javaapi.OptionConverters;
+
+/** INTERNAL API */
+@InternalStableApi
+public class OptionalUtil {
```
He-Pin: `final` class

actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
```diff
@@ -0,0 +1,64 @@
...
+    } else {
+      return scala.Option.empty();
+    }
+  }
+
+  // This is needed to be used in Java source code that calls Scala code which expects scala.Int
+  // since an implicit cast from java.lang.Int to scala.Int is not available in Java source
+  public static scala.Option<Object> convertOptionalToScala(OptionalInt o) {
+    if (o.isPresent()) {
+      return new scala.Some(o.getAsInt());
```
He-Pin, line 59: Why not `Option<`Java.lang.Integer`>?

actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
mdedetrich: Done and pushed

actor/src/main/java/org/apache/pekko/util/OptionalUtil.java
mdedetrich: Because it doesn't pass type inference, this is precisely the reason why these util methods need to exist.

mdedetrich:
> seems generally ok but a couple of regression tests would be useful

Regression tests would be pointless due to how trivial the code is and also its case of if it compiles then it works, its just that I have split this PR out of the main work I am doing as part of zstd. We know that this code works as it was directly copied from pekko-http and its working locally with my zstd work

To put it differently, once other code starts using this code we know it will work because of that other work being able to compile. If someone changes the code in this PR to create a regression, than that other code that are using these utils will no longer compile.

mdedetrich:
Thanks for the review, I created https://github.com/apache/pekko-http/issues/869 as a reminder so that we use these methods in pekko-http once it starts depending on pekko 2.0.0-M1 milestone

PR #2456, opened 2025-11-08
feat: Add unpersistent versions of persistent behaviors

He-Pin:
Motivation:
port https://github.com/akka/akka-core/pull/31464  https://github.com/akka/akka-core/pull/31686

PR #2465, opened 2025-11-09
feat: Add unpersistent versions of persistent behaviors (#2456)

He-Pin:
(cherry picked from commit 523c05f7f6b9ab7cc8b4e91e80b877055221e405)


He-Pin:
Finally, the last one~

pjfanning:
@mdedetrich what do you think about the class names? I think NonPersistent would be a better name.

pjfanning:
Unpersist means to undo a persist action. Non-Persistence is a better way to describe a persistence approach which doesn't really use physical persistence but that creates a dummy implementation useful for tests.

He-Pin:
I agree, non persistent seems better.

He-Pin:
@pjfanning I don't have a computer by my side, would you mind submit a pr to do the renaming ？

pjfanning:
Let's see what people think about the name. If NonPersist has doubts, we could just go for a more neutral name prefixed with Testkit or some such.

PR #2467, opened 2025-11-09
Make automatic flushing of compression configurable

mdedetrich:
One interesting facet when working on the compression code is that while the `compress` method in `CompressionUtils` is used in tests, its actually never actually used in production for any of the pekko-stream compresion methods. Only `compressAndFlush` is used in production, hence this PR makes the flush behaviour of pekko-stream compression configurable, solving https://github.com/akka/akka/issues/21849. 

Tests have been added to make sure that this behaviour works as expected. Also this PR is required for https://github.com/apache/pekko/pull/2409 as I also want the flush behaviour to be configurable in that PR too and so I need to reuse the newly added `autoFlush` parameter on `CompressionUtils`. Docs have also been modified to reflect the changes.

The PR can be safely backported to 1.3.0 branch hence the `@since 1.3.0`.

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/CoderSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/CoderSpec.scala
@@ -45,6 +45,8 @@ abstract class CoderSpec(codecName: String) extends AnyWordSpec with CodecSpecSu
   case object AllDataAllowed extends Exception with NoStackTrace
   protected def corruptInputCheck: Boolean = true
 
+  protected def autoFlush: Boolean = true
+
   def extraTests(): Unit = {}
 
   s"The $codecName codec" should {
@@ -145,6 +147,7 @@ abstract class CoderSpec(codecName: String) extends AnyWordSpec with CodecSpecSu
     }
 
     "be able to decode chunk-by-chunk (depending on input chunks)" in {
+      assume(autoFlush)
       val minLength = 100
       val maxLength = 1000
       val numElements = 1000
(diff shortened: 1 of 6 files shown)


stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/CoderSpec.scala
```diff
@@ -145,6 +147,7 @@ abstract class CoderSpec(codecName: String) extends AnyWordSpec with CodecSpecSu
     }
 
     "be able to decode chunk-by-chunk (depending on input chunks)" in {
+      assume(autoFlush)
```
mdedetrich, line 150: This test is only valid when you flush on each element on the stream as its testing that each individual element can be decompressed on its own and that is not the case when you have `autoFlush` disabled

stream-tests/src/test/scala/org/apache/pekko/stream/io/compression/CoderSpec.scala
```diff
@@ -45,6 +45,8 @@ abstract class CoderSpec(codecName: String) extends AnyWordSpec with CodecSpecSu
   case object AllDataAllowed extends Exception with NoStackTrace
   protected def corruptInputCheck: Boolean = true
 
+  protected def autoFlush: Boolean = true
```
He-Pin, line 48: does that need flush interval too?

mdedetrich: Nope, having an interval based flush with compression would yield wild results

Issue #2398, opened 2025-10-26
failed: AsyncDnsResolverIntegrationSpec


He-Pin:
```
[10-26 12:53:45.192] [info] - must resolve single A record *** FAILED *** (10 seconds, 26 milliseconds)
[10-26 12:53:45.192] [info]   The future returned an exception of type: org.apache.pekko.pattern.AskTimeoutException, with message: Ask timed out on [Actor[pekko://AsyncDnsResolverIntegrationSpec/system/IO-DNS#-138638123]] after [10000 ms]. Message of type [org.apache.pekko.io.dns.DnsProtocol$Resolve]. A typical reason for `AskTimeoutException` is that the recipient actor didn't send a reply.. (AsyncDnsResolverIntegrationSpec.scala:210)
[10-26 12:53:45.192] [info]   org.scalatest.exceptions.TestFailedException:
[10-26 12:53:45.192] [info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:333)
[10-26 12:53:45.193] [info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[10-26 12:53:45.193] [info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:259)
[10-26 12:53:45.193] [info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:281)
[10-26 12:53:45.193] [info]   at org.apache.pekko.io.dns.AsyncDnsResolverIntegrationSpec.resolve$1(AsyncDnsResolverIntegrationSpec.scala:210)
[10-26 12:53:45.193] [info]   at org.apache.pekko.io.dns.AsyncDnsResolverIntegrationSpec.f$proxy1$1(AsyncDnsResolverIntegrationSpec.scala:77)
[10-26 12:53:45.193] [info]   at org.apache.pekko.io.dns.AsyncDnsResolverIntegrationSpec.$init$$$anonfun$1$$anonfun$1(AsyncDnsResolverIntegrationSpec.scala:75)
[10-26 12:53:45.193] [info]   at org.scalatest.Transformer.apply$$anonfun$1(Transformer.scala:22)
[10-26 12:53:45.193] [info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[...]

PR #2471, opened 2025-11-10
chore: don't auto-update agrona

raboof:
As agrona according to the comment must align with the version used in aeron, so it should only be updated when we update aeron.

We already missed that in #2280, #2446 and #2461 so let's exclude it from scala-steward (and see if we can be more careful with aeron updates...)

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -5,6 +5,9 @@ updates.pin = [
   { groupId = "org.scala-lang", artifactId = "scala3-library", version = "3.3." }
   # sbt-assembly 2.3 causes build issues (https://github.com/apache/pekko/pull/1744)
   { groupId = "com.eed3si9n", artifactId = "sbt-assembly", version = "2.2." }
+  # agrona major+minor version should match the one brought
+  # in by aeron
+  { groupId = "org.agrona", artifactId = "agrona", version = "2.2." }
 ]
 
 updates.ignore = [
@@ -25,8 +28,6 @@ updates.ignore = [
   { groupId = "com.lightbend.sbt", artifactId = "sbt-java-formatter" }
   # lmdbjava 0.9.2 causes issues - https://github.com/apache/pekko/issues/2484
   { groupId = "org.lmdbjava" }
-  # agrona should only be updated when aeron is updated
-  { groupId = "org.agrona", artifactId = "agrona" }
 ]
 
 updatePullRequests = "always"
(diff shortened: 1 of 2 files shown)


.scala-steward.conf
```diff
@@ -23,6 +23,8 @@ updates.ignore = [
   { groupId = "tools.jackson.jaxrs" }
   # other ignored updates
   { groupId = "com.lightbend.sbt", artifactId = "sbt-java-formatter" }
+  # agrona should only be updated when aeron is updated
+  { groupId = "org.agrona", artifactId = "agrona" }
```
pjfanning: I think it would be better to pin to 2.2

raboof: As far as I remember we've historically keps agrona at *exactly* the version used by aeron, not deviating in patch versions either.

I'm OK with loosening this to only keeping the major+minor in sync

pjfanning: ok - let's merge this then

raboof:
failure seems to be at least #2398 - let's not merge this until we can make progress on that

pjfanning:
@raboof I think checking DNS related stuff in CI agents is just not going to be robust. Even if this has become flakier, I suspect that it is down to GitHub CI as opposed to any changes on our side.

raboof:
> Even if this has become flakier, I suspect that it is down to GitHub CI as opposed to any changes on our side.

Very likely 

> I think checking DNS related stuff in CI agents is just not going to be robust.

OTOH, the whole point of Pekko is to allow building systems that are robust even in challenging circumstances, so I think we should be able to make this more solid :)

raboof:
Other failure, filed #2476 and re-running

raboof:
Interestingly it was https://github.com/apache/pekko/issues/2476 again that failed, rerunning again

pjfanning:
@raboof I see that lmdbjava relies on agrona 1.22 - so it could be that we are now affected by a combination of this and #2483 

Could we merge the 2 of these and see if we get some stability? 

https://github.com/search?q=repo%3Almdbjava%2Flmdbjava%20agrona&type=code

pjfanning:
https://github.com/apache/pekko/pull/2483 is merged so I rebased this to use latest commits on main branch

raboof:
> @raboof I see that lmdbjava relies on agrona 1.22 - so it could be that we are now affected by a combination of this and #2483
> 
> Could we merge the 2 of these and see if we get some stability?
> 
> https://github.com/search?q=repo%3Almdbjava%2Flmdbjava%20agrona&type=code

oof..

PR #2477, opened 2025-11-11
Fix AsyncWriteJournalResponseOrderSpec with global dispatcher

migesok:
Follow up to this discussion: https://github.com/apache/pekko/issues/2026#issuecomment-3501310771

This fixes a random test failure when the global dispatcher used as a persistence plugin dispatcher.

This change should unblock https://github.com/apache/pekko/pull/2434
Also adjusted wording changes for write-response-global-order from https://github.com/apache/pekko/pull/2437

The fix explanation:
The test relied on an ordering guarantee which wasn't really there. Even though asyncWriteMessages promises were completed in some particular order, AsyncWriteJournal sent out messages in an onComplete callback scheduled on the actor dispatcher, yielding random message order. The test failure shouldn't be a problem for production, non-test, Pekko code, because nothing there relies on write response ordering guarantees between independent persistent actors. And for an individual persistent actor writes - the persistent actor implementation internals handle ordering there correctly regardless of the plugin response ordering.

### DIFF
--- a/persistence/src/main/resources/reference.conf
+++ b/persistence/src/main/resources/reference.conf
@@ -165,9 +165,7 @@ pekko.persistence {
       # When this setting is "off", the journal plugin is allowed to send back write responses
       # in any order. This can improve throughput and reduce latency under load.
       #
-      # The old behavior is still enabled by default ("on"). After more testing on existing applications,
-      # the default might be switched to "off", and eventually this option might be removed altogeter, leaving
-      # "off" the only behaviour available.
+      # The old behavior is still enabled by default ("on").
       write-response-global-order = on
     }
 


pjfanning:
Thanks for the PR. But I would prefer not to use conf files to promise anything. Right now, there are no plans to switch the default behaviour.

migesok:
> Thanks for the PR. But I would prefer not to use conf files to promise anything. Right now, there are no plans to switch the default behaviour.

@pjfanning sure. I removed the relevant part. Please check again.

migesok:
```
[11-11 20:47:20.836] [error] (cluster-sharding / Pr-validation / test) sbt.TestsFailedException: Tests unsuccessful
```
I haven't changed anything which can reach this place. Some random failure not connected with the change?
I ran tests for relevant modules locally, it worked and without the change.

mdedetrich:
@pjfanning Shall we backport this to 1.3.x since we also maintain branch?

pjfanning:
@migesok thay's my intention - I'm just waiting to see how #2482 gets on

PR #2479, opened 2025-11-12
DRAFT: Split ActorSystem creation into javadsl and scaladsl

mdedetrich:
This PR is an exploratory PR to see whether its worth it to resolve https://github.com/apache/pekko/issues/2093 as a non source breaking change when going from Pekko 1.3.0 to 2.0.0. If we decide that a source breaking change in Pekko 2.0.0 is acceptable then this PR is not necessary  

Explanation of methods

- `terminate`: This one is deprecated and is no longer meant to be used. I was forced to do this since the current `terminate` method returns a `Future` making it impossible to create the appropriate return type for `javadsl`/`scaladsl` (`javadsl` should return `CompletionStage` where as `scaladsl` should return `Future`). Instead we have `terminateAsync` which returns `Future` for `scaladsl` and `CompletionStage` for `javadsl`
- `close`: A method with a configurable timeout that blocks on termination. Also implements the Java `AutoCloseable` interface, meaning it will execute automatically if an `ActorSystem` is created in a try/catch clause.
  - Adding `AutoCloseable` may be too much as it is a behaviour change, might be better to do this in Pekko 2.0.0
- `closeAsync`: Essentially the same as old `terminate`/new `terminateAsync` but it doesn't return a `Future/`CompletionStage` but rather just `Unit`/`void`. This is necessary in cases where you have the root `pekko.actor.ActorSystem` and still need to terminate the system in a non blocking manner, typically when using `context` i.e.
    ```scala
    case TerminateMsg(Left(false)) =>
      context.system.closeAsync()
      stop()
    ```

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -1212,6 +1212,9 @@ pekko {
     # Terminate the ActorSystem in the last phase actor-system-terminate.
     terminate-actor-system = on
 
+    # The timeout that will be used when calling .close on an ActorSystem
+    close-actor-system-timeout = 60 s
+
     # Exit the JVM (System.exit(0)) in the last phase actor-system-terminate
     # if this is set to 'on'. It is done after termination of the
     # ActorSystem if terminate-actor-system=on, otherwise it is done
--- a/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
@@ -17,17 +17,13 @@ import java.io.Closeable
 import java.util.Optional
 import java.util.concurrent._
 import java.util.concurrent.atomic.AtomicReference
-
 import scala.annotation.tailrec
 import scala.collection.immutable
-import scala.concurrent.{ ExecutionContext, ExecutionContextExecutor, Future, Promise }
-import scala.concurrent.blocking
+import scala.concurrent.{ blocking, Await, ExecutionContext, ExecutionContextExecutor, Future, Promise }
 import scala.concurrent.duration.Duration
 import scala.util.{ Failure, Success, Try }
 import scala.util.control.{ ControlThrowable, NonFatal }
-
 import com.typesafe.config.{ Config, ConfigFactory }
-
 import org.apache.pekko
 import pekko.ConfigurationException
 import pekko.actor.dungeon.ChildrenContainer
@@ -189,6 +185,7 @@ object ActorSystem {
    * associated with the ActorSystem class.
    * Then it loads the default reference configuration using the ClassLoader.
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(): ActorSystem = apply()
 
   /**
@@ -198,18 +195,21 @@ object ActorSystem {
    * associated with the ActorSystem class.
    * Then it loads the default reference configuration using the ClassLoader.
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(name: String): ActorSystem = apply(name)
 
   /**
    * Java API: Creates a new actor system with the specified name and settings
    * The core actor system settings are defined in [[BootstrapSetup]]
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(name: String, setups: ActorSystemSetup): ActorSystem = apply(name, setups)
 
   /**
    * Java API: Shortcut for creating an actor system with custom bootstrap settings.
    * Same behavior as calling `ActorSystem.create(name, ActorSystemSetup.create(bootstrapSettings))`
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(name: String, bootstrapSetup: BootstrapSetup): ActorSystem =
     create(name, ActorSystemSetup.create(bootstrapSetup))
 
@@ -221,6 +221,7 @@ object ActorSystem {
    *
    * @see <a href="https://lightbend.github.io/config/latest/api/index.html" target="_blank">The Typesafe Config Library API Documentation</a>
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(name: String, config: Config): ActorSystem = apply(name, config)
 
   /**
@@ -228,6 +229,7 @@ object ActorSystem {
    *
    * @see <a href="https://lightbend.github.io/config/latest/api/index.html" target="_blank">The Typesafe Config Library API Documentation</a>
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(name: String, config: Config, classLoader: ClassLoader): ActorSystem = apply(name, config, classLoader)
 
   /**
@@ -245,6 +247,7 @@ object ActorSystem {
    *
    * @see <a href="https://lightbend.github.io/config/latest/api/index.html" target="_blank">The Typesafe Config Library API Documentation</a>
    */
+  @deprecated("Use org.apache.pekko.actor.javadsl.ActorSystem.create instead", "1.3.0")
   def create(
       name: String,
       config: Config,
@@ -259,6 +262,7 @@ object ActorSystem {
    * associated with the ActorSystem class.
    * Then it loads the default reference configuration using the ClassLoader.
    */
+  @deprecated("Use org.apache.pekko.actor.scaladsl.ActorSystem.apply instead", "1.3.0")
   def apply(): ActorSystem = apply("default")
 
   /**
@@ -268,12 +272,14 @@ object ActorSystem {
    * associated with the ActorSystem class.
    * Then it loads the default reference configuration using the ClassLoader.
    */
+  @deprecated("Use org.apache.pekko.actor.scaladsl.ActorSystem.apply instead", "1.3.0")
   def apply(name: String): ActorSystem = apply(name, None, None, None)
 
   /**
    * Scala API: Creates a new actor system with the specified name and settings
    * The core actor system settings are defined in [[BootstrapSetup]]
    */
+  @deprecated("Use org.apache.pekko.actor.scaladsl.ActorSystem.apply instead", "1.3.0")
   def apply(name: String, setup: ActorSystemSetup): ActorSystem = {
     val bootstrapSettings = setup.get[BootstrapSetup]
     val cl = bootstrapSettings.flatMap(_.classLoader).getOrElse(findClassLoader())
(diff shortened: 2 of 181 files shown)


actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
```diff
@@ -697,7 +713,56 @@ abstract class ActorSystem extends ActorRefFactory with ClassicActorSystemProvid
...
+      Duration(settings.config.getDuration("coordinated-shutdown.close-actor-system-timeout").toMillis,
+        TimeUnit.MILLISECONDS))
+  }
+
+  /**
+   * Asynchronously terminates this actor system by running [[CoordinatedShutdown]] with reason
+   * [[CoordinatedShutdown.ActorSystemTerminateReason]]. This method will block
+   * until either the actor system is terminated or
+   * `pekko.coordinated-shutdown.close-actor-system-timeout` timeout duration is
+   * passed, in which case a [[TimeoutException]] is thrown.
```
pjfanning, line 748: if a TimeoutException can be thrown, why not add it as an `@throws(classOf[TimeoutException])`?

actor/src/main/scala/org/apache/pekko/actor/scaladsl/ActorSystem.scala
```diff
@@ -0,0 +1,136 @@
...
+   * Creates a new ActorSystem with the specified name,
+   * obtains the current ClassLoader by first inspecting the current threads' getContextClassLoader,
+   * then tries to walk the stack to find the callers class loader, then falls back to the ClassLoader
+   * associated with the ActorSystem class.
+   * Then it loads the default reference configuration using the ClassLoader.
+   */
+  def apply(name: String): ActorSystem = apply(name, None, None, None)
+
+  /**
+   * Scala API: Creates a new actor system with the specified name and settings
```
pjfanning: Why do we need 'Scala API' in the scaladoc of ScalaDSL classes? This appears in other methods in this class too.

actor/src/main/scala/org/apache/pekko/actor/javadsl/ActorSystem.scala
```diff
@@ -0,0 +1,139 @@
...
+   * Creates a new ActorSystem with the specified name,
+   * obtains the current ClassLoader by first inspecting the current threads' getContextClassLoader,
+   * then tries to walk the stack to find the callers class loader, then falls back to the ClassLoader
+   * associated with the ActorSystem class.
+   * Then it loads the default reference configuration using the ClassLoader.
+   */
+  def create(name: String): ActorSystem = create(name, None, None, None)
+
+  /**
+   * Scala API: Creates a new actor system with the specified name and settings
```
pjfanning: This can't be a Scala API if it's in a Java DSL class - this appears in other places in this class too

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
```diff
@@ -17,17 +17,13 @@ import java.io.Closeable
 import java.util.Optional
 import java.util.concurrent._
 import java.util.concurrent.atomic.AtomicReference
-
```
pjfanning, line 20: can we keep the whitespace separating different groups of imports?

actor/src/main/resources/reference.conf
```diff
@@ -1212,6 +1212,9 @@ pekko {
     # Terminate the ActorSystem in the last phase actor-system-terminate.
     terminate-actor-system = on
 
+    # The timeout that will be used when calling .close on an ActorSystem
+    close-actor-system-timeout = 10 s
```
pjfanning: Is this long enough?

actor/src/main/scala/org/apache/pekko/actor/scaladsl/ActorSystem.scala
mdedetrich: This was accidentally copied over

actor/src/main/scala/org/apache/pekko/actor/javadsl/ActorSystem.scala
mdedetrich: This was accidentally copied over

actor/src/main/resources/reference.conf
mdedetrich: From all of the tests, yes. There isn't any timeout that runs longer than 10 seconds, and if there is you can always configure it to be higher

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
mdedetrich: It is https://github.com/apache/pekko/blob/2ed0aea6f18b0cec5ecbd5e4118d2cc2c9b43c42/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala#L738

actor/src/main/resources/reference.conf
pjfanning: I would favour a longer value. I think most users who choose to use the .close method would be happy enough to wait for longer, maybe 1 minute before a TimeoutException.

pjfanning:
In the PR description, would it be possible to provide a succinct description of what the close, terminate, closeAsync and terminateAsync methods do?

mdedetrich:
> In the PR description, would it be possible to provide a succinct description of what the close, terminate, closeAsync and terminateAsync methods do?

Will do, right now this is more of an exploratory PR to see if we should go ahead with it at all (see https://lists.apache.org/thread/0lzlhh4ll08t1o1gfh7w08635c5t6bw3)

mdedetrich:
> In the PR description, would it be possible to provide a succinct description of what the close, terminate, closeAsync and terminateAsync methods do?

Done

actor/src/main/scala/org/apache/pekko/actor/javadsl/ActorSystem.scala
```diff
@@ -0,0 +1,124 @@
...
+   * If no ExecutionContext is given, the system will fallback to the executor configured under "pekko.actor.default-dispatcher.default-executor.fallback".
+   * The system will use the passed in config, or falls back to the default reference configuration using the ClassLoader.
+   *
+   * @see <a href="https://lightbend.github.io/config/latest/api/index.html" target="_blank">The Typesafe Config Library API Documentation</a>
+   */
+  def create(
+      name: String,
+      config: Option[Config] = None,
+      classLoader: Option[ClassLoader] = None,
+      defaultExecutionContext: Option[ExecutionContext] = None): ActorSystem =
```
He-Pin: Optional

He-Pin (commented):
I think the shape is great, but what about the `AutoCloseable`?

actor/src/main/resources/reference.conf
mdedetrich: Updated with 60 seconds

mdedetrich:
> I think the shape is great, but what about the `AutoCloseable`?

Its implemented

He-Pin:
@mkurz FYI too

pjfanning:
Whatever we come up, we should make sure that the typed ActorSystem is then set up to have an equivalent API.
Same split into Java DSL / Scala DSL. Same method names and behaviours for the `close*` and `terminate*` methods. Autocloseable interface.
That can be done separately but we shouldn't release 1.3.0 or 2.0.0-M1 with changes in classic ActorSystem and no changes in typed ActorSystem.

mdedetrich:
> That can be done separately but we shouldn't release 1.3.0 or 2.0.0-M1 with changes in classic ActorSystem and no changes in typed ActorSystem.

I also plan to do typed `ActorSystem` but that will come later

Also can we please not do copilot AI review until PR is done, its just going to create a whole lot of noise at a time when its not useful

He-Pin:
@mdedetrich Is there an ETA, maybe we can do that in one PR. to make sure it consistentcy

mdedetrich:
So I have some bad/annoying news, it appears that its not possible to do this in Pekko 1.3.x because of how the inheritance between the various Actor classes is done and the fact that Java interfaces cannot extend abstract classes. I can get the `scaladsl` to work by having the `actor.scaladsl.ActorSystem`/`actor.javadsl.ActorSystem`be a `trait`, converting `actor.ActorSystem` to a trait but the same doesn't work because we need a constructor for `actor.ActorSystem` (some parts of Akka can instantiate a `actor.ActorSystem` manually by reflection, this also breaks MiMa).

I tried making `ActorSystemImpl` a trait to get around this, but we then get a cyclic inheritance issue. Ill have another try at it, but it looks like that in order to do this it would have to be with breaking changes in Pekko 2.0.0 where some slight adjustments to the inheritance need to be made

mdedetrich:
> @mdedetrich Is there an ETA, maybe we can do that in one PR. to make sure it consistentcy

There is no ETA as of yet, this is a proof of concept to see if its even possible (and given the previous comment it doesn't look like it is)

He-Pin:
then what about we delay this in 2.0.0, and let 1.3.x released soon?

pjfanning:
Thanks @mdedetrich, can we retarget at main branch and continue there?

mdedetrich:
> Thanks @mdedetrich, can we retarget at main branch and continue there?

Give me a couple of days to figure out, if that doesn't work then I will make a new PR against 2.0.0 as it will be slightly different and will have a better/cleaner API.

Also if anyone has any other alternatives feel free to checkout the branch and play with it locally, may have forgetten something

pjfanning:
One possible 1.x solution would to leave the existing ActorSystem class as is except maybe to deprecate its methods. You can add new Scala and Java DSL classes that delegate to the existing class but that implement their own APIs and Autocloseable. The new classes would not subclass the existing class.
I would still prefer to see this in main branch first and then something that implements something equivalent can be backported.
In the end of day, there isn't a big demand for this. I'd prefer to proceed with 1.3.0 without this and I expect we'll have a 1.4.0 at some stage.

mdedetrich:
> One possible 1.x solution would to leave the existing ActorSystem class as is except maybe to deprecate its methods. You can add new Scala and Java DSL classes that delegate to the existing class but that implement their own APIs and Autocloseable.

The issue is that even doing this doesn't work at least without a huge amount of code churn. Direct inheritance doesn't work (as described) and delegation also won't work for other reasons, a lot of Actor code expects to work with `pekko.actor.ActorSystem` (or one of its subclasses) and so delegation won't work here.

Regarding `AutoCloseable`, I made a PR at https://github.com/apache/pekko/pull/2486. The `AutoCloseable` interface is not reliant on scala/java dsl in any way so that can be implemented irrespective of https://github.com/apache/pekko/issues/2093

He-Pin (commented):
This PR targets `1.3.x` but the discussion in the comments indicates it may not be feasible without breaking changes. Key points from the thread:

1. The author (`mdedetrich`) discovered that Java interface/abstract class inheritance constraints make this impossible to implement cleanly in 1.3.x without MiMa breaks.
2. PJ Fanning suggested a delegation approach, but the author confirmed delegation won't work because internal Pekko code expects `pekko.actor.ActorSystem` specifically.
3. The consensus seems to be to defer this to 2.0.0.

**Recommendation:** This PR should either be closed in favor of a new PR targeting `main` (2.0.0), or the author should confirm whether they plan to continue work. It's been sitting since November 2025 with no recent activity.

The design direction (separate Scala/Java DSLs with proper return types, AutoCloseable support) is good, but the inheritance challenges are real and would need a clean-slate approach for 2.0.0.

PR #2486, opened 2025-11-12
Add close on ActorSystem

mdedetrich:
Adds the `close` method on `ActorSystem` (classic and typed) as well as implementing the `AutoCloseable` interface, work taken from https://github.com/apache/pekko/pull/2479. Regardless of what happens with https://github.com/apache/pekko/issues/2093, the changes in this PR are always valid (`close` is blocking and returns `Unit`/`Void` so its valid for both `javadsl` and `scaladsl`).

The PR also cleans up test code by removing the `Await.result(system.close,...)` pattern that is used everywhere.

An overloaded `close(duration: Duration)` method can be added in Pekko 2.0.0 if https://github.com/apache/pekko/pull/2479 works out, as the `scaladsl` will use the Scala `Duration` and the `javadsl` will use the Java `Duration`.

### DIFF
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/ActorSystemStub.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/ActorSystemStub.scala
@@ -113,6 +113,10 @@ import com.typesafe.config.{ Config, ConfigFactory }
   override def terminate(): Unit = terminationPromise.trySuccess(Done)
   override def whenTerminated: Future[Done] = terminationPromise.future
   override def getWhenTerminated: CompletionStage[Done] = whenTerminated.asJava
+  override def close(): Unit = {
+    terminate()
+    Await.result(whenTerminated, scala.concurrent.duration.Duration.Inf)
+  }
   override val startTime: Long = System.currentTimeMillis()
   override def uptime: Long = System.currentTimeMillis() - startTime
   override def threadFactory: java.util.concurrent.ThreadFactory = new ThreadFactory {
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
@@ -122,6 +122,8 @@ import org.slf4j.{ Logger, LoggerFactory }
   override lazy val getWhenTerminated: CompletionStage[pekko.Done] =
     whenTerminated.asJava
 
+  override def close(): Unit = system.close()
+
   override def systemActorOf[U](behavior: Behavior[U], name: String, props: Props): ActorRef[U] = {
     val ref = system.systemActorOf(
       PropsAdapter(
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -1244,6 +1244,12 @@ pekko {
     # Terminate the ActorSystem in the last phase actor-system-terminate.
     terminate-actor-system = on
 
+    # The timeout that will be used when calling .close on an ActorSystem.
+    # This timeout will also be used when ActorSystem's are automatically
+    # terminated by using Java's try-with-resources or Scala's
+    # scala.util.Using
+    close-actor-system-timeout = 60 s
+
     # Exit the JVM (System.exit(0)) in the last phase actor-system-terminate
     # if this is set to 'on'. It is done after termination of the
     # ActorSystem if terminate-actor-system=on, otherwise it is done
--- a/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
@@ -20,7 +20,7 @@ import java.util.concurrent.atomic.AtomicReference
 
 import scala.annotation.tailrec
 import scala.collection.immutable
-import scala.concurrent.{ ExecutionContext, ExecutionContextExecutor, Future, Promise }
+import scala.concurrent.{ Await, ExecutionContext, ExecutionContextExecutor, Future, Promise }
 import scala.concurrent.blocking
 import scala.concurrent.duration.Duration
 import scala.jdk.CollectionConverters._
@@ -524,7 +524,7 @@ object ActorSystem {
  * extending [[pekko.actor.ExtendedActorSystem]] instead, but beware that you
  * are completely on your own in that case!
  */
-abstract class ActorSystem extends ActorRefFactory with ClassicActorSystemProvider {
+abstract class ActorSystem extends ActorRefFactory with ClassicActorSystemProvider with AutoCloseable {
   import ActorSystem._
 
   /**
@@ -676,6 +676,26 @@ abstract class ActorSystem extends ActorRefFactory with ClassicActorSystemProvid
    * future completes.
    */
   def terminate(): Future[Terminated]
+
+  /**
+   * Terminates this actor system by running [[CoordinatedShutdown]] with reason
+   * [[CoordinatedShutdown.ActorSystemTerminateReason]]. This method will block
+   * until either the actor system is terminated or
+   * `pekko.coordinated-shutdown.close-actor-system-timeout` timeout duration is
+   * passed, in which case a [[TimeoutException]] is thrown.
+   *
+   * If `pekko.coordinated-shutdown.run-by-actor-system-terminate` is configured to `off`
+   * it will not run `CoordinatedShutdown`, but the `ActorSystem` and its actors
+   * will still be terminated.
+   *
+   * This will stop the guardian actor, which in turn
+   * will recursively stop all its child actors, and finally the system guardian
+   * (below which the logging actors reside) and then execute all registered
+   * termination handlers (see [[ActorSystem#registerOnTermination]]).
+   * @since 1.3.0
+   */
+  @throws(classOf[TimeoutException])
+  override def close(): Unit
 
   /**
    * Returns a Future which will be completed after the ActorSystem has been terminated
@@ -1078,6 +1098,14 @@ private[pekko] class ActorSystemImpl(
       finalTerminate()
     }
     whenTerminated
+  }
+
+  override def close(): Unit = {
+    terminate()
+    val duration =
+      Duration(settings.config.getDuration("pekko.coordinated-shutdown.close-actor-system-timeout").toMillis,
+        TimeUnit.MILLISECONDS)
+    Await.result(whenTerminated, duration)
   }
 
   override private[pekko] def finalTerminate(): Unit = {
--- a/actor-tests/src/test/java/org/apache/pekko/actor/ActorSystemTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/actor/ActorSystemTest.java
@@ -15,6 +15,7 @@ package org.apache.pekko.actor;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 import java.util.concurrent.CompletionStage;
 import org.apache.pekko.testkit.PekkoJUnitActorSystemResource;
@@ -46,5 +47,15 @@ public class ActorSystemTest extends JUnitSuite {
   @Test
   public void testGetWhenTerminatedWithoutTermination() {
     assertFalse(system.getWhenTerminated().toCompletableFuture().isDone());
+  }
+
+  @Test
+  public void testTryWithResources() throws Exception {
+    ActorSystem system = null;
+    try (ActorSystem actorSystem = ActorSystem.create()) {
+      system = actorSystem;
+    }
+    final CompletionStage<Terminated> cs = system.getWhenTerminated();
+    assertTrue(cs.toCompletableFuture().isDone());
   }
 }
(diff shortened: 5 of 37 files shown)


raboof:
> `close` is blocking and returns `Unit`/`Void`

Is there a requirement for close to block until the closing is 'complete'?

> The PR also cleans up test code by removing the `Await.result(system.close,...)` pattern that is used everywhere.

I can see the appeal, though blocking is a bit 'uncharacteristic' for us. I guess because you're stopping the ActorSystem and thus are anyway 'on your way out' it might be fine?

mdedetrich:
> Is there a requirement for close to block until the closing is 'complete'?

I believe so otherwise you can get weird behaviour but I will do research to confirm. From what I gather (and also when learning Kotlin and using stuff like Arrow Resource), `AutoCloseable` is kind of pointless if the `close` method is asynchronous, which is why you have stuff like https://docs.datastax.com/en/latest-java-driver-api/com/datastax/oss/driver/api/core/AsyncAutoCloseable.html

> I can see the appeal, though blocking is a bit 'uncharacteristic' for us. I guess because you're stopping the ActorSystem and thus are anyway 'on your way out' it might be fine?

To clarify, if I replace `Await.result` with `close` its because the behaviour of blocking is exactly the same. In those cases, the only thing that would have changed would be the timeout (i.e. sometimes it was 15 seconds, sometimes 20 but these are all tests so it shouldn't be an issue anyways).

raboof:
> > Is there a requirement for close to block until the closing is 'complete'?
> 
> I believe so otherwise you can get weird behaviour but I will do research to confirm. From what I gather (and also when learning Kotlin and using stuff like Arrow Resource), `AutoCloseable` is kind of pointless if the `close` method is asynchronous

Why? When I want to make sure the cleanup is triggered but don't care too much when it happens, it might be fine?

> which is why you have stuff like https://docs.datastax.com/en/latest-java-driver-api/com/datastax/oss/driver/api/core/AsyncAutoCloseable.html

It's an interesting data point that the `close` of that one is also blocking.
 
> > I can see the appeal, though blocking is a bit 'uncharacteristic' for us. I guess because you're stopping the ActorSystem and thus are anyway 'on your way out' it might be fine?
> 
> To clarify, if I replace `Await.result` with `close` its because the behaviour of blocking is exactly the same. In those cases, the only thing that would have changed would be the timeout (i.e. sometimes it was 15 seconds, sometimes 20 but these are all tests so it shouldn't be an issue anyways).

Sure, I meant blocking in the 'production' code is a bit uncharacteristic - in tests it's indeed definitely fine.

mdedetrich:
> Why? When I want to make sure the cleanup is triggered but don't care too much when it happens, it might be fine?

From what I recall, if there is an error in cleaning up on `close()` its expected that an exception be thrown which is not possible if its asynchronous but I will look into it.

I would say personally that having `close()` cleaning up resources in an asynchronous manner would be surprising behaviour

mdedetrich:
So regarding blocking on `close()`, the official documentation from https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html i.e.

> See the Javadoc of the [AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) and [Closeable](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html) interfaces for a list of classes that implement either of these interfaces. The Closeable interface extends the AutoCloseable interface. The close method of the Closeable interface throws exceptions of type IOException while the close method of the AutoCloseable interface throws exceptions of type Exception. Consequently, subclasses of the AutoCloseable interface can override this behavior of the close method to throw specialized exceptions, such as IOException, or no exception at all.

And more directly from https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

> While this interface method is declared to throw Exception, implementers are strongly encouraged to declare concrete implementations of the close method to throw more specific exceptions, or to throw no exception at all if the close operation cannot fail.

> Cases where the close operation may fail require careful attention by implementers. It is strongly advised to relinquish the underlying resources and to internally mark the resource as closed, prior to throwing the exception. The close method is unlikely to be invoked more than once and so this ensures that the resources are released in a timely manner. Furthermore it reduces problems that could arise when the resource wraps, or is wrapped, by another resource.

Implies that it is expected that if the `close()` method fails that an exception should be thrown. It is possible for `ActorSystem.terminate()` to fail and the only way to throw an exception in that case is to block on it.

Personally I was also apprehensive of implementing `close()` for this exact reason, blocking isn't entirely desirable but to me it seems expected that `close()` methods throw an exception if the closing of the resource fails.

pjfanning:
Would it be acceptable to not put this on ActorSystem but create a factory method like `ActorSystem.createWithBlockingClose()` or `ActorSystem.apply(withBlockingClose: Boolean = false)`? The returned ActorSystem would mix in a close method.

We could also support mixing in a non-blocking close.

mdedetrich:
> Would it be acceptable to not put this on ActorSystem but create a factory method like `ActorSystem.createWithBlockingClose()` or `ActorSystem.apply(withBlockingClose: Boolean = false)`?

Well I would still have to put it on `ActorSystem` otherwise `AutoCloseable` won't work, but I can make the blocking behaviour configurable (alternately it can go in `reference.conf` though given that `close` should block by default)

I think the point is, if you are going to use `close()` and don't want it to block, at that point you are already rolling your own resource cleaning mechanism and so `close()` + `AutoCloseable` is kind of pointless. There are also already methods that do close in an async manner (thats what `.terminate()` does).

This really does seem to be one of those cases where either we implement `close` as blocking by default, or we don't implement it at all as to "force"/encourage users to not block on `ActorSystem` termination. I still fall into the latter camp, but I do admit I haven't seen a case where a blocking `ActorSystem.close()` could cause issues in the common case (it can create deadlocks if you have blocking code in actors, which you shouldn't, but thats why the timeout is there).

pjfanning:
> > Would it be acceptable to not put this on ActorSystem but create a factory method like `ActorSystem.createWithBlockingClose()` or `ActorSystem.apply(withBlockingClose: Boolean = false)`?
> 
> Well I would still have to put it on `ActorSystem` otherwise `AutoCloseable` won't work, but I can make the blocking behaviour configurable (alternately it can go in `reference.conf` though given that `close` should block by default)>
> 
> I think the point is, if you are going to use `close()` and don't want it to block, at that point you are already rolling your own resource cleaning mechanism at which point `close()` + `AutoCloseable` is kind of pointless. There are also already methods that do close in an async manner (thats what `.terminate()` does).

A config option to decide on whether the close is blocking alongside a config to control the timeout for blocking seems ok.

mdedetrich:
> A config option to decide on whether the close is blocking alongside a config to control the timeout for blocking seems ok.

Cool, Ill see feedback from others as well

samueleresca:
Does this PR also need to update [ActorSystemLifecycle.scala#L54](https://github.com/apache/pekko/blob/main/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/ActorSystemLifecycle.scala#L54)?

mdedetrich:
> Does this PR also need to update [ActorSystemLifecycle.scala#L54](https://github.com/apache/pekko/blob/main/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/ActorSystemLifecycle.scala#L54)?

So for this I would need an overloaded version of close with a duration i.e. `close(duration)` and that version would come after creating a `scaladsl`/`javadsl` for `ActorSystem` since the `javadsl` will use a java duration and the `scaladsl` will use a scala duration.

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
```diff
@@ -1080,6 +1100,14 @@ private[pekko] class ActorSystemImpl(
     whenTerminated
   }
 
+  override def close(): Unit = {
```
He-Pin, line 1103: @throws(classOf[TimeoutException])

actor/src/main/resources/reference.conf
```diff
@@ -1244,6 +1244,9 @@ pekko {
     # Terminate the ActorSystem in the last phase actor-system-terminate.
     terminate-actor-system = on
 
+    # The timeout that will be used when calling .close on an ActorSystem
+    close-actor-system-timeout = 60 s
```
He-Pin, line 1251: ForkJoinPool's default is `1Day`

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
```diff
@@ -122,6 +122,8 @@ import org.slf4j.{ Logger, LoggerFactory }
   override lazy val getWhenTerminated: CompletionStage[pekko.Done] =
     whenTerminated.asJava
 
+  override def close(): Unit = system.close()
```
He-Pin, line 125: @throws(classOf[TimeoutException])

He-Pin:
Once get merged, this should be ported to 2.0.0

btw, @mdedetrich what's the status of kotlin dsl of Pekko?

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
He-Pin, line 125: add `@since 1.3.0`

actor/src/main/resources/reference.conf
```diff
@@ -1244,6 +1244,9 @@ pekko {
     # Terminate the ActorSystem in the last phase actor-system-terminate.
     terminate-actor-system = on
 
+    # The timeout that will be used when calling .close on an ActorSystem
```
He-Pin: add `since 1.3.0`

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
mdedetrich: This is the implementation, the parent which defines the abstract method (`ActorSystem`) has `@throws(classOf[TimeoutException])` which is what matters for users, I think thats enough?

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
mdedetrich: Same as https://github.com/apache/pekko/pull/2486#discussion_r2521369096, this is the implementation and end users won't see it (they will see the abstract method)

mdedetrich: Same as https://github.com/apache/pekko/pull/2486#discussion_r2521369096, this is not part of public API

actor/src/main/resources/reference.conf
mdedetrich: Same as https://github.com/apache/pekko/pull/2486#discussion_r2521369096, this is not part of public API

mdedetrich:
> btw, @mdedetrich what's the status of kotlin dsl of Pekko?

I have some good news on this front, there is an official sbt kotlin plugin by jetbrains so it should be largely trivial https://github.com/JetBrains/sbt-kotlin-plugin

The thing is that its blocked by https://github.com/apache/pekko/issues/2093 as kotlion also has its own way of handling async IO (it has suspend functions)

mdedetrich:
I will wait a few days to see if @raboof responds and also his views on making the blocking nature configurable

actor/src/main/resources/reference.conf
mdedetrich: 1Day is way too much, people calling `close()` will expect it to do so in a reasonable time especially if they are doing it before a pod/server/instance is shutting down.

1 minute is a good default in my view

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/ActorSystemStub.scala
```diff
@@ -113,6 +113,10 @@ import com.typesafe.config.{ Config, ConfigFactory }
   override def terminate(): Unit = terminationPromise.trySuccess(Done)
   override def whenTerminated: Future[Done] = terminationPromise.future
   override def getWhenTerminated: CompletionStage[Done] = whenTerminated.asJava
+  override def close(): Unit = {
+    terminate()
+    Await.result(whenTerminated, scala.concurrent.duration.Duration.Inf)
```
pjfanning, line 118: Not a big deal but could this use the timeout config discussed elsewhere?
I've never been a big fan of tests that potentially wait indefinitely, eg in CI runs.

mdedetrich: Actually in quite a few places we have `Await.result(whenTerminated, scala.concurrent.duration.Duration.Inf)`, not just here where I introduced it. I can do it, but this method isn't even executed/called so its quite benign and from what I can tell the purpose of this `ActorSystemStub` is to not use any config in order to make it easier to programmatically setup test scenarios.

I did look for any kind of timeout that can be used, but there isn't any in this context.

mdedetrich:
I have just updated the PR to actually add tests to make sure the `AutoCloseable` works as expected. There is a Java test that uses try-with-resources and there is a Scala test using `scala.util.Using` (which automatically uses `AutoCloseable`).

One note is that when I cherry pick this into Pekko 1.3.x, the `scala.util.Using` tests will only work for Scala 2.13/Scala 3 so I will need to modify the tests there.

actor/src/main/resources/reference.conf
mdedetrich: Do we have an example of what this is meant to look like?

He-Pin:
@mdedetrich feel free to merge this and send a backport to 1.3.x

He-Pin:
I will do the backporting

mdedetrich:
> I will do the backporting

Please let me do it as I have to modify a test

He-Pin:
@mdedetrich Nice you are online, hands over~

PR #2490, opened 2025-11-13
use switch in OnMessageIntroTest

pjfanning:
In main branch we can use Java 17 code (don't need to worry about Java 8 compilation)


pjfanning:
ok - that pattern matching switch was only preview in Java 17 - need Java 21 for full support

PR #2492, opened 2025-11-13
Update OnMessageIntroTest.java

pjfanning:
* copies over some of the changes from #2491 
* we still need to compile in Java 8 in 1.3.x but this brings some of the comments and commented out code examples up to date

### DIFF
--- a/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/OnMessageIntroTest.java
+++ b/actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/OnMessageIntroTest.java
@@ -137,7 +137,7 @@ public interface OnMessageIntroTest {
          *     return onPublishSessionMessage(psm);
          *   }
          *
-         * Java 21 onward: JEP 441 (https://openjdk.org/jeps/441 =>
+         * Java 21 onward: JEP 441 (https://openjdk.org/jeps/441) =>
          // #chatroom-behavior
         // uses Java 21-onward features
         switch(msg) {


He-Pin (approved):

actor-typed-tests/src/test/java/jdocs/org/apache/pekko/typed/OnMessageIntroTest.java
```diff
@@ -137,26 +137,19 @@ public Behavior<RoomCommand> onMessage(RoomCommand msg) throws UnsupportedEncodi
          *     return onPublishSessionMessage(psm);
          *   }
          *
-         * Java 17 onward: JEP 406 (https://openjdk.org/jeps/406) =>
+         * Java 21 onward: JEP 441 (https://openjdk.org/jeps/441 =>
```
He-Pin: actually， this can be used in Java 17 too, just lack some features.

pjfanning: * our existing docs have a section about pattern matching in Java but say Java 21 is when it was implemented - 441 is a Java 21 JEP
* there were some preview JEPs before that but my experience is that orgs are reluctant to set the preview flag when running Java
* it just is easier to say Java 21 than to give a long spiel about when which preview JEP became available when
* anyone who has enabled preview on their JVM will just know anyway - no need to confuse everyone else

He-Pin: that's true, just jump to Java 21/25

PR #2493, opened 2025-11-14
make Java sample more modern by using sealed and record

pjfanning:
now that main branch does not have to support Java 17


raboof (approved):
nice!

PR #2494, opened 2025-11-14
rename UnpersistentBehavior as PersistenceProbeBehavior

pjfanning:
new classes introduced in #2456

I don't think the class name makes sense in English.

The new name was chosen because I think it works better in this doc:

>  Unit testing of `EventSourcedBehavior` can be performed by converting it into an @apidoc[PersistenceProbeBehavior]. Instead of
> persisting events and snapshots, the `PersistenceProbeBehavior` exposes @apidoc[PersistenceProbe]s for events and snapshots which
> can be asserted on.


He-Pin:
We should backport this to 1.3.0 once get merged

PR #2495, opened 2025-11-15
Add OpenJDK 25 to supported JDK versions

pjfanning:


He-Pin:
We need Add  Java 25 to nightly build then

pjfanning:
> We need Add Java 25 to nightly build then

@He-Pin see #1887

PR #2507, opened 2025-11-15
test pekko 1.3 with java 25

pjfanning:
scala 2.12.20 does not support java 25 (thus the exclude)


He-Pin:
That's because we have `inline` optimizer enabled

He-Pin:
https://contributors.scala-lang.org/t/scala-2-12-21-release-planning/6753/10

2.12.21 support java 25

PR #2511, opened 2025-11-17
handle build issue in OSGi build code

### LINKED ISSUE #2509: OSGI build throws an exception if version value in sbt is not in x.y.z format
The change that caused this brittleness is associated with https://github.com/apache/pekko/issues/2312

```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at OSGi$.pekkoImport(OSGi.scala:147)
	at OSGi$.$anonfun$exports$1(OSGi.scala:136)
	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
	at sbt.internal.util.EvaluateSettings$MixedNode.evaluate0(INode.scala:229)
	at sbt.internal.util.EvaluateSettings$INode.evaluate(INode.scala:171)
```

The version is derived by sbt-dynver based on git tags. We might have a git tag that is not a version number. With our source release tarballs, there is no git at all, so the derived version is based a timestamp.

We should change OSGi.scala so that it has a failover strategy if the string split based on dots does not give us an array with enough elements.

pjfanning:
fixes #2509 

Not having a semver version and having sbt-dynver derive versions like HEAD+20251117-1039 when the code is not under git control leads to other problems with the build but this issue breaks the build.

Other issues that we get with HEAD+20251117-1039 style versions include:
```
[error] bnd: Invalid value for Bundle-Version, HEAD+20251117-1039 does not match \d{1,9}(\.\d{1,9}(\.\d{1,9}(\.[-\w]+)?)?)?
```

### DIFF
--- a/project/OSGi.scala
+++ b/project/OSGi.scala
@@ -137,6 +137,8 @@ object OSGi {
       "!scala.util.parsing.*",
       scalaImport(scalaVersion))
   def pekkoImport(version: String, packageName: String = "org.apache.pekko.*") = {
+    // see https://github.com/apache/pekko/issues/2509 for background
+    // on why the version that is input might not be in the expected format
     val versionComponents = version.split('.')
     if (versionComponents.length < 2) {
       useDefaultImportRange(version, packageName)


project/OSGi.scala
```diff
@@ -138,9 +138,22 @@ object OSGi {
       scalaImport(scalaVersion))
   def pekkoImport(version: String, packageName: String = "org.apache.pekko.*") = {
     val versionComponents = version.split('.')
-    val nextMinorVersion = versionComponents(1).toInt + 1
-    versionedImport(packageName, s"${versionComponents.head}.${versionComponents(1)}",
-      s"${versionComponents.head}.$nextMinorVersion")
+    if (versionComponents.length < 2) {
```
mdedetrich, line 143: Can we added a comment here explaining the context of what's going on (I.e. summary of PR)

pjfanning: is it ok if I add a link to https://github.com/apache/pekko/issues/2509 as the comment? - it's easier to keep the GitHub description updated than to potentially need PRs to update the comments

He-Pin: I think so.

He-Pin (approved):
lgtm

project/OSGi.scala
```diff
@@ -137,10 +137,25 @@ object OSGi {
...
+        versionedImport(packageName, s"${versionComponents.head}.${versionComponents(1)}",
+          s"${versionComponents.head}.$nextMinorVersion")
+      } catch {
+        case _: NumberFormatException =>
+          useDefaultImportRange(version, packageName)
+      }
+    }
+  }
+  private def useDefaultImportRange(version: String, packageName: String) = {
+    System.err.println(s"Invalid Pekko version: $version for OSGi import package versioning, defaulting to [0.0,10.0)")
```
He-Pin, line 157: nice to use `err` here

PR #2512, opened 2025-11-17
Delete KEYS

pjfanning:
The KEYS should be retrieved from
https://downloads.apache.org/pekko/KEYS


raboof:
I kinda like having the 'source of truth' for the KEYS here (so it is publicly version-controlled), and have https://downloads.apache.org/pekko/KEYS just follow this one. I know the KEYS in downloads are probably technically also version-controlled, but that's less transparent.

PR #2518, opened 2025-11-18
Use duration converters for close()

mdedetrich:
(cherry picked from commit 29b7d22e29625a59e3424f5352e4cd7c3b59d984)

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
@@ -41,6 +41,7 @@ import pekko.japi.Util.immutableSeq
 import pekko.serialization.SerializationExtension
 import pekko.util._
 import pekko.util.FutureConverters._
+import pekko.util.JavaDurationConverters._
 import pekko.util.OptionConverters._
 import pekko.util.Helpers.toRootLowerCase
 import pekko.util.ccompat.JavaConverters._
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
```diff
@@ -23,6 +23,7 @@ import scala.collection.immutable
 import scala.concurrent.{ Await, ExecutionContext, ExecutionContextExecutor, Future, Promise }
 import scala.concurrent.blocking
 import scala.concurrent.duration.Duration
+import scala.jdk.DurationConverters._
```
pjfanning: * we need to use pekko util because of Scala 2.12 res in 1.x
* is this a big win for 1.x?

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
```diff
@@ -40,6 +40,7 @@ import pekko.event.Logging.DefaultLogger
 import pekko.japi.Util.immutableSeq
 import pekko.serialization.SerializationExtension
 import pekko.util._
+import org.apache.pekko.util.JavaDurationConverters._
```
pjfanning: drop the org.apache bit

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
mdedetrich: Fixed and pushed

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
mdedetrich: Fixed and pushed

actor/src/main/scala/org/apache/pekko/actor/ActorSystem.scala
mdedetrich: > is this a big win for 1.x?

I would like to make sure that currently maintained branches have the same behaviour/code

PR #2532, opened 2025-11-25
Release notes for 2.0.0-M1

raboof:
Kept it fairly short, I think it doesn't make sense to have an exhaustive list but we should collect some of the highlights when we get closer to a non-milestone 2.0.0 release.

### DIFF
--- a/docs/src/main/paradox/release-notes/releases-2.0.md
+++ b/docs/src/main/paradox/release-notes/releases-2.0.md
@@ -1,0 +1,18 @@
+# Release Notes (2.0.x)
+
+Apache Pekko 2.0.x releases support Java 17 and above.
+
+## 2.0.0-M1
+
+This is milestone release and is aimed at testing this new major version
+by early adopters. This is experimental. This release should not be used in production.
+
+### Main changes
+
+* Java 17 is the new minimum
+* Scala 2.12 support dropped
+* A lot of deprecated code removed
+* A lot of pekko.util classes for Scala version compatibility have been removed
+* Big change for all Java DSL users due to the removal of `pekko.japi.Function` (and related classes) to use `pekko.japi.function.Function` instead (lambdas should recompile ok but if you declared variables or functions explicitly, then you may need to change your imports)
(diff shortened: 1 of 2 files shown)


mdedetrich:
I assume that this is still in draft stage as there are some final things to be completed before 2.0.0-M1?

raboof:
> I assume that this is still in draft stage as there are some final things to be completed before 2.0.0-M1?

The release notes don't need to be exhaustive, and don't necessarily need to be the _last_ thing thing we merge before cutting the release, so I think this can be reviewed/merged already even if there's some other things that are also still under consideration.

pjfanning:
We generally only merge after the release vote but these seem fine to me

raboof:
> We generally only merge after the release vote but these seem fine to me

Ah, I understood from https://github.com/apache/pekko-site/wiki/Pekko-Release-Process#checklist that we wanted to merge the release notes before tagging the release (and thus before voting on it) - I'm fine with merging release notes after the vote, will update the process.

docs/src/main/paradox/release-notes/releases-2.0.md
```diff
@@ -0,0 +1,17 @@
...
+This is milestone release and is aimed at testing this new major version
+by early adopters. This is experimental. This release should not be used in production.
+
+### Main changes
+
+* Java 17 is the new minimum
+* Scala 2.12 support dropped
+* A lot of deprecated code removed
+* A lot of pekko.util classes for Scala version compatibility have been removed
+* Big change for all Java DSL users due to the removal of `pekko.japi.Function` (and related classes) to use `pekko.japi.function.Function` instead (lambdas should recompile ok but if you declared variables or functions explicitly, then you may need to change your imports)
```
pjfanning, line 16: A couple of additional changes that we can mention.

* New pekko-serialization-jackson3 lib. Users who are happy with the pekko-serialization-jackson, which uses Jackson 2, can stick with that lib.
* Changed the pekko-serialization-jackson lz4-java dependency to `at.yawk.lz4:lz4-java`, a fork that has important bug fixes.

PR #2544, opened 2025-12-03
chore: test Scala 3 with `validatePullRequest`

raboof:
The Scala 3 build+test job takes almost an hour because it runs all tests, while the Scala 2 build+test job is sometimes much faster because it runs with `validatePullRequest` which selects which tests to run based on the files that were changed. I don't think we've regularly seen situations where a failure was missed by `validatePullRequest` and caught by the Scala 3 test, so it might be nice to run those with `validatePullRequest` as well.


.github/workflows/build-test-prValidation.yml
```diff
@@ -57,6 +57,9 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+    strategy:
+      matrix:
+        scalaVersion: ["2.12.x", "3.3.x"]
```
pjfanning: 2.13.x

raboof: woops

PR #2549, opened 2025-12-04
release notes for 1.4.0

pjfanning:
release vote passed


He-Pin:
I want to add some performance optimization around the stream this weekend, is that possible?

pjfanning:
> I want to add some performance optimization around the stream this weekend, is that possible?

Let's do a 1.5.0 later. There is a reason to get this out - discussed on our mailing lists.

PR #2556, opened 2025-12-06
chore: Optimize Source.from iterable.

He-Pin:
Optimize it when the size is little.

### DIFF
--- a/stream-tests/src/test/java/org/apache/pekko/stream/javadsl/SourceTest.java
+++ b/stream-tests/src/test/java/org/apache/pekko/stream/javadsl/SourceTest.java
@@ -85,6 +85,21 @@ public class SourceTest extends StreamTest {
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


He-Pin:
fixed: https://github.com/apache/pekko/issues/2558

He-Pin:
The current code will miss the `x :: Nil` case, so I upstream the changes to scala
https://github.com/scala/scala/pull/11188

mdedetrich:
> The current code will miss the `x :: Nil` case, so I upstream the changes to scala [scala/scala#11188](https://github.com/scala/scala/pull/11188)

You will also have to do Scala 2.12 and also Scala 3 (aka dotty)

He-Pin:
@mdedetrich No such method in Scala 2.12:) and Scala 3 will sync the stdlib from 2.13.x

PR #2560, opened 2025-12-07
chore: optimize Source#future and Source#futureSource

He-Pin:
Motivation:
The TraversalBuilder has some optimization around the `Single` and `Empty` source. so convert them it single/empty if possible will help performance.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
@@ -589,16 +589,11 @@ object Source {
    * Emits a single value when the given `Future` is successfully completed and then completes the stream.
    * The stream fails if the `Future` is completed with a failure.
    */
-  def future[T](futureElement: Future[T]): Source[T, NotUsed] = {
-    val maybeValue = futureElement.value
-    if (maybeValue eq None) {
-      fromGraph(new FutureSource[T](futureElement))
-    } else maybeValue match {
-      case Some(scala.util.Success(null)) => empty[T]
-      case Some(scala.util.Success(elem)) => single(elem)
-      case Some(scala.util.Failure(ex))   => failed[T](ex)
-      case _                              => throw new IllegalStateException("unexpected case") // will not happen, for exhaustiveness check
-    }
+  def future[T](futureElement: Future[T]): Source[T, NotUsed] = futureElement.value match {
+    case None                           => fromGraph(new FutureSource[T](futureElement))
+    case Some(scala.util.Success(null)) => empty[T]
+    case Some(scala.util.Success(elem)) => single(elem)
+    case Some(scala.util.Failure(ex))   => failed[T](ex)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


He-Pin:
@mdedetrich @pjfanning @raboof Would you like to give a review on this, thanks.

pjfanning:
I'm still suspicious about this and other recent performance oriented changes. They introduce more checks which reduces performance. These checks include instanceof checks which are not trivial. The performance gains are not proven. No microbenchmarks, etc.

He-Pin:
This primarily compares the process to materializing an Actor. Although some checks are added here, it essentially involves only a few bytecode instructions, making it significantly lighter than actually materializing an Actor. At the same time, simplifying a Stage to Empty or Single provides more optimization opportunities within the TraversalBuilder.

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
```diff
@@ -567,8 +561,12 @@ object Source {
    * Emits a single value when the given `Future` is successfully completed and then completes the stream.
    * The stream fails if the `Future` is completed with a failure.
    */
-  def future[T](futureElement: Future[T]): Source[T, NotUsed] =
-    fromGraph(new FutureSource[T](futureElement))
+  def future[T](futureElement: Future[T]): Source[T, NotUsed] = futureElement.value match {
+    case Some(scala.util.Success(null)) => empty[T]
```
pjfanning, line 594: Would be better to check once for if the future has already completed? If it has then check the various success cases. Also, how about checking if it's already failed?

pjfanning: The Scala compiler might produce an optimized if or switch but I'd prefer if the code here explicitly checked for future element.value == None first as this is the most likely result. Checking this last could lead to unnecessary extra checks.

He-Pin: Wow, Nice!

He-Pin:
@pjfanning I have updated with your suggestion

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
```diff
@@ -589,8 +589,17 @@ object Source {
    * Emits a single value when the given `Future` is successfully completed and then completes the stream.
    * The stream fails if the `Future` is completed with a failure.
    */
-  def future[T](futureElement: Future[T]): Source[T, NotUsed] =
-    fromGraph(new FutureSource[T](futureElement))
+  def future[T](futureElement: Future[T]): Source[T, NotUsed] = {
+    val maybeValue = futureElement.value
+    if (maybeValue eq None) {
+      fromGraph(new FutureSource[T](futureElement))
+    } else futureElement.value match {
```
pjfanning: Use maybeValue param

He-Pin: Opps，would you push directly on this,I am away from computure, thanks

mdedetrich:
> I'm still suspicious about this and other recent performance oriented changes. They introduce more checks which reduces performance. These checks include instanceof checks which are not trivial. The performance gains are not proven. No microbenchmarks, etc.

I haven't looked into the details of this PR, but from what I can tell its just doing specialization which in general is fine premise especially with data structures that time time to initialize/have an expensive initial cost. Having 0/1 sized streams is very common so this can provide real benefits

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
He-Pin: I  just found I should write all in pattern matching, instead of this wired if else.

He-Pin:
I just found I should have written it all in pattern matching, instead of this wired if else one....

pjfanning (approved):
lgtm

PR #2561, opened 2025-12-08
chore: Matching on seq

He-Pin:
The pr against Scala is rejected, so I need to handle it here.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
@@ -402,13 +402,15 @@ object Source {
    * stream will see an individual flow of elements (always starting from the
    * beginning) regardless of when they subscribed.
    */
-  def apply[T](iterable: immutable.Iterable[T]): Source[T, NotUsed] = {
-    (iterable.knownSize: @switch) match {
-      case 0 => empty
-      case 1 => single(iterable.head)
-      case _ =>
-        fromGraph(new IterableSource[T](iterable)).withAttributes(DefaultAttributes.iterableSource)
-    }
+  def apply[T](iterable: immutable.Iterable[T]): Source[T, NotUsed] = iterable match {
+    case immutable.Seq()                   => empty[T]
+    case immutable.Seq(elem: T @unchecked) => single(elem)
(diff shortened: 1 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
```diff
@@ -402,13 +402,15 @@ object Source {
...
    */
-  def apply[T](iterable: immutable.Iterable[T]): Source[T, NotUsed] = {
-    (iterable.knownSize: @switch) match {
-      case 0 => empty
-      case 1 => single(iterable.head)
-      case _ =>
-        fromGraph(new IterableSource[T](iterable)).withAttributes(DefaultAttributes.iterableSource)
-    }
+  def apply[T](iterable: immutable.Iterable[T]): Source[T, NotUsed] = iterable match {
+    case immutable.Seq()                   => empty[T]
```
pjfanning, line 406: I don't see how adding extra checks, including instanceof checks, improves performance. Is there any way that we could keep a variant of the old code from the recent other PR that introduced changes?

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
```diff
@@ -402,13 +402,15 @@ object Source {
...
+    case immutable.Seq(elem: T @unchecked) => single(elem)
+    case other                             => (other.knownSize: @switch) match {
```
pjfanning, line 408: Can you add a comment saying that if the Iterable has unknown size, that knownSize returns -1?

He-Pin: It's not unknown size,just the size can or not been cheaply cimputed

stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala
He-Pin: It's vs the materialization

pjfanning:
https://github.com/apache/pekko/pull/2562 is merged

He-Pin:
@pjfanning Cool~

PR #2562, opened 2025-12-08
add Source.apply that is optimised for Seq

pjfanning:
an alternative to #2561


He-Pin:
1. We need add testing about overloading
2. The set with one element should be optimized to a single

PR #2566, opened 2025-12-10
docs: document the Agrona update may need `--add-opens`

raboof:
Fixes https://github.com/apache/pekko-http/issues/887


He-Pin:
We can write a JPMSUtils like thing to read from a config file too, but that will complicate things

PR #2583, opened 2025-12-14
chore: optimize Source#future and Source#futureSource (#2560)

He-Pin:
* chore: optimize Source#future and Source#futureSource

---------

Co-authored-by: PJ Fanning <pjfanning@users.noreply.github.com>
(cherry picked from commit 01f2265fb2d69e4e82924680740d1b0e3e212f10)

PR #2584, opened 2025-12-14
chore: Remove CollectionUtil (#2582)

He-Pin:
(cherry picked from commit f735cca1a5dea740ea79ea643b9470991979cedd)


He-Pin:
@pjfanning This is ready for 1.5.x

pjfanning:
I can half understand backporting #2583 but I don't see any benefit to backporting this.

pjfanning:
In particular, I would be worried that some pekko module uses CollectionUtil. We have broken releases before by removing classes.
So removing any classes in 1.x is a no for me.
-1.

If I get outvoted on this, the one thing that is a blocker for me is removing the class. We could backport the code changes without deleting the class.

But my preference is not to bother. 2.0.0 will be finished faster if we don't waste so much time on 1.x.

He-Pin:
OK isee

Issue #2563, opened 2025-12-08
Make use of jspecify


He-Pin:
https://github.com/reactor/reactor-core/releases/tag/v3.8.0

Just migrated to jspecify, it helps javadsl.

PR #2617, opened 2026-01-10
add jspecify nullable annotations on streams Java DSL

pjfanning:
* linked to #2563
* probably best to stick with Java DSL only
* can look at other libs later

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -573,6 +573,7 @@ lazy val streamTyped = pekkoModule("stream-typed")
     streamTestkit % "test->test",
     actorTestkitTyped % "test->test",
     actorTypedTests % "test->test")
+  .settings(Dependencies.streamTyped)
   .settings(AutomaticModuleName.settings("pekko.stream.typed"))
   .enablePlugins(ScaladocNoVerificationOfDiagrams)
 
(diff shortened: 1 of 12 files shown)


He-Pin (approved):
lgtm

He-Pin (commented):
Good addition -- JSpecify annotations on the Java DSL will help IDE static analysis and help Java users avoid NPEs.

**Observations:**

1. The `@Nullable` annotations are applied consistently to parameters that can accept null (like `rest`, `those`, `log` parameters). This is correct.

2. The `package-info.java` with `@NullMarked` is the right approach -- it makes the entire Java DSL package null-safe by default, with explicit `@Nullable` overrides where needed.

3. JSpecify is added as an `Optional` dependency, which means it won't be required at runtime. This is correct since the annotations are only needed at compile time.

4. One thing to consider: this PR only covers the stream Java DSL. The PR description mentions 'probably best to stick with Java DSL only' -- but the actor and other Java DSLs would benefit from the same treatment. Consider filing a follow-up issue for the remaining modules.

No blocking issues.

pjfanning:
@He-Pin if we don't see any issues caused by this, I'll look into adding similar changes to other modules.

PR #2666, opened 2026-02-16
docs: replace ACM link with authors' site

raboof:
While ACM seems to be one of the less-awful publishers, still nice to link to the 'actual source'. And is fixes the link checker as ACM seems to 403 when accessed from it.


pjfanning (approved):
lgtm

PR #2631, opened 2026-01-19
fix: Fix recoverWith on Failed stage.

He-Pin:
Motivation:
Fix https://github.com/apache/pekko/issues/2620

@kstokoz fyi

PR #2674, opened 2026-02-23
Revert recent changes to recoverWith

pjfanning:
see #2620 and https://github.com/apache/pekko-http/issues/954

This keeps tests and doc changes but reverts the Ops.scala changes from #1775 and #2631


pjfanning:
Thanks @mdedetrich - it would be my aim to get this backported to 1.5 and appear in the next 1.x release. I agree that on the main branch that we can try to improve the code again but we do need to improve the test coverage to avoid breaking pekko-http http2 tests again. I hope to find some time this week to try to find specifically what is going wrong in pekko-http with the existing changes. I suspect that it is one of the 2 usages of recoverWithRetries that is being affected.
https://github.com/search?q=repo%3Aapache%2Fpekko-http%20recoverWithRetries&type=code

pjfanning:
I will revert this if there are objections but in the short term, my aim is to fix the pekko-http build - https://github.com/apache/pekko-http/issues/954

PR #2676, opened 2026-02-24
upgrade jackson, netty, protobuf-java

pjfanning:
* jackson 2.21.1
* protobuf-java 4.33.5
* netty 4.2.10
* config 1.4.6

all in use on main branch and show no issues - these are not big upgrades to what we already have in the branch but mean that we pick up latest bug fixes

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.11.7
+sbt.version=1.12.3


pjfanning:
SessionRenegotiationFirstOne tests failing - they fail intermittently and its possible that a Java upgrade could lead to the cipher suites being no longer supported.

I will investigate but I think it is unrelated to this PR. The tests are failing in another PR today too.

PR #2689, opened 2026-02-27
Create stage-release-candidate.yml

pjfanning:
* copied from main branch but we need to have the yml in the 1.5.x if we want to run it for that branch
* slightly modified to get the build instructions for the right branch

### DIFF
--- a/.github/workflows/stage-release-candidate.yml
+++ b/.github/workflows/stage-release-candidate.yml
@@ -67,7 +67,6 @@ jobs:
           fetch-depth: 0
           fetch-tags: true
           persist-credentials: false
-          ref: ${{ github.ref_name }}
 
       - name: Generate source archive
         run: |-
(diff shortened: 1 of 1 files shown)


raboof:
> * copied from main branch but we need to have the yml in the 1.5.x if we want to run it for that branch

:+1: 

>     * slightly modified to use ref in the checkout action and to get the build instructions for the right branch

Are you sure this is needed? Since we manually dispatch this workflow, we can select the exact tag we want the workflow to run for (per https://github.com/apache/pekko-site/wiki/Pekko-Release-Process-%E2%80%90-ATR-CI%E2%80%90assisted-release) which AFAICT the checkout action should pick up?

pjfanning:
Let me try testing if we need this at all for 1.x releases. One useful enhancement to the main yml would be to have it derive the branch name for the build instructions URL.

pjfanning:
@raboof I removed the ref change to the checkout action so this now matches what is in main branch
The only diff is that I fixed the URL for the build instructions.

You can't run the workflow if it isn't in the code marked by the git tag. I tried this and GitHub wouldn't let me run the workflow for a x1.5.0-RC0 tag.

PR #2690, opened 2026-02-27
Fix branch reference in stage-release-candidate.yml

pjfanning:
Updated the branch reference in the release candidate workflow to use the dynamic branch variable.

Also, explicitly say 72hrs because the token does not seem to be replaced.


pjfanning:
I can test this as part of prepping the 1.5.0 release. Hard to test without actually running the workflow.

pjfanning:
I'm going to merge this to 1.5.x branch because that is where I am testing for the upcoming 1.5.0 release.

I can forward fit the changes to main branch when they look like they work ok.

PR #2694, opened 2026-02-28
bring back most of the recoverWith changes

pjfanning:
* aimed at 2.0.0 only - it doesn't seem worth the risk with breaking stuff on 1.x
* omits the failure handling that was added before as this seems to be what caused the http2 testing issues in pekko-http
* I have tested this with pekko-http locally
* see #2674
* the original changes were by @He-Pin - I have just removed some parts that seemed to be problematic


He-Pin:
thanks ,I was planning looking at it tomorrow,  I will.check it in detail. 

The problem is we don't have a test to cover the edge cases...

pjfanning:
I'd prefer to add the changes back incrementally but any enhancement to the test cases is definitely a great start

pjfanning:
#2712 covers this

PR #2710, opened 2026-03-02
Unpin jackson-annotations

Philippus:
Jackson-annotations is unmoored from jackson-core from version 2.20, see https://github.com/FasterXML/jackson-annotations?tab=readme-ov-file#release-notes.


pjfanning:
thanks - merged

PR #2712, opened 2026-03-04
fix: Fix recoverWith with add missing attempt back.

He-Pin:
Motivation:
It was reverted and there is a bug in that code.

refs: https://github.com/apache/pekko/issues/2620
refs: https://github.com/apache/pekko/pull/2694

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
@@ -2190,20 +2190,20 @@ private[pekko] object TakeWithin {
                           attempt += 1
                           onFailure(ex)
                         case None =>
-                          switchTo(source)
                           attempt += 1
+                          switchTo(source)
                       }
                     case iterableSource: IterableSource[T @unchecked] =>
                       emitMultiple(out, iterableSource.elements, () => completeStage())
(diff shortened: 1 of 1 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala
```diff
@@ -2181,11 +2181,15 @@ private[pekko] object TakeWithin {
               TraversalBuilder.getValuePresentedSource(source) match {
                 case OptionVal.Some(graph) => graph match {
                     case singleSource: SingleSource[T @unchecked] => emit(out, singleSource.elem, () => completeStage())
-                    case failed: FailedSource[T @unchecked]       => onFailure(failed.failure)
+                    case failed: FailedSource[T @unchecked]       =>
+                      attempt += 1
```
He-Pin, line 2185: missed the `attempt += 1`

pjfanning (approved):
lgtm

I tried this on my laptop with the pekko-http tests that broke before and they work with this version.

I would appreciate if we just keep this as 2.0.0 change because I don't want to take any more risks with the 1.x releases.

He-Pin:
Yeah，let's just keep it in 2.0.0， thanks for the testing

He-Pin:
It was spring festival, sorry for the delay

PR #2726, opened 2026-03-13
fix(stream): Source.combine single source with type-transforming fan-in strategies

### LINKED ISSUE #2723: Source.combine and MergeLatest don't work correctly for a single source
`Stream.combine` combined with `MergeLatest` does not work as expected if a list with exactly one source is passed to it.

## Example

`Source.combine(Seq(Source.single(1)))(MergeLatest(_))` emits a single value `1`, instead of the expected `List(1)`.

## Cause

There is a special handling for a list with exactly one source:
https://github.com/apache/pekko/blob/66d7dc19464c1e082e759f0595cd8976249b7b00/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Source.scala#L806

This works for something like `Merge`, but not `MergeLatest`.

He-Pin:
### Motivation
Fixes #2723. `Source.combine(Seq(Source.single(1)))(MergeLatest(_))` emits `1` instead of `List(1)` because the single-source case bypasses the fan-in strategy with an unsafe `asInstanceOf` cast. The same bug exists in `Sink.combine`.

### Approach: TypePreserving marker traits

Instead of the fragile try-catch approach, this PR introduces two **marker traits**:

- **`TypePreservingFanIn`** — for fan-in stages where output type == input type (T → T)
- **`TypePreservingFanOut`** — for fan-out stages where output type == input type (T → T)

In `Source.combine` / `Sink.combine`, for the single-source/sink case:
1. Instantiate the strategy with `fanInStrategy(1)` / `fanOutStrategy(1)`
2. Check if the strategy has the `TypePreservingFanIn` / `TypePreservingFanOut` trait
3. **If yes** → bypass (safe because T == U)
4. **If no** → route through the strategy graph (ensures correct type transformation)

This design uses a **"safe default"**: unknown/third-party strategies without the trait always go through the full graph, preventing silent type errors.

### Stages marked

**Fan-in (TypePreservingFanIn):**
- `Merge`, `MergePrioritized` — bypass fires via factory methods
- `Concat`, `Interleave`, `MergeSequence`, `OrElse` — trait present on class but factory methods wrap via `withDetachedInputs`, so single-source cases route through the graph (correct, just slightly less optimal)

**Fan-out (TypePreservingFanOut):**
- `Broadcast`, `Balance` — bypass fires via factory methods
- `Partition` intentionally NOT marked — its `partitioner` function provides user-specified routing semantics that would be lost if bypassed

**NOT marked (type-transforming):**
- `MergeLatest` (T → List[T]), `ZipWithN` (A → O)

### Additional changes
- `Concat`, `Interleave`, `MergeSequence`: relaxed `require(inputPorts > 1)` to `require(inputPorts >= 1)` — single-input is semantically valid (pass-through) and enables routing through the graph for single-source cases
- `Source.fromGraph` / `Sink.fromGraph` used in bypass path for safe handling of non-Source/non-Sink `Graph` inputs

### Tests
- **12 Scala SourceSpec tests**: MergeLatest, ZipWithN, Merge, Concat, Interleave, MergeSequence, MergePrioritized, wrapped stages (.named), materialized values, empty sources
- **4 Scala SinkSpec tests**: Broadcast, Balance, Partition (routes through strategy), wrapped Broadcast (.named)
- **1 Java SourceTest**: MergeLatest single source

### Cross-review
This fix has been through **3 rounds of cross-review** with 5+ agents (GPT-5.4, Gemini 3 Pro, Claude Opus 4.6 ×2, Claude Sonnet 4.6), plus expert panel discussion. Key decisions validated:
- Safe default design (WON'T FIX: intentional behavioral change for third-party strategies)
- Partition excluded from TypePreservingFanOut (SHOULD FIX → fixed)
- Scaladoc updated to clarify factory method wrapping behavior (MUST FIX → fixed)

### Binary Compatibility
- Adding marker traits to `final` classes is binary compatible (empty interface, no method changes)
- Adding new public traits is additive/backward compatible
- `require` relaxation is behavioral, not binary

### DIFF
--- a/stream-tests/src/test/java/org/apache/pekko/stream/javadsl/SourceTest.java
+++ b/stream-tests/src/test/java/org/apache/pekko/stream/javadsl/SourceTest.java
@@ -1156,6 +1156,21 @@ public class SourceTest extends StreamTest {
     assertEquals(6, result.toCompletableFuture().get(3, TimeUnit.SECONDS).intValue());
   }
 
+  // Regression test for https://github.com/apache/pekko/issues/2723
+  // Verifies that Source.combine with a single source correctly applies
+  // type-transforming strategies (like MergeLatest), rather than bypassing
+  // them with an unsafe asInstanceOf cast.
+  @Test
+  public void mustBeAbleToCombineSingleSourceWithMergeLatest() throws Exception {
+    final List<Source<Integer, NotUsed>> sources = Collections.singletonList(Source.single(1));
+    final List<List<Integer>> result =
+        Source.<Integer, List<Integer>, NotUsed>combine(sources, MergeLatest::create)
+            .runWith(Sink.collect(Collectors.toList()), system)
+            .toCompletableFuture()
+            .get(3, TimeUnit.SECONDS);
+    assertEquals(Collections.singletonList(Collections.singletonList(1)), result);
+  }
+
   @SuppressWarnings("unchecked")
   @Test
   public void mustBeAbleToZipN() throws Exception {
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/SinkSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/SinkSpec.scala
@@ -25,6 +25,8 @@ import pekko.stream.testkit._
 import pekko.stream.testkit.scaladsl.{ TestSink, TestSource }
 import pekko.testkit.DefaultTimeout
 
+import scala.collection.immutable
+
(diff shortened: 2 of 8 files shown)


He-Pin:
I will try to add a trait to see if I can come up a better result

pjfanning:
@He-Pin this is marked as Draft. Is it ready to merge? I'd like back port this to the 1.5.0 RC.

He-Pin:
Not really I will come up with a better design

He-Pin:
It's now better with a marker trait

Issue #2730, opened 2026-03-16
port some Akka 2.8.0 changes to Pekko


pjfanning:
https://github.com/akka/akka-core/releases/tag/v2.8.0

These Akka 2.8.0 changes are now Apache licensed.

We probably don't want to copy over everything, just for the sake of it. Bug fixes should be the priority.

Philippus:
https://github.com/apache/pekko/pull/2736

Philippus:
https://github.com/apache/pekko/pull/2737

He-Pin:
I will do the last check, leeve these to you:)

pjfanning:
I think this is done

PR #2733, opened 2026-03-16
Port CRDT key expiry from akka-core #31721

pjfanning:
Ports the CRDT key expiry feature from [akka/akka-core#31721](https://github.com/akka/akka-core/pull/31721) with imports updated for Pekko. Allows `Replicator` entries to be automatically removed (without tombstones) after a configurable period of inactivity.

part of #2730 

## Configuration

```hocon
pekko.cluster.distributed-data.expire-keys-after-inactivity {
  "session-*" = 30 minutes
  "cache-key" = 5 minutes
}
```

Prefix wildcard (`*`) matching supported.

### Classic API (`distributed-data`)
- New `Expired[A]` subscribe response (alongside `Changed`/`Deleted`) sent to subscribers when a key is evicted
- `ReplicatorSettings` gains `expiryKeys: Map[KeyId, FiniteDuration]` with backward-compat constructor, `withExpiryKeys()` Scala/Java API, and `parseExpiry()` helper
- `dataEntries` extended from `(DataEnvelope, Digest)` to `(DataEnvelope, Digest, Timestamp)` to track last-used time per key
- Timestamp updated on `receiveGet`, `receiveUpdate`, `receiveDelete`
- `cleanupExpired()` runs on every gossip tick; expired entries removed and subscribers notified
- Gossip protocol (`Status`/`Gossip` messages) now carries `usedTimestamp` so all replicas share expiry clock state — entries received via gossip that are already expired are discarded

### Durable store
- New `Expire(keys: Set[KeyId])` message added to `DurableStore` protocol
- `LmdbDurableStore` handles `Expire` via new `dbDelete(keys)` method

### Serialization
- `ReplicatorMessages.proto`: optional `sint64 usedTimestamp = 3` added to `Status.Entry` and `Gossip.Entry`
- Generated `ReplicatorMessages.java` updated accordingly (zero value = unset; backward-compatible)
- `ReplicatorMessageSerializer` updated to read/write timestamps

### Typed API (`cluster-typed`)
- `Expired[A]` type alias added to `scaladsl.Replicator`
- `Expired[A]` case class added to `javadsl.Replicator`
- `ReplicatorBehavior` forwards classic `Expired` to Java API subscribers


He-Pin:
Should we wait this on 3.17？

pjfanning:
March 16 (today) is ok - https://github.com/akka/akka-core/blob/v2.8.0/LICENSE

PR #2750, opened 2026-03-17
Port akka-core optimizations for typed actor context and system adapter

pjfanning:
Ports two performance optimizations from [akka-core#31843](https://github.com/akka/akka-core/pull/31843) and [akka-core#31842](https://github.com/akka/akka-core/pull/31842).

Part of #2730 

### Remove redundant field in `ActorContextAdapter` (#31843)
- Drops `val classicContext` constructor parameter — it was a duplicate of `adapter.context`
- `classicActorContext` now delegates directly to `adapter.context`
- Saves one field allocation per typed actor instance
- `scheduleOnce` updated to pass executor explicitly (avoids Scala 2.13 fatal unused-import warning on method-result imports, unlike Akka's stable-`val` form)

### Cache typed system adapter in `ActorSystemImpl` (#31842)
- Adds `private[pekko] var typedSystem: OptionVal[AnyRef]` to `ActorSystemImpl` as an optimistic cache
- `ActorSystemAdapter.apply` checks this cache before going through the Extension infrastructure, cutting repeated adapter lookups on every typed actor start
- `system` in `ActorContextAdapter` changed from `final override val` → `override def` so the cached path is taken lazily rather than at construction time

```scala
// Before: AdapterExtension lookup on every call
def apply(system: classic.ActorSystem): ActorSystem[Nothing] = AdapterExtension(system).adapter

// After: O(1) cache hit after first resolution
def apply(system: classic.ActorSystem): ActorSystem[Nothing] = system match {
  case system: classic.ActorSystemImpl =>
    system.typedSystem match {
      case OptionVal.Some(typedSystem: ActorSystem[_]) => typedSystem
      case _ =>
        val typedSystem: ActorSystem[_] = AdapterExtension(system).adapter
        system.typedSystem = OptionVal.Some(typedSystem)
        typedSystem
    }
  case _ => AdapterExtension(system).adapter
}
```

The cache write is intentionally unsynchronized (same as Akka) — worst case is two threads race to set the same adapter reference.

### Test update
- `MailboxSelectorSpec` updated to use `classicActorContext` in place of the removed `classicContext` field.

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
@@ -148,7 +148,7 @@ private[pekko] object ActorSystemAdapter {
         // Optimization to cut out going through adapter lookup if possible
         system.typedSystem match {
           case OptionVal.Some(typedSystem: ActorSystem[_]) => typedSystem
-          case _ =>
+          case _                                           =>
             val typedSystem: ActorSystem[_] = AdapterExtension(system).adapter
             system.typedSystem = OptionVal.Some(typedSystem)
             typedSystem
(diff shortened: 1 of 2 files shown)


He-Pin:
This is needed, thanks.

He-Pin (approved):

actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorSystemAdapter.scala
```diff
@@ -141,7 +142,21 @@ import org.slf4j.{ Logger, LoggerFactory }
 }
 
 private[pekko] object ActorSystemAdapter {
-  def apply(system: classic.ActorSystem): ActorSystem[Nothing] = AdapterExtension(system).adapter
+  def apply(system: classic.ActorSystem): ActorSystem[Nothing] = {
+    system match {
+      case system: classic.ActorSystemImpl =>
+        // Optimization to cut out going through adapter lookup if possible
```
He-Pin, line 148: Should add // After: O(1) cache hit after first resolution comment here, which would be better

He-Pin (commented):
lgtm

PR #2757, opened 2026-03-18
chore: Rename Junit5 to JunitJupiter

He-Pin:
Will deprecate this in 1.5.x @pjfanning


pjfanning:
Can't we just deprecate them in 2.0?

PR #2758, opened 2026-03-18
Port akka-core#31777: O(1) mandatory attribute lookup via map index

pjfanning:
https://github.com/akka/akka-core/pull/31777

part of #2730 

`getMandatoryAttribute` previously performed O(n) linear traversal of the attribute list on every call — a hot path during stream materialization.

## Changes

- **`Attributes`**: Converted from `final case class` to `final class`, adding a `mandatoryAttributes: Map[Class[AnyRef], MandatoryAttribute]` field maintained alongside `attributeList` for O(1) lookup.
- **`getMandatoryAttribute`**: Now does a direct map lookup instead of linear search.
- **`and()` overloads**: Updated to merge mandatory attribute maps when combining `Attributes` instances.
- **Backward compatibility**: Added explicit `apply`/`unapply` factory methods and deprecated `copy`, `productArity`, `productElement`, `equals`, `hashCode`, `canEqual` shims to preserve case class contract.
- **`NestedMaterializationCancellationPolicy`**: Changed from open `class` with anonymous subclasses to `final class` with a `name` constructor parameter; removes unnecessary anonymous class allocation.
- Removed now-unused `OptionVal` import.

```scala
// Before: O(n) scan on every mandatoryAttribute[] call
find(attributeList) match {
  case OptionVal.Some(t) => t.asInstanceOf[T]
  case _                 => throw new IllegalStateException(...)
}

// After: O(1) map lookup
mandatoryAttributes(c.asInstanceOf[Class[AnyRef]]).asInstanceOf[T]
```

### DIFF
--- a/stream/src/main/mima-filters/2.0.x.backwards.excludes/optimized-mandatory-attributes.excludes
+++ b/stream/src/main/mima-filters/2.0.x.backwards.excludes/optimized-mandatory-attributes.excludes
@@ -17,3 +17,7 @@
 
 # Optimized mandatory attributes
 ProblemFilters.exclude[FinalClassProblem]("org.apache.pekko.stream.Attributes$NestedMaterializationCancellationPolicy")
+ProblemFilters.exclude[DirectMissingMethodProblem]("org.apache.pekko.stream.Attributes.fromProduct")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("org.apache.pekko.stream.Attributes.unapply")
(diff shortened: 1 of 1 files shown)


He-Pin:
It would be nice to have a jmh attached to it.

pjfanning:
> It would be nice to have a jmh attached to it.

This is a copy of an Akka PR. I have no plans to spend time profiling or benchmarking it. I don't see many of the performance related PRs from other contributors coming with detailed performance metrics.

He-Pin (approved):

He-Pin:
@pjfanning Yeah, that should be fine.

PR #2814, opened 2026-03-28
Add opt-in stream error log throttling

He-Pin:
## Motivation

When a stream stage fails rapidly and repeatedly (e.g., persistent network failure, bad data in a loop), each failure generates a log message. In high-throughput systems, this can result in thousands of log messages per second, overwhelming log aggregation systems and masking other important messages.

## Modification

Added a new configuration option:
```hocon
pekko.stream.materializer.stage-errors-log-throttle-period = off
```

When set to a positive duration (e.g. `10s`), only the first stage error within each time window is logged at ERROR level. Subsequent errors in the window are counted silently, and a summary warning is emitted when the next window opens or the interpreter finishes.

### Implementation details

| Aspect | Detail |
|--------|--------|
| **Scope** | Per-interpreter throttle state (not shared across streams) |
| **First error** | Uses `errorLogInitialized` flag to guarantee first error always logs, regardless of `System.nanoTime()` origin |
| **Config validation** | Rejects negative durations with `require()` |
| **Cleanup** | Flushes suppressed count in `finish()` for best-effort reporting |
| **Default** | `off` — zero behavior change for existing users |

### Files changed

- `stream/src/main/resources/reference.conf` — New config key
- `stream/.../GraphInterpreter.scala` — Throttle state fields + modified `reportStageError` + `finish()` flush
- `stream-tests/.../StageErrorLogThrottleSpec.scala` — 4 tests (enabled throttle with Broadcast fan-out, single error, disabled with fan-out, disabled single error)

## Result

- Operators can opt-in to throttle stage error logs in noisy environments
- Zero behavior change by default (`off`)
- Tests use Broadcast with 5 parallel failing stages to exercise the actual throttle code path (multiple `reportStageError` calls per interpreter)
- All new tests pass, all existing `ActorGraphInterpreterSpec` and `InterpreterSpec` tests pass (no regression)

## References

- Complements PR #2805 (fine-grained stream error log level control)

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/impl/fusing/StageErrorLogThrottleSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/impl/fusing/StageErrorLogThrottleSpec.scala
@@ -1,0 +1,120 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.stream.impl.fusing
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+import org.apache.pekko
+import pekko.stream._
+import pekko.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import pekko.stream.scaladsl._
+import pekko.stream.stage.GraphStageLogic
+import pekko.stream.stage.InHandler
+import pekko.stream.stage.OutHandler
+import pekko.stream.testkit.StreamSpec
+import pekko.testkit.EventFilter
+
+class StageErrorLogThrottleSpec extends StreamSpec("""
+    pekko.stream.materializer.stage-errors-log-throttle-period = 2s
+    pekko.loglevel = DEBUG
+  """) {
+
+  private def mkFailingStage: SimpleLinearGraphStage[Int] = new SimpleLinearGraphStage[Int] {
+    override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+      new GraphStageLogic(shape) with InHandler with OutHandler {
+        setHandlers(in, out, this)
+        override def onPush(): Unit = throw new IllegalArgumentException("test failure")
+        override def onPull(): Unit = pull(in)
+      }
+  }
+
+  "Stage error log throttling" must {
+
+    "suppress repeated errors and flush count on stream finish" in {
+      // Use Broadcast to fan-out to 5 independent failing stages sharing one interpreter.
+      // Only the first error should be logged; the remaining 4 are suppressed by throttling.
+      EventFilter[IllegalArgumentException](pattern = "Error in stage.*", occurrences = 1).intercept {
+        val done = RunnableGraph
+          .fromGraph(GraphDSL.createGraph(Sink.ignore) { implicit b => sink =>
+            import GraphDSL.Implicits._
+            val bcast = b.add(Broadcast[Int](5, eagerCancel = false))
+            Source.single(1) ~> bcast
+            bcast.out(0)     ~> b.add(mkFailingStage) ~> sink
+            for (i <- 1 until 5) {
+              bcast.out(i) ~> b.add(mkFailingStage) ~> b.add(Sink.ignore)
+            }
+            ClosedShape
+          })
+          .run()
+        Await.ready(done, 3.seconds)
+      }
+    }
+
+    "always log a single error even with throttling enabled" in {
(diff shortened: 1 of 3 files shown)


He-Pin (commented):
Very useful addition for production deployments dealing with noisy stream failures.

**Positive:**
- Default `off` means zero breaking change
- Per-interpreter scope is the right granularity
- The `errorLogInitialized` flag correctly handles `System.nanoTime()` origin edge cases
- `finish()` flush ensures suppressed errors aren't silently lost
- Tests cover both enabled and disabled scenarios with fan-out

**Concerns:**

1. `suppressedErrorCount` is reset in `finish()` but only flushed if `> 0`. If the throttle period expires between errors (not at stream finish), the count is reset inside the `shouldLog` branch but never flushed as a warning. This means errors that fall at the boundary of throttle windows could be silently lost without any summary. Consider flushing the count at the start of each new window:

```scala
if (suppressedErrorCount > 0) {
  log.warning("{} additional stage error(s) were suppressed during throttle period", suppressedErrorCount)
}\n```

This is already done when the stream finishes -- it should also happen when a new window starts.

2. `config.getDuration(path, TimeUnit.NANOSECONDS)` could throw `ConfigException` for invalid values (e.g. `"abc"`). The `try/catch` only catches `UnsupportedOperationException`. Consider catching the broader `Exception` or validating the config format more explicitly.

3. `ThreadLocalRandom` import was added but doesn't appear to be used in the diff. Likely a leftover from development.

**Review by He-Pin — COMMENTED** [review]
## Deep CR: PR #2814 - Add opt-in stream error log throttling

### Architecture Review

This PR adds per-interpreter throttle state for stage error logging in `GraphInterpreter`. When enabled, only the first error within a time window is logged at ERROR level, with subsequent errors counted and summarized.

**Design is well-scoped**:
- Per-interpreter granularity is correct - each stream has independent throttle state
- Default `off` means zero behavior change for existing users
- `finish()` flush ensures suppressed counts are not silently lost

**Implementation details:**

The throttle state uses three fields:
- `throttlePeriodNanos` (lazy val) - parsed once from config
- `lastErrorLogTime` (var) - tracks when the last error was logged
- `errorLogInitialized` (boolean flag) - handles `System.nanoTime()` origin edge cases
- `suppressedErrorCount` (var) - counts suppressed errors

The `errorLogInitialized` flag is critical and correctly implemented. `System.nanoTime()` on some JVMs can have arbitrary origins (not necessarily 0), so checking `!errorLogInitialized` ensures the first error always logs regardless of the nanoTime origin.

### Binary Compatibility

No public API changes. This is purely internal behavior controlled by a new config key. No MiMa impact.

### Config Design

`pekko.stream.materializer.stage-errors-log-throttle-period = off` follows the established pattern for duration configs (similar to `akka.http.server.request-timeout = off`). The `Helpers.toRootLowerCase(value) == "off"` check is the standard Pekko pattern for disabling duration-based features.

**Config parsing concern**: The `try/catch` only handles `UnsupportedOperationException` (from `NoMaterializer` in tests). If a user provides an invalid value like `"abc"`, `config.getDuration(path, TimeUnit.NANOSECONDS)` throws `ConfigException.WrongType` or `ConfigException.BadValue`, which would propagate as an unchecked exception during interpreter creation. The `require(nanos > 0L)` catches negative durations but not malformed strings.

**Recommendation**: Catch the broader `Exception` in the outer try/catch, or add a specific catch for `ConfigException` with a clear error message.

### Throttle Window Boundary Issue

There is a subtle issue with throttle window boundaries. When the throttle period expires and a new error triggers a new window:

```scala
if (!errorLogInitialized || now - lastErrorLogTime >= throttlePeriodNanos) {
  if (suppressedErrorCount > 0) {
    log.warning("{} additional stage error(s) were suppressed during throttle period",
      suppressedErrorCount)
  }
  errorLogInitialized = true
  lastErrorLogTime = now
  suppressedErrorCount = 0
  true
}
```

This correctly flushes the suppressed count when a new window starts. However, if the stream finishes (`finish()`) during a window where the suppressed count is already 0 (first error just logged), no flush warning is needed. This is correct.

But consider: if errors occur at t=0, t=1, t=2, and the throttle period is 5 seconds, and the stream finishes at t=3:
- t=0: Error logged (first error)
- t=1: suppressedErrorCount = 1
- t=2: suppressedErrorCount = 2
- t=3: finish() flushes "2 additional stage error(s) were suppressed"

This is correct behavior. The `finish()` flush covers the case where the stream ends before the throttle period expires.

### Test Coverage

The tests are well-designed using `Broadcast` with 5 parallel failing stages:
1. **Throttle enabled, fan-out**: Verifies only 1 error logged (not 5)
2. **Throttle enabled, single error**: Verifies single error is always logged
3. **Throttle disabled, fan-out**: Verifies all 5 errors are logged
4. **Throttle disabled, single error**: Verifies single error behavior

**Missing tests**:
- **Window expiration**: No test verifies that after the throttle period expires, the next error is logged and a summary is emitted
- **Multiple windows**: No test exercises multiple throttle windows within a single stream lifetime
- **Config validation**: No test for invalid config values

### Code Quality

**Unused import**: `java.util.concurrent.ThreadLocalRandom` was already imported before this change. The diff shows it being replaced with `{ ThreadLocalRandom, TimeUnit }`. The `TimeUnit` is used in `config.getDuration(path, TimeUnit.NANOSECONDS)`. However, `ThreadLocalRandom` appears to be unused in the current file after this change. Let me verify...

Actually, checking the full file, `ThreadLocalRandom` may be used elsewhere. If not, it should be removed.

### Java API Coverage

N/A - this is an internal configuration change. Java users configure it via HOCON config just like Scala users.

### Suggestions

1. **Add config validation**: Catch `ConfigException` in the try/catch and provide a clear error message for invalid values.
2. **Add a test for throttle window expiration**: Verify that after the period expires, the next error triggers a summary and a new window.
3. **Verify ThreadLocalRandom import**: Check if it is still used elsewhere in the file; if not, remove it.

PR #2828, opened 2026-03-30
Update Java version to 26

pjfanning:
just an experiment to see what breaks

### DIFF
--- a/.github/workflows/build-test-prValidation.yml
+++ b/.github/workflows/build-test-prValidation.yml
@@ -36,11 +36,11 @@ jobs:
       - name: Cache Coursier cache
         uses: coursier/cache-action@90c37294538be80a558fd665531fcdc2b467b475 # 8.1.0
 
-      - name: Setup Java 17
+      - name: Setup Java 26
         uses: actions/setup-java@v5
         with:
           distribution: temurin
-          java-version: 17
+          java-version: 26
 
       - name: Install sbt
         uses: sbt/setup-sbt@508b753e53cb6095967669e0911487d2b9bc9f41 # v1.1.22
@@ -69,11 +69,11 @@ jobs:
           fetch-tags: 0
           persist-credentials: false
 
-      - name: Setup Java 17
+      - name: Setup Java 26
         uses: actions/setup-java@v5
         with:
           distribution: temurin
-          java-version: 17
+          java-version: 26
 
       - name: Install sbt
         uses: sbt/setup-sbt@508b753e53cb6095967669e0911487d2b9bc9f41 # v1.1.22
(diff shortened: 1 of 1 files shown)


He-Pin (commented):
This PR changes CI to exclusively use Java 26, dropping Java 17 and 21. Given that Pekko targets multiple Java versions, this could break compatibility testing for users still on Java 17/21.

**Recommendation:** This should not replace the existing Java 17/21 matrix entries. Instead, Java 26 should be added as an additional matrix entry alongside the existing versions, or this should be a separate experimental workflow. The PR description says 'just an experiment' -- consider whether this should be converted to a draft or closed in favor of a proper matrix expansion.

He-Pin (commented):
## Deep CR: PR #2828 - Update Java version to 26

### Architecture Review

This PR changes CI to exclusively use Java 26, replacing Java 17 and Java 21 in all workflow jobs. The PR description explicitly states "just an experiment to see what breaks."

### Concerns

**1. Single Java Version Testing is Insufficient**
Pekko 2.0 targets Java 17+ as its minimum. Testing ONLY on Java 26 means:
- Regression on Java 17 (the actual minimum supported version) will not be caught
- Java 21 LTS (the most widely deployed LTS between 17 and 26) will not be tested
- Users on Java 17-25 have no CI guarantee of compatibility

**2. Experiment vs Production CI**
This PR modifies the `build-test-prValidation.yml` workflow which is the primary CI pipeline for PR validation. Experiments should use a separate workflow file (e.g., `build-test-java26-experimental.yml`) that does not block regular PRs.

**3. Missing Context on Why Java 26**
Java 26 is the latest release at the time of this PR (April 2026). Testing on it is valuable, but replacing existing LTS testing is not. The correct approach is a matrix:
- Java 17 (minimum supported)
- Java 21 (LTS, most common production version)
- Java 26 (latest, forward compatibility)

### Recommendations

1. **Do not merge this as-is**. Convert to a separate experimental workflow or add Java 26 as a matrix entry alongside Java 17/21.
2. If the intent is to eventually raise the minimum Java version to 26, that should be a separate discussion with community consensus, documented in the mailing list.
3. Consider using `--matrix-exclude` to reduce CI cost while maintaining multi-version coverage (e.g., run Java 17 on all PRs, Java 21 and 26 on nightly builds).

### Binary Compatibility

CI-only change. No impact on binary compatibility or the public API.

### Test Coverage

The workflows themselves are being tested - they will fail if Java 26 introduces incompatibilities. However, the lack of Java 17/21 testing means incompatibilities with those versions will go undetected until users report them.

pjfanning:
seems that we still compile with java 26

PR #2838, opened 2026-04-04
VarHandle multi-byte reads for ByteString lastIndexOf

pjfanning:
* Relates to #2150 (there is non SWAR PR too that may work out better)

## SWARUtil — extended VarHandle infrastructure
- Use `byteArrayViewVarHandle` instances for `short` (BE+LE), `int` (BE+LE), `long` (LE)
- Added `getLastIndex(word)` — finds rightmost byte match in SWAR result using `numberOfTrailingZeros`

## ByteString
- Use SWARUtil for multi byte processing

Significant improvement in benchmark.
```
With Changes
[info] Benchmark                                                              Mode  Cnt          Score           Error  Units
[info] ByteString_lastIndexOf_Benchmark.bs1_lastIndexOf                      thrpt    3  107880125.372 ± 247014962.175  ops/s
[info] ByteString_lastIndexOf_Benchmark.bs1_lastIndexOf_byte                 thrpt    3   88478677.724 ± 181685386.699  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_best_case            thrpt    3   58313896.505 ± 806477112.611  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_far_index_case       thrpt    3   17472854.418 ±  94251978.262  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_far_index_case_byte  thrpt    3   19332603.332 ±   2245521.678  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_worst_case           thrpt    3   20053173.893 ±   2385070.269  ops/s

Without Changes
[info] Benchmark                                                              Mode  Cnt          Score          Error  Units
[info] ByteString_lastIndexOf_Benchmark.bs1_lastIndexOf                      thrpt    3   33298670.109 ± 63631788.200  ops/s
[info] ByteString_lastIndexOf_Benchmark.bs1_lastIndexOf_byte                 thrpt    3   27021365.677 ±  5557597.483  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_best_case            thrpt    3   34686168.040 ±  7080975.674  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_far_index_case       thrpt    3     635863.919 ±    69057.413  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_far_index_case_byte  thrpt    3     597510.828 ±    29481.618  ops/s
[info] ByteString_lastIndexOf_Benchmark.bss_lastIndexOf_worst_case           thrpt    3    2125744.038 ±   196176.723  ops/s
```

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
@@ -405,6 +405,56 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       ByteString1.fromString("0123456789").take(3).drop(1) should ===(ByteString("12"))
       ByteString1.fromString("0123456789").take(10).take(8).drop(3).take(5) should ===(ByteString("34567"))
     }
+    "lastIndexOf (specialized)" in {
+      ByteString.empty.lastIndexOf(5.toByte, -1) should ===(-1)
+      ByteString.empty.lastIndexOf(5.toByte, 0) should ===(-1)
+      ByteString.empty.lastIndexOf(5.toByte, 1) should ===(-1)
+      ByteString.empty.lastIndexOf(5.toByte) should ===(-1)
+      val byteString1 = ByteString1.fromString("abb")
+      byteString1.lastIndexOf('d'.toByte) should ===(-1)
+      byteString1.lastIndexOf('d'.toByte, -1) should ===(-1)
+      byteString1.lastIndexOf('d'.toByte, 4) should ===(-1)
+      byteString1.lastIndexOf('d'.toByte, 1) should ===(-1)
+      byteString1.lastIndexOf('d'.toByte, 0) should ===(-1)
+      byteString1.lastIndexOf('a'.toByte, -1) should ===(-1)
+      byteString1.lastIndexOf('a'.toByte) should ===(0)
+      byteString1.lastIndexOf('a'.toByte, 0) should ===(0)
+      byteString1.lastIndexOf('a'.toByte, 1) should ===(0)
+      byteString1.lastIndexOf('b'.toByte) should ===(2)
+      byteString1.lastIndexOf('b'.toByte, 2) should ===(2)
+      byteString1.lastIndexOf('b'.toByte, 1) should ===(1)
+      byteString1.lastIndexOf('b'.toByte, 0) should ===(-1)
+
+      val byteStrings = ByteStrings(ByteString1.fromString("abb"), ByteString1.fromString("efg"))
+      byteStrings.lastIndexOf('e'.toByte) should ===(3)
+      byteStrings.lastIndexOf('e'.toByte, 6) should ===(3)
+      byteStrings.lastIndexOf('e'.toByte, 4) should ===(3)
+      byteStrings.lastIndexOf('e'.toByte, 1) should ===(-1)
+      byteStrings.lastIndexOf('e'.toByte, 0) should ===(-1)
+      byteStrings.lastIndexOf('e'.toByte, -1) should ===(-1)
+
+      byteStrings.lastIndexOf('b'.toByte) should ===(2)
+      byteStrings.lastIndexOf('b'.toByte, 6) should ===(2)
+      byteStrings.lastIndexOf('b'.toByte, 4) should ===(2)
+      byteStrings.lastIndexOf('b'.toByte, 1) should ===(1)
+      byteStrings.lastIndexOf('b'.toByte, 0) should ===(-1)
+      byteStrings.lastIndexOf('b'.toByte, -1) should ===(-1)
+
+      val compact = byteStrings.compact
+      compact.lastIndexOf('e'.toByte) should ===(3)
+      compact.lastIndexOf('e'.toByte, 6) should ===(3)
+      compact.lastIndexOf('e'.toByte, 4) should ===(3)
+      compact.lastIndexOf('e'.toByte, 1) should ===(-1)
+      compact.lastIndexOf('e'.toByte, 0) should ===(-1)
+      compact.lastIndexOf('e'.toByte, -1) should ===(-1)
+
+      compact.lastIndexOf('b'.toByte) should ===(2)
+      compact.lastIndexOf('b'.toByte, 6) should ===(2)
+      compact.lastIndexOf('b'.toByte, 4) should ===(2)
+      compact.lastIndexOf('b'.toByte, 1) should ===(1)
+      compact.lastIndexOf('b'.toByte, 0) should ===(-1)
+      compact.lastIndexOf('b'.toByte, -1) should ===(-1)
+    }
     "copyToArray" in {
       val byteString = ByteString1(Array[Byte](1, 2, 3, 4, 5), startIndex = 1, length = 3)
       def verify(f: Array[Byte] => Unit)(expected: Byte*): Unit = {
--- a/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
@@ -888,6 +888,12 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       zeros.lastIndexOf(0.toByte) should ===(2)
       val neg = ByteString(Array[Byte](-1, 0, -1))
       neg.lastIndexOf((-1).toByte) should ===(2)
+
+      val concat0 = makeMultiByteStringsSample()
+      concat0.lastIndexOf(0xFF.toByte) should ===(18)
+      concat0.lastIndexOf(0xFF.toByte, 18) should ===(18)
+      concat0.lastIndexOf(0xFF.toByte, 17) should ===(0)
+      concat0.lastIndexOf(0xFE.toByte) should ===(-1)
     }
     "indexOf (specialized)" in {
       ByteString.empty.indexOf(5.toByte) should ===(-1)
@@ -920,6 +926,11 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       compact.indexOf('e'.toByte) should ===(3)
       compact.indexOf('f'.toByte) should ===(4)
       compact.indexOf('g'.toByte) should ===(5)
+
+      val concat0 = makeMultiByteStringsSample()
+      concat0.indexOf(0xFF.toByte) should ===(0)
+      concat0.indexOf(16.toByte) should ===(17)
+      concat0.indexOf(0xFE.toByte) should ===(-1)
     }
     "indexOf (specialized) from offset" in {
       ByteString.empty.indexOf(5.toByte, -1) should ===(-1)
@@ -986,6 +997,11 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       byteStringLong.indexOf('m', 2) should ===(12)
       byteStringLong.indexOf('z', 2) should ===(25)
       byteStringLong.indexOf('a', 2) should ===(-1)
+
+      val concat0 = makeMultiByteStringsSample()
+      concat0.indexOf(0xFF.toByte, 0) should ===(0)
+      concat0.indexOf(0xFF.toByte, 17) should ===(18)
+      concat0.indexOf(0xFE.toByte, 17) should ===(-1)
     }
     "contains" in {
       ByteString.empty.contains(5) should ===(false)
@@ -1999,5 +2015,18 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
         }
       }
     }
+  }
+
+  private def makeMultiByteStringsSample(): ByteString = {
+    val byteStrings = Vector(
+      ByteString1(Array[Byte](0xFF.toByte)),
+      ByteString1(Array[Byte](0, 1, 2, 3)),
+      ByteString1(Array[Byte](4, 5)),
+      ByteString1(Array[Byte](6, 7, 8, 9)),
+      ByteString1(Array[Byte](10)),
+      ByteString1(Array[Byte](11, 12, 13, 14, 15, 16)),
+      ByteString1(Array[Byte](0xFF.toByte))
+    )
+    ByteStrings(byteStrings)
   }
 }
--- a/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
@@ -405,56 +405,6 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
       ByteString1.fromString("0123456789").take(3).drop(1) should ===(ByteString("12"))
       ByteString1.fromString("0123456789").take(10).take(8).drop(3).take(5) should ===(ByteString("34567"))
     }
-    "lastIndexOf (specialized)" in {
-      ByteString.empty.lastIndexOf(5.toByte, -1) should ===(-1)
-      ByteString.empty.lastIndexOf(5.toByte, 0) should ===(-1)
-      ByteString.empty.lastIndexOf(5.toByte, 1) should ===(-1)
-      ByteString.empty.lastIndexOf(5.toByte) should ===(-1)
-      val byteString1 = ByteString1.fromString("abb")
-      byteString1.lastIndexOf('d'.toByte) should ===(-1)
-      byteString1.lastIndexOf('d'.toByte, -1) should ===(-1)
-      byteString1.lastIndexOf('d'.toByte, 4) should ===(-1)
-      byteString1.lastIndexOf('d'.toByte, 1) should ===(-1)
-      byteString1.lastIndexOf('d'.toByte, 0) should ===(-1)
-      byteString1.lastIndexOf('a'.toByte, -1) should ===(-1)
-      byteString1.lastIndexOf('a'.toByte) should ===(0)
-      byteString1.lastIndexOf('a'.toByte, 0) should ===(0)
-      byteString1.lastIndexOf('a'.toByte, 1) should ===(0)
-      byteString1.lastIndexOf('b'.toByte) should ===(2)
-      byteString1.lastIndexOf('b'.toByte, 2) should ===(2)
-      byteString1.lastIndexOf('b'.toByte, 1) should ===(1)
-      byteString1.lastIndexOf('b'.toByte, 0) should ===(-1)
-
-      val byteStrings = ByteStrings(ByteString1.fromString("abb"), ByteString1.fromString("efg"))
-      byteStrings.lastIndexOf('e'.toByte) should ===(3)
-      byteStrings.lastIndexOf('e'.toByte, 6) should ===(3)
-      byteStrings.lastIndexOf('e'.toByte, 4) should ===(3)
-      byteStrings.lastIndexOf('e'.toByte, 1) should ===(-1)
-      byteStrings.lastIndexOf('e'.toByte, 0) should ===(-1)
-      byteStrings.lastIndexOf('e'.toByte, -1) should ===(-1)
-
-      byteStrings.lastIndexOf('b'.toByte) should ===(2)
-      byteStrings.lastIndexOf('b'.toByte, 6) should ===(2)
-      byteStrings.lastIndexOf('b'.toByte, 4) should ===(2)
-      byteStrings.lastIndexOf('b'.toByte, 1) should ===(1)
-      byteStrings.lastIndexOf('b'.toByte, 0) should ===(-1)
-      byteStrings.lastIndexOf('b'.toByte, -1) should ===(-1)
-
-      val compact = byteStrings.compact
-      compact.lastIndexOf('e'.toByte) should ===(3)
-      compact.lastIndexOf('e'.toByte, 6) should ===(3)
-      compact.lastIndexOf('e'.toByte, 4) should ===(3)
-      compact.lastIndexOf('e'.toByte, 1) should ===(-1)
-      compact.lastIndexOf('e'.toByte, 0) should ===(-1)
-      compact.lastIndexOf('e'.toByte, -1) should ===(-1)
-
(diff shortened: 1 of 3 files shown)


He-Pin (commented):
Good performance work, but a few concerns:

**Concerns:**
1. `ByteString1C.lastIndexOf` and `ByteString1.lastIndexOf` -- The SWAR-based specialized `lastIndexOf(elem: Byte, end: Int)` looks correct but the `unrolledLastIndexOf` tail handling is complex. The tests cover basic cases but I'd recommend adding a test that specifically targets the boundary between tail and full 8-byte chunks (e.g. length=15, element at position 14).

2. `SWARUtil.getLastIndex` -- the doc says `Currently only supports big endian`. However it's used by the `lastIndexOf` implementation which reads via `SWARUtil.getLong(bytes, chunkStart)` (the default big-endian path). This is correct, but worth explicitly documenting to prevent future misuse.

3. The PR description says WIP and needs benchmarking. I'd recommend not merging until benchmarks confirm the SWAR `lastIndexOf` is actually faster than the simple loop for typical ByteString sizes (many ByteStrings are small, where SWAR overhead may not pay off).

4. `ByteIterator` overrides -- using `byteOrder == ByteOrder.BIG_ENDIAN` boolean flag is cleaner than the if/else chain in #2847. Consider whether these two PRs should be coordinated to avoid duplicate implementations.

pjfanning:
I'm breaking this up into smaller PRs

pjfanning:
I removed the parts of the original PR and focused on SWAR lastIndexOf. The other parts were merged already in other PRs.

He-Pin:
I will check this in detail soon

He-Pin:
Do we need something like Netty's VarHandleFactory to handle different platform byte orders

pjfanning:
> Do we need something like Netty's VarHandleFactory to handle different platform byte orders

in our code, we only need the 2 ByteOrders that are part of the Java ByteOrder enum - Big Endian and Little Endian. We have no code that uses exotic byte orders.

SwarUtil has the various var handles that we need, I don't think we need to grab anything else from Netty's VarHandleFactory.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -921,6 +1030,67 @@ object ByteString {
...
+          if (bsIdx < 0) -1
+          else {
+            val bs = bytestrings(bsIdx)
+            val bsStartIndex = len - bs.length
+
+            if (relativeIndex < bsStartIndex || bs.isEmpty) {
+              if (bsIdx == 0) -1
+              else find(bsIdx - 1, relativeIndex, bsStartIndex)
+            } else {
+              val subIndexOf = bs.lastIndexOf(elem, relativeIndex)
```
He-Pin: ### Bug: `relativeIndex` is in global coordinates, `bs.lastIndexOf` expects local

`relativeIndex` tracks the overall ByteStrings position, but `bs.lastIndexOf(elem, end)` interprets `end` as a local index within that sub-bytestring.

**Failing case:**
```scala
val bs = ByteStrings(ByteString1.fromString("ab"), ByteString1.fromString("dd"))
// Total: "abdd", 'd' at global indices 2 and 3
bs.lastIndexOf('d'.toByte, 2) // Expected: 2, Actual: 3
```

Trace: `relativeIndex = 2`, sub-bs = `"dd"`, `bsStartIndex = 2`.
`"dd".lastIndexOf('d', 2)` → endIdx clamps to `min(2, 1) = 1` → finds last `'d'` at **local 1** → returns `1 + 2 = 3` (beyond `end = 2`).

**Fix:**
```scala
val subIndexOf = bs.lastIndexOf(elem, relativeIndex - bsStartIndex)
```

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -921,6 +1030,67 @@ object ByteString {
...
+            if (bsIdx < 0) -1
+            else {
+              val bs = bytestrings(bsIdx)
+              val bsStartIndex = len - bs.length
+
+              if (relativeIndex < bsStartIndex || bs.isEmpty) {
+                if (bsIdx == 0) -1
+                else find(bsIdx - 1, relativeIndex, bsStartIndex)
+              } else {
+                val subIndexOf = bs.lastIndexOf(elem, relativeIndex)
```
He-Pin: Same coordinate-translation bug as the generic version above.

```suggestion
                val subIndexOf = bs.lastIndexOf(elem, relativeIndex - bsStartIndex)
```

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -921,6 +1030,67 @@ object ByteString {
...
+        }
+
+        find(byteStringsLast, math.min(end, length), length)
+      }
+    }
+
+    override def lastIndexOf(elem: Byte, end: Int): Int = {
+      if (end < 0) -1
+      else {
+        if (end < 0) -1
```
He-Pin: Redundant check -- `end < 0` is already handled at line 1063. This inner check is dead code.

```suggestion
        {
```

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -921,6 +1030,67 @@ object ByteString {
...
+              val subIndexOf = bs.lastIndexOf(elem, relativeIndex)
+              if (subIndexOf < 0) {
+                if (bsIdx == 0) -1
+                else find(bsIdx - 1, relativeIndex, bsStartIndex)
+              } else subIndexOf + bsStartIndex
+            }
+          }
+        }
+
+        find(byteStringsLast, math.min(end, length), length)
```
He-Pin: Nit: should be `math.min(end, length - 1)` -- valid indices are `0..length-1`. Currently masked by inner clamping in `ByteString1C`/`ByteString1`, but inconsistent with those implementations which all use `length - 1`.

Same applies to line 1089.

actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
```diff
@@ -821,6 +821,56 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
...
+      compact.lastIndexOf('e'.toByte, 4) should ===(3)
+      compact.lastIndexOf('e'.toByte, 1) should ===(-1)
+      compact.lastIndexOf('e'.toByte, 0) should ===(-1)
+      compact.lastIndexOf('e'.toByte, -1) should ===(-1)
+
+      compact.lastIndexOf('b'.toByte) should ===(2)
+      compact.lastIndexOf('b'.toByte, 6) should ===(2)
+      compact.lastIndexOf('b'.toByte, 4) should ===(2)
+      compact.lastIndexOf('b'.toByte, 1) should ===(1)
+      compact.lastIndexOf('b'.toByte, 0) should ===(-1)
```
He-Pin, line 881: Test coverage suggestions:

1. **Exercise the SWAR path** -- use strings >=16 bytes so `SWARUtil.getLong` is actually called:
```scala
val long1 = ByteString1.fromString("abcdefghijklmnop") // 16 bytes
long1.lastIndexOf('a'.toByte) should ===(0)
long1.lastIndexOf('p'.toByte) should ===(15)
long1.lastIndexOf('h'.toByte, 7) should ===(7)
long1.lastIndexOf('h'.toByte, 6) should ===(-1)
```

2. **Catch the ByteStrings relativeIndex bug** -- duplicate bytes spanning `end`:
```scala
val bs = ByteStrings(ByteString1.fromString("ab"), ByteString1.fromString("dd"))
bs.lastIndexOf('d'.toByte, 2) should ===(2)  // not 3
bs.lastIndexOf('d'.toByte, 3) should ===(3)
```

3. **Non-zero startIndex** -- via `.drop()`:
```scala
val sliced = ByteString1.fromString("xxabcdefghijk").drop(2)
sliced.lastIndexOf('k'.toByte) should ===(8)
```

4. **Edge byte values**:
```scala
val zeros = ByteString(Array[Byte](0, 1, 0, 1))
zeros.lastIndexOf(0.toByte) should ===(2)
val neg = ByteString(Array[Byte](-1, 0, -1))
neg.lastIndexOf((-1).toByte) should ===(2)
```

He-Pin (changes requested):
## Review Summary

`SWARUtil.getLastIndex` and the `ByteString1C`/`ByteString1` implementations are correct and well-structured. The SWAR chunking strategy (tail-first, then 8-byte blocks right-to-left) mirrors the existing `indexOf` and the `startIndex` offset handling is sound.

However, **`ByteStrings.lastIndexOf` has a coordinate-translation bug** that causes incorrect results when `end` falls within a sub-bytestring containing duplicate target bytes. Both the generic and specialized overloads are affected.

### Test Coverage Gaps

| Scenario | Covered? |
|---|---|
| ByteString1C / ByteStrings / compact basic cases | Yes |
| **ByteStrings: `end` cuts within a sub-bs with duplicate target bytes** | No (misses the bug above) |
| **SWAR code path (>=8 bytes in a single ByteString1)** | No (all test strings <=6 chars, `SWARUtil.getLong`+`applyPattern`+`getLastIndex` is never exercised) |
| **ByteString1 with non-zero `startIndex`** (e.g. via `.drop()`) | No |
| **Edge byte values (0x00, 0xFF)** | No |

The most important gap: no test has >=8 bytes, so the entire SWAR path -- the point of this PR -- is never actually tested.

He-Pin (commented):
## Review: ByteStrings.lastIndexOf Coordinate Translation Bug (Both Overloads)

### 🔴 Blocker: `ByteStrings.lastIndexOf` returns wrong result when `end` falls within a fragment containing duplicate target bytes

**File:** `actor/src/main/scala/org/apache/pekko/util/ByteString.scala` — lines 1033-1097 (generic) and 1065-1097 (specialized)

**Root cause:** The `relativeIndex` passed to `bs.lastIndexOf(elem, relativeIndex)` is a **global coordinate**, not a local index within the current ByteString fragment. When `relativeIndex` exceeds `bs.length - 1`, the fragment's own `lastIndexOf` clamps it to `bs.length - 1`, effectively searching the **entire fragment** including bytes beyond the intended global `end`.

**Reproduction:**
```scala
// bytestrings = ["abb", "beb"], total length = 6
// Global indices: a(0), b(1), b(2), b(3), e(4), b(5)
val bs = ByteStrings(ByteString1.fromString("abb"), ByteString1.fromString("beb"))
bs.lastIndexOf('b'.toByte, 4)
// Expected: 2 (last 'b' at or before global index 4)
// Actual:   5 (the method searches all of bs1="beb", finds 'b' at local index 2, returns 2+3=5)
```

**Bug trace** (specialized overload, `ByteString.scala:1065-1097`):
1. `find(1, math.min(4, 6), 6)` → `find(1, 4, 6)`
2. bsIdx=1, bs="beb" (length=3), `bsStartIndex = 6 - 3 = 3`
3. `relativeIndex(4) < bsStartIndex(3)` → false, so we enter the else branch
4. `bs.lastIndexOf('b', 4)` → inside ByteString1: `endIdx = math.min(4, 2) = 2` → searches entire "beb" → returns 2
5. Returns `2 + bsStartIndex(3) = 5` — **but global index 5 > end(4)!**

**Both the generic `lastIndexOf[B >: Byte]` and specialized `lastIndexOf(elem: Byte)` overloads have this bug.** The fix must convert the global `relativeIndex` to a local index before calling `bs.lastIndexOf`:

```scala
// Instead of:
val subIndexOf = bs.lastIndexOf(elem, relativeIndex)
// Should be:
val localEnd = relativeIndex - bsStartIndex
val subIndexOf = bs.lastIndexOf(elem, localEnd)
```

---

### 🟡 Suggestion: Redundant `if (end < 0)` check

**File:** `actor/src/main/scala/org/apache/pekko/util/ByteString.scala` — lines 1065-1069

The specialized `ByteStrings.lastIndexOf(elem: Byte, end: Int)` has a duplicate guard:
```scala
if (end < 0) -1
else {
  if (end < 0) -1  // ← redundant, outer check already handles this
  else { ... }
}
```

The outer `if (end < 0)` already covers this case. The inner check is dead code.

---

### 🟡 Suggestion: Missing test for ByteString1 with non-zero `startIndex` via `.drop()`

The existing tests create ByteString1 directly via `fromString` (startIndex=0). The SWAR path in `ByteString1.lastIndexOf` should be tested with a dropped/sliced ByteString to verify the `startIndex` offset arithmetic is correct:

```scala
// Test SWAR path (>=8 bytes) with non-zero startIndex
val dropped = ByteString1.fromString("xxxxxxab").drop(6) // "ab", startIndex=6
dropped.lastIndexOf('a'.toByte) should ===(0)
dropped.lastIndexOf('b'.toByte) should ===(1)
```

---

### Verification commands

```bash
# Run the ByteStringSpec tests
sbt "actor-tests/testOnly org.apache.pekko.util.ByteStringSpec"

# Verify the ByteStrings bug with a quick scala snippet
sbt "actor/console" <<< '
import org.apache.pekko.util._
val bs = ByteStrings(ByteString1.fromString("abb"), ByteString1.fromString("beb"))
println(s"lastIndexOf(b, 4) = ${bs.lastIndexOf("b".toByte, 4)}")  // Expected: 2
'
```

actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
pjfanning: thanks @He-Pin for the helpful review and change suggestions. The 3rd test here seems wrong. If you drop the first 2 chars ('xx') - 'k' is still the 10th element (0 based indexing)

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: applied with most of the suggested changes - thanks

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: changed

He-Pin (commented):
## Review: Remaining Issues After Fix Verification

The ByteStrings coordinate-translation fix is correct — verified with trace analysis. The redundant `if (end < 0)` nesting has also been cleaned up. Good work.

Below are **3 new findings** not covered in previous reviews.

---

### 🔴 Critical: `getLastIndex(0)` returns garbage value — asymmetry with `getIndex` is a footgun

**File:** `actor/src/main/scala/org/apache/pekko/util/SWARUtil.scala:157`

`getIndex` returns `8` (a clear sentinel) when `word == 0`:
```scala
def getIndex(word: Long): Int = java.lang.Long.numberOfLeadingZeros(word) >>> 3
// word=0 → numberOfLeadingZeros(0)=64 → 64>>>3 = 8 ✅ sentinel
```

But `getLastIndex` has no such guard:
```scala
def getLastIndex(word: Long): Int = (java.lang.Long.SIZE - 1 - java.lang.Long.numberOfTrailingZeros(word)) >>> 3
// word=0 → numberOfTrailingZeros(0)=64 → (63-64)>>>3 = (-1)>>>3 = 2305843009213693951 ❌ garbage
```

In Scala/Java, `(-1) >>> 3` treats `-1` as unsigned `0xFFFFFFFFFFFFFFFF`, shifting right 3 gives `0x1FFFFFFFFFFFFF` — a **massive positive number**, not a sentinel.

All current callers correctly guard with `if (result != 0)`, so this is safe **today**. But the asymmetry means a future developer who copies the `getIndex` pattern (which does not need a guard because it returns 8) will get silently wrong results instead of an obvious sentinel.

**Recommendation:** Add a guard to match `getIndex` sentinel behavior:
```scala
def getLastIndex(word: Long): Int =
  if (word == 0) 8
  else (java.lang.Long.SIZE - 1 - java.lang.Long.numberOfTrailingZeros(word)) >>> 3
```

This way both functions return `8` for "no match", making them interchangeable and self-documenting.

---

### 🟡 Suggestion: ByteStrings relies on implicit clamping contract with inner `bs.lastIndexOf`

**File:** `actor/src/main/scala/org/apache/pekko/util/ByteString.scala` — lines 1049-1050 (generic) and 1079-1080 (specialized)

The initial call passes `relativeIndex - bsStartIndex` to `bs.lastIndexOf`:
```scala
val subIndexOf = bs.lastIndexOf(elem, relativeIndex - bsStartIndex)
```

When `end >= length`, the initial `relativeIndex = math.min(end, length - 1) = length - 1`. For the last fragment with `bs.length = 3` and `bsStartIndex = 7`, this means calling `bs.lastIndexOf(elem, (length-1) - 7)`. If the fragment spans the full remaining range, `relativeIndex - bsStartIndex` could equal `bs.length - 1`, which is fine. But the **implicit contract** is that the inner `lastIndexOf` will clamp its own `end` parameter — the outer code does not clamp before delegating.

**Current behavior:** Correct, because ByteString1C/ByteString1 do `val endIdx = math.min(end, length - 1)` internally.

**Risk:** If the inner implementation ever changes its clamping strategy, this breaks silently. The outer code should be explicit:

```scala
val localEnd = math.min(relativeIndex - bsStartIndex, bs.length - 1)
val subIndexOf = bs.lastIndexOf(elem, localEnd)
```

This makes the contract explicit and removes the dependency on inner implementation details.

---

### ⚪ Nit: API asymmetry — `lastIndexOf` lacks the three-argument form that `indexOf` has

**File:** `actor/src/main/scala/org/apache/pekko/util/ByteString.scala` — base class at line 1264

`ByteString` has:
- `indexOf(elem: Byte, from: Int)` ✅
- `indexOf(elem: Byte, from: Int, to: Int)` ✅

But the new `lastIndexOf` only has:
- `lastIndexOf(elem: Byte)` ✅
- `lastIndexOf(elem: Byte, end: Int)` ✅
- `lastIndexOf(elem: Byte, from: Int, to: Int)` ❌ missing

Users who need to find the last occurrence within a bounded range `[from, to)` have no optimized path. Not a bug — but worth considering for API symmetry in a follow-up PR.

---

### Verification

The fix for the ByteStrings bug is verified correct:

```
// bytestrings = ["abb", "beb"], length=6, lastIndexOf(b, 4)
// With fix: find(1, 4, 6) → bs1.lastIndexOf(b, 4-3=1) → endIdx=min(1,2)=1 → finds b at local 0 → returns 0+3=3 ✓ (3 ≤ 4)
```

`sbt "actor-tests/testOnly org.apache.pekko.util.ByteStringSpec"` should pass with the new tests.

He-Pin (approved):
## Post-Fix Verification — LGTM

Verified all three `lastIndexOf` implementations against the fixes from previous review rounds.

### Correctness Verification

**ByteStrings coordinate-translation fix** — confirmed correct. Traced through:
- `ByteStrings(["ab", "dd"]).lastIndexOf('d', 2)`: `find(1, 2, 4)` → bs="dd", bsStartIndex=2, `"dd".lastIndexOf('d', 2-2=0)` → local 0 → returns 0+2=**2** ✓
- `ByteStrings(["abb", "efg"]).lastIndexOf('b', 4)`: `find(1, 4, 6)` → bs="efg", bsStartIndex=3, `"efg".lastIndexOf('b', 4-3=1)` → -1 → `find(0, 4, 3)` → bs="abb", bsStartIndex=0, `"abb".lastIndexOf('b', 4-0=4)` → endIdx=min(4,2)=2 → returns **2** ✓

**ByteString1 startIndex offset handling** — verified for the SWAR path. For `ByteString1.fromString("xxabcdefghijk").drop(2)` (startIndex=2, length=11):
- `unrolledLastIndexOf(startIndex + tailStart, ...)` correctly adds the offset
- `return index - startIndex` correctly converts back to logical coordinates
- `SWARUtil.getLong(bytes, startIndex + chunkStart, ...)` reads from the correct physical position

**ByteString1C SWAR boundary conditions** — all combinations verified:
- `searchLength=7`: tail-only path (no SWAR chunks) ✓
- `searchLength=8`: single SWAR chunk (no tail) ✓
- `searchLength=15`: 7-byte tail + 1 SWAR chunk ✓
- `searchLength=16`: 2 SWAR chunks (no tail) ✓

### Remaining Items (Non-blocking)

1. **`getLastIndex(0)` sentinel** (from previous review) — still returns undefined value for `word=0`, while `getIndex(0)` returns sentinel `8`. The Scaladoc makes the precondition clear and all callers guard correctly. Suggest adding a `word == 0` guard in a follow-up for API consistency — not blocking.

2. **Minor: `compilePattern` computed before short-circuit** — `SWARUtil.compilePattern(elem)` is called unconditionally in `ByteString1C`/`ByteString1.lastIndexOf`, even when `searchLength < 8` where only `unrolledLastIndexOf` runs. Consistent with existing `indexOf` implementation, so not a regression. Micro-optimization opportunity for a follow-up.

3. **Test coverage suggestion: ≥3 ByteStrings fragments** — All `ByteStrings` tests use exactly 2 fragments. Adding a test with 3+ fragments would verify the recursive `find` across multiple `bsStartIndex = len - bs.length` calculations. Low risk since the recursion is straightforward, but would be thorough.

### Test Coverage Assessment

| Scenario | Status |
|---|---|
| ByteString1C basic cases | ✅ Covered |
| ByteString1 with non-zero startIndex (`.drop()`) | ✅ Covered |
| ByteStrings with coordinate translation | ✅ Covered (the regression test) |
| SWAR code path (≥8 bytes) | ✅ Covered (16-byte test) |
| Edge byte values (0x00, -1) | ✅ Covered (short strings only) |
| SWAR boundary (tail + chunk transitions) | ✅ Implicitly covered (11-byte and 16-byte tests) |
| ≥3 ByteStrings fragments | ⚠️ Not covered |
| SWAR with edge bytes (0x00/0xFF in ≥8 byte string) | ⚠️ Not covered (low risk, SWAR math is byte-value agnostic) |

Overall the implementation is correct and well-structured. The SWAR `lastIndexOf` mirrors the existing `indexOf` pattern cleanly.

pjfanning:
Thanks @He-Pin . I'll look at the other review suggestions before proceeding.

PR #2850, opened 2026-04-08
add jspecify to persistence libs (javadsl/japi packages only)

pjfanning:
follow up to #2617 

* couldn't find much that explicitly support nulls
* could be more but in the end of the day, this is best effort
* if users report more cases where nulls are supported, we can update the annotations


He-Pin (approved):
lgtm

PR #2853, opened 2026-04-10
actor-typed: Fix ProducerController crash on restart (#2762)

pjfanning:
cherry pick 1c633dae0b423a77c745ac211bb3691af5ec511d #2762 

I am still wary of this change. It does seem like a bug fix to me but I'm equally concerned that it took this long to come to light. I'd really appreciate if we could get more reviews of it.


He-Pin (approved):
lgtm

He-Pin:
Can we have this on 1.5.x ?

pjfanning:
@He-Pin this is targeted at 1.6.0. I prefer the idea of the next release of 1.x being 1.6.0 as opposed to 1.5.1.

PR #2855, opened 2026-04-11
ByteString lastIndexOfSlice

pjfanning:
### New: `lastIndexOfSlice`

- **ByteString (abstract)**: `override def lastIndexOfSlice[B >: Byte](slice: Seq[B], end: Int): Int` — uses `lastIndexOf` on the **last byte** of the slice to find candidate end positions, then verifies the rest of the slice.
- **ByteString (abstract)**: `def lastIndexOfSlice(slice: Array[Byte], end: Int): Int` and `def lastIndexOfSlice(slice: Array[Byte]): Int` — same approach for `Array[Byte]` slices.

**Key optimization**: both implementations cap `end` to `min(end, length - sliceLength)` before computing the search start, preventing `Int` overflow when the parent class calls `lastIndexOfSlice(that, Int.MaxValue)`.

## Tests

- Added `"lastIndexOf (specialized)"` test block covering `ByteString1`, `ByteStrings`, and compact variants.
- Added `"lastIndexOfSlice"` test block (using `ByteString` seq type) and `"lastIndexOfSlice (specialized)"` (using `Array[Byte]`), covering `ByteString1`, `ByteStrings`, offset `ByteString1`, and edge cases.
- Fixes some issues with edge cases in `indexOfSlice`

## Benchmark

```
With Changes
[info] Benchmark                                                       Mode  Cnt         Score          Error  Units
[info] ByteString_lastIndexOfSlice_Benchmark.bs_lastIndexOfSlice_abc  thrpt    3  28105765.600 ± 76735569.699  ops/s
[info] ByteString_lastIndexOfSlice_Benchmark.bs_lastIndexOfSlice_hv   thrpt    3  30468747.357 ±  3858323.420  ops/s
[info] ByteString_lastIndexOfSlice_Benchmark.bss_lastIndexOfSlice     thrpt    3   2732170.943 ±  3114888.476  ops/s

Without Changes
[info] Benchmark                                                       Mode  Cnt         Score         Error  Units
[info] ByteString_lastIndexOfSlice_Benchmark.bs_lastIndexOfSlice_abc  thrpt    3   9101755.721 ± 3333167.328  ops/s
[info] ByteString_lastIndexOfSlice_Benchmark.bs_lastIndexOfSlice_hv   thrpt    3   7757911.197 ± 3422112.843  ops/s
[info] ByteString_lastIndexOfSlice_Benchmark.bss_lastIndexOfSlice     thrpt    3   1054397.115 ±  536082.169  ops/s
```

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -352,7 +352,7 @@ object ByteString {
       else if (bytes(fromIndex) == value) fromIndex
       else -1
     }
-    
+
     private def unrolledFirstIndexOf(fromIndex: Int, byteCount: Int, value: Byte): Int = {
       if (bytes(fromIndex) == value) fromIndex
       else if (byteCount == 1) -1
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -1295,7 +1295,7 @@ sealed abstract class ByteString
       true
     }
     val sliceLength = slice.length
-    if (sliceLength == 0) if (from >= length) -1 else math.max(from, 0)
+    if (sliceLength == 0) if (from > length) -1 else math.max(from, 0)
     else {
       val headByte = slice.head.asInstanceOf[Byte]
       @tailrec def rec(from: Int): Int = {
@@ -1333,7 +1333,7 @@ sealed abstract class ByteString
       true
     }
     val sliceLength = slice.length
-    if (sliceLength == 0) if (from >= length) -1 else math.max(from, 0)
+    if (sliceLength == 0) if (from > length) -1 else math.max(from, 0)
     else if (sliceLength > length) -1
     else {
       val headByte = slice.head
@@ -1454,7 +1454,7 @@ sealed abstract class ByteString
    *          to `slice`, or `-1`, if none exists.
    * @since 2.0.0
    */
-  def lastIndexOfSlice(slice: Array[Byte]): Int = lastIndexOfSlice(slice, length - 1)
+  def lastIndexOfSlice(slice: Array[Byte]): Int = lastIndexOfSlice(slice, length)
 
   override def contains[B >: Byte](elem: B): Boolean = indexOf(elem, 0) != -1
 
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -340,6 +340,19 @@ object ByteString {
       -1
     }
 
+    // Searches byteCount bytes (1-7) starting at fromIndex from highest to lowest index,
+    // returning the rightmost (last) match, or -1 if not found.
+    private def unrolledLastIndexOf(fromIndex: Int, byteCount: Int, value: Byte): Int = {
+      if (byteCount >= 7 && bytes(fromIndex + 6) == value) fromIndex + 6
+      else if (byteCount >= 6 && bytes(fromIndex + 5) == value) fromIndex + 5
+      else if (byteCount >= 5 && bytes(fromIndex + 4) == value) fromIndex + 4
+      else if (byteCount >= 4 && bytes(fromIndex + 3) == value) fromIndex + 3
+      else if (byteCount >= 3 && bytes(fromIndex + 2) == value) fromIndex + 2
+      else if (byteCount >= 2 && bytes(fromIndex + 1) == value) fromIndex + 1
+      else if (bytes(fromIndex) == value) fromIndex
+      else -1
+    }
+    
     private def unrolledFirstIndexOf(fromIndex: Int, byteCount: Int, value: Byte): Int = {
       if (bytes(fromIndex) == value) fromIndex
       else if (byteCount == 1) -1
@@ -354,19 +367,6 @@ object ByteString {
       else if (bytes(fromIndex + 5) == value) fromIndex + 5
       else if (byteCount == 6) -1
       else if (bytes(fromIndex + 6) == value) fromIndex + 6
-      else -1
-    }
-
-    // Searches byteCount bytes (1-7) starting at fromIndex from highest to lowest index,
-    // returning the rightmost (last) match, or -1 if not found.
-    private def unrolledLastIndexOf(fromIndex: Int, byteCount: Int, value: Byte): Int = {
-      if (byteCount >= 7 && bytes(fromIndex + 6) == value) fromIndex + 6
-      else if (byteCount >= 6 && bytes(fromIndex + 5) == value) fromIndex + 5
-      else if (byteCount >= 5 && bytes(fromIndex + 4) == value) fromIndex + 4
-      else if (byteCount >= 4 && bytes(fromIndex + 3) == value) fromIndex + 3
-      else if (byteCount >= 3 && bytes(fromIndex + 2) == value) fromIndex + 2
-      else if (byteCount >= 2 && bytes(fromIndex + 1) == value) fromIndex + 1
-      else if (bytes(fromIndex) == value) fromIndex
       else -1
     }
 
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -1294,17 +1294,19 @@ sealed abstract class ByteString
       }
       true
     }
-    val headByte = slice.head.asInstanceOf[Byte]
-    @tailrec def rec(from: Int): Int = {
-      val startPos = indexOf(headByte, from, length - slice.length + 1)
-      if (startPos == -1) -1
-      else if (check(startPos)) startPos
-      else rec(startPos + 1)
-    }
     val sliceLength = slice.length
-    if (sliceLength == 0) 0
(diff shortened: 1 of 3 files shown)


actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1353,6 +1353,101 @@ sealed abstract class ByteString
...
+      @tailrec def rec(currEnd: Int): Int = {
+        val endPos = lastIndexOf(tailByte, currEnd)
+        if (endPos < sliceLength - 1) -1
+        else {
+          val startPos = endPos - sliceLength + 1
+          if (check(startPos)) startPos
+          else rec(endPos - 1)
+        }
+      }
+      if (sliceLength == 1) lastIndexOf(tailByte, effectiveEnd)
```
He-Pin, line 1392: The  and  implementations are nearly identical — the only differences are:\n\n1.  vs  (no cast needed for Array)\n2.  access is the same syntax\n\nConsider extracting the common logic into a private helper method to avoid duplication. Something like:\n\n```scala\nprivate def lastIndexOfSliceCore(\n  sliceLength: Int, \n  tailByte: Byte, \n  end: Int, \n  getByte: Int => Byte\n): Int = {\n  // shared implementation\n}\n```\n\nThis would reduce maintenance burden and ensure bug fixes apply to both paths.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1353,6 +1353,101 @@ sealed abstract class ByteString
    */
   def indexOfSlice(slice: Array[Byte]): Int = indexOfSlice(slice, 0)
 
+  override def lastIndexOfSlice[B >: Byte](slice: scala.collection.Seq[B], end: Int): Int = {
+    val sliceLength = slice.length
```
He-Pin, line 1363: The  method is removed from lines 343-353 and re-added at lines 360-371 with **zero changes** to the implementation.\n\nThis appears to be an accidental reordering. Unless there's a specific reason for this move (e.g., it needs to be closer to a related method), consider reverting this change to keep the diff clean and focused on the actual feature.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1353,6 +1353,101 @@ sealed abstract class ByteString
...
+    if (sliceLength == 0) if (end < 0) -1 else math.min(end, length)
+    else if (sliceLength > length) -1
+    else {
+      val tailByte = slice(sliceLength - 1).asInstanceOf[Byte]
```
He-Pin, line 1367: The empty slice handling: `if (sliceLength == 0) if (end < 0) -1 else math.min(end, length)`\n\nThis returns `length` when `end >= length`. Is this consistent with Scala's standard library `Seq.lastIndexOfSlice` behavior? \n\nFor an empty slice, Scala's `Seq` typically returns the length of the collection (the position after the last element). But here we're capping it to `end` if `end < length`. Double-check this matches the parent class semantics to avoid subtle behavioral differences.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1353,6 +1353,101 @@ sealed abstract class ByteString
...
+      // Check all bytes of the slice except the last one (which was matched by lastIndexOf)
+      def check(startPos: Int): Boolean = {
+        var i = startPos
+        var j = 0
+        while (j < sliceLength - 1) {
+          if (apply(i) != slice(j)) return false
+          i += 1
+          j += 1
```
He-Pin, line 1376: Consider adding a bounds check or assertion in the `check` function for safety.\n\nThe comment says "Check all bytes of the slice except the last one (which was matched by lastIndexOf)" — but `apply(i)` could theoretically access out-of-bounds if `endPos` calculation is wrong.\n\nThe current code trusts that `endPos - sliceLength + 1` is valid, which should be guaranteed by the `endPos < sliceLength - 1` check below, but a defensive comment or an explicit bound would help future maintainers understand why this is safe.

bench-jmh/src/main/scala/org/apache/pekko/util/ByteString_lastIndexOfSlice_Benchmark.scala
```diff
@@ -0,0 +1,39 @@
...
+class ByteString_lastIndexOfSlice_Benchmark {
+  val start = ByteString("abcdefg") ++ ByteString("hijklmno") ++ ByteString("pqrstuv")
+  val bss = start ++ start ++ start ++ start ++ start ++ ByteString("xyz")
+
+  val bs = bss.compact // compacted
+  val abc = "abc".getBytes(StandardCharsets.UTF_8)
+  val hv = "hijklmnopqrstuv".getBytes(StandardCharsets.UTF_8)
+
+  @Benchmark
+  def bss_lastIndexOfSlice: Int = bss.lastIndexOfSlice(abc, 1)
```
He-Pin: The benchmark only tests cases with `end = 1`, which means the search is heavily constrained. \n\nConsider adding benchmarks that test:\n1. **Unconstrained search** (no `end` parameter, or `end = length - 1`) — worst case for the tail-byte optimization\n2. **Large ByteString** (e.g., 1MB+) — to show real-world performance\n3. **No-match scenario** — slice not present, forcing full scan\n4. **Match at beginning vs end** — to demonstrate best/worst case behavior\n\nThe PR description shows ~12x improvement for `bs_lastIndexOfSlice_abc`, but the constrained benchmark (`end=1`) may not represent typical usage patterns.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: changed

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: the new lastIndexOfSlice code seems ok - I added extra tests to verify it - indexOfSlice which is already added has issues and I've updated the tests and fixed the bugs in this area with that method

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: I added a comment instead, copied from indexOfSlice. These are private methods and the point is to be efficient and not add checks that are not needed in practice. I have added extra edge case tests.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: the code in ByteString has lots of duplication - I might revisit but this code right now is more in keeping with the existing code style in this class

bench-jmh/src/main/scala/org/apache/pekko/util/ByteString_lastIndexOfSlice_Benchmark.scala
pjfanning: the benchmark was not great - I've removed the index part in the benchmark

He-Pin (approved):
lgtm,  looking forward to see performacne update in pekko-http and pekko-grpc

pjfanning:
Thanks @He-Pin - I don't think this change will have much impact for Pekko other than you helping me find bugs in the indexOfSlice method that was added before but missing some edge case tests.

I still think it is a nice to have to fill in more performant implementations of various Seq functions that ByteString indirectly implements (but that the inherited implementations from Seq are inefficient for ByteString).

One other method that I would like to look at is endsWith but that is not urgent. endsWith has a default impl that uses drop and it would be more efficient to have it work by just looking at the bytes in the original ByteString.

He-Pin:
Yeah, btw the Scala3 stdlib contribution is open for new features, not sure if its a good addition there. https://contributors.scala-lang.org/t/standard-library-now-open-for-improvements-and-suggestions/7337/138
@pjfanning I think it would be nice to have the SWAR in stdlib

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1353,6 +1359,103 @@ sealed abstract class ByteString
...
 
+  override def lastIndexOfSlice[B >: Byte](slice: scala.collection.Seq[B], end: Int): Int = {
+    val sliceLength = slice.length
+    if (sliceLength == 0) if (end < 0) -1 else math.min(end, length)
+    else if (sliceLength > length) -1
+    else {
+      val tailByte = slice(sliceLength - 1).asInstanceOf[Byte]
+      // Check all bytes of the slice except the last one (which was matched by lastIndexOf)
+      def check(startPos: Int): Boolean = {
+        var i = startPos
```
He-Pin, line 1370: ⚡ **性能优化思路**: `lastIndexOfSlice` 从尾部字节开始搜索（`lastIndexOf(tailByte)`），这与 `indexOfSlice` 从头部字节搜索（`indexOf(headByte)`）形成对称。

对于高频率尾部重复字节（如日志中的 `\\n` 或二进制协议尾部填充），这种策略显著减少了 `check()` 的调用次数——benchmark 显示 3x 提升验证了这一点。

一个小观察：`effectiveEnd = math.min(end, length - sliceLength)` 的 overflow 防护很关键。父类 `Seq.lastIndexOfSlice` 传入 `Int.MaxValue`，如果没有这个 clamp，`end + sliceLength - 1` 会溢出。✅ 设计周到。

PR #2862, opened 2026-04-15
Add ByteString.endsWith and expand startsWith test coverage

pjfanning:
`ByteString` had `startsWith(Array[Byte])` but no `endsWith` equivalent, and the existing `startsWith` tests lacked edge-case coverage.

## New API

`endsWith(bytes: Array[Byte]): Boolean` added to the `ByteString` base class, mirroring `startsWith` semantics:

```scala
ByteString("abcdefghijklmnopqrstuvwxyz").endsWith("uvwxyz".getBytes) // true
ByteString("abcdefghijklmnopqrstuvwxyz").endsWith("12345".getBytes)  // false
ByteString.empty.endsWith(Array.emptyByteArray)                      // true
```

## Test improvements

- **`startsWith`**: added coverage for empty byte array (always `true`), exact-length match, bytes longer than the string (`false`), `ByteString1C`, and empty `ByteString`
- **`endsWith`**: full suite covering `ByteString1`, `ByteString1C`, `ByteStrings` (multi-segment, including cross-boundary suffix), `ByteString1` with internal offset, empty byte array, exact match, oversized input, and empty `ByteString`

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -374,6 +374,44 @@ object ByteString {
       else -1
     }
 
+    override def startsWith(bytes: Array[Byte], offset: Int): Boolean = {
+      val needleLen = bytes.length
+      if (length - offset < needleLen) return false
+      var hIdx = offset
+      var nIdx = 0
+      while (needleLen - nIdx >= 8) {
+        if (SWARUtil.getLong(this.bytes, hIdx, ByteOrder.BIG_ENDIAN) !=
+            SWARUtil.getLong(bytes, nIdx, ByteOrder.BIG_ENDIAN)) return false
+        hIdx += 8
+        nIdx += 8
+      }
+      while (nIdx < needleLen) {
+        if (this.bytes(hIdx) != bytes(nIdx)) return false
+        hIdx += 1
+        nIdx += 1
+      }
+      true
+    }
+
+    override def endsWith(bytes: Array[Byte]): Boolean = {
+      val needleLen = bytes.length
+      if (length < needleLen) return false
+      var hIdx = length - needleLen
+      var nIdx = 0
+      while (needleLen - nIdx >= 8) {
+        if (SWARUtil.getLong(this.bytes, hIdx, ByteOrder.BIG_ENDIAN) !=
+            SWARUtil.getLong(bytes, nIdx, ByteOrder.BIG_ENDIAN)) return false
+        hIdx += 8
+        nIdx += 8
+      }
+      while (nIdx < needleLen) {
+        if (this.bytes(hIdx) != bytes(nIdx)) return false
+        hIdx += 1
+        nIdx += 1
+      }
+      true
+    }
+
     override def slice(from: Int, until: Int): ByteString =
       if (from <= 0 && until >= length) this
       else if (from >= length || until <= 0 || from >= until) ByteString.empty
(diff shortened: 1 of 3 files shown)


pjfanning:
I intend to look at adding a set of JMH benchmarks.
The existing impl inherited from Seq class involves creating iterators and doing drops and I would expect that this comes with a performance penalty for ByteString.

pjfanning:
```
With Changes
[info] Benchmark                                          Mode  Cnt         Score          Error  Units
[info] ByteString_startEnd_Benchmark.bs_endsWith         thrpt    3  70943370.602 ± 18942924.748  ops/s
[info] ByteString_startEnd_Benchmark.bs_endsWithBytes    thrpt    3  96973711.680 ±  7171877.362  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWith       thrpt    3  64399277.889 ±  6238656.895  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWithBytes  thrpt    3  91607927.878 ± 16655608.705  ops/s
[info] ByteString_startEnd_Benchmark.bss_endsWith        thrpt    3   5540020.658 ±   384005.497  ops/s
[info] ByteString_startEnd_Benchmark.bss_startsWith      thrpt    3  26458399.812 ±  3425486.309  ops/s

Without Changes
[info] Benchmark                                          Mode  Cnt         Score           Error  Units
[info] ByteString_startEnd_Benchmark.bs_endsWith         thrpt    3  60443574.756 ±  37556938.021  ops/s
[info] ByteString_startEnd_Benchmark.bs_endsWithBytes    thrpt    3  44665163.692 ±  31893558.974  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWith       thrpt    3  62253099.716 ±  11052746.775  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWithBytes  thrpt    3  99282620.436 ± 247522926.430  ops/s
[info] ByteString_startEnd_Benchmark.bss_endsWith        thrpt    3    354955.743 ±     30325.695  ops/s
[info] ByteString_startEnd_Benchmark.bss_startsWith      thrpt    3   1961839.823 ±     81235.164  ops/s
```

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1473,6 +1473,16 @@ sealed abstract class ByteString
    */
   def contains(elem: Byte): Boolean = indexOf(elem, 0) != -1
 
+  override def startsWith[B >: Byte](iterable: scala.collection.IterableOnce[B], offset: Int): Boolean = {
```
He-Pin: The `startsWith(iterable, offset)` override doesn't validate negative `offset` values. If `offset < 0`, the while loop condition `i >= length` won't catch it immediately, and `apply(i)\) will throw IndexOutOfBoundsException. 

The existing `startsWith(bytes: Array[Byte], offset: Int)\) method at line 1494 also doesn't guard against negative offset - if `offset < 0\), the expression `length - offset < bytes.length` could overflow or give incorrect results.

Consider adding: `if (offset < 0) throw new IndexOutOfBoundsException(...)` or at least document the behavior.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1506,6 +1516,41 @@ sealed abstract class ByteString
    */
   def startsWith(bytes: Array[Byte]): Boolean = startsWith(bytes, 0)
 
+  override def endsWith[B >: Byte](iterable: scala.collection.Iterable[B]): Boolean = {
```
He-Pin: The `endsWith(iterable)\) override uses `iterable.size` which traverses the entire iterable for non-sized collections. For `IterableOnce` this would require traversal, but you're using `Iterable[B]` here which does have `size` - however, for lazy iterables or streams passed to this method, it could trigger full evaluation.

More importantly: the parent trait signature from `scala.collection.IterableOps` expects `def endsWith[B >: Byte](that: Iterable[B]): Boolean`. If there's a performance-sensitive use case where someone passes a very long iterable, consider whether this should be documented or if early-exit optimization when `size > length` is beneficial before calling `iterable.size`.

bench-jmh/src/main/scala/org/apache/pekko/util/ByteString_startEnd_Benchmark.scala
```diff
@@ -0,0 +1,48 @@
...
+  @Benchmark
+  def bss_startsWith: Boolean = bss.startsWith("abc")
+
+  @Benchmark
+  def bss_endsWith: Boolean = bss.endsWith("xyz")
+
+  @Benchmark
+  def bs_startsWith: Boolean = bss.startsWith("abc")
+
+  @Benchmark
```
He-Pin: Benchmark methods `bs_startsWith` and `bs_endsWith` use `bss` (uncompacted) instead of `bs` (compacted). The variable names suggest they should test the compacted `bs` version, but they're actually testing `bss`. This appears to be a copy-paste error.

Lines 38-41:
```scala
@Benchmark
def bs_startsWith: Boolean = bss.startsWith("abc")  // should be bs.startsWith?

@Benchmark  
def bs_endsWith: Boolean = bs.endsWith("xyz")       // this one is correct
```

pjfanning:
@He-Pin changes made and benchmarks rerun

He-Pin (commented):
Resolved

He-Pin (approved):

PR #2863, opened 2026-04-15
small optimisation on indexOf

pjfanning:
* small change from #2859
* lastIndexOf already works this way
* We have an overload for indexOf that accepts Byte input directly and the compiler will prefer this if you have a Byte that you want to match
* but a number of Seq methods use indexOf under the hood and they will use the Seq API call so it is good to pattern match and forward Byte inputs to the dedicated overload method

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -234,14 +234,18 @@ object ByteString {
       else toByteString1.drop(n)
 
     override def indexOf[B >: Byte](elem: B, from: Int): Int = {
-      if (from >= length) -1
-      else {
-        var i = math.max(from, 0)
-        while (i < length) {
-          if (bytes(i) == elem) return i
-          i += 1
-        }
-        -1
+      elem match {
+        case byte: Byte => indexOf(byte, from)
+        case _          =>
+          if (from >= length) -1
+          else {
+            var i = math.max(from, 0)
+            while (i < length) {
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -234,14 +234,18 @@ object ByteString {
...
-      if (from >= length) -1
-      else {
-        var i = math.max(from, 0)
-        while (i < length) {
-          if (bytes(i) == elem) return i
-          i += 1
-        }
-        -1
+      elem match {
+        case byte: Byte => indexOf(byte, from)
```
He-Pin, line 238: Will this reduce the performance? as it add another instance Of  and checkedcast call

pjfanning: So if your code uses indexOf directly, the compiler will bind directly to `indexOf(byte, from)`.
I don't think this will have a big slowdown for other use cases. We already have this code in lastIndexOf so in the end, we need to keep them consistent. That means adding this change or removing the code from lastIndexOf. In the end, it was Copilot/Claude that suggested this in #2859 and it documented it as 'improving performance for Byte inputs (avoids boxing)'.

He-Pin (approved):
LGTM. The pattern match routing Byte inputs to the SWAR-optimized indexOf(byte, from) is a valid optimization, consistent with lastIndexOf's existing approach. The isInstanceOf overhead for non-Byte inputs is negligible compared to the 8x speedup for the common Byte case.

PR #2865, opened 2026-04-16
Remove implicit JUnit4 dependency from LogCapturingExtension

### LINKED ISSUE #2864: LogCapturingExtension depends on JUnit4
JUnit5 extension LogCapturingExtension uses Logger for LogCapturing that is for JUnit4. So if you use the extension you have to add extra dependecies on JUnit4.

He-Pin:
## Summary

Fixes #2864

`LogCapturingExtension` is a JUnit 5 `InvocationInterceptor`, but it used `classOf[LogCapturing]` to initialize its logger. Since both classes are in the same package (`javadsl`), this resolves to `javadsl.LogCapturing` — which is the JUnit 4 `TestRule` class. This forces pure JUnit 5 users to also include JUnit 4 on their classpath.

## Changes

- Changed `classOf[LogCapturing]` to `classOf[LogCapturingExtension]` in `LogCapturingExtension.scala` line 27
- This removes the implicit JUnit 4 class reference from the JUnit 5 extension

## Testing

- [x] Code formatted with scalafmt
- [ ] Full CI tests pending


pjfanning (approved):
lgtm

PR #2871, opened 2026-04-18
feat: make ForkJoinPool minimum-runnable configurable and improve pool documentation

He-Pin:
## Summary

Two targeted improvements to the ForkJoinPool executor configuration — no behavior change by default, purely additive.

### 1. Configurable `minimum-runnable`

`actor/src/main/resources/reference.conf`

```hocon
fork-join-executor {
  # Default = 1 (matches JDK ForkJoinPool prior behavior)
  minimum-runnable = 1
}
```

`actor/src/main/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfigurator.scala`

- `PekkoForkJoinPool` now accepts `minimumRunnable: Int = 1`
- `ForkJoinExecutorServiceFactory` reads `minimum-runnable` from config
- Default `1` is identical to the previously hardcoded value — **no behavior change**
- Users on JDK 21+ experiencing ForkJoinPool starvation (see #2870) can increase this (e.g. to `parallelism / 2`) to prompt faster compensation-thread creation

**Binary compatibility**: new params carry default values; class is `@InternalApi`, so MiMa exclusion is not required.

### 2. Improved `maximum-pool-size` documentation

The existing comment was minimal. The new comment explains that on JDK 21+:
- Compensation threads are created up to `maximum-pool-size`  
- Tuning this alongside `minimum-runnable` helps under heavy actor-round-trip workloads

## What this PR does NOT change

- Default dispatcher mode remains FIFO (no LIFO change)
- Stream test dispatcher is unchanged
- Default `minimum-runnable = 1` preserves existing behaviour

## Related

- #2870 — root-cause tracking issue (JDK-8300995 ForkJoinPool regression)
- #2869 — test stability fixes (already merged)
- apache/pekko#2872 — opt-in `virtualize = on` investigation

Issue #2870, opened 2026-04-18
ForkJoinPool compensation-thread regression (JDK-8300995) causes test flakiness on JDK 21+


He-Pin:
## Summary

The Pekko nightly CI has observed intermittent test failures on JDK 21 and JDK 25 that trace to a ForkJoinPool behavioral regression in JDK 21+ ([JDK-8300995](https://bugs.openjdk.org/browse/JDK-8300995), [JDK-8321335](https://bugs.openjdk.org/browse/JDK-8321335)).

This issue documents the root cause and proposed mitigations for Pekko's stream tests. Related: #2573.

---

## Root Cause

Pekko's default `fork-join-executor` uses `asyncMode=true` (FIFO, via `task-peeking-mode = "FIFO"`). In JDK 21+, this mode interacts poorly with actor round-trip workloads:

1. When actor A sends a message to actor B and waits for a reply, the reply task is placed at the **back** of the FIFO work queue.
2. With a fixed-size pool (e.g. 8 threads) under load, the reply waits behind all pending tasks before being executed.
3. The compensation-thread mechanism (which should unblock this) was [regressed in JDK 21](https://bugs.openjdk.org/browse/JDK-8300995) for asyncMode=FIFO pools — compensation threads are not reliably created when needed.
4. This cascades for workloads with many actor hand-offs (e.g. `MergeHub buffer=1`: one full round-trip per element).

JDK 25 exacerbates the issue because work-stealing heuristics under high CI load diverge further from expected throughput.

---

## Evidence

- `ForkJoinPoolStarvationSpec` has been marked `pending` on JDK 17+ acknowledging this class of issue
- `stream-typed-tests/MapAsyncPartitionedSpec.scala:110-116` already has a comment referencing #2573: "JDK 25 makes 1000-sample property checks noticeably more expensive"
- Nightly CI for JDK 25 was failing for 30+ consecutive days before the fix in PR #2869

---

## Mitigations Applied (PR #2869)

1. **Stream test dispatcher**: Changed `task-peeking-mode` from `FIFO` to `LIFO` in `stream-testkit/src/test/resources/reference.conf`. With LIFO, response tasks go to the **front** of each worker's deque and are processed immediately. Actor mailbox ordering is unaffected (`UnboundedMailbox` is always FIFO).

[...]

PR #2869, opened 2026-04-18
fix: stabilise stream tests on JDK 25 nightly (timeout scaling, element counts)

He-Pin:
## Summary

Fixes 3 categories of stream-test flakiness observed on JDK 25 nightly CI (30+ consecutive days failing). See #2870 for root-cause analysis; see #2871 for the configuration-level mitigation.

## Failing tests fixed

| Test | Root cause | Fix |
|------|-----------|-----|
| `HubSpec` – "must work with long streams" (×4 variants) | 20 K elements × FJP FIFO delay = 60+ s timeout | Reduced element counts; remove `throttle` in favour of `Thread.sleep(1)` |
| `AggregateWithTimeBoundaryAndSimulatedTimeSpec` | `interval = 1.milli` races with scheduler | Changed to `interval = 1.second` |
| TCK `stochastic_spec103_mustSignalOnMethodsSequentially` | Hardcoded 1 s timeout ignores timefactor | Reads `pekko.test.timefactor` system property |
| `FlowMapAsyncPartitionedSpec` – "must ignore null-completed futures" | `Random.nextInt(10)` can produce 0, skipping null path | Shift to `Random.nextInt(10) + 1` |

## Files changed

- `stream-tests/…/HubSpec.scala` — PatienceConfig scaled by timefactor; element counts reduced; throttle → sleep
- `stream-tests/…/AggregateWithBoundarySpec.scala` — interval 1.milli → 1.second (2 locations)
- `stream-tests-tck/…/Timeouts.scala` — read `pekko.test.timefactor` JVM property
- `stream-tests/…/FlowMapAsyncPartitionedSpec.scala` — always exercise null-future path

## Related

- #2870 — JDK-8300995 root cause tracking issue
- #2871 — configuration-level improvements (LIFO dispatcher + configurable minimum-runnable)

PR #2872, opened 2026-04-18
feat: enable virtualize=on in stream test dispatcher to bypass JDK 21+ FJP scheduling regression

He-Pin:
## Summary

Enables `virtualize = on` in the Pekko stream test dispatcher so that on JDK 21+ each dispatcher task runs in its own virtual thread (Project Loom). This **bypasses** the ForkJoinPool compensation-thread scheduling regression (JDK-8300995) at the root rather than working around it.

> ⚠️ **Depends on #2869** (test stability fixes). Retarget this PR to `main` after #2869 merges.

---

## Root Cause Recap

ForkJoinPool with `asyncMode=FIFO` on JDK 21+ places response tasks at the back of the deque, causing actor reply futures to queue behind unrelated tasks. Compensation threads that should keep worker count stable became more conservative in JDK-8321335 (JDK 21) and have not improved in JDK 25. This manifests as cascading latency spikes in any test with tight actor round-trips (MergeHub, HubSpec, etc.).

## How `virtualize = on` Fixes It

With `virtualize = on`, each dispatcher `Runnable` is scheduled as a virtual thread (JDK 21+ Project Loom). Virtual threads **unmount** their carrier when they block (e.g., awaiting an actor reply), freeing the carrier for other work. The FJP pool's FIFO-ordering starvation cannot happen because:

- There are no blocked workers holding carriers
- Compensation threads are never needed for actor-reply waits
- The FJP pool is used only as a **carrier scheduler**, not to queue Runnables

## Changes

`stream-testkit/src/test/resources/reference.conf`

```hocon
pekko.test.stream-dispatcher {
  fork-join-executor {
    # Enable virtual threads on JDK 21+. Silently ignored on JDK < 21.
    virtualize = on
  }
}
```

## Safety Notes

- On JDK < 21: `VirtualThreadSupport.isSupported = false` → flag silently ignored, pool behaves identically to today
- `--add-opens` flags required by Loom are already supplied by `JdkOptions.scala` for JDK 9+
- Mailbox ordering is unaffected (UnboundedMailbox is always FIFO)
- Verified: 48/48 HubSpec tests pass in ~20 seconds on JDK 25 (was 2+ minutes timeout previously)

## Relation to Other PRs

| PR | What it does |
|----|-------------|
| #2869 | Test stability fixes (patience scaling, element counts, TCK timefactor) |
| #2871 | FJP config improvements (LIFO, configurable minimum-runnable, max-pool-size docs) |
| **this PR** | Root-cause fix: virtual threads bypass FJP scheduling entirely |
| #2870 | Tracking issue documenting the JDK-8300995 root cause |

### DIFF
--- a/stream-testkit/src/test/resources/reference.conf
+++ b/stream-testkit/src/test/resources/reference.conf
@@ -14,6 +14,13 @@ pekko.test.stream-dispatcher {
   fork-join-executor {
     parallelism-min = 8
     parallelism-max = 8
+    # Enable virtual threads on JDK 21+. Virtual threads (Project Loom) bypass the
+    # ForkJoinPool compensation-thread starvation issue (JDK-8300995) that causes
+    # spurious test timeouts on JDK 21+ when actors block on reply futures in FIFO mode.
+    # On JDK < 21 this flag is silently ignored (VirtualThreadSupport.isSupported = false).
+    # The required --add-opens flags (java.base/jdk.internal.misc, java.base/java.lang)
+    # are already supplied by JdkOptions.scala for JDK 9+.
+    virtualize = on
   }
   mailbox-requirement = "org.apache.pekko.dispatch.UnboundedMessageQueueSemantics"
 }
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -148,10 +148,10 @@ jobs:
 
       - name: Compile and Test
         # note that this is not running any multi-jvm tests because multi-in-test=false
-        # JDK 25 ForkJoinPool scheduling changes need a higher timefactor (see #2573)
+        # JDK 25 ForkJoinPool scheduling changes need a higher timefactor (see #2573, #2870)
         run: |-
           if [ "${{ matrix.javaVersion }}" -ge 25 ]; then
-            TIMEFACTOR=3
+            TIMEFACTOR=4
           else
             TIMEFACTOR=2
           fi
--- a/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/Timeouts.scala
+++ b/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/Timeouts.scala
@@ -18,10 +18,14 @@ package org.apache.pekko.stream.tck
  */
 object Timeouts {
 
+  // Scale timeouts by pekko.test.timefactor (set to 3 on JDK 25 nightly builds).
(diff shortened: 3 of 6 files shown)


He-Pin:
## 🔬 Key Finding & Decision

### What we measured

Enabling `virtualize = on` in the stream test dispatcher produced a dramatic speedup:

| Test suite | Without virtual threads | With virtual threads |
|-----------|------------------------|---------------------|
| HubSpec (48 tests) | 2+ min timeout (3 failures) | **~20 seconds, 48/48 pass** |

Virtual threads completely bypass the ForkJoinPool FIFO-starvation issue (JDK-8300995) because blocked virtual threads **unmount** their carrier, so compensation threads are never needed for actor-reply waits.

### Why we're NOT merging this as a default

The concern is sound: silently switching the stream test dispatcher to virtual threads changes the runtime model — any test that implicitly relies on FJP scheduling behavior (thread affinity, carrier pinning, blocking detection) could produce false positives or mask real issues. Enabling it only in CI for JDK 21+ would be a hidden configuration divergence.

### Recommended path

1. **Opt-in per test config**: teams investigating virtual-thread performance can add `virtualize = on` to their own `reference.conf` override. The support is already present in Pekko v1.2.0+.
2. **Future work**: a dedicated `VirtualThreadDispatcherSpec` that explicitly opts in and asserts virtual-thread behavior is a better way to validate this path without affecting all stream tests. Tracked in #2870.

Closing in favour of documenting in #2870.

He-Pin:
@pjfanning I want to turn this on in the nightly build for Java 25

stream-testkit/src/test/resources/reference.conf
```diff
@@ -14,6 +14,13 @@ pekko.test.stream-dispatcher {
   fork-join-executor {
     parallelism-min = 8
     parallelism-max = 8
+    # Enable virtual threads on JDK 21+. Virtual threads (Project Loom) bypass the
```
pjfanning, line 17: could this be moved to the application.conf file(s) where we might still have test issues - instead of making it a global setting that affects any test that uses the stream-testkit?

He-Pin: Yes, this can be done. we can enable it just for Java 25 and Java 21

He-Pin:
Superseded by PR #2881 which includes enhanced virtual thread support with environment variable conditioning and developer documentation.

Issue #2860, opened 2026-04-14
Rewritten more to graphstage


He-Pin:
I think we should rewrite more stream operators to graphstages, there are some are not.

He-Pin:
## Analysis: Legacy Stream Components That Can Be Refactored to GraphStage

After analyzing the codebase, here is a structured inventory of components still using the legacy `Pump`/`InputBunch`/`OutputBunch`/`TransferPhase` infrastructure that could be refactored to modern `GraphStage` equivalents.

---

### Component Analysis Table

| # | Component | File | Current Pattern | Legacy Dependencies | Shape | Refactor Priority | Complexity | Fusion Benefit |
|---|-----------|------|-----------------|---------------------|-------|-------------------|------------|----------------|
| 1 | **TLSActor** | `impl/io/TLSActor.scala` | `Actor + Pump` | `InputBunch` (2), `OutputBunch` (2), `TransferPhase`, custom `TransferState` | `BidiShape` | **High** | Very High | Low |
| 2 | **Unzip (Legacy)** | `impl/FanOut.scala:328` | `FanOut + Pump` | `OutputBunch`, `BatchingInputBuffer`, `TransferPhase` | `FanOutShape2` | **High** | Low | High |
| 3 | **FanoutProcessorImpl** | `impl/FanoutProcessor.scala:135` | `ActorProcessorImpl + Pump` | `BatchingInputBuffer`, `FanoutOutputs`, `TransferPhase` | `FlowShape` | **High** | Medium | Medium |
| 4 | **FanoutPublisherSink** | `impl/Sinks.scala:114` | `SinkModule` → creates `FanoutProcessorImpl` | Depends on #3 | `SinkShape` | Medium | Medium | Medium |
| 5 | **FanIn (Base Class)** | `impl/FanIn.scala` | `Actor + Pump` | `InputBunch`, `SimpleOutputs` | Variable | Low | Medium | High |
| 6 | **FanOut (Base Class)** | `impl/FanOut.scala:273` | `Actor + Pump` | `OutputBunch`, `BatchingInputBuffer` | Variable | Low | Medium | High |
| 7 | **ActorProcessorImpl** | `impl/ActorProcessor.scala:272` | `Actor + Pump` | `BatchingInputBuffer`, `SimpleOutputs` | `FlowShape` | Low | Medium | Low |
[...]

PR #2873, opened 2026-04-18
chore(stream): remove internal FanOut Unzip

He-Pin:
## Summary
This is the first small slice for #2860.

It removes the dead internal `org.apache.pekko.stream.impl.Unzip` implementation from `FanOut.scala` and adds the matching `2.0.x` MiMa exclusions for the deleted binary-visible symbols.

The public Scala/Java DSL `Unzip` is already GraphStage-based, so this PR only drops the stale internal actor-backed implementation. This cleanup is part of Apache Pekko, which is now Apache licensed.

## Motivation
- make #2860 progress in the smallest safe step
- remove verified dead internal legacy code before larger runtime migrations
- keep the cleanup independently reviewable and independently releasable

## Modification
- delete the internal `impl.Unzip` object/class from `stream/src/main/scala/org/apache/pekko/stream/impl/FanOut.scala`
- add `stream/src/main/mima-filters/2.0.x.backwards.excludes/remove-internal-unzip.excludes`

## Result
- the dead internal Unzip path is removed
- public `Unzip` behavior remains unchanged
- this slice stands on its own as a tiny cleanup PR before the larger fanout/TLS work

## Validation
- `sbt scalafmtAll`
- `sbt "stream/test:compile"`
- `sbt "stream/mimaReportBinaryIssues"`
- `sbt "stream-tests/testOnly org.apache.pekko.stream.scaladsl.GraphUnzipSpec"`

## Upstream / references
- Issue: https://github.com/apache/pekko/issues/2860

### DIFF
--- a/stream/src/main/mima-filters/2.0.x.backwards.excludes/remove-internal-unzip.excludes
+++ b/stream/src/main/mima-filters/2.0.x.backwards.excludes/remove-internal-unzip.excludes
@@ -1,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #2874, opened 2026-04-18
feat(stream): replace fanout publisher runtime with GraphStage bridge

He-Pin:
## Summary
This is the second #2860 slice.

It replaces the actor-backed runtime behind `Sink.asPublisher(fanout = true)` with `FanoutPublisherBridgeStage`, removes the legacy `FanoutProcessor.scala`, replaces the old implementation-bound spec with `FanoutPublisherBehaviorSpec`, and adds the matching `2.0.x` MiMa exclusions.

The runtime stays on a dedicated internal fanout bridge and preserves the existing terminal-signal contract already exercised by `FlowSpec`. 

## Motivation
- remove the remaining legacy actor-backed fanout runtime from the fanout publisher path
- keep `Sink.asPublisher(fanout = true)` behavior-compatible while moving the runtime to GraphStage-based stream infrastructure
- replace implementation-bound actor tests with black-box behavior coverage and TCK-backed validation

## Modification
- rewire `FanoutPublisherSink.create` in `stream/src/main/scala/org/apache/pekko/stream/impl/Sinks.scala`
- add `stream/src/main/scala/org/apache/pekko/stream/impl/FanoutPublisherBridgeStage.scala`
- delete `stream/src/main/scala/org/apache/pekko/stream/impl/FanoutProcessor.scala`
- replace `stream-tests/src/test/scala/org/apache/pekko/stream/impl/FanoutProcessorSpec.scala` with `stream-tests/src/test/scala/org/apache/pekko/stream/impl/FanoutPublisherBehaviorSpec.scala`
- add `stream/src/main/mima-filters/2.0.x.backwards.excludes/remove-fanout-processor.excludes`
- add a follow-up scalafmt-only commit on `FanoutPublisherBridgeStage.scala` to satisfy the GitHub `Code is formatted` check

## Result
- the fanout publisher path no longer depends on `FanoutProcessorImpl`
- timeout, late-subscriber, downstream-failure, and multi-subscriber delivery behavior remain covered
- the slice is independently reviewable and leaves the TLS migration as the next focused step
- the PR branch now includes the CI formatting-only follow-up needed by the GitHub scalafmt diff-ref workflow

## Validation
- `sbt scalafmtAll`
- `sbt "stream/test:compile"`
- `sbt "stream/mimaReportBinaryIssues"`
- `sbt "stream-tests/testOnly org.apache.pekko.stream.impl.FanoutPublisherBehaviorSpec org.apache.pekko.stream.impl.TimeoutsSpec org.apache.pekko.stream.scaladsl.FlowSpec org.apache.pekko.stream.scaladsl.SinkSpec"`
- `sbt "stream-tests/testOnly org.apache.pekko.stream.javadsl.SinkTest"`
- `sbt "stream-tests-tck/testOnly org.apache.pekko.stream.tck.FanoutPublisherTest"`
- `sbt "scalafmtOnly stream/src/main/scala/org/apache/pekko/stream/impl/FanoutPublisherBridgeStage.scala"`

## Upstream / references
- Issue: https://github.com/apache/pekko/issues/2860
- Related prior exploration: https://github.com/apache/pekko/pull/1705 and https://github.com/akka/akka-core/issues/21261


pjfanning (approved):
lgtm

stream/src/main/scala/org/apache/pekko/stream/impl/FanoutPublisherBridgeStage.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  private val pendingSubscribers = new AtomicReference[immutable.Seq[Subscriber[_ >: T]]](Nil)
+  private val shutdownStarted = new AtomicBoolean(false)
+
+  @volatile private var shutdownReason: Option[Throwable] = None
+
+  override def subscribe(subscriber: Subscriber[_ >: T]): Unit = {
+    requireNonNullSubscriber(subscriber)
+
+    @tailrec def doSubscribe(): Unit = {
```
He-Pin, line 196: Long-term safety risk: \`FanoutPublisherBridgePublisher\` holds a reference to \`registerPendingSubscribers: AsyncCallback[Unit]\`, which internally holds a reference to the \`GraphStageLogic\`. If the \`Publisher\` is retained by user code after the stage terminates, this creates a reference chain that prevents GC of the entire stage logic and all its state (buffer, subscriptions, callbacks). Consider using a weak reference pattern or clearing the callback reference in \`shutdown()\` to break this chain for long-running applications.

pjfanning: would this be possible to add?

He-Pin: I think there is no need to fix this.

PR #2875, opened 2026-04-19
fix: eliminate it.isEmpty and double-traversal risk in CompactByteString.apply(IterableOnce[Byte])

pjfanning:
`CompactByteString.apply(bytes: IterableOnce[Byte])` called `it.isEmpty` (backed by `IterableOnce.isEmpty` → `!iterator.hasNext`) before `it.toArray`, creating fragility for custom `IterableOnce` implementations where `isEmpty` could consume or corrupt iterator state, and leaving a latent double-traversal footgun.

## Changes

- **`ByteString.scala`** — Rewrite `CompactByteString.apply(IterableOnce[Byte])` to materialize the iterator into an array in a single pass, then check emptiness on the array:

```scala
// Before
val it = bytes.iterator
if (it.isEmpty) empty
else ByteString.ByteString1C(it.toArray)

// After
val iter = bytes.iterator
if (iter.hasNext) ByteString.ByteString1C(iter.toArray) else empty
```

- **`ByteStringSpec.scala`** — Add a dedicated `CompactByteString.apply(IterableOnce[Byte])` test section covering:
  - Empty `Seq`, `List`, and `Iterator` inputs return the canonical `CompactByteString.empty` instance
  - Non-empty inputs produce correct bytes
  - Result satisfies `isCompact`
  - Custom single-use `IterableOnce` (both empty and non-empty) verifies `iterator` is called exactly once

Agent-Logs-Url: https://github.com/pjfanning/incubator-pekko/sessions/2f07e80b-5dd8-4504-9af4-8bcc5692f3d6

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -2031,8 +2031,8 @@ object CompactByteString {
    * Creates a new CompactByteString by traversing bytes.
    */
   def apply(bytes: IterableOnce[Byte]): CompactByteString = {
-    val arr = bytes.iterator.toArray
-    if (arr.isEmpty) empty else ByteString.ByteString1C(arr)
+    val iter = bytes.iterator
+    if (iter.hasNext) ByteString.ByteString1C(iter.toArray) else empty
   }
 
   /**
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -2031,9 +2031,8 @@ object CompactByteString {
    * Creates a new CompactByteString by traversing bytes.
    */
   def apply(bytes: IterableOnce[Byte]): CompactByteString = {
-    val it = bytes.iterator
-    if (it.isEmpty) empty
-    else ByteString.ByteString1C(it.toArray)
+    val arr = bytes.iterator.toArray
+    if (arr.isEmpty) empty else ByteString.ByteString1C(arr)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


He-Pin (approved):

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -2031,9 +2031,8 @@ object CompactByteString {
    * Creates a new CompactByteString by traversing bytes.
    */
   def apply(bytes: IterableOnce[Byte]): CompactByteString = {
-    val it = bytes.iterator
-    if (it.isEmpty) empty
-    else ByteString.ByteString1C(it.toArray)
+    val arr = bytes.iterator.toArray
```
He-Pin: maybe we can just test iterator.hasNext and avoid the toArray when it''s empty

pjfanning: there are constants for things like empty arrays so I would expect that there is no overhead like creating an empty array instance is going to happen here

He-Pin: `bytes.iterator.toArray` creates a fresh array each time, so passing it directly to `ByteString1C` is safe (no aliasing). However, `Iterator.toArray` on Scala 2.13 uses a `ArrayBuilder` under the hood which may over-allocate and then trim. For very large inputs this adds GC pressure. Consider whether the hot path should have a specialized overload for `Array[Byte]` or `IndexedSeq[Byte]` that avoids the builder overhead. This is not a correctness issue, just a performance consideration for high-throughput scenarios.

pjfanning: there is already a dedicated `apply` overload for `Array[Byte]`

pjfanning: change made

PR #2878, opened 2026-04-20
feat(stream): add TLS GraphStage engine

He-Pin:
Adds an opt-in GraphStage TLS engine for Pekko Streams while keeping the existing legacy actor path as the default.

### Motivation

The current Streams TLS route depends on the legacy actor/FanoutProcessor infrastructure. This PR adds a GraphStage route for the stream internals, but keeps the long-standing TLSActor route as the default compatibility baseline.

The new GraphStage implementation is a migration of Pekko's existing TLS pump state machine into GraphStage lifecycle and port handling. Shared low-level SSLEngine helpers are factored out so the legacy actor and GraphStage path keep the same record-buffer preparation and delegated-task behavior.

### Implementation

- Add `TlsGraphStage`, a `BidiShape[SslTlsOutbound, ByteString, ByteString, SslTlsInbound]` GraphStage.
- Preserve the existing TLS pump phases in GraphStage form: bidirectional transfer, outbound flushing, close-awaiting, outbound-closed, and inbound-closed.
- Reuse Pekko's TCP direct `BufferPool` for TLS transport packet buffers via `Tcp(materializer.system).bufferPool`.
- Size application buffers from the active `SSLSession` packet/application buffer sizes instead of fixed magic constants.
- Keep the legacy actor path as the default; legacy actor changes are limited to shared helper use and TLS 1.3 close/unwrap correctness.
- Add two-level batching for the GraphStage path:
  - small plaintext writes are batched before wrapping;
  - transport writes are consolidated with a size guard so large TLS records flush immediately.
- Apply an async boundary and input buffer around the GraphStage engine from `TLS.apply`.
- Add `pekko.stream.materializer.tls.engine` with valid values:
  - `legacy-actor` (default)
  - `graph-stage` (opt-in)

### Tests

- `TlsSpec` continues to exercise the legacy actor path, including TLSv1.3.
- `TlsGraphStageSpec` runs the same shared TLS regression matrix against the GraphStage path in the same JVM.
- `TlsGraphStageEdgeCasesSpec` covers production async/input-buffer attributes, async-boundary routing, fragmented cipher input, small-write batching, no-upstream-completion flushes for small and large plaintext writes, plainOut backpressure, per-materialization `SSLEngine` isolation, and empty-source completion.
- `TlsGraphStageIsolatedSpec` covers early failures, empty/non-empty alternation, fragmented large payloads, eager close, immediate completion, and TLS 1.2 renegotiation.
- `TlsBenchmark` remains as a JMH fixture for future comparison work; this PR does not switch the default engine based on benchmark numbers.

### Benchmarks

Local JMH run on macOS 26.3 with Java 21.0.5 HotSpot. Throughput is `ops/ms`; higher is better.

Commands:

- `bench-jmh/Jmh/run -wi 2 -i 3 -f1 -t1 -p implementation=legacy,graphstage -p payloadSize=64,256,1024,4096,65536 .*TlsBenchmark.warmRoundTrip.*`
- `bench-jmh/Jmh/run -wi 1 -i 2 -f1 -t1 -p implementation=legacy,graphstage -p payloadSize=256,4096 .*TlsBenchmark.coldHandshake.*`

Warm round-trip:

| payload | legacy | graph-stage | delta |
| ---: | ---: | ---: | ---: |
| 64 B | 248.011 | 616.426 | +148.5% |
| 256 B | 195.364 | 466.311 | +138.7% |
| 1 KiB | 230.728 | 314.497 | +36.3% |
| 4 KiB | 130.534 | 158.900 | +21.7% |
| 64 KiB | 17.021 | 16.929 | -0.5% |

Cold handshake:

| payload | legacy | graph-stage | delta |
| ---: | ---: | ---: | ---: |
| 256 B | 0.672 | 0.700 | +4.2% |
| 4 KiB | 0.720 | 0.696 | -3.3% |

### Validation

- `actor / scalafmt`
- `stream / scalafmt`
- `stream-tests / scalafmt`
- `bench-jmh / scalafmt`
- `actor / scalafmtCheck`
- `stream / scalafmtCheck`
- `stream-tests / scalafmtCheck`
- `bench-jmh / scalafmtCheck`
- `stream / Test / compile`
- JDK 21 latest targeted run after the directional tests: `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageEdgeCasesSpec`
  - 13 tests succeeded, 0 failed
- JDK 21 full TLS regression: `TlsSpec`, `TlsGraphStageSpec`, `TlsGraphStageEdgeCasesSpec`, `TlsGraphStageIsolatedSpec`
  - 239 tests succeeded, 0 failed
- JDK 17 full TLS regression: `TlsSpec`, `TlsGraphStageSpec`, `TlsGraphStageEdgeCasesSpec`, `TlsGraphStageIsolatedSpec`
  - 239 tests succeeded, 0 failed
- JDK 25 GraphStage regression: `TlsGraphStageSpec`, `TlsGraphStageEdgeCasesSpec`, `TlsGraphStageIsolatedSpec`
  - 128 tests succeeded, 0 failed
- `git diff --check`
- Red-flag scan for prior suspicious port markers: no matches in the new/changed TLS GraphStage, config, test, or benchmark files

### Related

- Part of #2860
- Covers TLSv1.3 on both the legacy actor path and the new GraphStage path


stream/src/main/resources/reference.conf
```diff
@@ -175,6 +175,18 @@ pekko {
...
       //#stream-ref
+
+      # TLS (SSL/TLS stream cipher) configuration.
+      # Config path: pekko.stream.materializer.tls.*
+      #
+      # This is nested inside materializer so the full key is:
+      #   pekko.stream.materializer.tls.use-legacy-actor
+      # which is what TlsGraphStage reads at class-initialization time.
+      tls {
+      # Select the TLS implementation used by pekko.stream.scaladsl.TLS.
```
pjfanning: fix indent of these lines

stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsGraphStage.scala
```diff
@@ -0,0 +1,1010 @@
...
+@InternalApi private[stream] object TlsGraphStage {
+
+  /**
+   * JVM-level feature flag. Read once at class initialization to avoid inconsistent
+   * state within a single JVM (different ActorSystems cannot use different implementations).
+   *
+   * Package-private to allow access from TLS.scala in the scaladsl package.
+   */
+  private[stream] lazy val useLegacyActor: Boolean =
+    ConfigFactory.load().getBoolean("pekko.stream.materializer.tls.use-legacy-actor")
```
pjfanning: it would be better to read this from actorsystem - we have lots of tests that create actor systems with custom config and they benefit if our code does not load configs with `ConfigFactory.load()`

He-Pin: Yes.

pjfanning: `final class Tcp(system: ExtendedActorSystem)` is the class that creates TLS instances - so we can change the code to either pass the system as a param to the TLS class instance or a boolean value for useLegacyActor that has been evaluated for the system settings as opposed to using `ConfigFactory.load()`

He-Pin:
I run it with GPT5.5 xhigh, and seems ok to me now.

He-Pin:
Huge thanks @mingyang91 for gpt 5.5

pjfanning (dismissed):
lgtm - I guess if this is optional and non-default then it seems like rolling this out is less risky

pjfanning (approved):
lgtm

PR #2879, opened 2026-04-20
ByteStringBuilder.putInt/putLong: eliminate closure allocation via SWARUtil VarHandle writes

pjfanning:
`ByteStringBuilder.putInt` and `putLong` passed lambdas to `fillArray`, allocating a closure object on every call. This replaces those closures with direct inlined writes using new `SWARUtil.putInt`/`putLong` helpers.

### `SWARUtil`
- Added `putInt(array, index, value, byteOrder)` and `putLong(array, index, value, byteOrder)` — VarHandle `set`-based writes mirroring the existing `getInt`/`getLong` readers
- Added byte-by-byte fallbacks (`putIntBEWithoutMethodHandle`, `putIntLEWithoutMethodHandle`, `putLongBEWithoutMethodHandle`, `putLongLEWithoutMethodHandle`) for environments without VarHandle support

### `ByteStringBuilder`
- `putInt` and `putLong` now call `ensureTempSize` + `SWARUtil.putInt/putLong` directly, updating `_tempLength`/`_length` inline — same structure as `fillArray` but without the closure

```scala
// Before — allocates a lambda on every call
def putInt(x: Int)(implicit byteOrder: ByteOrder): this.type =
  fillArray(4) { (target, offset) => /* byte shifts */ }

// After — no allocation
def putInt(x: Int)(implicit byteOrder: ByteOrder): this.type = {
  ensureTempSize(_tempLength + 4)
  SWARUtil.putInt(_temp, _tempLength, x, byteOrder)
  _tempLength += 4; _length += 4
  this
}
```

`putFloat` and `putDouble` benefit automatically since they delegate to `putInt`/`putLong`.

### `SWARUtilSpec`
- Added tests for all six new methods (`putInt`, `putIntBEWithoutMethodHandle`, `putIntLEWithoutMethodHandle`, `putLong`, `putLongBEWithoutMethodHandle`, `putLongLEWithoutMethodHandle`) covering specific byte layouts at multiple offsets, boundary values (`Int`/`Long` `MaxValue`/`MinValue`), offset correctness (surrounding bytes not clobbered), and round-trip consistency with the corresponding `get*` fallbacks.

### `ByteStringSpec`
- Added dedicated `putInt` and `putLong` test blocks under `"A ByteStringBuilder"` covering: known byte sequences for both byte orders, boundary values, length assertions, multi-put accumulation, and property-based round-trips against `ByteBuffer` for both big-endian and little-endian.

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
@@ -2634,6 +2634,140 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
         }
       }
     }
+
+    "putInt" when {
+      "big-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(0x01020304)(BIG_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0x01, 0x02, 0x03, 0x04))
+      }
+      "little-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(0x01020304)(LITTLE_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0x04, 0x03, 0x02, 0x01))
+      }
+      "max value big-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(Int.MaxValue)(BIG_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0x7F, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte))
+      }
+      "min value big-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(Int.MinValue)(BIG_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0x80.toByte, 0x00, 0x00, 0x00))
+      }
+      "max value little-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(Int.MaxValue)(LITTLE_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0x7F))
+      }
+      "min value little-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(Int.MinValue)(LITTLE_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0x00, 0x00, 0x00, 0x80.toByte))
+      }
+      "result has length 4" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(42)(BIG_ENDIAN)
+        builder.length should ===(4)
+        builder.result().length should ===(4)
+      }
+      "multiple puts accumulate correctly" in {
+        val builder = ByteString.newBuilder
+        builder.putInt(0x01020304)(BIG_ENDIAN)
+        builder.putInt(0x05060708)(BIG_ENDIAN)
+        builder.result().toSeq should ===(Seq[Byte](0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08))
+      }
+      "round-trips with ByteBuffer big-endian" in {
+        check { (value: Int) =>
+          val builder = ByteString.newBuilder
+          builder.putInt(value)(BIG_ENDIAN)
+          val reference = new Array[Byte](4)
+          ByteBuffer.wrap(reference).order(BIG_ENDIAN).putInt(value)
+          builder.result().toSeq == reference.toSeq
+        }
+      }
+      "round-trips with ByteBuffer little-endian" in {
+        check { (value: Int) =>
+          val builder = ByteString.newBuilder
+          builder.putInt(value)(LITTLE_ENDIAN)
+          val reference = new Array[Byte](4)
+          ByteBuffer.wrap(reference).order(LITTLE_ENDIAN).putInt(value)
+          builder.result().toSeq == reference.toSeq
+        }
+      }
+    }
+
+    "putLong" when {
+      "big-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(0x0102030405060708L)(BIG_ENDIAN)
+        builder.result().toSeq should ===(
+          Seq[Byte](0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08))
+      }
+      "little-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(0x0102030405060708L)(LITTLE_ENDIAN)
+        builder.result().toSeq should ===(
+          Seq[Byte](0x08, 0x07, 0x06, 0x05, 0x04, 0x03, 0x02, 0x01))
+      }
+      "max value big-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(Long.MaxValue)(BIG_ENDIAN)
+        builder.result().toSeq should ===(
+          Seq[Byte](0x7F, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte))
+      }
+      "min value big-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(Long.MinValue)(BIG_ENDIAN)
+        builder.result().toSeq should ===(
+          Seq[Byte](0x80.toByte, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00))
+      }
+      "max value little-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(Long.MaxValue)(LITTLE_ENDIAN)
+        builder.result().toSeq should ===(
+          Seq[Byte](0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0xFF.toByte, 0x7F))
+      }
+      "min value little-endian" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(Long.MinValue)(LITTLE_ENDIAN)
+        builder.result().toSeq should ===(
+          Seq[Byte](0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x80.toByte))
+      }
+      "result has length 8" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(42L)(BIG_ENDIAN)
+        builder.length should ===(8)
+        builder.result().length should ===(8)
+      }
+      "multiple puts accumulate correctly" in {
+        val builder = ByteString.newBuilder
+        builder.putLong(0x0102030405060708L)(BIG_ENDIAN)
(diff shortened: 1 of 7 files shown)


pjfanning:
I'll look into creating a JMH benchmark

He-Pin:
nice improvement

pjfanning:
```
With Changes
[info] Benchmark                                Mode  Cnt       Score        Error   Units
[info] ByteString_putInt_Benchmark.putIntBE    thrpt    3  239684.591 ± 102494.674  ops/ms
[info] ByteString_putInt_Benchmark.putIntLE    thrpt    3  272971.614 ±  43025.163  ops/ms
[info] ByteString_putInt_Benchmark.putLongBE   thrpt    3  127419.705 ±   6651.774  ops/ms
[info] ByteString_putInt_Benchmark.putLongLE   thrpt    3  137667.126 ±  25170.882  ops/ms
[info] ByteString_putInt_Benchmark.putShortBE  thrpt    3  266982.432 ±  69828.845  ops/ms
[info] ByteString_putInt_Benchmark.putShortLE  thrpt    3  309618.809 ±   5523.773  ops/ms


Without Changes
[info] Benchmark                                Mode  Cnt       Score       Error   Units
[info] ByteString_putInt_Benchmark.putIntBE    thrpt    3  239961.626 ± 24055.818  ops/ms
[info] ByteString_putInt_Benchmark.putIntLE    thrpt    3  238674.817 ± 47275.478  ops/ms
[info] ByteString_putInt_Benchmark.putLongBE   thrpt    3  131730.867 ±  4642.640  ops/ms
[info] ByteString_putInt_Benchmark.putLongLE   thrpt    3  105625.382 ±  5410.259  ops/ms
[info] ByteString_putInt_Benchmark.putShortBE  thrpt    3  271689.337 ±  7647.398  ops/ms
[info] ByteString_putInt_Benchmark.putShortLE  thrpt    3  260050.052 ± 16497.148  ops/ms
```

persistence/src/main/scala/org/apache/pekko/persistence/serialization/SnapshotSerializer.scala
```diff
@@ -81,13 +81,18 @@ class SnapshotSerializer(val system: ExtendedActorSystem) extends BaseSerializer
     Snapshot(snapshotFromBinary(bytes))
 
   private def headerToBinary(snapshot: AnyRef, snapshotSerializer: Serializer): Array[Byte] = {
-    val out = new ByteArrayOutputStream
-    writeInt(out, snapshotSerializer.identifier)
-
     val ms = migrateManifestIfNecessary(Serializers.manifestFor(snapshotSerializer, snapshot))
```
He-Pin: PR 标题和描述只说消除 `putInt/putLong` 闭包分配，但实际还包含了 `SnapshotSerializer` 的重构（移除 `ByteArrayOutputStream` + `writeInt`）。虽然字节序检查过是一致的（原始 `writeInt` 写 `i >>> 0, i >>> 8, i >>> 16, i >>> 24` = little-endian，新代码用 `LITTLE_ENDIAN`，`headerFromBinary` 也用 `LITTLE_ENDIAN` 读取），但这个改动应单独一个 PR 或至少在 PR 描述中说明。

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -2245,60 +2245,32 @@ final class ByteStringBuilder extends Builder[Byte, ByteString] {
...
   def putShort(x: Int)(implicit byteOrder: ByteOrder): this.type = {
-    if (byteOrder == ByteOrder.BIG_ENDIAN) {
-      this += (x >>> 8).toByte
-      this += (x >>> 0).toByte
-    } else if (byteOrder == ByteOrder.LITTLE_ENDIAN) {
-      this += (x >>> 0).toByte
-      this += (x >>> 8).toByte
-    } else throw new IllegalArgumentException("Unknown byte order " + byteOrder)
+    ensureTempSize(_tempLength + 2)
+    SWARUtil.putShort(_temp, _tempLength, x.toShort, byteOrder)
```
He-Pin: The `putShort` signature is `def putShort(x: Int)`, but the new implementation calls `x.toShort` before passing to SWARUtil. The original implementation used `x >>> 8` and `x >>> 0` on the full `Int` value.

For values outside `Short` range (e.g. `x = 65535` = 0x0000FFFF), `x.toShort` truncates to `-1` before the bit shifts occur in SWARUtil. While the final byte output happens to be the same for this case, the semantics have changed because the parameter type is `Int` but the behavior is now constrained to `Short` range.

The added tests only cover `Short` range values (via `check { (value: Short) => ... }`), so this truncation behavior is not explicitly tested. Either add tests for out-of-range `Int` values to verify the behavior is preserved, or document this as an intentional semantic change.

persistence/src/main/scala/org/apache/pekko/persistence/serialization/SnapshotSerializer.scala
He-Pin, line 84: The `SnapshotSerializer` changes refactor `headerToBinary` and the serialization method to use direct byte arrays + `SWARUtil.putInt` instead of `ByteArrayOutputStream` + `writeInt`. This is a significant change to the serialization format logic.

However, no tests are added or modified for `SnapshotSerializer` to verify that:
1. The serialized binary format is identical (same byte order, same structure)
2. Round-trip serialization/deserialization works correctly
3. The `migrateManifestIfNecessary` path still produces correct output

Please add tests for `SnapshotSerializer` to cover these cases, or point to existing tests that would catch regressions here.

actor/src/main/scala/org/apache/pekko/util/SWARUtil.scala
```diff
@@ -235,6 +235,84 @@ private[pekko] object SWARUtil {
...
+   * @param array     the byte array to write to
+   * @param index     the index to write at
+   * @param value     the int value to write
+   * @param byteOrder the byte order to use (big-endian or little-endian)
+   */
+  def putInt(array: Array[Byte], index: Int, value: Int, byteOrder: ByteOrder): Unit = {
+    if (byteOrder == ByteOrder.BIG_ENDIAN) {
+      if (intBeArrayViewSupported) {
+        intBeArrayView.set(array, index, value)
+      } else {
```
He-Pin, line 252: **Correctness concern:** The `putInt`/`putLong`/`putShort` methods silently treat any non-BIG_ENDIAN byte order as LITTLE_ENDIAN. The original `ByteStringBuilder.putInt` threw `IllegalArgumentException` for unknown byte orders:

```scala
} else throw new IllegalArgumentException("Unknown byte order " + byteOrder)
```

While `ByteOrder` only has BE/LE instances in practice, this is still a behavioral regression. Consider adding an explicit validation or at least documenting this change.

bench-jmh/src/main/scala/org/apache/pekko/util/ByteString_putInt_Benchmark.scala
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+@Fork(2)
+@Warmup(iterations = 5)
```
He-Pin, line 25: **Performance concern:** The benchmark only measures the new implementation. There is no baseline `@Benchmark` method using the old `fillArray` approach for before/after comparison.

To validate the performance claim (eliminating closure allocation), the benchmark should either:
1. Include a method that replicates the old `fillArray` pattern for direct comparison, or
2. Document that reviewers should run this benchmark against both `main` and the PR branch.

Without a baseline, the benchmark cannot demonstrate the claimed improvement.

actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
```diff
@@ -2634,6 +2634,203 @@ class ByteStringSpec extends AnyWordSpec with Matchers with Checkers {
         }
       }
     }
+
+    "putInt" when {
```
He-Pin, line 2638: **Test coverage gap:** The PR description states that `putFloat` and `putDouble` "benefit automatically since they delegate to `putInt`/`putLong`". However, no tests are added for `putFloat` or `putDouble` to verify this delegation still works correctly after the refactoring.

Please add at least basic round-trip tests for `putFloat` and `putDouble` against `ByteBuffer` (similar to the existing `putInt`/`putLong` property tests) to catch any regression in the delegation path.

actor-tests/src/test/scala/org/apache/pekko/util/SWARUtilSpec.scala
```diff
@@ -57,6 +57,200 @@ class SWARUtilSpec extends AnyWordSpec with Matchers {
       SWARUtil.getShortBEWithoutMethodHandle(testData, 2) should ===(0x0203.toShort)
       SWARUtil.getShortLEWithoutMethodHandle(testData, 2) should ===(0x0302.toShort)
     }
```
He-Pin, line 59: **Endianness testing gap:** The round-trip and inverse tests are comprehensive, but the direct layout tests are missing explicit coverage for all-zero (`0`) and all-ones (`-1`) byte patterns:

- `putInt(arr, 0, 0, BE)` should produce `[0x00, 0x00, 0x00, 0x00]`
- `putInt(arr, 0, -1, BE)` should produce `[0xFF, 0xFF, 0xFF, 0xFF]`

Similarly for `putLong` (8 bytes of `0x00` or `0xFF`).

These patterns exercise every byte uniformly and are common edge cases in serialization. The inverse tests cover `-1` as a value, but explicit byte-layout assertions would make it clear the byte order is correct for all-bits-set values.

actor/src/main/scala/org/apache/pekko/util/SWARUtil.scala
pjfanning: there are only 2 ByteOrders in the Java class - https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/nio/ByteOrder.html

bench-jmh/src/main/scala/org/apache/pekko/util/ByteString_putInt_Benchmark.scala
pjfanning: this is the case with all our benchmarks - it is not hard to run the benchmark with and without the main source changes

actor-tests/src/test/scala/org/apache/pekko/util/ByteStringSpec.scala
pjfanning: added more tests

actor-tests/src/test/scala/org/apache/pekko/util/SWARUtilSpec.scala
pjfanning: added more tests

persistence/src/main/scala/org/apache/pekko/persistence/serialization/SnapshotSerializer.scala
pjfanning: #2891 has extra tests and they appear in this PR too

He-Pin (approved):
lgtm

PR #2889, opened 2026-04-23
fix: stabilise JDK 21+ / JDK 25 nightly test runs

He-Pin:
## Summary

Targets the recurring JDK 21+ / JDK 25 nightly flakes observed in the latest [nightly run 24810320571](https://github.com/apache/pekko/actions/runs/24810320571/). Two complementary moves: (1) bump `fork-join-executor.minimum-runnable` for the JDK 21+ test JVMs to fight the JDK-8300995 / JDK-8321335 compensation regression, and (2) plug four hardcoded test timeouts that ignore `pekko.test.timefactor`.

## Why `minimum-runnable`, not LIFO

LIFO `task-peeking-mode` improves cache locality but causes head-of-line blocking and scheduling-fairness regressions across mailboxes — a poor default for actor workloads. Raising `minimum-runnable` keeps FIFO semantics and only changes how eagerly the pool spawns compensation threads when workers block, which is exactly the behaviour change needed to counter the JDK 21+ regression. The override is applied **only** in the workflow; production defaults stay at `1`.

## Changes

| File | Change | Why |
|------|--------|-----|
| `.github/workflows/nightly-builds.yml` | When `javaVersion >= 21`, append `-Dpekko.actor.default-dispatcher.fork-join-executor.minimum-runnable=4` and same for `pekko.remote.default-remote-dispatcher` | Eager compensation-thread creation under JDK 21+ FJP regression; targets artery / remoting flakes |
| `stream-tests/.../TlsSpec.scala` | Add `.dilated` to the `15.seconds` and `17.seconds` timeouts in the `ServerInitiatesViaTcp / CancellingRHSIgnoresBoth` scenario | Hardcoded timeouts ignored `timefactor` — failed at exactly 17s on JDK 25 |
| `persistence-typed-tests/.../EventSourcedStashOverflowSpec.scala` | Pass an explicit `30.seconds` budget to `probe.receiveMessages(stashCapacity)` | Default 12s on JDK 25 not enough to drain 20 K messages on slow runners |
| `persistence/.../journal/SteppingInmemJournal.scala` | Bump `step()` ask timeout from `3.seconds.dilated` to `10.seconds.dilated` | `PersistentActorRecoveryTimeoutSpec` failed at 6s (3s × timefactor=2) on JDK 17 |

## Out of scope (separate follow-ups)

[...]

PR #2890, opened 2026-04-23
fix: auto-tune ForkJoinPool minimum-runnable on JDK 25+

He-Pin:
## Summary

- Add `ForkJoinExecutorConfigurator.resolveMinimumRunnable` that auto-scales the `minimum-runnable` setting on JDK 25+ (`min(8, max(1, parallelism / 2))`) while keeping the legacy value of `1` on older JDKs. Explicit user values (including `0`) are still honoured.
- Switch the default in `reference.conf` from `1` to the sentinel `-1` (auto) and update the doc block.
- New `ForkJoinExecutorConfiguratorSpec` covering: pure-function matrix; directional checks (auto value strictly higher on JDK 25+, never exceeds documented cap of 8); wiring integration that proves the resolved value reaches the `ForkJoinExecutorServiceFactory`.

## Context

JDK-8300995 / JDK-8321335 changed `ForkJoinPool` asyncMode (FIFO) compensation-thread creation to be much more conservative. Pekko fork-join dispatchers using the prior default `minimum-runnable = 1` are then prone to starvation under blocking workloads on JDK 21+, which is the root cause behind the workflow override added in #2889.

That workflow override only protects nightly CI. This PR pushes the same protection into the library so production users on JDK 25+ get the mitigation automatically.

## Behaviour matrix

| `minimum-runnable` | JDK | Effective value |
|---|---|---|
| `-1` (default) | < 25 | `1` (legacy) |
| `-1` (default) | >= 25 | `min(8, max(1, parallelism / 2))` |
| `0` | any | `0` (compensation disabled) |
| any positive `N` | any | `N` |

Escape hatch for users on JDK 21+ who want to keep the prior behaviour: set `pekko.actor.default-dispatcher.fork-join-executor.minimum-runnable = 1` (or any explicit value).

## Compatibility

- No constructor signatures changed (`PekkoForkJoinPool` and `ForkJoinExecutorServiceFactory` keep their `minimumRunnable: Int = 1` defaults).
- No MiMa filter required.

## Follow-up

Once this PR is merged and a nightly run confirms green, a small follow-up PR will drop the `EXTRA_JVM_OPTS` block that #2889 added to `.github/workflows/nightly-builds.yml` (the workflow override becomes redundant).

## Test plan

- [x] `sbt scalafmtAll`
- [x] `sbt actor/compile actor-tests/Test/compile`
- [x] `sbt 'actor-tests/testOnly org.apache.pekko.dispatch.ForkJoinExecutorConfiguratorSpec'` — all 10 tests pass (1 pending on JDK 21+ for the legacy branch)
- [x] `sbt 'actor-tests/testOnly org.apache.pekko.dispatch.ForkJoinPoolStarvationSpec'` — still pending as before, no regression
- [ ] Nightly Builds workflow on `main` after merge confirms JDK 17 / 21 / 25 matrix is green

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfiguratorSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfiguratorSpec.scala
@@ -1,0 +1,165 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project.
+ */
+
+package org.apache.pekko.dispatch
+
+import java.util.concurrent.ThreadFactory
+
+import org.apache.pekko
+import pekko.testkit.PekkoSpec
+import pekko.util.JavaVersion
+
+import com.typesafe.config.{ Config, ConfigFactory }
+
+object ForkJoinExecutorConfiguratorSpec {
+  // Keep the root config explicit so the dispatcher-config integration checks below
+  // see exactly the values this spec is asserting on (and not whatever the project's
+  // global test configuration happens to set).
+  val config: Config = ConfigFactory.parseString("""
+      |fj-auto-default-dispatcher {
+      |  executor = "fork-join-executor"
+      |  fork-join-executor {
+      |    parallelism-min = 8
+      |    parallelism-factor = 1.0
+      |    parallelism-max = 64
+      |  }
+      |}
+      |fj-auto-small-dispatcher {
+      |  executor = "fork-join-executor"
+      |  fork-join-executor {
+      |    parallelism-min = 1
+      |    parallelism-factor = 1.0
+      |    parallelism-max = 1
+      |  }
+      |}
+      |fj-explicit-zero-dispatcher {
+      |  executor = "fork-join-executor"
+      |  fork-join-executor {
+      |    parallelism-min = 8
+      |    parallelism-factor = 1.0
+      |    parallelism-max = 64
+      |    minimum-runnable = 0
+      |  }
+      |}
+      |fj-explicit-seven-dispatcher {
+      |  executor = "fork-join-executor"
+      |  fork-join-executor {
+      |    parallelism-min = 8
+      |    parallelism-factor = 1.0
+      |    parallelism-max = 64
+      |    minimum-runnable = 7
+      |  }
+      |}
+    """.stripMargin)
+}
+
+class ForkJoinExecutorConfiguratorSpec extends PekkoSpec(ForkJoinExecutorConfiguratorSpec.config) {
+
+  import ForkJoinExecutorConfigurator.resolveMinimumRunnable
+
+  "ForkJoinExecutorConfigurator.resolveMinimumRunnable" must {
+
+    "honour explicit zero (compensation disabled)" in {
+      resolveMinimumRunnable(configured = 0, parallelism = 16, jdkMajorVersion = 21) shouldBe 0
+      resolveMinimumRunnable(configured = 0, parallelism = 16, jdkMajorVersion = 17) shouldBe 0
+    }
+
+    "honour explicit positive overrides verbatim" in {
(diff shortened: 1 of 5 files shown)


He-Pin:
@pjfanning I think this should be the only way to make JDK25 pass

actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfiguratorSpec.scala
```diff
@@ -0,0 +1,165 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
pjfanning, line 2: Looks like the wrong license header. If this needs an Akka derivation hear, it also needs Lighbend copyright headers.

pjfanning:
@He-Pin can you update the dispatcher docs? I'm also wary of changing the default Java 21 behaviour. We've got strong evidence from the nightly tests that Java 25 could really do with the change you are making here but Java 21 seems fine.

pjfanning:
Maybe it is best as the jdk issue affects Java 21 but just doesn't seem to have quite the same impact there

He-Pin:
@pjfanning thanks for the review — addressed all three comments in 904eef2:

1. **License header on `ForkJoinExecutorConfiguratorSpec.scala`** — replaced the abbreviated header with the canonical Apache 2.0 header used by other clean-room test files in the project (e.g. `ByteStringBuilderSpec`, `ByteBufferCleanerSpec`). Since this is a brand-new file with no derivation, no Lightbend copyright is needed.

2. **Narrowed the auto-policy from JDK 21+ to JDK 25+**, per your guidance — nightly evidence is strongest on the JDK 25 line; JDK 21 has been running fine on the legacy default of `1` for years and there's no need for a silent behaviour change there. Updated `resolveMinimumRunnable`, the reference.conf doc comment, the spec matrix, and the directional/pending guards to match.

3. **Updated dispatcher docs** — added a note in both `docs/src/main/paradox/dispatchers.md` (classic) and `docs/src/main/paradox/typed/dispatchers.md` describing the new auto-tuning behaviour and how to opt out (set `minimum-runnable` explicitly).

Local `actor-tests/testOnly ForkJoinExecutorConfiguratorSpec` is green (10 passed, 1 pending — the JDK<25 branch is naturally pending on a JDK 25 host).

pjfanning (approved):
lgtm

snuyanzin:
hi @He-Pin , @pjfanning
Is there any chance to have this backported to 1.x?

Asking since 
I'm working on java25 support in Flink and it looks i faced same issue for a couple of Flink tests

He-Pin:
@snuyanzin Feel free to submit a PR

pjfanning:
I created https://github.com/apache/pekko/pull/3037 for discussion but I am far from convinced that this is
* safe for a 1.x release
* there is no major plan for a 1.7.0 release and a big backlog of other releases that we need to do in other pekko modules
* I'd really prefer to concentrate on proven bugs and getting 2.0.0 finished

pjfanning:
@snuyanzin if this change is so useful, why don't you try testing with pekko 2.0.0-M3 which contains this?
Earlier 2.0 milestones should be avoided due to issues that have have been fixed in 2.0.0-M3.

snuyanzin:
@pjfanning we are still on scala 2.12.x and Flink still should be built with java 11, 17, 21 and 25(in progress)
based on Pekko's 2.0 nightlies it doesn't run with java 11 and scala 2.12 or do I understand it wrong?

pjfanning:
@snuyanzin if you must have this, you can create your custom dispatcher for Pekko. You don't need to use the built-in ones.

snuyanzin:
I see
would be great if you can give some pointers to examples/tests where I can take a look

pjfanning:
> I see would be great if you can give some pointers to examples/tests where I can take a look

@snuyanzin I would suggest starting with AI or reading Pekko docs. If you find issues with the docs, report them.

pjfanning:
@snuyanzin it's possible that we will merge a version of #3037. If we do so, there will be snapshot jars that can be used to test the change. It just could take a while for the actual 1.7.0 release to happen.

snuyanzin:
Thanks, I will look into docs (with AI)
no rush for now, I'm doing it in some spare time and there are still other issues to fix

PR #2891, opened 2026-04-23
add regression tests to SnapshotSerializerSpec

pjfanning:
ports some tests from #2879 to show that the existing code produces the same results as the new code in 2879


He-Pin (approved):
lgtm

PR #2893, opened 2026-04-23
upgrade license headers

pjfanning:
Tracked down source files with the Apache Software License withe the modification mentioning Akka but where we have no Lightbend copyright. I added the copyright.

I removed one file, JUnit5IntegrationExampleTest, because we have a newer copy called JUnitJupiterIntegrationExampleTest. No need for both.


He-Pin (approved):
lgtm

PR #2895, opened 2026-04-24
fix: stabilize async DNS dispatchers under low minimum-runnable

He-Pin:
## Summary
- keep the internal dispatcher on an explicit compensation floor
- route async DNS routees for both IO-DNS and SD-DNS through the configured DNS dispatcher
- add a regression assertion for the built-in internal dispatcher floor

## Motivation
The nightly JDK 17 / Scala 3.3.x failure came from `docs.discovery.DnsDiscoveryDocSpec` timing out and going pending after `minimum-runnable` moved to the new auto policy. The isolated SD-DNS manager and resolver already used the DNS dispatcher, but the actual async DNS routees were still left on the default dispatcher, making lookups much more sensitive to starvation.

## Result
With `-Dpekko.actor.default-dispatcher.fork-join-executor.minimum-runnable=-1`, the docs DNS discovery regression now passes again instead of pending.

Fixes the failure from:
https://github.com/apache/pekko/actions/runs/24863583862/job/72794553815


pjfanning (approved):
lgtm

PR #2896, opened 2026-04-24
optimize ByteStrings startswith/endswith

pjfanning:
Some more copilot suggestions. The changes that have been added so far have mainly targeted the byte array backed ByteString implementations but the implementation for the concatenated ByteStrings can still exhibit O(n^2) issues due to the use of apply(int) function which is inefficient for that implementation.
Without this change, users are best advised to to compact their ByteString if they need startsWith/endsWith and suspect that they might be dealing with a concatenated ByteString.
 
```
With Changes
Benchmark                (cellCount)   Mode  Cnt    Score    Error  Units
[info] Benchmark                                          Mode  Cnt         Score           Error  Units
[info] ByteString_startEnd_Benchmark.bs_endsWith         thrpt    3  41146243.815 ±  89199318.989  ops/s
[info] ByteString_startEnd_Benchmark.bs_endsWithBytes    thrpt    3  57772963.058 ± 113861001.097  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWith       thrpt    3  51618538.730 ±  50209295.437  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWithBytes  thrpt    3  61139506.870 ±  43826147.198  ops/s
[info] ByteString_startEnd_Benchmark.bss_endsWith        thrpt    3   3763017.312 ±   1874477.984  ops/s
[info] ByteString_startEnd_Benchmark.bss_startsWith      thrpt    3  22659830.804 ±   4867250.514  ops/s

Without Changes
[info] Benchmark                                          Mode  Cnt         Score          Error  Units
[info] ByteString_startEnd_Benchmark.bs_endsWith         thrpt    3  42643024.729 ± 39682206.774  ops/s
[info] ByteString_startEnd_Benchmark.bs_endsWithBytes    thrpt    3  48849001.581 ± 36760529.610  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWith       thrpt    3  41864132.036 ± 57823779.206  ops/s
[info] ByteString_startEnd_Benchmark.bs_startsWithBytes  thrpt    3  54296545.089 ± 33013510.355  ops/s
[info] ByteString_startEnd_Benchmark.bss_endsWith        thrpt    3   3428387.507 ±  1458843.947  ops/s
[info] ByteString_startEnd_Benchmark.bss_startsWith      thrpt    3  14300744.704 ±   369312.100  ops/s
```

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -402,6 +402,24 @@ object ByteString {
       java.util.Arrays.equals(this.bytes, hIdx, hIdx + (needleLen - nIdx), bytes, nIdx, needleLen)
     }
 
+    /**
+     * INTERNAL API: compare `len` bytes from this ByteString starting at `haystackOffset`
+     * against `needle[needleOffset..needleOffset+len)`.
+     */
+    private[pekko] def matchesAt(haystackOffset: Int, needle: Array[Byte], needleOffset: Int, len: Int): Boolean = {
+      var hIdx = haystackOffset
+      var nIdx = needleOffset
+      var remaining = len
+      while (remaining >= 8) {
+        if (SWARUtil.getLong(bytes, hIdx, ByteOrder.BIG_ENDIAN) !=
+            SWARUtil.getLong(needle, nIdx, ByteOrder.BIG_ENDIAN)) return false
+        hIdx += 8
+        nIdx += 8
+        remaining -= 8
+      }
+      java.util.Arrays.equals(bytes, hIdx, hIdx + remaining, needle, nIdx, nIdx + remaining)
+    }
+
     override def slice(from: Int, until: Int): ByteString =
       if (from <= 0 && until >= length) this
       else if (from >= length || until <= 0 || from >= until) ByteString.empty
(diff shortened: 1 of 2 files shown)


He-Pin:
I think we can dig more performance optimization:) hope next time the grpc-bench we can close with vert.x

pjfanning:
@He-Pin do you think this is ok to merge?

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1199,6 +1235,71 @@ object ByteString {
...
+      }
+      true
+    }
+
+    override def endsWith(bytes: Array[Byte]): Boolean = {
+      val needleLen = bytes.length
+      if (length < needleLen) return false
+      if (needleLen == 0) return true
+      // Locate the fragment that contains position `length - needleLen`
+      var fragIdx = 0
```
He-Pin: Non-blocking performance thought: because `endsWith` is anchored at the tail, this could avoid the upfront walk from `bytestrings(0)` when the suffix is small and the `ByteStrings` has many fragments. A follow-up could compare from the last fragment backward, or otherwise locate the suffix start from the tail, so `endsWith(Array(lastByte))` only touches the suffix fragments instead of every fragment. The current version is still a useful improvement since it avoids compacting.

He-Pin (approved):
LGTM. I did not find a correctness issue in the ByteStrings changes. `actor-tests/testOnly org.apache.pekko.util.ByteStringSpec` passed locally (203 tests). I left one non-blocking performance thought for a possible follow-up.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: https://github.com/apache/pekko/pull/2896/commits/1c55fa97fda0f5e08793cb114043393d10de916c addresses this

PR #2897, opened 2026-04-24
Fix copyToBuffer: guard negative return in ByteString1C.writeToBuffer

pjfanning:
early-exit in ByteStrings.copyToBuffer; add comprehensive tests


He-Pin (approved):

He-Pin:
We need to add direction testing too, you can try my skill https://github.com/share-skills/pi

PR #2899, opened 2026-04-24
fix: MapAsyncPartitioned drainQueue concurrent-modification race

He-Pin:
## Summary

Fixes a hang in `MapAsyncPartitioned` (ordered and unordered) caused by `drainQueue` iterating `buffer` while `processElement` could re-enter `pushNextIfPossible` and mutate the buffer, silently skipping buffered elements so their futures were never started.

## Symptom

The property test `should process elements in parallel preserving order in partition` in `MapAsyncPartitionedSpec` hung past the 4-minute `futureValue` timeout on JDK 25 CI ([example failure](https://github.com/apache/pekko/actions/runs/24878067348/job/72839362438)). On JDK 21 the same race exists but is rarely tickled because the JIT/scheduler do not surface the synchronous-future window as often.

Local reproduction:
- **Before this PR**: 0 of 5 attempts completed within 240 s.
- **With this PR**: 4 of 5 attempts pass in well under a second; the full spec (22 tests) finishes in ~17 s.

## Root cause

`drainQueue` looked like this:

```scala
private def drainQueue(): Unit = {
  if (buffer.nonEmpty) {
    buffer.foreach {
      case (partition, wrappedInput) =>
        if (canStartNextElement(partition)) {
          wrappedInput.resume()
          processElement(partition, wrappedInput)
        }
    }
  }
}
```

`processElement` calls the user function `f` and, when the resulting `Future` is already complete, takes the #20217 fast path: it sets the holder result and synchronously calls `pushNextIfPossible()`. In ordered mode that path does `buffer.dequeue()` from the front; in unordered mode it does `buffer = buffer.filter(...)`. Both invalidate the iteration that `drainQueue.foreach` is in the middle of.

`mutable.Queue` is backed by `ArrayDeque`, whose `foreach` captures `length` and indexes into the live array (`array((start + i) & mask)`). Once `dequeue` advances `start`, every subsequent step of the captured loop reads the *next* element instead of the intended one — silently skipping queued items. A skipped item never has its future registered, so the stream stalls forever.

## Fix

Snapshot the buffer before iterating, and guard each item with `holder.out eq NotYetThere` so an item that was already started (or already pushed) during synchronous re-entry is never processed twice:

```scala
val snapshot = buffer.toList
snapshot.foreach {
  case (partition, wrappedInput) =>
    val holder = wrappedInput.element
    if ((holder.out eq NotYetThere) && canStartNextElement(partition)) {
      wrappedInput.resume()
      processElement(partition, wrappedInput)
    }
}
```

The `NotYetThere eq` check is the load-bearing piece: when re-entry pops or filters items out of the buffer those items have `holder.out` set, so they are skipped on the outer iteration. Items still pending have `out eq NotYetThere` and continue to be processed in FIFO order. The #20217 fast path is preserved.

The buffer never grows beyond `parallelism`, so the snapshot is bounded.

## Test plan

- [x] `sbt 'stream/scalafmtAll'` — formatting validated.
- [x] `sbt 'stream-typed-tests/Test/testOnly org.apache.pekko.stream.MapAsyncPartitionedSpec'` — all 22 tests pass in ~17 s.
- [x] Targeted property test repeated 5 times — passes in well under a second per run (vs. baseline timeout at 240 s on every run).
- [ ] CI green across all JDK / Scala combinations.

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/MapAsyncPartitioned.scala
@@ -252,9 +252,17 @@ private[stream] final class MapAsyncPartitioned[In, Out, Partition](
 
       private def drainQueue(): Unit = {
         if (buffer.nonEmpty) {
-          buffer.foreach {
+          // Snapshot the buffer because processElement may complete a Future synchronously
+          // (the #20217 optimization), which re-enters pushNextIfPossible and either
+          // dequeues from the buffer (ordered) or rebuilds it via filter (unordered),
+          // invalidating any in-flight iterator over `buffer`. Items already started or
+          // already pushed during re-entry have holder.out set, so the NotYetThere
+          // guard ensures we never invoke processElement twice for the same Holder.
+          val snapshot = buffer.toList
+          snapshot.foreach {
             case (partition, wrappedInput) =>
-              if (canStartNextElement(partition)) {
+              val holder = wrappedInput.element
+              if ((holder.out eq NotYetThere) && canStartNextElement(partition)) {
                 wrappedInput.resume()
                 processElement(partition, wrappedInput)
               }


pjfanning:
is this something that affects pekko 1.6.0 and should I consider cancelling that and waiting for this?

pjfanning (approved):
lgtm

pjfanning:
> is this something that affects pekko 1.6.0 and should I consider cancelling that and waiting for this?

@He-Pin is this something you could comment on? I don't know whether to cancel the vote.

He-Pin:
I think so, maybe this operator is not widely used, so no one reports

PR #2901, opened 2026-04-24
refactor: inline onEvent into MergeHub.tryProcessNext

He-Pin:
## Summary
- Inline the `Boolean`-returning `onEvent` helper directly into `MergedSourceLogic.tryProcessNext`
- Pattern-match the event in place: `Element` pushes and exits, `Register`/`Deregister` continue the `@tailrec` loop to drain control messages from the queue
- Same wakeup protocol, same demand handling — purely a clarity refactor

## Motivation
`onEvent` returned a `Boolean` flag whose only purpose was to tell `tryProcessNext` whether to continue looping. Folding the dispatch into the call site removes the indirection and makes the loop / exit conditions read in one place.

## Test plan
- [x] `sbt scalafmtAll`
- [x] `sbt stream/compile` — `@tailrec` accepted, no warnings
- [x] `sbt "stream-tests/testOnly org.apache.pekko.stream.scaladsl.HubSpec"` — 48/48 in 18s

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Hub.scala
@@ -201,21 +201,6 @@ private[pekko] class MergeHub[T](perProducerBufferSize: Int, drainingEnabled: Bo
 
     setHandler(out, this)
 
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #2910, opened 2026-04-25
fix(stream): preserve abrupt fanout publisher shutdown signal

He-Pin:
### Motivation
FanoutPublisherBridgeStage.postStop always completed the exposed publisher with ActorPublisher.NormalShutdownReason, even when the stage was stopped abruptly without an upstream or subscriber terminal signal. That made active and late subscribers observe a normal shutdown instead of the GraphStage abrupt termination failure.

### Modification
Track whether the bridge is stopping from an intentional terminal path. Only synthesize AbruptStageTerminationException from postStop when no terminal path was already signalled, and add a bridge-level regression test for active and late subscribers.

### Result
Controlled completion, failure, timeout, and last-subscriber cancellation keep their existing reasons, while abrupt bridge shutdown is now reported as AbruptStageTerminationException.

### Tests
- sbt "stream-tests/testOnly org.apache.pekko.stream.impl.FanoutPublisherBehaviorSpec"
- sbt "stream-tests-tck/testOnly org.apache.pekko.stream.tck.FanoutPublisherTest"
- sbt "stream / Compile / scalafmtCheck" "stream-tests / Test / scalafmtCheck"

### References
apache/pekko#2874

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/impl/FanoutPublisherBehaviorSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/impl/FanoutPublisherBehaviorSpec.scala
@@ -22,14 +22,19 @@ import scala.concurrent.duration._
 
 import org.apache.pekko
 import pekko.testkit.EventFilter
-import pekko.stream.Attributes
-import pekko.stream.ActorAttributes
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
Lgtm

PR #2917, opened 2026-04-26
Enhance DurableStateStore TCK with OCC, serialization, and soft-delete tests

drse:
## Summary

Enhances the AI-drafted `DurableStateStore` TCK from #2833 to address the gaps
discussed in #2831 — specifically optimistic-concurrency checks, serialization
roundtrip, and the deprecated soft-delete overload.

This PR supersedes #2833. The first three commits are @pjfanning's original
work cherry-picked unchanged; the fourth commit layers the enhancements on top.

### Capability flags (`CapabilityFlags.scala`)

`DurableStateStoreCapabilityFlags` extended with three new flags:

- **`supportsUpsertWithRevisionCheck`** — when on, asserts that an upsert with
  a stale revision is rejected and the prior state is preserved.
- **`supportsSerialization`** — when on, asserts that values roundtrip through
  the configured Pekko serializer using the same `TestPayload(probe.ref)`
  pattern as `JournalSpec`. Defaults to **`on`** in `DurableStateStoreSpec` to
  match the `JournalSpec` convention; the in-memory testkit reference TCK
  explicitly overrides to `off` since the testkit holds raw refs.
- **`supportsSoftDelete`** — when on, asserts the deprecated single-argument
  `deleteObject(persistenceId)` overload deletes regardless of revision.

### Spec additions (`DurableStateStoreSpec.scala`)

- New required test: **"upsert again after a deletion"** — ensures the store
  accepts a fresh upsert at a higher revision after a delete.
- New optional tests behind each new capability flag.
- New `preparePersistenceId(pid: String)` hook called from `beforeEach`,
  mirroring `JournalSpec`. Plugins backed by external state (JDBC, NATS, etc.)
  can override this to clean residual records between tests.
- Test `timeout` promoted to `protected` so subclasses can override for slow
  stores.

### Reference TCK (`PersistenceTestKitDurableStateStoreTCKSpec.scala`)

Added an explicit `supportsSerialization = off` override since the in-memory
testkit doesn't serialize values.

### Docs (`persistence-journals.md` + doc-test files)

Added a brief subsection in the `## Plugin TCK` section mirroring the
Journal/Snapshot pattern, with `#durable-state-tck-scala` and
`#durable-state-tck-java` snippet markers.

## Deferred for follow-up

These ideas surfaced during this work but were deliberately scoped out:

- **Query-by-tag roundtrip** — verify a tag passed to `upsertObject` is
  retrievable via `DurableStateStoreQuery.changes(...)`. Requires extending the
  TCK to cover the read-side query SPI.
- **Get-after-delete revision contract** — current tests assert `value=None`
  after delete but not what `revision` should be (impls differ; needs a
  committer decision).
- **Idempotent upsert** — repeating `upsertObject(pid, N, sameValue, tag)` for
  replay safety.
- **Slice-based query roundtrip** (`DurableStateStoreBySliceQuery`).
- **Paged persistence-ID enumeration** (`DurableStateStorePagedPersistenceIdsQuery`).

Happy to open follow-up PRs once this lands.

## Test plan

- [x] `sbt scalafmtCheckAll` — clean
- [x] `sbt persistence-tck/Test/compile docs/Test/compile` — clean
- [x] `sbt "persistence-testkit/Test/testOnly *PersistenceTestKitDurableStateStoreTCKSpec"` —
  6/6 required tests pass; 4 optional skip as expected (testkit doesn't enforce
  OCC, doesn't serialize, has a no-op `deleteObject(pid)`).

Closes #2831.

### DIFF
--- a/persistence-tck/src/main/scala/org/apache/pekko/persistence/state/DurableStateStoreSpec.scala
+++ b/persistence-tck/src/main/scala/org/apache/pekko/persistence/state/DurableStateStoreSpec.scala
@@ -7,6 +7,10 @@
  * This file is part of the Apache Pekko project, which was derived from Akka.
  */
 
+/*
+ * Copyright (C) 2014-2022 Lightbend Inc. <https://www.lightbend.com>
+ */
+
 package org.apache.pekko.persistence.state
 
 import scala.concurrent.Await
--- a/persistence-tck/src/main/scala/org/apache/pekko/persistence/state/DurableStateStoreSpec.scala
+++ b/persistence-tck/src/main/scala/org/apache/pekko/persistence/state/DurableStateStoreSpec.scala
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+package org.apache.pekko.persistence.state
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+import org.apache.pekko
+import pekko.actor.ActorSystem
+import pekko.persistence.CapabilityFlag
+import pekko.persistence.DurableStateStoreCapabilityFlags
+import pekko.persistence.PluginSpec
+import pekko.persistence.scalatest.{ MayVerb, OptionalTests }
+import pekko.persistence.state.scaladsl.DurableStateUpdateStore
+
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+
+object DurableStateStoreSpec {
+  val config: Config = ConfigFactory.empty()
+}
+
+/**
+ * This spec aims to verify custom pekko-persistence [[DurableStateStore]] implementations.
+ * Plugin authors are highly encouraged to include it in their plugin's test suites.
+ *
+ * In case your durable state store plugin needs some kind of setup or teardown, override the `beforeAll`
+ * or `afterAll` methods (don't forget to call `super` in your overridden methods).
+ *
+ * For a Java and JUnit consumable version of the TCK please refer to
+ * [[pekko.persistence.japi.state.JavaDurableStateStoreSpec]].
+ *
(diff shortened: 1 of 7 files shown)


persistence-tck/src/main/scala/org/apache/pekko/persistence/state/DurableStateStoreSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+/*
+ * Copyright (C) 2014-2022 Lightbend Inc. <https://www.lightbend.com>
```
He-Pin: Not 2023?

He-Pin (approved):
lgtm

persistence-tck/src/main/scala/org/apache/pekko/persistence/state/DurableStateStoreSpec.scala
pjfanning: it is copied from a file that has 2022 in it

He-Pin: but IIRC akka will update the year every year

pjfanning: we shouldn't update the date if we don't grab newer akka code from 2023 - this PR does not grab any 2023 akka code

pjfanning:
Merged, thanks

pjfanning:
@drse the TCK you added fails with pekko-persistence-jdbc. It looks like deleteObject there is implemented incorrectly.
Could you have a look?
Copilot suggests this change to fix deleteObject.
https://github.com/apache/pekko-persistence-jdbc/pull/482/changes/4cee1c466a783696825a7fa55b0154acfb8173d9

The Copilot summary for this change is
https://github.com/pjfanning/incubator-pekko-persistence-jdbc/pull/2#issuecomment-4444870449

@ptrdom @He-Pin would either of you be able to review this too?

drse:
Thanks, fix looks correct. Looks like the tck was useful afterall.

PR #2923, opened 2026-04-27
Initialize DistributedData extension from ExternalShardAllocationStra…

Philippus:
…tegy eagerly. Ports https://github.com/akka/akka-core/pull/31907, part of https://github.com/akka/akka-core/releases/tag/v2.8.1 which is now available under the Apache License 2.0.


He-Pin:
The description should be updated with more detail.

He-Pin (approved):
lgtm

PR #2924, opened 2026-04-27
Optimize two-part ByteString concatenation

He-Pin:
Motivation:
Small ByteString concatenation currently promotes two fragments into ByteStrings backed by a Vector. This is expensive for hot paths that build a frame from a header and payload before reading or copying it.

Modification:
- Add an internal two-fragment ByteString representation for simple non-empty fragments.
- Keep ByteStrings for three or more fragments by flattening ByteString2 only when needed.
- Add unchecked internal byte access for already bounds-checked reads.
- Extend ByteString append benchmarks to cover append, read, copy, and cross-boundary header reads.

Result:
- ByteStringSpec: 203 tests passed.
- git diff --check: passed.
- actor / scalafmtCheck and bench-jmh / scalafmtCheck: passed.
- mimaReportBinaryIssues: passed.
- +mimaReportBinaryIssues: passed for Scala 2.13 and Scala 3.3.7.

Focused JMH, JDK 25, short local run with -prof gc:

| Benchmark | main | this branch |
| --- | ---: | ---: |
| appendTwo | 67,579 ops/ms, 160 B/op | about 503k ops/ms, 32 B/op |
| appendTwoAndReadHeader | 150,154 ops/ms | about 482k ops/ms |
| readTwoPartHeader | 249,745 ops/ms | 545k-556k ops/ms |
| appendTwoAndCopyToBuffer | 8,596 ops/ms, 112 B/op | 12,604 ops/ms, about 0 B/op |

Additional A/B on this branch showed byteAtUnchecked improves cross-fragment reads:

| Benchmark | without unchecked access | with unchecked access |
| --- | ---: | ---: |
| readTwoPartCrossBoundaryHeader | 244,277 ops/ms | 285,612 ops/ms |
| appendTwoAndReadCrossBoundaryHeader | 306,544 ops/ms | 338,764 ops/ms |

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -187,6 +187,8 @@ object ByteString {
   final class ByteString1C private (private val bytes: Array[Byte]) extends CompactByteString {
     def apply(idx: Int): Byte = bytes(idx)
 
+    private[pekko] override def byteAtUnchecked(offset: Int): Byte = bytes(offset)
+
     override def length: Int = bytes.length
 
     // Avoid `iterator` in performance sensitive code, call ops directly on ByteString instead
@@ -217,7 +219,13 @@ object ByteString {
     override def ++(that: ByteString): ByteString = {
       if (that.isEmpty) this
       else if (this.isEmpty) that
-      else toByteString1 ++ that
+      else
+        that match {
+          case b: ByteString1C => ByteString2.fromNonEmptySimpleFragments(this, b)
+          case b: ByteString1  => ByteString2.fromNonEmptySimpleFragments(this, b)
+          case bs: ByteString2 => ByteStrings(this, bs)
+          case bs: ByteStrings => ByteStrings(toByteString1, bs)
+        }
     }
 
     override def take(n: Int): ByteString =
@@ -518,6 +526,8 @@ object ByteString {
 
     def apply(idx: Int): Byte = bytes(checkRangeConvert(idx))
 
+    private[pekko] override def byteAtUnchecked(offset: Int): Byte = bytes(startIndex + offset)
+
     // Avoid `iterator` in performance sensitive code, call ops directly on ByteString instead
     override def iterator: ByteIterator.ByteArrayIterator =
       ByteIterator.ByteArrayIterator(bytes, startIndex, startIndex + length)
@@ -633,11 +643,12 @@ object ByteString {
       else if (this.isEmpty) that
       else
         that match {
-          case b: ByteString1C => ByteStrings(this, b.toByteString1)
+          case b: ByteString1C => ByteString2.fromNonEmptySimpleFragments(this, b)
           case b: ByteString1  =>
             if ((bytes eq b.bytes) && (startIndex + length == b.startIndex))
               new ByteString1(bytes, startIndex, length + b.length)
-            else ByteStrings(this, b)
+            else ByteString2.fromNonEmptySimpleFragments(this, b)
+          case bs: ByteString2 => ByteStrings(this, bs)
           case bs: ByteStrings => ByteStrings(this, bs)
         }
     }
@@ -886,14 +897,27 @@ object ByteString {
 
   private[pekko] object ByteStrings extends Companion {
     def apply(bytestrings: Vector[ByteString1]): ByteString =
-      new ByteStrings(bytestrings, bytestrings.foldLeft(0)(_ + _.length))
+      apply(bytestrings, bytestrings.foldLeft(0)(_ + _.length))
 
-    def apply(bytestrings: Vector[ByteString1], length: Int): ByteString = new ByteStrings(bytestrings, length)
+    def apply(bytestrings: Vector[ByteString1], length: Int): ByteString =
+      bytestrings.length match {
+        case 0 => ByteString.empty
+        case 1 => bytestrings.head
+        case 2 => ByteString2(bytestrings(0), bytestrings(1))
+        case _ => new ByteStrings(bytestrings, length)
+      }
 
     def apply(b1: ByteString1, b2: ByteString1): ByteString = compare(b1, b2) match {
-      case 3 => new ByteStrings(Vector(b1, b2), b1.length + b2.length)
+      case 3 => ByteString2(b1, b2)
       case 2 => b2
       case 1 => b1
+      case 0 => ByteString.empty
+    }
+
+    def apply(b: ByteString, bs: ByteString2): ByteString = compare(b, bs) match {
+      case 3 => new ByteStrings(addFragments(b, bs), b.length + bs.length)
+      case 2 => bs
+      case 1 => b
       case 0 => ByteString.empty
     }
 
@@ -904,10 +928,38 @@ object ByteString {
       case 0 => ByteString.empty
     }
 
+    def apply(bs: ByteString2, b: ByteString): ByteString = compare(bs, b) match {
+      case 3 => new ByteStrings(addFragments(bs, b), bs.length + b.length)
+      case 2 => b
+      case 1 => bs
+      case 0 => ByteString.empty
+    }
+
     def apply(bs: ByteStrings, b: ByteString1): ByteString = compare(bs, b) match {
       case 3 => new ByteStrings(bs.bytestrings :+ b, bs.length + b.length)
       case 2 => b
       case 1 => bs
+      case 0 => ByteString.empty
+    }
+
+    def apply(bs: ByteStrings, b: ByteString2): ByteString = compare(bs, b) match {
+      case 3 => new ByteStrings(addFragments(bs, b), bs.length + b.length)
+      case 2 => b
+      case 1 => bs
+      case 0 => ByteString.empty
+    }
+
+    def apply(b: ByteString2, bs: ByteStrings): ByteString = compare(b, bs) match {
+      case 3 => new ByteStrings(addFragments(b, bs), b.length + bs.length)
+      case 2 => bs
+      case 1 => b
+      case 0 => ByteString.empty
+    }
+
+    def apply(b1: ByteString2, b2: ByteString2): ByteString = compare(b1, b2) match {
+      case 3 => new ByteStrings(addFragments(b1, b2), b1.length + b2.length)
+      case 2 => b2
+      case 1 => b1
       case 0 => ByteString.empty
     }
 
@@ -924,6 +976,21 @@ object ByteString {
         if (b2.isEmpty) 0 else 2
       else if (b2.isEmpty) 1
       else 3
+
+    private[ByteString] def addFragments(first: ByteString, second: ByteString): Vector[ByteString1] = {
+      val builder = new VectorBuilder[ByteString1]
+      addFragments(builder, first)
+      addFragments(builder, second)
+      builder.result()
+    }
+
+    private[ByteString] def addFragments(builder: VectorBuilder[ByteString1], byteString: ByteString): Unit =
+      byteString match {
+        case b: ByteString1C => builder += b.toByteString1
+        case b: ByteString1  => builder += b
+        case b: ByteString2  => b.addFragmentsTo(builder)
+        case b: ByteStrings  => builder ++= b.bytestrings
+      }
 
     val SerializationIdentity = 2.toByte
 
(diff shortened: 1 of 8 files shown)


He-Pin:
@pjfanning I think this will help the pekko-grpc case

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -902,6 +969,338 @@ object ByteString {
...
+        (byteAt(offset + 1, firstLength) & 0xFF) << 16 |
+        (byteAt(offset + 2, firstLength) & 0xFF) << 8 |
+        (byteAt(offset + 3, firstLength) & 0xFF)
+    }
+
+    private[pekko] override def readIntLEUnchecked(offset: Int): Int = {
+      if (offset + java.lang.Integer.BYTES <= firstLength) first.readIntLEUnchecked(offset)
+      else if (offset >= firstLength) second.readIntLEUnchecked(offset - firstLength)
+      else
+        (byteAt(offset, firstLength) & 0xFF) |
```
pjfanning: can you use SWARUtil instead of having all this bitwise code?

pjfanning:
@hepin this code does not appear to support JDK Serialization of the new class

pjfanning:
@He-Pin there are probably a few cases where we use ++ to append 2 ByteStrings but where we could build up the binary data ourselves in a byte array and create a ByteString.fromArrayUnsafe.
Might help us avoid the complications of adding a ByteString2 class.
Or we could look at having the ++ method overridden on byte array backed ByteString impls so that it tries to return a new byte array backed ByteString where possible instead of ByteStrings instance.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -1236,61 +1236,69 @@ object ByteString {
...
     private[pekko] override def readIntLEUnchecked(offset: Int): Int = {
       if (offset + java.lang.Integer.BYTES <= firstLength) first.readIntLEUnchecked(offset)
       else if (offset >= firstLength) second.readIntLEUnchecked(offset - firstLength)
-      else
-        (byteAt(offset, firstLength) & 0xFF) |
-        (byteAt(offset + 1, firstLength) & 0xFF) << 8 |
-        (byteAt(offset + 2, firstLength) & 0xFF) << 16 |
-        (byteAt(offset + 3, firstLength) & 0xFF) << 24
+      else {
+        SWARUtil.getIntLE(
```
pjfanning, line 1381: Could we use the public method, getLong? This supports var handles that should be faster than byte by byte manipulation?

pjfanning: We could at least check if the first underlying byte string has enough bytes so that we can delegate the call to it.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -902,6 +969,346 @@ object ByteString {
...
+  /**
+   * A ByteString with exactly two simple fragments.
+   */
+  private[pekko] final class ByteString2 private (
+      private val first: ByteString,
+      private val second: ByteString,
+      val length: Int)
+      extends ByteString
+      with Serializable {
+    if (first.isEmpty) throw new IllegalArgumentException("first must not be empty")
```
pjfanning: Are these checks worth the overhead. You could use a Scala assert to enable this in tests but leave it disabled otherwise. scala has 'require' if we want to have runtime checks - less boilerplate than this code.

He-Pin:
Addressed the review comments in 3a2bca8b6d:

- Added a dedicated ByteString2 Java serialization identity/read path and serializer-binding coverage.
- Enforced ByteString2 as exactly two non-empty simple fragments through the assert-based factory path.
- Reused the supplied precomputed length in ByteStrings.apply(..., length), including the 0/1/2 fragment fast paths.
- Cached firstLength on ByteString2 hot paths and delegated primitive reads to the underlying fragment when the value is contiguous; cross-fragment reads still use SWARUtil byte helpers to avoid copying.
- Replaced the bare ByteString2 cast in the test with an explicit expectation helper.

Local verification passed:

- sbt "actor-tests/testOnly org.apache.pekko.util.ByteStringSpec" "actor-tests/testOnly org.apache.pekko.serialization.PrimitivesSerializationSpec org.apache.pekko.serialization.ReferenceSerializationSpec org.apache.pekko.serialization.AllowJavaSerializationSpec"
- sbt "++3.3.7 actor-tests/testOnly org.apache.pekko.util.ByteStringSpec org.apache.pekko.serialization.PrimitivesSerializationSpec org.apache.pekko.serialization.ReferenceSerializationSpec org.apache.pekko.serialization.AllowJavaSerializationSpec"
- sbt "bench-jmh/compile"
- sbt scalafmtCheckAll
- git diff --check

He-Pin:
Follow-up pushed in 147572b054ed458ae6e5924adac8ec6863e8325e:

- Moved the ByteString2 simple-fragment/non-empty/length assertions to the private constructor, so the invariant is checked at the construction point while remaining assertion-only.
- Reused known result lengths in ByteString2 take/drop/dropRight paths.
- Resolved the review threads that are covered by the current implementation.

Local verification passed:

- sbt "actor-tests/testOnly org.apache.pekko.util.ByteStringSpec" "actor-tests/testOnly org.apache.pekko.serialization.PrimitivesSerializationSpec org.apache.pekko.serialization.ReferenceSerializationSpec org.apache.pekko.serialization.AllowJavaSerializationSpec" "++3.3.7 actor-tests/testOnly org.apache.pekko.util.ByteStringSpec org.apache.pekko.serialization.PrimitivesSerializationSpec org.apache.pekko.serialization.ReferenceSerializationSpec org.apache.pekko.serialization.AllowJavaSerializationSpec" "bench-jmh/compile" scalafmtCheckAll
- git diff --check

pjfanning:
For me, this is a lot of new code to suit one or 2 use cases where we build a ByteString from precisely 2 underlying byte strings. 
I only know of this potential use case:
https://github.com/apache/pekko-grpc/pull/686/changes#diff-8c921b2d5583d7729d19bc797839f0f5e811a008becb746926a33003ea6a50e6R106
In the end of the day, what this saves us in Vector overhead is probably dwarved many times by the I/O time.

He-Pin:
Merged current main into this branch in e6e9f567f4 and resolved the two conflicts:

- MetricsBasedResizerSpec: kept the second async send stabilization and retained main's awaitAssert guard before recording the performance log.
- ByteStringSpec: kept the ByteString2 serialization coverage and main's SerializationProxy round-trip coverage.
- Also removed one trailing whitespace issue from the merged ClusterShardingCoordinatorRoleSpec so diff checks stay clean.

On the design concern: I left the ByteString2 implementation in place for this draft because the simpler copy-to-compact fast path would trade the Vector allocation for copying payload bytes on header+payload appends. The current approach keeps that use case zero-copy while promoting to ByteStrings as soon as a third fragment is added. If the preferred direction is to keep ByteString smaller and accept the copy for two-fragment appends, I can rework the PR that way.

Local verification passed:

- sbt "actor-tests / Test / testOnly org.apache.pekko.util.ByteStringSpec org.apache.pekko.routing.MetricsBasedResizerSpec"
- sbt "actor-tests / Test / testOnly org.apache.pekko.serialization.PrimitivesSerializationSpec org.apache.pekko.serialization.ReferenceSerializationSpec org.apache.pekko.serialization.AllowJavaSerializationSpec"
- sbt "bench-jmh / Compile / compile"
- sbt "actor-tests / Test / scalafmt" "actor / Compile / scalafmt" "bench-jmh / Compile / scalafmt"
- git diff --check
- git diff --cached --check

All review threads are currently resolved.

pjfanning:
Having an extra, 4th, ByteString impl comes with too large a maintenance overhead. Lots more code and lots more tests and easy to miss having test coverage. I don't think there are too many places where we benefit - instances where we concat exactly 2 byte strings that we don't later concat or that have been previously concatenated.

He-Pin:
This is for the case pekko-grpc， where body + headers

pjfanning:
> This is for the case pekko-grpc， where body + headers

I know the use case and it don't think this is worth it. I documented my reasoning in https://github.com/apache/pekko/pull/2924#issuecomment-4378345386

He-Pin:
In Scala stdlib there are seq1234 ,map1234 and set1234, we need make it been well tested covered.

He-Pin:
@pjfanning Fair maintenance concern — let me put numbers behind the use case so we can decide on data, not vibes.

Pushed in c962213cbf:

1. **Removed a `LazyList` allocation in `ByteString2.iterator`** — replaced `LazyList(...)` with a 2-element `List`. That alone ~30× the read-side throughput on `iterator.getInt` / `readIntBE` paths and was the reason the pre-built downstream numbers looked unconvincing earlier.
2. **Added a directional JMH benchmark** (`ByteString_concat3way_Benchmark`) that mirrors the gRPC frame shape (5-byte header ++ payload, 64 B / 1 KB / 16 KB / 64 KB / 256 KB) and compares two strategies:
   - `copy` — assemble into a fresh array, wrap with `fromArrayUnsafe` (the alternative you suggested in https://github.com/apache/pekko/pull/2924#issuecomment-4326704648)
   - `bs2` — `ByteString2.apply(header, payload)` (this PR)
3. **Added a cross-check property test** in `ByteStringSpec` — the same logical bytes built as `ByteString1C` (reference) and as `ByteString2` with **every possible split point**, asserting equivalence across `apply`, full byte-by-byte iteration, `iterator.getShort/Int/Long` (BE+LE) at every offset, `readShortBE/LE`/`readIntBE/LE`/`readLongBE/LE` at every offset, `take/drop/takeRight/dropRight/slice` at every cut, `copyToBuffer`, `copyToArray`, `indexOf`, `compact`, equality and hashCode. That should cover the "easy to miss having test coverage" concern — a regression in any read/transform path that diverges from `ByteString1C` will be caught.

JDK 25, G1GC, single fork, 5×1s measurement. Numbers are ops/μs (higher is better).

**Build the frame** (the operation pekko-grpc actually does per outgoing message):

| size | copy | bs2 |
|---:|---:|---:|
| 64 B    | 163.60 | **501.90**  (3.07×) |
| 1 KB    |  22.82 | **527.13**  (23.10×) |
| 16 KB   |   1.67 | **516.72**  (308×) |
| 64 KB   |   0.63 | **510.53**  (815×) |
| 256 KB  |   0.15 | **444.32**  (3038×) |

**Build the frame and read the header** (parser hot path):

| size | copy | bs2 |
|---:|---:|---:|
| 64 B    | 147.03 | **616.02**  (4.19×) |
| 1 KB    |  23.67 | **609.47**  (25.75×) |
| 16 KB   |   1.97 | **608.79**  (308×) |
| 64 KB   |   0.59 | **616.68**  (1053×) |
| 256 KB  |   0.14 | **612.79**  (4367×) |

**Build the frame and write to a `ByteBuffer`** (Netty / NIO sink):

| size | copy | bs2 |
|---:|---:|---:|
| 64 B   | 102.93 | **154.73**  (1.50×) |
| 1 KB   |  17.23 | **49.49**   (2.87×) |
| 16 KB  |   1.29 | **2.99**    (2.31×) |
| 64 KB  |   0.34 | **0.65**    (1.95×) |
| 256 KB |   0.08 | **0.25**    (2.93×) |

**Build the frame and `drop(5)` (strip header)** — the inbound side:

| size | copy | bs2 |
|---:|---:|---:|
| 64 B   | 150.22 | **909.84**  (6.06×) |
| 1 KB   |  23.67 | **917.58**  (38.76×) |
| 256 KB |   0.14 | **913.26**  (6607×) |

**Where bs2 is *slower* than copy** — pre-built downstream reads on the merged result:

| benchmark | size | copy | bs2 | bs2 / copy |
|---|---:|---:|---:|---:|
| `readIntBE(0) + apply(4)` | 1 KB | 1196.56 | 730.76 | 0.61× |
| `iterator.getInt + apply(4)` | 1 KB | 909.99 | 400.68 | 0.44× |
| `asByteBuffers` | 1 KB | 200.44 | 98.90 | 0.49× |

This is the honest cost: when *every other op* is amortized away, contiguous wins because the bounds check / vtable dispatch is cheaper. But the "build, then do one downstream op" workload — which is what real frame producers and consumers do — wins on `bs2` by 2–4000× because we never pay the `System.arraycopy`. And per [pekko-grpc#686](https://github.com/apache/pekko-grpc/pull/686/files#diff-8c921b2d5583d7729d19bc797839f0f5e811a008becb746926a33003ea6a50e6R106), that arraycopy *is* the alternative we're being compared against.

**Re: "I/O dwarfs the difference"** — for a streaming RPC at ~1 M small messages/s, the per-frame `System.arraycopy(payload)` is exactly proportional to throughput in CPU and allocation. On a 16 KB payload that's 16 GB/s of additional memory traffic at 1 M frames/s; on a 64 KB payload, 64 GB/s. That's not nothing even when an I/O wait sits on top.

Happy to drop this if you still don't think the maintenance is worth it — but I wanted to make sure the decision was informed by the post-fix numbers and a reproducible benchmark, since the pre-fix iterator regression made the pre-built ops look much worse than they actually are.

pjfanning:
@hepin there are commits in this PR for 'stabilising' unrelated tests - can you remove any commits not related to ByteString changes

He-Pin:
Yes， will do

pjfanning:
the code is a bit outdated as it doesn't take into account some recent ByteString changes for copyToBuffer, endsWith/startsWith and map. https://github.com/pjfanning/incubator-pekko/pull/51 has suggested changes.

pjfanning (approved):
lgtm

PR #2928, opened 2026-04-27
add tests to see if old format ByteStrings serialized using JDK can be deserialized

pjfanning:
an experiment but could be useful as a regression test


pjfanning:
I'll look again at this. It does seem like a good idea to put a serialVersionId on the problem class but I will need to get the right value from the 1.6.0 release and to update the test bin files.

pjfanning:
Actually after #2929, Pekko 2 won't be able to deserialize Pekko 1.x ByteString1C instances with or without this change.
At this stage, JDK Serialization of ByteStrings in Pekko 1.x and Akka is not a good idea and it might be worth waiting for a real would world complaint before making any changes in Pekko 1.x.

PR #2937, opened 2026-05-05
aeron 1.50.4

pjfanning:
https://github.com/apache/pekko/issues/2329 is still an issue - this works around the small behaviour change

Needed to remove some unused code from a test source file, AeronStats, because the unused code in that class needed an Agrona class that has been removed in Agrona 2.4.x.

#2939 is open for the follow up changes needed for Aeron 1.51+

### DIFF
--- a/.github/workflows/nightly-builds-aeron.yml
+++ b/.github/workflows/nightly-builds-aeron.yml
@@ -12,6 +12,7 @@ on:
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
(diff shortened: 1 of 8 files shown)


pjfanning:
We have some code (AeronStat) that uses https://javadoc.io/doc/org.agrona/agrona/2.3.0/org/agrona/concurrent/SigInt.html and it has been removed in Agrona 2.4.x.

He-Pin (approved):
lgtm

PR #2938, opened 2026-05-05
remove unnecessary code from AeronStats

pjfanning:
* issues with CI in #2937 
* but it would be good to prune the unnecessary code from AeronStats because of the use of deprecated and now removed class, SigInt.


He-Pin (approved):
lgtm

pjfanning:
#2937 merged with this change

PR #2947, opened 2026-05-08
feat: Make it possible to define appVersion later

pjfanning:
Part of https://github.com/akka/akka-core/releases/tag/v2.8.2 which is now available under 
Apache License v2.0.

### Motivation
Alternative to the config `app-version` that is defined before `ActorSystem` startup.
If the version is read from an external system (e.g. Kubernetes) this feature makes it possible to define it after system startup but before joining.

Port of https://github.com/akka/akka-core/pull/31934.

### Modification
- Added `SetAppVersionLater` / `SetAppVersion` messages to `ClusterUserAction` in `ClusterDaemon.scala`
- Added `laterAppVersion` state and `setAppVersionLater` / `setAppVersion` private methods to `ClusterCoreDaemon`
- Updated `uninitialized`, `tryingToJoin`, and `initialized` receive handlers to handle the new messages
- Updated `join()` to defer sending a `Join` message when the appVersion `Future` is not yet completed
- Updated `joining()` to use the resolved `laterAppVersion` when adding self as a member
- Added `setAppVersionLater(Future[Version])` (Scala) and `setAppVersionLater(CompletionStage[Version])` (Java) API to `Cluster.scala`
- Added `SetAppVersionLater` case class with Scala and Java constructors to `cluster-typed/Cluster.scala`
- Handled `SetAppVersionLater` in `AdaptedClusterImpl.managerBehavior`
- Added `AppVersionSpec` multi-JVM test
- Uses `scala.jdk.FutureConverters` (Pekko standard) instead of `scala.compat.java8.FutureConverters` (Akka)

### Result
Users can now supply `appVersion` from an external source (e.g. Kubernetes) after `ActorSystem` startup but before joining the cluster. The cluster join will wait for the `Future`/`CompletionStage` to complete before proceeding.

### References
Refs https://github.com/akka/akka-core/pull/31934

### DIFF
--- a/cluster-typed/src/main/scala/org/apache/pekko/cluster/typed/Cluster.scala
+++ b/cluster-typed/src/main/scala/org/apache/pekko/cluster/typed/Cluster.scala
@@ -13,7 +13,10 @@
 
 package org.apache.pekko.cluster.typed
 
+import java.util.concurrent.CompletionStage
+
 import scala.collection.immutable
+import scala.concurrent.Future
 
 import org.apache.pekko
 import pekko.actor.Address
@@ -24,6 +27,7 @@ import pekko.cluster._
 import pekko.cluster.ClusterEvent.{ ClusterDomainEvent, CurrentClusterState }
 import pekko.cluster.typed.internal.AdapterClusterImpl
 import pekko.japi.Util
+import pekko.util.Version
 
 /**
  * Messages for subscribing to changes in the cluster state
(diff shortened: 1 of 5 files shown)


cluster-typed/src/main/scala/org/apache/pekko/cluster/typed/Cluster.scala
```diff
@@ -128,6 +132,30 @@ final case class JoinSeedNodes(seedNodes: immutable.Seq[Address]) extends Cluste
...
+object SetAppVersionLater {
+
+  /**
+   * Java API: If the `appVersion` is read from an external system (e.g. Kubernetes) it can be defined after
+   * system startup but before joining by completing the `appVersion` `CompletionStage`. In that case,
+   * `SetAppVersionLater` should be sent before [[Join]] or [[JoinSeedNodes]]. It's fine to send
+   * `Join` or `JoinSeedNodes` immediately afterwards (before the `CompletionStage` is completed). The join will
+   * then wait for the `appVersion` to be completed.
+   */
+  def create(appVersion: CompletionStage[Version]): SetAppVersionLater = {
```
He-Pin, line 153: Maybe better with _<: Version

He-Pin (approved):
lgtm

PR #2945, opened 2026-05-08
Stabilize JDK 25 virtualized stream tests

He-Pin:
### Motivation

The JDK 25 nightly run with virtualized dispatchers can hit load-sensitive timeouts in several stream tests and in `MixedProtocolClusterSpec` shutdown. The observed failures include `actor-system-terminate` timing out after 10 seconds and stream tests timing out while waiting for graph progress or abrupt materializer cleanup.

### Modification

* Increase `ClusterTestUtil.shutdownAll` and `MixedProtocolClusterSpec` actor-system termination windows for virtualized JDK 25 nightly runs.
* Make exposed stream test waits honor `pekko.test.timefactor` via dilated timeouts.
* Ensure `FlowMapWithResourceSpec` abrupt materializer termination tests wait until the resource has actually been created before shutting down the materializer.

### Result

This keeps the tests strict about completion and cleanup, but avoids failing under JDK 25 virtual-thread scheduling contention before the stream or actor system has had enough time to make progress.

### Verification

Ran locally with JDK 25.0.1 and the nightly virtual-thread dispatcher settings:

* `++ 3.3.x stream-tests/testOnly org.apache.pekko.stream.scaladsl.FlowScanSpec org.apache.pekko.stream.scaladsl.FlowMapAsyncUnorderedSpec org.apache.pekko.stream.scaladsl.FlowMapWithResourceSpec -- -z "restart properly" -z "not run more futures than configured" -z "abrupt materializer termination"`
* `++ 2.13.x stream-tests/testOnly org.apache.pekko.stream.scaladsl.FlowScanSpec org.apache.pekko.stream.scaladsl.FlowMapAsyncUnorderedSpec org.apache.pekko.stream.scaladsl.FlowMapWithResourceSpec -- -z "restart properly" -z "not run more futures than configured" -z "abrupt materializer termination"`
* `++ 3.3.x stream-tests/testOnly org.apache.pekko.stream.scaladsl.HubSpec`
* `cluster/testOnly org.apache.pekko.cluster.MixedProtocolClusterSpec -- -z "allow a node using the pekko protocol to join the cluster (udp)"`
* `git diff --check`

PR #2949, opened 2026-05-09
test: stabilize two flaky tests on JDK21+/virtualized nightly

He-Pin:
Two unrelated flaky tests surfaced again on the nightly schedule. Reference run [25586760215](https://github.com/apache/pekko/actions/runs/25586760215) failed; the manually triggered [25580301468](https://github.com/apache/pekko/actions/runs/25580301468) on the same `main` commit passed — confirming intermittent behavior on JDK 21/25 with virtualized dispatchers, not a regression.

---

## 1. MetricsBasedResizerSpec — "record the performance log with the correct pool size"

**Motivation**
Test fails after ~12s on JDK 25 / Scala 2.13.x with `None was empty (MetricsBasedResizerSpec.scala:245)`. Between the two `reportMessageCount` checkpoints the test sent a single extra `mockSend` per routee. With virtual-thread scheduling those messages can finish processing before the second checkpoint runs, leaving routees idle (`currentMessage = null`) so `messagesInRoutees` underestimates the queue, the resizer's `fullyUtilized` predicate falls through, no performance log is recorded, and the final assertion fails.

**Modification**
Replace the inter-checkpoint `mockSend` pair with a second `sendToAll(await = false)` call and `Await.ready` on its first latch. Both routees stay actively holding their `currentMessage` while `reportMessageCount` runs.

**Result**
The resizer always observes the routees as fully utilized across both checkpoints. Test runs deterministically on virtualized dispatchers; no longer relies on routees still being busy by chance.

---

## 2. FlowMapWithResourceSpec — "will close the autocloseable resource on abrupt materializer termination"

**Motivation**
Test fails in 22–115 ms on JDK 21 / Scala 3.3.x with `1 was not equal to 0 (FlowMapWithResourceSpec.scala:538)`. `mapWithResource` carries the `IODispatcher` attribute, so it lands on a different dispatcher than `Source.single`/`Sink.never` and the materializer introduces an async-island bridge whose `SubSink` can pull eagerly. Once that pull propagates back, `Source.single` pushes its element and completes, `StatefulMap.onUpstreamFinish` fires, `closeStateAndComplete` runs the close callback and `closedCounter` reaches 1 *before* the test thread issues `mat.shutdown()`. The pre-shutdown `closedCounter.get shouldBe 0` is a timing assumption that doesn't reflect the operator's contract.

**Modification**
Remove the pre-shutdown `closedCounter` assertion. The `Await` on `created.future` (added in #2945) still proves `create()` ran before shutdown.

**Result**
The abrupt-termination guarantee is still verified by the `AbruptTerminationException` check on the materialized value, the `Await` on `promise.future`, and the post-shutdown `closedCounter == 1` assertion. The fragile timing-dependent pre-check is gone.

---

## Test plan
- [x] `actor-tests/testOnly org.apache.pekko.routing.MetricsBasedResizerSpec` passes locally
- [x] `stream-tests/testOnly org.apache.pekko.stream.scaladsl.FlowMapWithResourceSpec` passes locally
- [x] `sbt scalafmtAll` clean
- [ ] Nightly virtualized run on JDK 21 / 25 stays green

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/routing/MetricsBasedResizerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/routing/MetricsBasedResizerSpec.scala
@@ -235,15 +235,12 @@ class MetricsBasedResizerSpec extends PekkoSpec(ResizerSpec.config) with Default
     "record the performance log with the correct pool size" in {
       val resizer = DefaultOptimalSizeExploringResizer()
       val router = TestRouter(routees(2))
-      val msgs = router.sendToAll(await = true)
+      val msgs1 = router.sendToAll(await = true)
+      val msgs2 = router.sendToAll(await = false)
       resizer.reportMessageCount(router.routees, router.msgs.size)
-      msgs.head.second.open()
+      msgs1.head.second.open()
 
-      router.mockSend(await = true, routeeIdx = 0)
-      router.mockSend(await = false, routeeIdx = 1)
-      awaitAssert {
-        resizer.updatedStats(router.routees, router.msgs.size)._1.get(2) should not be empty
-      }
+      Await.ready(msgs2.head.first, timeout.duration)
       resizer.reportMessageCount(router.routees, router.msgs.size)
       resizer.performanceLog.get(2) should not be empty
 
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #2952, opened 2026-05-09
perf: fast path getShort/Int/Long in MultiByteArrayIterator

He-Pin:
### Motivation

`MultiByteArrayIterator` (the iterator returned by multi-fragment `ByteString`s such as `ByteStrings` and the new `ByteString2`) inherits its `getShort/Int/Long` from `ByteIterator`, which always reads byte-by-byte through `getByte`. This skips the SWAR fast path that `ByteArrayIterator` already implements for the common case where the read fits entirely inside the current fragment.

For callers that hold a `MultiByteArrayIterator` and pull several primitives from it (codecs, framing parsers), every read pays the byte-loop overhead even though most reads do not actually cross a fragment boundary.

### Modification

Override `getShort`, `getInt`, and `getLong` on `MultiByteArrayIterator`:

- If the current `ByteArrayIterator` has at least the requested primitive's `BYTES` available, delegate to its SWAR-backed read and then call `normalize()` to advance over any now-empty current iterator.
- Otherwise fall back to `super.getShort/Int/Long`, which already handles cross-fragment reads correctly.

Mirrored to both `actor/src/main/scala-2.13/.../ByteIterator.scala` and `actor/src/main/scala-3/.../ByteIterator.scala`.

### Result

Reused multi-fragment iterators avoid the byte-by-byte loop when the read does not cross a fragment boundary. Cross-fragment reads keep the existing semantics. No public surface change; pure internal override of an existing method.

### Tests

Added `ByteIteratorSpec` case `match ByteArrayIterator semantics for getShort/Int/Long across both fast and cross-fragment paths` which, for a 16-byte source and every split point in `1..15`, asserts that `combined.iterator.drop(off).getShort/Int/Long` equals the `ByteArrayIterator` reference value at every offset and in both byte orders. This exercises both the new fast path (read fits in current fragment) and the inherited byte-by-byte path (read straddles fragments).

```
sbt 'project actor-tests' 'testOnly org.apache.pekko.util.ByteIteratorSpec'
[info] - should correctly implement indexOf
[info] - should match ByteArrayIterator semantics for getShort/Int/Long across both fast and cross-fragment paths
[info] Tests: succeeded 2, failed 0
```

`sbt 'project actor' 'mimaReportBinaryIssues'` clean.
`sbt scalafmtAll` and `sbt headerCreateAll` clean.

### References

Refs #2924 - follow-up extracted from the ByteString2 PR review.

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/util/ByteIteratorSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/ByteIteratorSpec.scala
@@ -13,6 +13,8 @@
 
 package org.apache.pekko.util
 
+import java.nio.ByteOrder
+
 import org.apache.pekko.util.ByteIterator.ByteArrayIterator
 
 import org.scalatest.matchers.should.Matchers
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

PR #2919, opened 2026-04-26
Port akka-core PR #31906 and #31926: DnsResolutionActor and in-flight deduplication

pjfanning:
Part of https://github.com/akka/akka-core/releases/tag/v2.8.1 - now available under Apache License v2.0

## Summary

Ports changes from [akka-core #31906](https://github.com/akka/akka-core/pull/31906) and [akka-core #31926](https://github.com/akka/akka-core/pull/31926) to pekko, adapted for pekko's existing `IdGenerator` infrastructure.

### `AsyncDnsResolver.scala` — actor-based per-request resolution

**PR #31906 — "Refactor concurrency in AsyncDnsResolver":**

- Replaces the future-chain-based DNS resolution (which accessed actor state like `cache.put()` and `log` from future callbacks) with a per-request `DnsResolutionActor` pattern
- Each `DnsResolutionActor` is spawned per resolution request and manages its own lifecycle: applies search domains, fails over to other nameservers, handles `DuplicateId` retries using pekko's existing thread-safe `IdGenerator`
- `DnsResolutionActor` reports results back to `AsyncDnsResolver` via a `ResolutionAnswer` message, so all cache updates and sender notifications happen on the actor thread
- Adds **in-flight deduplication**: when a second `Resolve(name, mode)` arrives while one is already in-progress for the same `(name, mode)`, the second sender is added as a waiter and receives the same result when the first completes — no second DNS question is sent

**Pekko-specific adaptations:**
- Uses pekko's existing `IdGenerator` (random, thread-safe) instead of akka's sequential `RequestIdInjector` actor
- Keeps `DuplicateId` handling (pekko-specific feature) inside `DnsResolutionActor.sendQuestion`
- `ResolutionAnswer` replaces akka's `CachePut` + direct-to-sender pattern, routing everything through `AsyncDnsResolver` for proper in-flight tracking

### `AsyncDnsResolverSpec.scala` — new test coverage

**PR #31926 — additional test coverage:**

Three new tests added (adapted for pekko's random ID generator):
1. **"attempt to drop a failed question on timeout"** — verifies `DropRequest` is sent to the DNS client when an ask times out
2. **"not reuse the request ids of pending requests"** — verifies that concurrent resolutions for different names use unique DNS question IDs
3. **"reuse in-progress resolutions"** — verifies that two concurrent resolves for the same name share a single DNS lookup; only one DNS question is sent and both callers receive the answer

## Test Results

All 18 tests pass (15 existing + 3 new).

PR #2960, opened 2026-05-09
fix: harden async DNS request id handling

He-Pin:
Motivation:
The async DNS resolver should avoid reusing request ids that are still active, should only release failed ids after the DNS client confirms the drop, and should ignore UDP DNS responses from unexpected remotes.

Modification:
Track active request ids in the request-id injector, keep duplicate ids reserved, add an explicit Dropped acknowledgement for DNS client drops, match drops by the full DNS question including record type, and reject UDP responses from non-nameserver remotes with a security marker.

Result:
Concurrent async DNS requests avoid active id reuse, failed requests clean up ids only when confirmed safe, and spoofed UDP responses from unexpected remotes are ignored.

Tests:
- scalafmt --mode diff-ref=port-ef33e71-async-dns-id-injector / passed
- scalafmt --list --mode diff-ref=port-ef33e71-async-dns-id-injector / passed
- git diff --check / passed
- sbt "actor-tests / Test / testOnly org.apache.pekko.io.dns.internal.AsyncDnsResolverSpec org.apache.pekko.io.dns.internal.DnsClientSpec org.apache.pekko.io.dns.DnsSettingsSpec" / passed

References:
Upstream commit: https://github.com/akka/akka-core/commit/7434794bab6fec95cb52d810d95126f90efb8c34, which is now Apache licensed

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolverSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/io/dns/internal/AsyncDnsResolverSpec.scala
@@ -26,7 +26,7 @@ import pekko.io.dns.{ AAAARecord, ARecord, DnsSettings, IdGenerator, SRVRecord }
 import pekko.io.dns.CachePolicy.Ttl
 import pekko.io.dns.DnsProtocol._
 import pekko.io.dns.internal.AsyncDnsResolver.ResolveFailedException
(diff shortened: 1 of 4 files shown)


He-Pin:
Just look up the code and adjust, not a line by line port.

pjfanning:
extends #2919

pjfanning (approved):
lgtm

PR #2961, opened 2026-05-09
docs: clarify security guidance

He-Pin:


docs/src/main/paradox/security/index.md
```diff
@@ -14,12 +14,12 @@ We strongly encourage people to report such problems to our private security mai
 
 Please follow the [guidelines](https://www.apache.org/security/) laid down by the Apache Security team.
 
-Ideally, any issues affecting Apache Pekko and Akka should be reported to Apache team first. We will share the
-report with the Lightbend Akka team.
+Ideally, any issues affecting Apache Pekko and its predecessor project should be reported to the Apache Pekko team
```
pjfanning, line 17: I don't see why we shouldn't name Akka here.

raboof (approved):

docs/src/main/paradox/security/index.md
raboof: I agree there's no need to avoid it. Perhaps we can even drop this whole paragraph, though?

PR #2962, opened 2026-05-09
docs: document serializer constructor alternatives

He-Pin:
Motivation:
Pekko already supports additional serializer constructor shapes, but the public Scaladoc still documented only the older constructor alternatives.

Modification:
Update Serializer and SerializerWithStringManifest documentation to list the supported constructor shapes in the same order as the current loading implementation.

Result:
The serializer API documentation now matches the behavior covered by the existing constructor tests.

Tests:
- scalafmt --mode diff-ref=origin/main / passed
- scalafmt --list --mode diff-ref=origin/main / passed
- git diff --check / passed
- sbt "actor / Test / compile" "actor-tests / Test / testOnly org.apache.pekko.serialization.SerializeSpec" / passed

References:
Upstream commit: https://github.com/akka/akka-core/commit/a2ab65d501a94d93a592445df569f76825a78704, which is now Apache licensed
Refs #29765


pjfanning (approved):
lgtm

PR #2977, opened 2026-05-17
optimize: FlattenMerge avoids substream materialization for value-presented sources

He-Pin:
## Motivation

`FlattenMerge` previously only fast-pathed `SingleSource`. Every other
inner source — `Source.empty`, `Source(List)`, `Source.fromJavaStream`,
`Source.future(Future.successful(...))`, range / iterator / repeat
sources — paid the cost of materializing a `SubSinkInlet` plus a full
`subFusingMaterializer.materialize(...)` round-trip. `FlattenConcat`
already eliminates this via `TraversalBuilder.getValuePresentedSource`
(introduced in 1.2.0); `FlattenMerge` should benefit too.

This is especially important because the single-argument
`flatMapConcat(f)` is implemented as `via(new FlattenMerge(1))`, so all
default `flatMapConcat` users go through `FlattenMerge` and miss out on
the value-presented optimization until now.

## Modification

- Switch `FlattenMerge.addSource` from `getSingleSource` to
  `getValuePresentedSource`, dispatching on `SingleSource`,
  `IterableSource`, `IteratorSource`, `RangeSource`, `RepeatSource`,
  `JavaStreamSource`, `FutureSource`, `FailedSource`, and empty sources
  in place — mirroring `FlattenConcat`.
- Introduce an `InflightSource[T]` family in a new `FlattenMerge`
  companion to occupy a breadth slot for multi-element value-presented
  sources without materializing a substream. Tracked via a new
  `pendingInflightSources` counter so `activeSources` still bounds the
  breadth budget correctly.
- Preserve merge semantics: after each push from a multi-element
  inflight source, re-enqueue it so other concurrent sources keep
  interleaving (rather than draining one source first as `FlattenConcat`
  does).
- Completed `Future`s / `FailedSource` are folded directly: success
  pushes or queues a single element; failure calls `failStage`.
- Pending `Future`s register an `getAsyncCallback` and occupy a slot
  until completion.
- Empty inner sources are discarded in place and consume no slot.

`FlattenMerge` is `@InternalApi private[pekko]` so this is purely an
internal performance change; MiMa is clean.

## Result

- `flatMapMerge(breadth, ...)` and the default `flatMapConcat(...)`
  (which routes through `FlattenMerge(1)`) skip substream materialization
  for value-presented inner sources, cutting per-source GC and stage
  overhead.
- Behaviour is unchanged for non-value-presented sources (regular
  `Source`s built from graphs, lazy futures, etc.) — they still
  materialize a `SubSinkInlet` as before.

## Tests

- All existing `FlowFlattenMergeSpec` tests pass.
- All existing `FlowFlatMapConcatParallelismSpec` tests pass (covers the
  `FlattenMerge(1)` path used by single-arg `flatMapConcat`).
- New tests in `FlowFlattenMergeSpec` mirror
  `FlowFlatMapConcatParallelismSpec`:
  - `work with value presented sources with breadth: {1,2,4,8,16,32,64,128}`
    — covers empty / single / iterable / completed-future / lazy-future /
    delayed-future inner sources.
  - `work with generated value presented sources with breadth: ...` —
    randomised mix of value-presented sources at scale.
  - `work with value presented failed sources` — failure inside the
    optimized path.
  - `avoid pre-materialization for value-presented sources` and
    `not materialize value-presented sources` — assert the fast path
    actually fires.

Local commands run:

- `sbt "stream-tests/testOnly org.apache.pekko.stream.scaladsl.FlowFlattenMergeSpec"` → 37/37 pass
- `sbt "stream-tests/testOnly *FlatMap* *Flatten* *PrefixAndTail* *Concat*Spec"` → 232/232 pass
- `sbt scalafmtAll headerCreateAll`
- `sbt "stream/mimaReportBinaryIssues"` → no issues

## References

- Mirrors the value-presented-source optimization added to
  `FlattenConcat` in 1.2.0 (`TraversalBuilder.getValuePresentedSource`).

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowFlatMapConcatParallelismSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowFlatMapConcatParallelismSpec.scala
@@ -165,6 +165,37 @@ class FlowFlatMapConcatParallelismSpec extends StreamSpec("""
         .futureValue should ===(Seq[Integer](1, 2))
     }
 
+    "propagate inner-source failure queued behind a Success(null) pending future" in {
+      // Regression: tryPullNextSourceInQueue used to call src.tryPull() for
+      // InflightSource heads, which is a no-op for already-settled sources
+      // (e.g. InflightCompletedFutureSource(Failure(_))). Once the pending
+      // Success(null) head was removed, the failure sat in the queue forever
+      // because nothing surfaced it. Drive the head via pushOut instead.
+      val ex = new BoomException
+      val probe = Source(
+        List[Source[Integer, NotUsed]](
+          Source.future(after(1.millis)(Future.successful[Integer](null))),
+          Source.failed[Integer](ex)))
+        .flatMapConcat(parallelism = 4, identity)
+        .runWith(TestSink())
+      probe.request(1)
+      probe.expectError() should ===(ex)
+    }
+
+    "emit completed future queued behind a Success(null) pending future" in {
+      // Same root cause as above: an InflightCompletedFutureSource(Success(_))
+      // queued behind a Success(null) pending future was stranded because
+      // tryPull() is a no-op on already-settled sources.
+      val toIntegerSeq = Flow[Integer].grouped(1000).toMat(Sink.head)(Keep.right)
+      Source(
+        List[Source[Integer, NotUsed]](
+          Source.future(after(1.millis)(Future.successful[Integer](null))),
+          Source.future(Future.successful[Integer](42))))
+        .flatMapConcat(parallelism = 4, identity)
+        .runWith(toIntegerSeq)
+        .futureValue should ===(Seq[Integer](42))
+    }
+
     "work with value presented sources when demands slow" in {
       val prob = Source(
         List(Source.empty[Int], Source.single(1), Source(List(2, 3, 4)), Source.lazyFuture(() => Future.successful(5))))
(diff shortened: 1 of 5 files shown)


He-Pin:
Followup pushed:

- New commit `refactor: share InflightSource family between FlattenConcat and FlattenMerge` extracts the duplicated `InflightSource[T]` family (Iterator/Range/Repeat/CompletedFuture/PendingFuture) into a shared `pekko.stream.impl.fusing.InflightSources` package-private object. Net -176 lines of duplication; both stages now reference one canonical home, so any future additions to the value-presented optimization (e.g. supporting more inner-source types) only need to land in one place.
- Aligns `InflightCompletedFutureSource.isClosed` to FlattenConcat's `true` semantics (behaviorally equivalent across both stages, but more faithful to the source being a one-shot cached value).
- `tryPull` / `cancel` / `materialize` move from abstract to concrete no-op defaults so value-presented subclasses no longer carry empty overrides; the only place that needs to override them is FlattenConcat's `attachAndMaterializeSource` anonymous subclass.

### Why this likely matters in practice

Beyond the synthetic benchmarks, Apache Pekko HTTP uses `flatMapMerge` on hot paths (chunked entity assembly, response data processing). Those usages frequently feed `FlattenMerge` with `Source.empty` / `Source.single` / `Source(List(...))` / `Source.future(Future.successful(...))` shaped inner sources — exactly the shapes that previously paid the per-source `SubSinkInlet` + `subFusingMaterializer.materialize(...)` cost and now don't. So this should translate to a measurable reduction in per-request stage and GC overhead on Pekko HTTP workloads, not just streams microbenchmarks.

Same verification as before:
- `sbt "stream-tests/testOnly *FlowFlattenMergeSpec *FlowFlatMapConcatParallelismSpec *FlowFlatMapConcatSpec"` — 64/64 pass.
- `sbt scalafmtAll headerCreateAll` — applied.
- `sbt "stream/mimaReportBinaryIssues"` — clean (`@InternalApi private[fusing]`).

He-Pin:
I will do a follow up optimization around the `++` operation once this got merged.

He-Pin:
Hope this can improve some performance of pekko-http

He-Pin:
@pjfanning seems you are rate limited,

He-Pin:
Not sure why the action is never finished :(

He-Pin:
@pjfanning  I think this one is ready

pjfanning (approved):
lgtm

PR #2978, opened 2026-05-17
optimize: extend internalConcat dispatch for value-presented sources

He-Pin:
## Motivation

`FlowOps#internalConcat` previously had only one fast-path: a `SingleSource` on the right-hand side of `concat` / `concatLazy` was rerouted through the lightweight `SingleConcat` stage instead of the general two-port `Concat[U](2, detached)` fan-in graph (which materializes the whole substream plus a detacher buffer). All other value-presented sources (`IterableSource`, `IteratorSource`, `RangeSource`, `RepeatSource`, `JavaStreamSource`, `FutureSource`, `FailedSource`) still took the heavy `concatGraph` path even though their data is already in memory or trivially producible — the fan-in machinery and substream materialization were pure overhead. Heavy `concat` users (pekko-http and others) carry that cost on every materialization.

This mirrors the optimization shipped for `FlattenConcat` / `FlattenMerge` in #2977 (now merged), applied to the `concat` operator chain.

## Modification

Add specialized `GraphStage[FlowShape[E, E]]` siblings of `SingleConcat`, each passing through elements while upstream is alive and draining its captured value-presented payload on `onUpstreamFinish`:

| New stage | Handles | Drain behavior |
|-----------|---------|----------------|
| `IterableConcat[E](createIterator)` | `IterableSource`, `IteratorSource`, `RangeSource` | `emitMultiple(out, createIterator(), () => completeStage())` |
| `JavaStreamConcat[E](open)` | `JavaStreamSource` | Open `BaseStream`, drain its iterator, close in `postStop` (mirrors `InflightJavaStreamSource` resource hygiene from #2977) |
| `RepeatConcat[E](elem)` | `RepeatSource` | Swap `OutHandler` so each `onPull` pushes `elem` |
| `FailedConcat[E](failure)` | `FailedSource` | `failStage(failure)` |
| `FutureConcat[E](future)` | `FutureSource` | If completed: emit/fail. If pending: swap `OutHandler` to a no-op (avoid pulling closed `in`), register parasitic async callback that emits/fails when resolved. `Success(null)` calls `completeStage()` — lock-step with `Source.future(Future.successful(null))` rerouting to `Source.empty` and `InflightSources.hasFutureElement` treating `Success(null)` as no-element (no null is ever forwarded — Reactive Streams compliance). |

`internalConcat` is extended to dispatch via `TraversalBuilder.getValuePresentedSource` and pattern-match the eight value-presented source types (existing `SingleSource` path is preserved). The `detached` flag is irrelevant for these stages — the right-hand data is already present, so the one-element pre-fetch buffer that `detached=true` provides has nothing to fetch (matching `SingleConcat`'s precedent).

> Note: `Source.future(Future.successful(x))` is itself optimized upstream into `SingleSource`, so completed futures dispatch via `SingleConcat`. Same for `Source.future(Future.failed(ex))` → `FailedSource` → `FailedConcat`. Only genuinely-pending futures land on `FutureConcat`.

## Result

For the eight value-presented source types, `concat` and `concatLazy` no longer pay for substream materialization or the two-port fan-in graph. Observable behavior is unchanged for all other sources, which still take the existing `concatGraph` path. The Java-stream `BaseStream` is closed deterministically on exhaustion, downstream cancellation, and exceptions. Pending-future `Success(null)` is consistent with the materialized `FutureSource` path and the flatten inflight path.

## Tests

Directional tests added to `AbstractFlowConcatSpec` (cover both `concat` and `concatLazy` paths):

- `optimize iterable concat` / `range concat` / `iterator concat` / `java-stream concat` — assert correct stage in the traversal builder, output correct.
- `optimize repeat concat` — assert `RepeatConcat(0)`, bounded with `.take(6)`.
- `optimize failed concat` — assert `FailedConcat`, error propagated.
- `optimize completed-future concat` — confirms it routes through `SingleConcat` (upstream optimization).
- `optimize pending-future concat` — uses an unresolved `Promise`, asserts `FutureConcat` and correct delivery after resolution.
- `treat pending-future resolved with null as completion in concat` — `Promise.success(null)` ⇒ stream completes without emitting null.
- `fail the stream when pending-future resolves with failure in concat` — `Promise.failure(ex)` ⇒ stream fails with `ex`.
- `optimize failed-future concat` — confirms it routes through `FailedConcat` (upstream optimization).
- `close the underlying java-stream after java-stream concat completes` — `BaseStream.close()` runs on stage exhaustion (`postStop`).
- `close the underlying java-stream when downstream cancels mid java-stream concat` — `take(2)` mid-stream still triggers `BaseStream.close()`.
- `avoid downstream substream materialization for value-presented sources` — counter-based assertion of zero substream materialization.

Verification commands:

```
sbt "stream-tests/testOnly *FlowConcat*Spec"               # 90/90 pass
sbt "stream-tests/testOnly *FlattenMerge* *FlatMapConcat*" # 70/70 pass (regression)
sbt "stream/mimaReportBinaryIssues"                        # MiMa clean
sbt scalafmtCheckAll headerCheckAll                        # clean
```

All new stages are `private[pekko]` / `@InternalApi` so MiMa stays green.

## References

- Builds on #2977 (value-presented optimization for `FlattenConcat` / `FlattenMerge`, merged).

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowConcatSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowConcatSpec.scala
@@ -515,7 +515,7 @@ abstract class AbstractFlowConcatSpec extends BaseTwoStreamsSetup {
       val s2 = Source(List(2, 3))
       val eagerStr = s1.concat(s2).traversalBuilder.pendingBuilder.toString
       val lazyStr = s1.concatLazy(s2).traversalBuilder.pendingBuilder.toString
-      eagerStr should not include "IterableConcat"
+      (eagerStr should not).include("IterableConcat")
       lazyStr should include("IterableConcat")
     }
 
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowConcatSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowConcatSpec.scala
@@ -13,14 +13,21 @@
 
 package org.apache.pekko.stream.scaladsl
 
+import java.util.Collections
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
 
 import scala.concurrent.Await
+import scala.concurrent.Future
 import scala.concurrent.Promise
 import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
 
 import org.apache.pekko
 import pekko.NotUsed
+import pekko.stream.ActorAttributes.supervisionStrategy
+import pekko.stream.KillSwitches
+import pekko.stream.Supervision.{ restartingDecider, resumingDecider }
 import pekko.stream.testkit.BaseTwoStreamsSetup
 import pekko.stream.testkit.TestPublisher
 import pekko.stream.testkit.TestSubscriber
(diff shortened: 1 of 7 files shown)


He-Pin:
Hope this can improve some performance of pekko-http

He-Pin:
Follow-up on Comment 1 (`getValuePresentedSource` drops attributes): force-pushed `dd8e3c6e2a` with a proper fix.

**What changed**

`Flow.internalConcat` now propagates the inlined source's wrapping attributes (`other.traversalBuilder.attributes`) onto the new optimized stage via `addAttributes(...)`, so anything the user attached via `.withAttributes(...)` on the inlined source — `SupervisionStrategy`, dispatcher hint, log level, stage name — survives the fast path.

The only stage where this is observably load-bearing is `IterableConcat` (covers `IterableSource` / `IteratorSource` / `RangeSource`). Its `createLogic` now reads `inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider` and mirrors the exact `IteratorSource` Resume / Restart / Stop semantics:

- **Stop** (default decider): `failStage(ex)`
- **Resume**: skip the throwing element, continue iteration via `tryPushNextOrComplete()`
- **Restart**: re-invoke the iterator factory (`createIterator()`) and resume emission with the fresh iterator

`JavaStreamSource` itself does not honour supervision (it propagates iterator throws via the framework default), so `JavaStreamConcat` matches that behaviour and does not need decider logic. The other optimized stages (`SingleConcat`, `RepeatConcat`, `FailedConcat`, `FutureConcat`) cannot throw mid-emission, so supervision is moot for them — but they still receive the source attributes for cosmetic / dispatcher consistency.

**Directional tests added** in `FlowConcatSpec` (run on both `eager` and `concatLazy` matrices):

- `Source.fromIterator(...).withAttributes(supervisionStrategy(resumingDecider))` — Resume skips the throwing element; expect `Seq(0, 1, 2, 4, 5)` for an iterator that throws on `3`.
- `Source.fromIterator(factory).withAttributes(supervisionStrategy(restartingDecider))` — Restart re-invokes the factory; expect emission to continue with the fresh iterator.
- Default decider (Stop) — failure propagates as before.

Each test asserts dispatch via `traversalBuilder.pendingBuilder.toString should include("IterableConcat")` to guarantee the optimized path is exercised.

**Note on #2977**

The same gap (attributes dropped during `getValuePresentedSource` unwrap) exists in `FlattenConcat` / `FlattenMerge` paths in the already-merged #2977. Happy to open a follow-up that applies the same attribute-propagation + decider-aware handling to `InflightIteratorSource` if maintainers want it tracked separately — kept out of this PR to minimise scope.

All FlowConcat suites (85 tests) pass, FlattenMerge / FlatMapConcat regression tests pass, MiMa is clean.

He-Pin:
@pjfanning I think this one should be ready now

pjfanning (approved):
lgtm

PR #2985, opened 2026-05-21
test: fix InterpreterBenchmark so it produces trustworthy numbers

He-Pin:
### Motivation
`InterpreterBenchmark` had two independent bugs that made its results unreliable, which becomes a problem the moment anyone wants to evaluate a `GraphInterpreter`-touching change against it.

1. The benchmark body was `new GraphInterpreterSpecKit { new TestSetup { ... } }`. Because that ran inside `@Benchmark`, every invocation built (and never tore down) a fresh `ActorSystem`. Long iterations exhausted native threads and JMH ended up reporting empty results once the JVM ran out of resources.
2. Inside the assembly we used `GraphStages.identity[Int]` once per slot, but `GraphStages.identity` is a singleton whose `Inlet`/`Outlet` shape is shared across every reference. Chaining N copies (`numberOfIds = 5/10`) collapses to a single shape and mis-wires the connections; the run logged a flood of `Cannot pull port twice` errors and ended up reporting nonsense throughput (5/10-stage configs faster than the 1-stage one).

### Modification
- Make `InterpreterBenchmark` itself extend `GraphInterpreterSpecKit` so JMH's `@State(Scope.Benchmark)` lifecycle reuses one `ActorSystem` across invocations, and add `@TearDown(Level.Trial)` to terminate it cleanly.
- Define a local `IdentityStage[T]` with its own `Inlet`/`Outlet` per instance and use `Vector.fill(numberOfIds)(new IdentityStage[Int])` so each slot in the chain is a distinct stage with a distinct shape.

No changes to production code — this PR only fixes the benchmark.

### Result
The benchmark now runs to completion without leaking actor systems and produces stable, monotonic numbers (throughput decreases as `numberOfIds` grows, as expected). This restores it as a usable baseline for subsequent `GraphInterpreter` work.

JMH on this branch (JDK 25, G1, single thread, `-i 5 -wi 3 -f 1 -t 1`):

```
Benchmark                                             (numberOfIds)   Mode  Cnt      Score      Error   Units
InterpreterBenchmark.graph_interpreter_100k_elements              1  thrpt    5  45238.227 ± 3143.242  ops/ms
InterpreterBenchmark.graph_interpreter_100k_elements              5  thrpt    5  10526.376 ±  151.239  ops/ms
InterpreterBenchmark.graph_interpreter_100k_elements             10  thrpt    5   5350.558 ±  192.965  ops/ms
```

Pre-fix the 5/10-stage rows were both higher than the 1-stage row (i.e. wrong direction) because the singleton-shape bug meant the chain wasn't actually N stages long.

This is a benchmark-correctness fix, not a performance improvement. There is no production-code change here.

### Tests
- `sbt 'bench-jmh/compile'`
- `sbt 'bench-jmh/headerCheck; bench-jmh/scalafmtCheck'`
- `sbt 'bench-jmh/Jmh/run -i 5 -wi 3 -f 1 -t 1 -rf json -rff /tmp/jmh.json .*InterpreterBenchmark.*'` — completes cleanly, scores above.

### References
None - benchmark-only fix surfaced while preparing to evaluate `GraphInterpreter` micro-optimizations against a trustworthy baseline.

### DIFF
--- a/bench-jmh/src/main/scala/org/apache/pekko/stream/InterpreterBenchmark.scala
+++ b/bench-jmh/src/main/scala/org/apache/pekko/stream/InterpreterBenchmark.scala
@@ -15,10 +15,12 @@ package org.apache.pekko.stream
 
 import java.util.concurrent.TimeUnit
 
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
 import org.openjdk.jmh.annotations._
 
 import org.apache.pekko
-import pekko.event._
 import pekko.stream.impl.fusing.GraphInterpreter.{ DownstreamBoundaryStageLogic, UpstreamBoundaryStageLogic }
 import pekko.stream.impl.fusing.GraphInterpreterSpecKit
 import pekko.stream.impl.fusing.GraphStages
@@ -27,7 +29,7 @@ import pekko.stream.stage._
 @State(Scope.Benchmark)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
 @BenchmarkMode(Array(Mode.Throughput))
-class InterpreterBenchmark {
+class InterpreterBenchmark extends GraphInterpreterSpecKit {
   import InterpreterBenchmark._
 
   // manual, and not via @Param, because we want @OperationsPerInvocation on our tests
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #2998, opened 2026-05-27
test: relax SendQueue fuzzing deadlines

He-Pin:
### Motivation
SendQueueSpec preserves high-volume wakeup coverage but its default per-expect timeout is too tight under fuzzing-mode and virtualized nightly/CI runners. Recent failures timed out while waiting for a burst or the next element even though the stress workload itself should remain unchanged.

### Modification
Keep the existing burst size, 100000 rounds, and first-message ordering variants, but use explicit dilated deadlines for the slow fuzzing assertions.

### Result
The test still exercises the same SendQueue wakeup scenarios while allowing scheduler jitter on constrained JDK 25 and Scala 3 CI runners. This does not reduce the test workload or weaken message-order assertions.

### Tests
- JDK 25 nightly-style: `sbt "remote / Test / testOnly org.apache.pekko.remote.artery.SendQueueSpec"` - 5 tests succeeded / 0 failed
- JDK 25 nightly-style Scala 3.3.7: `sbt "++ 3.3.7" "remote / Test / testOnly org.apache.pekko.remote.artery.SendQueueSpec"` - 5 tests succeeded / 0 failed
- `scalafmt --mode diff-ref=origin/main`
- `git diff --check`

### References
None - nightly-builds.yml and PR Scala 3 SendQueueSpec timeout

### DIFF
--- a/remote/src/test/scala/org/apache/pekko/remote/artery/SendQueueSpec.scala
+++ b/remote/src/test/scala/org/apache/pekko/remote/artery/SendQueueSpec.scala
@@ -28,6 +28,7 @@ import pekko.stream.testkit.TestSubscriber
 import pekko.stream.testkit.scaladsl.TestSink
 import pekko.testkit.ImplicitSender
 import pekko.testkit.PekkoSpec
+import pekko.testkit.TestDuration
 
 object SendQueueSpec {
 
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #2994, opened 2026-05-25
test: stabilize nightly stream and TLS specs

He-Pin:

Issue #2904, opened 2026-04-25
failed: InputStreamSourceTest


He-Pin:
[WARN] [04/24/2026 18:01:22.170] [InputStreamSourceTest-pekko.actor.internal-dispatcher-19] [CoordinatedShutdown(pekko://InputStreamSourceTest)] Coordinated shutdown phase [actor-system-terminate] timed out after 10000 milliseconds
[04-24 18:01:52.155] [info] org.apache.pekko.stream.tck.InputStreamSourceTest *** ABORTED ***
[04-24 18:01:52.157] [info]   java.lang.RuntimeException: Failed to stop [InputStreamSourceTest] within [40000 milliseconds] 
[04-24 18:01:52.157] [info] -> / LocalActorRefProvider$$anon$3 class org.apache.pekko.actor.LocalActorRefProvider$Guardian status=0 1 children
[04-24 18:01:52.157] [info]    ⌊-> system LocalActorRef class org.apache.pekko.actor.LocalActorRefProvider$SystemGuardian status=4 Terminating(Termination)
[04-24 18:01:52.157] [info]        |    toDie: Actor[pekko://InputStreamSourceTest/system/Materializers#1264249758]
[04-24 18:01:52.157] [info]        ⌊-> Materializers RepointableActorRef class org.apache.pekko.stream.impl.MaterializerGuardian status=4 Terminating(Termination)
[04-24 18:01:52.157] [info]            |    toDie: Actor[pekko://InputStreamSourceTest/system/Materializers/StreamSupervisor-0#1254292714]
[04-24 18:01:52.157] [info]            ⌊-> StreamSupervisor-0 LocalActorRef class org.apache.pekko.stream.impl.StreamSupervisor status=4 Terminating(Termination)
[04-24 18:01:52.157] [info]                |    toDie: Actor[pekko://InputStreamSourceTest/system/Materializers/StreamSupervisor-0/flow-69-0-take#-951041886]
[04-24 18:01:52.157] [info]                ⌊-> flow-69-0-take RepointableActorRef class org.apache.pekko.stream.impl.fusing.ActorGraphInterpreter status=2 no children
[04-24 18:01:52.158] [info]   at org.apache.pekko.stream.tck.ActorSystemLifecycle.shutdownActorSystem(ActorSystemLifecycle.scala:61)
[...]

PR #2999, opened 2026-05-28
test: make InputStreamSource TCK publisher finite

He-Pin:
### Motivation
JDK 25 nightly builds abort InputStreamSourceTest when the Reactive Streams TCK cancellation scenario waits for completion from an infinite, CPU-busy InputStream publisher.

### Modification
Make the test publisher finite and emit one byte per ByteString so the TCK element count maps directly to stream elements without relying on take(elements) to stop a busy reader.

### Result
The TCK publisher completes deterministically under JDK 25 nightly-style virtualized stream-dispatcher settings.

### Tests
- JDK 25 nightly-style virtualized stream-dispatcher flags: stream-tests-tck / Test / testOnly org.apache.pekko.stream.tck.InputStreamSourceTest
- git diff --check

### References
Refs #2994


pjfanning:
related to #2904 ?

He-Pin:
I extracted it from there, we can check this later,as we recently fix the bug with varhandle

pjfanning (approved):
lgtm

PR #3000, opened 2026-05-28
test: clean up TLS edge streams after early cancellation

He-Pin:
### Motivation
JDK 25 nightly builds time out in repeated TlsGraphStageEdgeCasesSpec early-cancellation scenarios because earlier materializations can keep draining after the expected bytes have been collected.

### Modification
Materialize collectExactly with a KillSwitch and watch stream termination, then shut down and await the stream in finally after the expected bytes are collected.

### Result
Repeated TLS edge-case checks do not leave prior materializations running in the same actor system.

### Tests
- JDK 25 nightly-style virtualized stream-dispatcher flags: stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageEdgeCasesSpec
- git diff --check

### References
Refs #2994


pjfanning (approved):
lgtm

PR #3001, opened 2026-05-28
test: ignore late identities while awaiting rotating-keys ping

He-Pin:
### Motivation
RotatingKeysSSLEngineProviderSpec can receive a delayed ActorIdentity from an earlier Identify attempt after the target actor ref has already been resolved. Nightly retry policy still fails builds when that stale identity arrives before the ping reply.

### Modification
Wait for the expected ping with fishForMessage and ignore late ActorIdentity messages while keeping the same per-attempt timeout budget.

### Result
The test accepts the intended ping reply without being failed by harmless delayed Identify responses.

### Tests
- JDK 21: remote / Test / testOnly org.apache.pekko.remote.artery.tcp.ssl.RotatingProviderWithChangingKeysSpec
- JDK 25: remote / Test / testOnly org.apache.pekko.remote.artery.tcp.ssl.RotatingProviderWithChangingKeysSpec (fails on existing JDK 25 EKU certificate validation behavior)
- git diff --check

### References
Refs #2994


pjfanning (approved):
lgtm

PR #3002, opened 2026-05-28
test: select the active FileSource actor in dispatcher checks

He-Pin:
### Motivation
FileSourceSpec dispatcher checks can select a stale fileSource child left briefly by a previous materialization, which makes the dispatcher override assertion flaky.

### Modification
Capture existing fileSource children before starting each dispatcher-check stream, wait for exactly the newly materialized child, and watch it terminate after cancellation.

### Result
The checks assert the dispatcher of the stream started by the current test instead of a leftover child from an earlier materialization.

### Tests
- JDK 17 Scala 3.3.6: stream-tests / Test / testOnly org.apache.pekko.stream.io.FileSourceSpec
- git diff --check

### References
Refs #2994


pjfanning (approved):
lgtm

PR #3003, opened 2026-05-28
test: avoid pre-materialized substream in SplitWhen TCK

He-Pin:
### Motivation
SplitWhenTest can time out in the reactive-streams stochastic completion check on JDK 25 when the publisher is created from a pre-materialized tail substream.

### Modification
Create the TCK publisher directly from splitWhen followed by concatSubstreams, avoiding the intermediate Sink.head materialization and cancellation of the outer stream.

### Result
The publisher exercises splitWhen while keeping completion tied to the returned publisher stream.

### Tests
- JDK 25 Scala 3.3.6 with nightly-style virtualized dispatcher flags: stream-tests-tck / Test / testOnly org.apache.pekko.stream.tck.SplitWhenTest
- git diff --check

### References
Refs #2994

### DIFF
--- a/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/SplitWhenTest.scala
+++ b/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/SplitWhenTest.scala
@@ -13,11 +13,7 @@
 
 package org.apache.pekko.stream.tck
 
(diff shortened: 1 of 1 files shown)


pjfanning:
With the TCK, is there a harm with keeping the old version of the test along side the new one? More tests the better.

Issue #2573, opened 2025-12-12
nightly tests (CI): java 25 runs have a lot of stream test failures


pjfanning:
https://github.com/apache/pekko/actions/workflows/nightly-builds.yml

We have some flakiness with other Java versions but we see far more test failures with Java 25 on average. Different stream tests fail every night. Sometimes dozens, sometimes just a few.

pjfanning:
Recent changes don't seem to have helped much. The Java 25 runs in the nightly plan fail every night.
https://github.com/apache/pekko/actions/workflows/nightly-builds.yml

He-Pin:
Opened #2821 to stabilize a few remaining Java 25 failures. It increases patience in the slow stream suites and updates `RotatingKeysSSLEngineProviderSpec` to accept the stricter TLS handshake failure mode seen on newer JDKs.

pjfanning:
The Scala 2.13 run on Java 25 only had 1 test failure last night.
The Scala 3.3 run is still performing poorly with lots of test timeouts - especially in FlowFlatMapConcatParallelismSpec.
But there are lots of other failures too.
https://github.com/apache/pekko/actions/runs/23774704773/job/69273862700

He-Pin:
https://github.com/apache/pekko/actions/workflows/nightly-builds.yml

recently fixed and green again with two run.

PR #3004, opened 2026-05-29
test: scale TCK Await timeouts by pekko.test.timefactor

He-Pin:
### Motivation

`GroupByTest` and `PrefixAndTailTest` abort the TCK suite on JDK 25 nightly runs with:

```
java.lang.AssertionError: Failed in iteration 18 of 100.
Expected completion signal after signalling 10 elements (signalled 0),
yet did not receive it within 32000 ms
  at org.reactivestreams.tck.PublisherVerification.stochastic_spec103_mustSignalOnMethodsSequentially(...)
```

after roughly 18 iterations of `stochastic_spec103_mustSignalOnMethodsSequentially`.

Both `createPublisher` implementations materialize an intermediate stream and call `Await.result(future, 3.seconds)` to obtain the `Source` wrapped by the test `Publisher`. The `3.seconds` value is a hardcoded constant that does not honour `pekko.test.timefactor`, so the JDK 25 nightly's compounded GC + ForkJoinPool pressure can push a single materialization beyond 3s during the 100-iteration stochastic test, which surfaces as a `Publisher` that signals zero elements (and then aborts the whole TCK suite).

`Timeouts` already scales every other field by `timeFactor`. `publisherShutdownTimeoutMillis` is the only existing field that wasn't scaled.

### Modification

- Add `Timeouts.materializerTimeoutMillis`, scaled by `timeFactor`.
- Route the `Await.result` calls in `GroupByTest` and `PrefixAndTailTest` through the new helper instead of the hardcoded `3.seconds`.
- Apply the same `timeFactor` scaling to the existing `publisherShutdownTimeoutMillis` so the whole `Timeouts` object behaves consistently.

### Result

On local runs (`timeFactor=1`) the effective budget stays at 3s, so the test still flags genuine materialization regressions. On the JDK 25 nightly (`timeFactor=4`) the budget grows to 12s, which covers the observed worst-case materialization delays without masking real bugs.

### Tests

- `sbt stream-tests-tck/Test/compile` (Scala 2.13) succeeded locally.

### References

- Failing nightly: https://github.com/apache/pekko/actions/runs/26571377673
- Failing nightly: https://github.com/apache/pekko/actions/runs/26611472055
- Tracking issue: #2573

### DIFF
--- a/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/GroupByTest.scala
+++ b/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/GroupByTest.scala
@@ -30,7 +30,7 @@ class GroupByTest extends PekkoPublisherVerification[Int] {
     else {
       val futureGroupSource =
         Source(iterable(elements)).groupBy(1, _ => "all").prefixAndTail(0).map(_._2).concatSubstreams.runWith(Sink.head)
-      val groupSource = Await.result(futureGroupSource, 3.seconds)
+      val groupSource = Await.result(futureGroupSource, Timeouts.materializerTimeoutMillis.millis)
       groupSource.runWith(Sink.asPublisher(false))
 
     }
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

PR #3006, opened 2026-05-29
test: await async alsoTo/wireTap side sink in context specs

He-Pin:
### Motivation

`SourceWithContextSpec` and `FlowWithContextSpec` fail on JDK 25 nightly with a **real assertion error (not a timeout)**:

```
Vector(Message(A,1), Message(B,2), Message(D,3)) was not equal to
Vector(Message(A,1), Message(B,2), Message(D,3), Message(C,4))
(SourceWithContextSpec.scala:94)  *** FAILED *** (15 milliseconds)
```

`alsoTo` / `alsoToContext` / `wireTap` / `wireTapContext` feed an **asynchronous** side `Sink`. The tests collected the side stream output into a buffer and asserted on it immediately after the main stream's `expectComplete()`. There is no happens-before between the main stream completing and the side `Sink` draining its last element, so the assertion can run before the side `Sink` has observed every element. The enclosing `within(10.seconds) { ... }` only bounds how long the block may take — it does **not** retry, so the buffer was read exactly once. On JDK 17 the side `Sink` usually drained first by luck; JDK 25's different scheduling surfaces the race as a hard failure in ~15ms.

The shared `scala.collection.mutable.ListBuffer` was also written from the side-stream dispatcher thread and read from the test thread without synchronization — a second latent data race.

### Modification

- Collect side-stream output into a `java.util.concurrent.ConcurrentLinkedQueue` (thread-safe, preserves arrival order).
- Replace the single `within(10.seconds) { assert }` with `awaitAssert(assert, 10.seconds)`, which polls until the async side `Sink` has caught up (or the deadline elapses).

### Result

The assertions now wait for the asynchronous side `Sink` to finish instead of sampling it once, removing both the timing race and the cross-thread buffer race. The checks still fail fast if the side `Sink` genuinely drops data, since `awaitAssert` rethrows the last assertion error at the deadline.

### Tests

- `sbt "stream-tests/testOnly org.apache.pekko.stream.scaladsl.SourceWithContextSpec org.apache.pekko.stream.scaladsl.FlowWithContextSpec"` — 24 tests pass (Scala 2.13).

### References

- Failing nightly: https://github.com/apache/pekko/actions/runs/26571377673
- Tracking issue: #2573

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowWithContextSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowWithContextSpec.scala
@@ -13,8 +13,10 @@
 
 package org.apache.pekko.stream.scaladsl
 
-import scala.collection.mutable.ListBuffer
+import java.util.concurrent.ConcurrentLinkedQueue
+
 import scala.concurrent.duration._
+import scala.jdk.CollectionConverters._
 import scala.util.control.NoStackTrace
 
 import org.apache.pekko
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3023, opened 2026-05-31
feat: replace CountMinSketch with FastFrequencySketch in Artery compression

He-Pin:
### Motivation

The Artery compression system uses a `CountMinSketch` (128KB per connection, 16×1024×8 bytes) for heavy hitter detection in actor ref and class manifest compression tables. Pekko already has `FastFrequencySketch` in `pekko-actor` (used for cluster sharding entity passivation) which is:

- **~32x more memory efficient** (~4KB vs ~128KB per inbound connection)
- **Includes TinyLFU reset** — periodic halving of all counters provides natural aging, enabling the heavy hitters table to adapt to changing traffic patterns over time
- **Already battle-tested** — used in production cluster sharding passivation strategies

This addresses [akka/akka-core#31093](https://github.com/akka/akka-core/issues/31093).

### Modification

- **`InboundCompressions.scala`**: Replace `CountMinSketch` with `FastFrequencySketch[T]` in `InboundCompression`. The `increment` method now calls `frequencySketch.increment(value)` per occurrence and uses `frequencySketch.frequency(value)` as the estimated weight for heavy hitter detection.
- **`TopHeavyHitters.scala`**: 
  - `update()` now always allows weight updates for existing heavy hitters (previously only allowed increases). This is necessary because `FrequencySketch`'s periodic reset halves all counters.
  - `updateExistingHeavyHitter()` supports both weight increase (push down) and decrease (bubble up) in the heap.
  - New `fixHeapUp()` method for bidirectional heap property restoration.
  - `isHeavy()` check now only gates insertion of **new** entries.
- **`CountMinSketch.java`**: Removed — was `INTERNAL API` with no external users.
- **`CountMinSketchBenchmark.scala`**: Updated to benchmark `FastFrequencySketch` instead.
- **Legal files**: Removed CountMinSketch/stream-lib attribution (no longer bundled).

### Result

Artery compression uses ~4KB instead of ~128KB per inbound connection. Heavy hitters now adapt to changing traffic patterns via TinyLFU aging rather than tracking monotonically increasing absolute counts.

### Tests

- `remote / testOnly ...HeavyHittersSpec`: 13/13 passed (includes 2 new tests for weight decrease and heap upward restoration)
- `remote / testOnly ...CompressionTableSpec`: 3/3 passed
- `remote / testOnly ...OutboundCompressionSpec`: 2/2 passed

### References

Refs akka/akka-core#31093

### DIFF
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -1046,6 +1046,12 @@ pekko {
         # compression of common strings in remoting messages, like actor destinations, serializers etc
         compression {
 
+          # Frequency sketch implementation used for heavy hitter detection.
+          # Options:
+          #   - "count-min-sketch": Original CountMinSketch implementation (legacy, ~128KB per connection)
+          #   - "fast-frequency-sketch": FastFrequencySketch with TinyLFU aging (default, ~4KB per connection)
+          frequency-sketch-implementation = "fast-frequency-sketch"
+
           actor-refs {
             # Max number of compressed actor-refs
             # Note that compression tables are "rolling" (i.e. a new table replaces the old
--- a/remote/src/main/scala/org/apache/pekko/remote/artery/ArterySettings.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/artery/ArterySettings.scala
@@ -278,6 +278,12 @@ private[pekko] object ArterySettings {
 
     private[pekko] final val Enabled = ActorRefs.Enabled || Manifests.Enabled
 
+    val FrequencySketchImplementation: String = toRootLowerCase(getString("frequency-sketch-implementation")) match {
+      case "count-min-sketch"      => "count-min-sketch"
+      case "fast-frequency-sketch" => "fast-frequency-sketch"
+      case other                   => throw new IllegalArgumentException(s"Unknown frequency-sketch-implementation: $other")
+    }
+
     object ActorRefs {
       val config: Config = getConfig("actor-refs")
       import config._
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -423,6 +424,9 @@ lazy val remote =
     .settings(OSGi.remote)
     .settings(Protobuf.settings)
     .settings(Test / parallelExecution := false)
+    .settings(mimaBinaryIssueFilters ++= Seq(
+      // CountMinSketch was INTERNAL API, replaced by FastFrequencySketch
+      ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.remote.artery.compress.CountMinSketch")))
```
pjfanning: can you move this to its own excludes file?

pjfanning:
I'd prefer if the old code was kept and made configurable - as in let user choose the compression scheme

build.sbt
He-Pin: Yes, this should be done.

He-Pin: The MiMa excludes were already removed in commit 83cd9ae3a6 — CountMinSketch is now kept as a configurable legacy option, so no MiMa exclusion is needed.

remote/src/main/resources/reference.conf
```diff
@@ -1046,6 +1046,12 @@ pekko {
         # compression of common strings in remoting messages, like actor destinations, serializers etc
         compression {
 
+          # Frequency sketch implementation used for heavy hitter detection.
+          # Options:
+          #   - "count-min-sketch": Original CountMinSketch implementation (legacy, ~128KB per connection)
+          #   - "fast-frequency-sketch": FastFrequencySketch with TinyLFU aging (default, ~4KB per connection)
+          frequency-sketch-implementation = "fast-frequency-sketch"
```
pjfanning, line 1053: could we add an integration test that sets frequency-sketch-implementation = "count-min-sketch" ?

pjfanning (approved):
lgtm

PR #3024, opened 2026-05-31
feat: Add BidirectionalGracefulShutdown CancellationStrategy

### LINKED ISSUE #17997: Clarify stream closing semantics for bidirection components
As shown in #17992 a bidirectional components can easily provoke a race when the component is about to finish. When that happens completing the downstream side (regularly or with an error) is potentially racing against cancelling the upstream side.

Eventually, both must be done, but a better solution might be to complete the downstream side first, then wait for a grace period for the error to bubble through the counterpart in some form and only then cancel the upstream. The reasoning is that, especially in cases of an error, cancelling isn't able to transport any explanation about *why* the cancellation happened.

He-Pin:
### Motivation

Addresses issue #17997 - Clarify stream closing semantics for bidirectional components.

When a bidirectional component is about to finish, completing the downstream side (regularly or with an error) can race against cancelling the upstream side. If the cancellation arrives first, the error signal is lost because the connection is already cancelled.

### Modification

- Added new  strategy to 
- This strategy first completes all output ports, then waits for a grace period to allow the error to propagate through the counterpart, and finally cancels all input ports
- Added Java API  method
- Added comprehensive tests to verify the new behavior
- The strategy prevents race conditions in BidiFlow stacks where error propagation is important

### Result

- Bidirectional components can now gracefully shut down by completing outputs first
- Error signals have a chance to propagate through the counterpart before upstream cancellation
- Addresses the race condition described in akka/akka-core#17997

### Tests

- stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.CancellationStrategySpec: All 15 tests passed

### References

- Fixes akka/akka-core#17997

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
@@ -565,6 +565,44 @@ object Attributes {
      */
     @ApiMayChange
     def afterDelay(delay: java.time.Duration, strategy: Strategy): Strategy = AfterDelay(delay.toScala, strategy)
+
+    /**
+     * Strategy that ensures graceful shutdown for bidirectional components.
+     *
+     * When `cancelStage` is invoked, this strategy first completes all output ports (regularly or with an error),
+     * then waits for a grace period to allow the completion/error signal to propagate through the counterpart,
+     * and finally cancels all input ports.
+     *
+     * This addresses the race condition in bidirectional components where cancelling the upstream side might
+     * prevent the error from being properly propagated downstream. By completing outputs first and waiting,
+     * the error has a chance to bubble through the counterpart before the upstream is cancelled.
+     *
+     * This strategy is particularly useful in stacks of BidiFlows where different layers are connected
+     * through both inputs and outputs, and error propagation is important for proper diagnostics.
+     *
+     * @param delay the grace period to wait after completing outputs before cancelling inputs
+     */
+    @ApiMayChange
+    final case class BidirectionalGracefulShutdown(delay: FiniteDuration) extends Strategy
+
+    /**
+     * Java API
+     *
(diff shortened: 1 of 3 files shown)


stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
```diff
@@ -565,6 +565,44 @@ object Attributes {
...
+     *
+     * This addresses the race condition in bidirectional components where cancelling the upstream side might
+     * prevent the error from being properly propagated downstream. By completing outputs first and waiting,
+     * the error has a chance to bubble through the counterpart before the upstream is cancelled.
+     *
+     * This strategy is particularly useful in stacks of BidiFlows where different layers are connected
+     * through both inputs and outputs, and error propagation is important for proper diagnostics.
+     *
+     * @param delay the grace period to wait after completing outputs before cancelling inputs
+     */
```
pjfanning, line 584: `@since 2.0.0`

stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
```diff
@@ -629,6 +667,26 @@ object Attributes {
...
+   * When `cancelStage` is invoked, this strategy first completes all output ports (regularly or with an error),
+   * then waits for a grace period to allow the completion/error signal to propagate through the counterpart,
+   * and finally cancels all input ports.
+   *
+   * This addresses the race condition in bidirectional components where cancelling the upstream side might
+   * prevent the error from being properly propagated downstream. By completing outputs first and waiting,
+   * the error has a chance to bubble through the counterpart before the upstream is cancelled.
+   *
+   * @param delay the grace period to wait after completing outputs before cancelling inputs
+   */
```
pjfanning, line 684: `@since 2.0.0`

PR #3029, opened 2026-05-31
Fix sbt load warnings by excluding unused lint keys

He-Pin:
### Motivation
sbt load was showing 165 warnings about unused keys, which cluttered the output and made it harder to identify real issues.

### Modification
Added  to  to exclude the following keys from  checks:
-  (used for linking to API docs)
-  (used to add timestamps to log output)
-  (used to filter projects for unidoc generation)
-  (used for PR validation)
-  (used to disable doclint for tests)

### Result
sbt load now completes without any warnings about unused keys.

### Tests
- Verified sbt load completes without warnings

### References
None - housekeeping

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,6 +22,14 @@ ThisBuild / reproducibleBuildsCheckResolver := Resolver.ApacheMavenStagingRepo
 
 ThisBuild / pekkoCoreProject := true
 
+Global / excludeLintKeys ++= Set(
+  projectInfoVersion,
+  logManager,
+  unidocProjectFilter,
+  fork,
+  javacOptions
+)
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

He-Pin:
@pjfanning I will backport this to 1.7.x

pjfanning:
@He-Pin why do we need this in 1.x? It is useful in main branch but 1.x branches are not very active and used for occasional patches.
I backported the test fix because the CI tests were failing too often on 1.7 branch.

He-Pin:
Because I don't want to see the warning @pjfanning

PR #1234, opened 2024-03-24
fix evaluation in nightly-builds.yml

pjfanning:
bug in #1231

PR #3030, opened 2026-05-31
fix: release references to completed stage logics in GraphInterpreter

He-Pin:
### Motivation
When multiple stages are fused, a single alive stage keeps the GraphInterpreter alive and with it references to all already completed GraphStage logics, which may keep a significant amount of memory.

This issue was originally reported in akka/akka-core#23439.

### Modification
- Modified GraphInterpreter.finish() to release references to all stage logics after finalization
- Modified GraphInterpreter.toSnapshot() to handle null logics gracefully
- Added null checks in finish() to avoid NPE when logic is already null

### Result
GraphInterpreter no longer keeps references to completed stage logics, allowing them to be garbage collected even if the interpreter is still alive due to other fused stages.

### Tests
- Added directional test to verify logics are released after finish()
- Added test to verify logics are released when stages complete early

### References
- Fixes #1234 (GraphInterpreter should not keep references to shut-down logics)
- Related to akka/akka-core#23439

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
@@ -244,6 +244,28 @@ import pekko.stream.stage._
   // Marks whether a stage has been finalized (finalizeStage been called) or not
   private val finalizedMark = Array.fill(logics.length)(false)
 
+  // Number of stages whose preStart has been invoked. Zero-port stages are completed from construction, so
+  // finalization scans during init must not include later stages that have not been started yet.
+  private var initializedStages = 0
+
+  // Connection owners are released when their stages complete, but snapshots still need stable endpoint identities.
+  private val connectionInStageIds = connectionStageIds(isInput = true)
+  private val connectionOutStageIds = connectionStageIds(isInput = false)
+
+  private def connectionStageIds(isInput: Boolean): Array[Int] = {
+    val stageIds = new Array[Int](connections.length)
+    var i = 0
+    while (i < connections.length) {
+      val connection = connections(i)
+      if (connection ne null) {
+        val owner = if (isInput) connection.inOwner else connection.outOwner
+        stageIds(i) = if (owner ne null) owner.stageId else -1
+      } else stageIds(i) = -1
+      i += 1
+    }
+    stageIds
+  }
+
   private var _subFusingMaterializer: Materializer = _
   private lazy val defaultErrorReportingLogLevel = LogLevels.defaultErrorLevel(materializer.system)
 
@@ -330,6 +352,7 @@ import pekko.stream.stage._
           log.error(e, "Error during preStart in [{}]: {}", logic.toString, e.getMessage)
           logic.failStage(e)
       }
+      initializedStages = i + 1
       afterStageHasRun(logic)
       i += 1
     }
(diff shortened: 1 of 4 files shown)


pjfanning:
@He-Pin there are merge conflicts and also open comments from Copilot that need actions.

He-Pin:
oops, will rebase and fix soon. the origin problem was reported by @jrudolph

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -342,10 +342,13 @@ import pekko.stream.stage._
...
     while (i < logics.length) {
       val logic = logics(i)
-      if (!isStageCompleted(logic) && !isStageFinalized(logic)) {
+      if ((logic ne null) && !isStageCompleted(logic) && !isStageFinalized(logic)) {
         markStageFinalized(logic)
         finalizeStage(logic)
       }
+      // Release reference to the stage logic so it can be garbage collected
+      // even if the GraphInterpreter is still alive due to other references
+      logics(i) = null
```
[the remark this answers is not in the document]
pjfanning: @He-Pin can you comment on why this should be treated as invalid or what changes that you have made to fix this?

He-Pin:
@pjfanning I have address all the problems

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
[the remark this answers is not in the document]
pjfanning: @He-Pin ping - for me, I'd like to see a response on this before approving this PR

He-Pin:
@pjfanning wdyt

stream/src/main/scala/org/apache/pekko/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -13,6 +13,7 @@
 
 package org.apache.pekko.stream.impl.fusing
 
+import java.util.BitSet
```
pjfanning: java.util.BitSet can have absolutely terrible performance

java.util.BitSet has a few well-known performance characteristics that bite people in specific scenarios:
Where it struggles

Sparse bit patterns. BitSet allocates a long[] word array sized to the highest set bit. If you set bit 10,000,000 while only having a handful of bits set overall, you still pay for ~1.25MB of backing array. Iteration (nextSetBit) then has to skip over huge runs of zero words.
Growth/resizing. Setting a bit beyond the current capacity triggers array reallocation and copy, similar to ArrayList — can be costly if you don't pre-size it via the constructor.
Not thread-safe, so concurrent access requires external synchronization, which people sometimes don't realize until it's a bottleneck.
Boxing/indirection overhead relative to raw primitive array manipulation, if you're doing very hot-loop bit twiddling — though this is usually minor compared to the sparse-layout issue.

Common alternatives with sparse layouts

RoaringBitmap — the most widely used one in the JVM ecosystem. It partitions the bit space into 2^16-sized "chunks" and picks the best internal representation per chunk (array of shorts for sparse, bitmap for dense, run-length encoding for runs). Excellent for both sparse and clustered-dense data, very fast unions/intersections/differences, and has great real-world adoption (Lucene, Spark, Druid, Elasticsearch all use it or a variant).
JavaEWAH (Compressed bitmap) — word-aligned hybrid run-length encoding, good when you have long runs of zeros/ones, common in bitmap-index use cases (analytics/OLAP).
Sparse hash-set-of-longs approaches (e.g. using a LongOpenHashSet from fastutil, or a custom compressed trie) if you genuinely just have a sparse handful of bits and don't need bitmap set-algebra.
fastutil / Eclipse Collections primitive collections — not bitmaps per se, but relevant if BitSet was being (mis)used just as a compact int-set.
Roaring64Bitmap / Roaring64NavigableMap — for when your bit indices exceed 32 bits.

Rule of thumb

Dense, bounded-range bits (e.g., flags over a known small universe) → BitSet is fine and often the fastest option due to JIT-friendly word operations.
Sparse or very large/unbounded index space, or you need fast boolean set-algebra (AND/OR/XOR/ANDNOT) at scale → RoaringBitmap is the standard choice today.
Sequential/long-run bitmaps (columnar/analytics indexes) → EWAH-style compression tends to win.

He-Pin: Maybe we can just use a plain loop as it before?

pjfanning: I'm struggling to review this PR. I much prefer when coders commit review changes separately and don't keep squashing the code. We squash it when we merge anyway. We lose all the history of the review when the the coder forces a new squashed commit. It is sometimes useful when there are lots of commits to squash some or all of the commits but generally, it is easier for reviewers if this does not happen.

So in short, I don't know what this PR looked like before the BitSet was introduced.

He-Pin: fixed now. maybe we can defer this to later realse.

pjfanning (approved):
lgtm

PR #3031, opened 2026-05-31
fix: release reference to initial interpreter shell after it is shut down

He-Pin:
### Motivation
An ActorGraphInterpreter starts with an initial shell but can host additional ones created using the subfusing materializer. The lifetime of ActorGraphInterpreter is not tied to the lifetime of the initial interpreter. However, a reference to the initial GraphInterpreterShell is kept around in the _initial constructor parameter, which can lead to unexpected memory usage because the initial GraphInterpreter and all its logics might be kept referenced much longer than expected.

This issue was originally reported in akka/akka-core#23535.

### Modification
- Changed _initial from val to private var
- Set _initial to null after initialization in preStart()
- Fixed debug logging to use shell parameter instead of _initial

### Result
ActorGraphInterpreter no longer keeps a reference to the initial interpreter shell after it has been shut down, allowing it and its logics to be garbage collected even when the actor is still alive hosting other subfused interpreters.

### Tests
- Added test to verify initial shell reference is released after completion
- Added test to verify actor can host additional interpreters after initial shell completes

### References
- Fixes #1235 (ActorGraphInterpreter keeps reference to initial shell)
- Related to akka/akka-core#23535

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/impl/fusing/ActorGraphInterpreterSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/impl/fusing/ActorGraphInterpreterSpec.scala
@@ -436,5 +436,20 @@ class ActorGraphInterpreterSpec extends StreamSpec {
       done.future.futureValue // would throw on failure
     }
 
+    "continue working with subfused interpreters after initial shell reference is released" in {
+      val mat = Materializer(system)
+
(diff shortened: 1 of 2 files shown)


pjfanning:
@He-Pin there are open comments from Copilot that need actions.

pjfanning (approved):
lgtm

PR #3037, opened 2026-06-02
fix: support setting the ForkJoinPool minimum-runnable value (#2890)

pjfanning:
cherry pick 980d2bd844fc37d9c75b5ed0f53c2203c67abfcb #2890 

Motivation:
JDK-8300995 / JDK-8321335 changed compensation-thread creation in ForkJoinPool asyncMode (FIFO) to be much more conservative. Pekko fork-join dispatchers using the default `minimum-runnable = 1` are then prone to starvation under blocking workloads on JDK 21+, which has shown up as flaky nightly runs (#2870) and is the root cause behind the workflow override added in #2889.

Modification:
* Introduce `ForkJoinExecutorConfigurator.resolveMinimumRunnable`, an internal helper that computes the effective `minimum-runnable` value from the configured value, the dispatcher parallelism, and the running JDK major version. A negative configured value (the new default `-1`) triggers the JDK-aware policy: on JDK 21+ the value becomes `min(8, max(1, parallelism / 2))`; on JDK < 21 it stays at `1`. Non-negative values are honoured verbatim, so explicit `0` still disables compensation entirely and explicit positive values (including `1`) keep their existing meaning.
* Add `pekko.actor.default-dispatcher.fork-join-executor.minimum-runnable` in `reference.conf` but keep to the default at `1` to match the existing Pekko 1.x behaviour.
* The default value in Pekko 2.x is `-1` to enable auto-tune behaviour
* Add `ForkJoinExecutorConfiguratorSpec` with three groups of assertions: (1) pure-function matrix on `resolveMinimumRunnable`; (2) directional checks asserting the auto policy strictly raises the value on JDK 21+ and never exceeds the documented cap of 8; (3) wiring integration that builds a `ForkJoinExecutorServiceFactory` from a real dispatcher config and verifies the resolved value reaches the factory (guarding against regressions of the resolver wiring).

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfiguratorSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/dispatch/ForkJoinExecutorConfiguratorSpec.scala
@@ -1,0 +1,177 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


He-Pin (approved):

pjfanning:
I created this for discussion but I am far from convinced that this is
* safe for a 1.x release
* there is no major plan for a 1.7.0 release and a big backlog of other releases that we need to do in other pekko modules
* I'd really prefer to concentrate on proven bugs and getting 2.0.0 finished

He-Pin:
Actually, a user can configure it with a custom fork-join executor too.

pjfanning:
@He-Pin what do you think about the option of changing the default on the 1.x targeted PR to 1 so that the default is to work like Pekko 1.6 and below. Users can opt into enabling auto-tune by setting the config to -1 or they also override the 1 setting with an explicit value.

He-Pin:
default 1 is ok, as the user can change it then.

PR #3059, opened 2026-06-13
Add -Yfuture-lazy-vals for Scala 3.3.x builds

He-Pin:
### Motivation

The Scala 3.3.8 compiler introduced the `-Yfuture-lazy-vals` flag ([scala/scala3-lts#637](https://github.com/scala/scala3-lts/pull/637)) to optionally use `VarHandle` instead of `sun.misc.Unsafe` for lazy val implementation. This prepares libraries for upcoming JDK releases that restrict `sun.misc.Unsafe` access.

### Modification

- Add `-Yfuture-lazy-vals` to `DefaultScalacOptions` in `project/PekkoBuild.scala`, conditionally applied only for Scala 3.3.x via `CrossVersion.partialVersion`
- Add MiMa filter in `remote/src/main/mima-filters/2.0.x.backwards.excludes/future-lazy-vals.excludes` to exclude the expected `<clinit>` bytecode change for `RemoteWatcher#Stats` (the new VarHandle-based lazy val implementation no longer generates the static initializer that the old Unsafe-based implementation produced)

### Result

Pekko's Scala 3.3.x build uses the future-proof `VarHandle`-based lazy val implementation, compatible with all JDK 9+ versions including future releases that restrict `sun.misc.Unsafe`.

### Tests

- `sbt "++3.3.8; actor/compile"` — passed (exit 0, only pre-existing warnings)
- `sbt "++3.3.8; remote/mimaReportBinaryIssues"` — passed (exit 0)

### References

Refs scala/scala3-lts#637


project/PekkoBuild.scala
```diff
@@ -97,7 +97,11 @@ object PekkoBuild {
         "-feature",
         "-unchecked",
         // 'blessed' since 2.13.1
-        "-language:higherKinds")
+        "-language:higherKinds") ++
+      (CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, 3)) => Seq("-Yfuture-lazy-vals")
```
pjfanning: we don't need this to be so complicated - just add it the Seq starting on line 94

He-Pin: Yes! will do it.

pjfanning (approved):
lgtm

PR #3060, opened 2026-06-13
Fix Scala 3.8.4 compilation compatibility while preserving 2.13 and 3.3.x cross-build

He-Pin:
### Motivation
The codebase needs to compile with both Scala 3.3.x (current production) and Scala 3.8.4 (forward compatibility for the upcoming 3.9 LTS). Several patterns in the codebase break under Scala 3.8 due to changes in how context bounds are desugared (they become `using` clauses), stricter deprecation warnings, and syntax changes.

### Modification
171 files changed across the codebase:

- **ClassTag context bound fixes**: In Scala 3.8, `[T: ClassTag]` desugars to a `using` clause, making explicit `ClassTag` passing positional-incompatible. Fixed by using `implicit val ct: ClassTag[T] = ClassTag(clazz)` in scope and relying on implicit resolution instead of explicit passing. Key files: `PersistencePlugin.scala`, `EventSourcedBehaviorTestKit.scala`, `ActorContextSpec.scala`, `BehaviorTestKitSpec.scala`, `Behavior.scala`, `EventStream.scala`, `Topic.scala`, `ShardingProducerController.scala`, `ShardedDaemonProcessImpl.scala`, `StreamTestKit.scala`.

- **PersistencePlugin implicit parameter merge**: Merged `[T: ClassTag]` context bound and `(implicit ev: PluginProvider[...])` into a single `implicit` parameter list `(implicit ct: ClassTag[T], ev: PluginProvider[T, ScalaDsl, JavaDsl])` to avoid context bound + separate implicit parameter mismatch in Scala 3.8.

- **`.apply()` trick**: Used `decoration[T].apply(behavior)` instead of `decoration[T](behavior)` to separate implicit resolution from function application (Scala 2.13 interprets `(behavior)` as the implicit ClassTag argument).

- **`private[this]` → `private`**: Scala 3.8 deprecates `private[this]` in favor of `private`.

- **`= _` initializers → explicit defaults**: Replaced `= _` with `= null`, `= None`, `= 0`, etc. as Scala 3.8 deprecates uninitialized vars.

- **`SortedSet.empty(ordering)` fix**: Bound ordering as `implicit val ord` before calling `.empty[Member]` to avoid implicit resolution issues.

- **`@unchecked` annotations**: Added for type-test patterns that can't be checked at runtime under Scala 3.

- **`@nowarn` filters**: Used message-based `@nowarn("msg=never used")` instead of category-based filters for cross-version compatibility.

- **`PekkoDisciplinePlugin.scala`**: Added `-Wconf` suppression rules for Scala 3.8 deprecation/syntax warnings in both main and test scopes.

- **No `using` keyword in shared sources**: All changes use cross-compatible patterns (`implicit val`, `.apply()` trick). The `using` keyword only appears in `scala-3/` directories.

### Result
All modules compile with both Scala 2.13.18 and Scala 3.8.4. Production Scala 3 version remains at 3.3.8 — this change only ensures forward compatibility. No behavioral changes to the runtime.

### Tests
- `sbt "++3.8.4" "Test / compile"` — success across all modules
- `sbt "Test / compile"` (Scala 2.13.18) — success across all modules

### References
None - proactive Scala 3.8 forward compatibility

### DIFF
--- a/actor-testkit-typed/src/test/scala/org/apache/pekko/actor/testkit/typed/scaladsl/BehaviorTestKitSpec.scala
+++ b/actor-testkit-typed/src/test/scala/org/apache/pekko/actor/testkit/typed/scaladsl/BehaviorTestKitSpec.scala
@@ -126,7 +126,8 @@ object BehaviorTestKitSpec {
               replyTo ! Done
               Behaviors.same
             case CreateMessageAdapter(messageClass, f, replyTo) =>
-              val adaptor = context.messageAdapter(f)(ClassTag(messageClass))
+              implicit val ct: ClassTag[Any] = ClassTag(messageClass)
+              val adaptor = context.messageAdapter[Any](f.asInstanceOf[Any => Command])
               replyTo.foreach(_ ! adaptor.unsafeUpcast)
               Behaviors.same
             case Log(what) =>
--- a/actor-tests/src/test/scala/org/apache/pekko/io/dns/DockerBindDnsService.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/io/dns/DockerBindDnsService.scala
@@ -19,6 +19,7 @@ import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
 import scala.util.Try
 import scala.util.control.NonFatal
+
 import com.github.dockerjava.api.DockerClient
 import com.github.dockerjava.api.async.ResultCallback
 import com.github.dockerjava.api.command.CreateContainerCmd
@@ -29,10 +30,11 @@ import com.github.dockerjava.httpclient5.ApacheDockerHttpClient
 import org.apache.pekko
 import pekko.actor.Props
 import pekko.io.dns.internal.DnsClient
+import pekko.testkit.PekkoSpec
 import pekko.util.Timeout
 
-import pekko.testkit.PekkoSpec
 import org.scalatest.concurrent.Eventually
+
 import com.typesafe.config.Config
 
 abstract class DockerBindDnsService(config: Config) extends PekkoSpec(config) with Eventually {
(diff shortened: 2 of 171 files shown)


actor-testkit-typed/src/test/scala/org/apache/pekko/actor/testkit/typed/scaladsl/BehaviorTestKitSpec.scala
```diff
@@ -126,7 +126,8 @@ object BehaviorTestKitSpec {
               replyTo ! Done
               Behaviors.same
             case CreateMessageAdapter(messageClass, f, replyTo) =>
-              val adaptor = context.messageAdapter(f)(ClassTag(messageClass))
+              implicit val ct: ClassTag[Any] = ClassTag(messageClass)
+              val adaptor = context.messageAdapter[Any](f.asInstanceOf[Any => Command])
```
pjfanning, line 130: it's ok to make this change but the casting seems weird and could be a bug in scala 3.8

He-Pin:
The 3.9.0-RC1 has just been released, hope this will make the code compiles with 3.9.0-RC1

actor-testkit-typed/src/test/scala/org/apache/pekko/actor/testkit/typed/scaladsl/BehaviorTestKitSpec.scala
He-Pin: Agreed, the `asInstanceOf` cast was a workaround for Scala 3.8 context bound desugaring. It has been reverted to the original `context.messageAdapter(f)(ClassTag(messageClass))` pattern in commit 683f8740c3 — the explicit ClassTag passing still works correctly in Scala 3.8 since `using` parameters can be passed positionally. The refactored `ClassTag[U]` version did not compile on Scala 2.13 because `U` is erased in pattern matching.

He-Pin:
Closing to reopen with a rebased branch that resolves merge conflicts with main (Hub.scala private[this] changes from #3063).

PR #3064, opened 2026-06-14
Fix unresolvable scaladoc links to the ShardRegion actor #353

arun2dot0:
### Motivation
`sbt unidoc` emits `Could not find any member to link for "ShardRegion"` warnings across the `cluster-sharding` module. The bare `[[ShardRegion]]` links refer to the `ShardRegion` *actor*, which is the `private[pekko] class ShardRegion` and therefore is not part of the generated API docs, so scaladoc cannot resolve the link.

### Modification
Replace the bare `[[ShardRegion]]` actor references with backticked `ShardRegion` in `ClusterSharding.scala`, `ShardRegion.scala`, and `ShardCoordinator.scala`. This matches the existing local style (these files already refer to the actor as `ShardRegion` in prose) and the approach taken in #2780. Qualified links that resolve correctly (e.g. `[[ShardRegion.Passivate]]`, `[[ShardRegion.MessageExtractor]]`) are left untouched.

### Result
The 18 `Could not find any member to link for "ShardRegion"` warnings in the `cluster-sharding` module are gone.

### Tests
- `sbt -Dpekko.scaladoc.diagrams=false "cluster-sharding/doc"` - success, 0 `ShardRegion` link warnings remaining (docs only)
- `sbt "cluster-sharding/scalafmtCheck"` - success
- `git diff --check` - clean

### References
Refs #353

### DIFF
--- a/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ClusterSharding.scala
+++ b/cluster-sharding/src/main/scala/org/apache/pekko/cluster/sharding/ClusterSharding.scala
@@ -72,7 +72,7 @@ import pekko.util.ByteString
  * '''ShardRegion''':
  * Each entity actor runs only at one place, and messages can be sent to the entity without
  * requiring the sender to know the location of the destination actor. This is achieved by
- * sending the messages via a [[ShardRegion]] actor, provided by this extension. The [[ShardRegion]]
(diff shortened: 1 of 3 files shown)


He-Pin (approved):

He-Pin:
Thanks

PR #3073, opened 2026-06-15
fix: ClassTag context bound compat for Scala 3.8 cross-build

He-Pin:
### Motivation
In Scala 3.8, `[T: ClassTag]` context bounds desugar to `using` clauses, making explicit `ClassTag` passing positional-incompatible. Code patterns like `method(args)(ClassTag(clazz))` break when the method uses a context bound.

### Modification
- Replace explicit `ClassTag` passing with `implicit val ct: ClassTag[T] = ClassTag(clazz)` followed by method calls that rely on implicit resolution, in `Behavior.scala`, `ProducerController.scala`, `WorkPullingProducerController.scala`, `Topic.scala`, `ConsistentHash.scala`, `ShardingProducerController.scala`, `EventSourcedBehaviorTestKit.scala`, `StreamTestKit.scala`
- Use `.apply()` trick in `ActorContextSpec.scala`: `decoration[T].apply(behavior)` to separate implicit resolution from function application
- Convert context bound `[T: ClassTag]` to explicit implicit parameter `(implicit ev: ClassTag[T])` in `Supervision.scala`
- Merge context bound with separate implicit parameter list in `PersistencePlugin.scala`: `(implicit ct: ClassTag[T], ev: PluginProvider[...])`
- Refactor `ShardedDaemonProcessImpl.scala` Java API methods to avoid explicit classTag passing
- Fix `EventStream.scala` secondary constructor formatting

### Result
All ClassTag-dependent code compiles with both Scala 2.13.18 and Scala 3.8.4. No behavioral changes.

### Tests
- Not run - compilation fix, verified by cross-compilation

### References
None - proactive Scala 3.8 forward compatibility

### DIFF
--- a/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/ActorContextSpec.scala
+++ b/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/ActorContextSpec.scala
@@ -83,7 +83,7 @@ abstract class ActorContextSpec extends ScalaTestWithActorTestKit with AnyWordSp
   def decoration[T: ClassTag]: Behavior[T] => Behavior[T]
 
   implicit class BehaviorDecorator[T](behavior: Behavior[T])(implicit ev: ClassTag[T]) {
-    def decorate: Behavior[T] = decoration[T](ev)(behavior)
+    def decorate: Behavior[T] = decoration[T].apply(behavior)
   }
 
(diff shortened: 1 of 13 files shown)


He-Pin:
Better switch to using I think

pjfanning (approved):
lgtm

PR #3076, opened 2026-06-16
Add TCK tests for replay bounds across a deletion gap

mucciolo:
### Motivation
[PR #517 on `pekko-persistence-jdbc`](https://github.com/apache/pekko-persistence-jdbc/pull/517) added regression tests for replaying messages whose range overlaps a deleted prefix (the gap left by `DeleteMessagesTo`). @pjfanning asked for the same coverage in the shared TCK so all journal implementations meet one contract.

### Test cases
Add five `JournalSpec` cases. Each deletes messages 1..3 (4..5 survive) and then replays a different window:
1. lower bound = 3 (last deleted): replays 4, 5
2. window 1..4 (spans the gap): replays 4
3. window 1..2 (ends inside the gap): no replay, `RecoverySuccess` only
4. window 1..3 (ends at last deleted): no replay, `RecoverySuccess` only
5. lower bound = 4 (first survivor): replays 4, 5


He-Pin (approved):
lgtm thanks

PR #3082, opened 2026-06-18
fix: Scala 3.9 context bound and trait field compatibility

He-Pin:
### Motivation
Scala 3.9.0-RC1 introduces breaking changes:
1. **Context bound desugaring**: `[T: ClassTag]` now desugars differently, making `method(args)(ClassTag(clazz))` fail with "No ClassTag available for T"
2. **Trait field bytecode**: `private val` in traits generates abstract setter methods with generic type signatures that mismatch between trait and implementing class, causing Java subclass compilation failures

### Modification
- **actor-typed/Behaviors.scala**: Use `implicit val ct: ClassTag[T] = ClassTag(clazz)` + implicit resolution instead of explicit `ClassTag(clazz)` passing for `monitor`, `transformMessages`, and `withMdc` methods
- **stream/Ops.scala**: Use `implicit val ls: LogSource[Materializer] = fromMaterializer` + implicit resolution instead of explicit passing for `Log` and `LogWithMarker` preStart methods
- **actor/FSM.scala**, **persistence/PersistentFSMBase.scala**: Change `private val handleEventDefault` to `private def handleEventDefault` to avoid abstract setter generation for this specific field

### Result
- ClassTag and LogSource context bound errors in actor-typed and stream modules are fully resolved
- `handleEventDefault` setter issue is partially mitigated; remaining trait setter issues are tracked in scala/scala3#26372 as a Scala 3.9 compiler regression

### Tests
- `sbt "++3.9.0-RC1!; compile"` — ClassTag/LogSource errors resolved

### References
- scala/scala3#26372 — Trait `private val` setter generates generic signature that Java subclasses cannot implement (Scala 3.9 regression)
- Related to #3073 which fixed similar ClassTag context bound issues for Scala 3.8

### DIFF
--- a/persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
+++ b/persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
@@ -466,7 +466,7 @@ trait PersistentFSMBase[S, D, E] extends Actor with Listeners with ActorLogging 
   /*
    * unhandled event handler
    */
-  private val handleEventDefault: StateFunction = {
+  private def handleEventDefault: StateFunction = {
     case Event(value, _) =>
       log.warning("unhandled event " + value + " in state " + stateName)
       stay()
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/javadsl/Behaviors.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/javadsl/Behaviors.scala
@@ -234,8 +234,10 @@ object Behaviors {
    * @param monitor The messages will also be sent to this `ActorRef`
    * @param behavior The inner behavior that is decorated
    */
-  def monitor[T](interceptMessageClass: Class[T], monitor: ActorRef[T], behavior: Behavior[T]): Behavior[T] =
-    scaladsl.Behaviors.monitor(monitor, behavior)(ClassTag(interceptMessageClass))
+  def monitor[T](interceptMessageClass: Class[T], monitor: ActorRef[T], behavior: Behavior[T]): Behavior[T] = {
+    implicit val ct: ClassTag[T] = ClassTag(interceptMessageClass)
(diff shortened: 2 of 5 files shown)


persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
```diff
@@ -466,7 +466,7 @@ trait PersistentFSMBase[S, D, E] extends Actor with Listeners with ActorLogging
   /*
    * unhandled event handler
    */
-  private val handleEventDefault: StateFunction = {
+  private def handleEventDefault: StateFunction = {
```
He-Pin, line 469: Will wait scala 3 fix

pjfanning (dismissed):
lgtm

persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
pjfanning: are we going to wait or will we log an issue to revert this when Scala team fix it?
We should link to the Scala bug.

He-Pin: https://github.com/scala/scala3/pull/26374 @pjfanning maybe we can wait the RC2 ?

pjfanning: let's wait till scala RC2

He-Pin: Yes, have to wait,  I want to make Pekko project 3.9.0 ready, then we can help capture new bugs

He-Pin: And I found Pekko's projects do help capture some bugs during the RC release

PR #3086, opened 2026-06-18
fix(actor-typed): discard stale ReceiveTimeout after cancelReceiveTimeout to avoid NPE

He-Pin:
### Motivation

When `context.cancelReceiveTimeout()` is called while a classic `ReceiveTimeout` message has already been enqueued in the mailbox by the scheduler, the typed actor crashes with a `NullPointerException` inside `InterceptorImpl.receive`.

**Root cause trace:**

1. `setReceiveTimeout(d, msg)` → sets `receiveTimeoutMsg = msg`, schedules `classic.ReceiveTimeout` via `scheduleOnce`
2. Before the actor processes the enqueued `ReceiveTimeout`, another message triggers `cancelReceiveTimeout()`
3. `cancelReceiveTimeout()` sets `receiveTimeoutMsg = null` and calls `classicActorContext.setReceiveTimeout(Duration.Undefined)`
4. The already-enqueued `classic.ReceiveTimeout` **cannot be retracted** from the mailbox (the `Cancellable.cancel()` only prevents future scheduling, not already-delivered messages)
5. `ActorAdapter.aroundReceive` matches `classic.ReceiveTimeout` and calls `handleMessage(ctx.receiveTimeoutMsg)` — **with null**
6. The null message propagates through `Behavior.interpretMessage` → `InterceptorImpl.receive` → **`msg.getClass` throws NPE**

This bug was observed in both 2.8.3 and 2.6.21 of the upstream project. A reproduction is available at https://github.com/lolboxen/akka-receive-timeout-bug.

### Modification

- `ActorAdapter.aroundReceive`: null-check `ctx.receiveTimeoutMsg` before forwarding to `handleMessage`. If null (timeout was cancelled), the stale `classic.ReceiveTimeout` is silently discarded.
- Added `CancelReceiveTimeoutSpec`: a deterministic regression test that directly sends `classic.ReceiveTimeout` to a typed actor after `cancelReceiveTimeout()` has been called, verifying the actor survives without NPE. The test uses an interceptor in the behavior stack to exercise the exact `InterceptorImpl.receive` code path where the NPE manifests.

### Result

Typed actors no longer crash when `cancelReceiveTimeout()` races with an already-enqueued stale `ReceiveTimeout`. The fix is a single null-guard at the adapter layer — the typed behavior stack never sees a null message.

### Tests

```
sbt "actor-typed-tests / Test / testOnly org.apache.pekko.actor.typed.CancelReceiveTimeoutSpec"
→ All tests passed

sbt "actor-typed-tests / Test / testOnly org.apache.pekko.actor.typed.ActorContextSpec org.apache.pekko.actor.typed.TimerSpec org.apache.pekko.actor.typed.CancelReceiveTimeoutSpec"
→ 14 tests passed, 0 failed
```

The regression test was verified to **fail without the fix** (NPE at `InterceptorImpl.scala:94`) and **pass with the fix**.

### References

- Reproduction: https://github.com/lolboxen/akka-receive-timeout-bug

### DIFF
--- a/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/CancelReceiveTimeoutSpec.scala
+++ b/actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/CancelReceiveTimeoutSpec.scala
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+package org.apache.pekko.actor.typed
+
+import scala.concurrent.duration._
+import scala.reflect.ClassTag
+
+import org.apache.pekko
+import pekko.actor.testkit.typed.scaladsl.LogCapturing
+import pekko.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
+import pekko.actor.testkit.typed.scaladsl.TestProbe
+import pekko.actor.typed.scaladsl.Behaviors
+import pekko.actor.typed.scaladsl.adapter._
+
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class CancelReceiveTimeoutSpec extends ScalaTestWithActorTestKit with AnyWordSpecLike with LogCapturing {
+
+  sealed trait Command
+  case object SetAndCancelTimeout extends Command
+  case object Timeout extends Command
+  case object Ping extends Command
+  case object Pong extends Event
+
+  sealed trait Event
+  case object Done extends Event
+
+  // A no-op interceptor that deepens the behavior stack so the message
+  // passes through InterceptorImpl.receive — the exact code path where
+  // msg.getClass throws NPE on a null message.
+  private def noopInterceptor[T: ClassTag] = new BehaviorInterceptor[T, T] {
+    override def aroundReceive(
+        ctx: TypedActorContext[T],
+        msg: T,
+        target: BehaviorInterceptor.ReceiveTarget[T]): Behavior[T] =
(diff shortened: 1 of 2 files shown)


actor-typed/src/main/scala/org/apache/pekko/actor/typed/internal/adapter/ActorAdapter.scala
```diff
@@ -100,7 +100,11 @@ import pekko.util.OptionVal
             } else Terminated(ActorRefAdapter(ref))
           handleSignal(msg)
         case classic.ReceiveTimeout =>
-          handleMessage(ctx.receiveTimeoutMsg)
+          // cancelReceiveTimeout() sets receiveTimeoutMsg to null, but a classic ReceiveTimeout
+          // that was already enqueued in the mailbox before the cancel cannot be retracted.
+          // Discard the stale timeout to avoid passing null into the typed behavior stack (#3084).
+          val timeoutMsg = ctx.receiveTimeoutMsg
+          if (timeoutMsg != null) handleMessage(timeoutMsg)
```
He-Pin, line 107: should use `ne`

actor-typed-tests/src/test/scala/org/apache/pekko/actor/typed/CancelReceiveTimeoutSpec.scala
```diff
@@ -0,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
```
pjfanning: wrong license header?

He-Pin: yes,working on it

pjfanning (approved):
lgtm

PR #3087, opened 2026-06-18
docs: add mandatory header workflow reminder to Licensing Rules

He-Pin:
### Motivation

The existing Licensing Rules already state "Do not hand-write or invent license headers. Let sbt manage them" and "For new files, run `sbt headerCreateAll`", but agents can still miss these rules and hand-write a header (e.g. accidentally using the Akka-derived header for new code).

### Modification

Add a prominent blockquote reminder at the top of the Licensing Rules section that spells out the mandatory workflow in one line: create file without header → `sbt headerCreateAll` → done.

### Result

Agents see the workflow before reading the detailed rules, reducing the chance of hand-writing an incorrect header.

### Tests

Not run - docs only

### References

None - process improvement

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,6 +16,8 @@ Follow `CONTRIBUTING.md`. If this file conflicts with `CONTRIBUTING.md`, follow 
 
 ## Licensing Rules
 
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #3041, opened 2026-06-03
fix: widen RemoteSendConsistencySpec await window to 30s

He-Pin:
## Motivation

`ArteryTcp/TlsTcp SendConsistencyWithOneLaneSpec` flake on busy CI:
- `must be able to send messages concurrently preserving order` — times out at the `within` window
- `must be able to send messages with actorSelection concurrently preserving order` — same pattern

Observed locally on Darwin 25.3.0 with timefactor=2 (so `within(10.seconds)` → actual 20s budget):

```
- must be able to send messages concurrently preserving order *** FAILED *** (20 seconds, 5 milliseconds)
  java.lang.AssertionError: assertion failed: timeout (19999978648 nanoseconds) during expectMsg while waiting for success
- must be able to send messages with actorSelection concurrently preserving order *** FAILED *** (20 seconds)
  java.lang.AssertionError: assertion failed: timeout (19999978257 nanoseconds) during expectMsg while waiting for success2
```

The test drives ~4004 round-trips through TCP/TLS-TCP across four sender pairs (1000 ping-pong messages per pair × 4 pairs + acks). TLS handshake plus JIT warmup on a loaded CI runner easily exceed the original 10s budget without indicating any real regression.

## Modification

Bump `within(10.seconds)` → `within(30.seconds)` at lines 179 and 219 of `RemoteSendConsistencySpec.scala`. Two-line diff.

## Result

Up to 60s of dilated wall-clock headroom on CI (timefactor=2 × 30s). Still fails fast on genuine deadlocks — does not mask real regressions.

## Tests

- `scalafmt --test` on the changed file: passes
- Did not run `remote/test` locally; CI `Check / Test` will exercise both ArteryTcp and ArteryTlsTcp variants.

## References

- Failing classes: `ArteryTcpSendConsistencyWithOneLaneSpec`, `ArteryTlsTcpSendConsistencyWithOneLaneSpec` (concrete subclasses of `AbstractRemoteSendConsistencySpec`)
- Test file: `remote/src/test/scala/org/apache/pekko/remote/artery/RemoteSendConsistencySpec.scala`

PR #3090, opened 2026-06-19
fix: widen ActorSelection SendConsistencySpec timeout to 60s

He-Pin:
### Motivation

The "must be able to send messages with actorSelection concurrently preserving order" test flakes on CI: 4 sender actors each drive 1000 round trips via `ActorSelection` (4004 messages total), wrapped in `ActorSelectionMessage` with remote-side path traversal overhead. On CI with `-Dpekko.test.timefactor=2`, `within(30.seconds)` dilates to 60s which is insufficient under load — 3 of 4 senders complete but the 4th times out waiting for the final `success2` message.

Investigation confirmed no recent code change caused this:
- Artery does not use stage actors → `LazyDispatch` (ba4e950edb) has zero impact
- Materializer wiring optimization (03ebaf59bb) affects materialization only, not runtime message flow
- `GraphInterpreter` `pendingFinalization` (a5058437af) is a hot-path optimization, no behavioral change

### Modification

Bump `within(30.seconds)` → `within(60.seconds)` at line 219 for the ActorSelection test only. The ActorRef variant (line 179) is unchanged as it passes within the original budget.

### Result

Up to 120s of dilated wall-clock headroom on CI for the ActorSelection variant; still fails fast on genuine deadlocks.

### Tests

- `sbt "remote / Test / compile"` — passes
- CI Check / Test will exercise the artery variants

### References

Refs #3041 (previous timeout widening from 10s to 30s)

### DIFF
--- a/remote/src/test/scala/org/apache/pekko/remote/artery/RemoteSendConsistencySpec.scala
+++ b/remote/src/test/scala/org/apache/pekko/remote/artery/RemoteSendConsistencySpec.scala
@@ -216,7 +216,7 @@ abstract class AbstractRemoteSendConsistencySpec(config: Config)
       system.actorOf(senderProps(selC))
       system.actorOf(senderProps(selA))
 
-      within(30.seconds) {
+      within(60.seconds) {
         expectMsg("success2")
         expectMsg("success2")
         expectMsg("success2")


pjfanning:
test now failing in scala 3.3

He-Pin:
Closing in favor of #3092 which fixes the root cause.

The timeout widening only masks the symptom. The actual issue is that `selectQueue` in `Association.send` uses the anchor's UID (root guardian, always 0) as the distribution key for outbound lanes. With `outbound-lanes > 1`, all ActorSelection messages concentrate on a single lane (`math.abs(0 % N) = 0`), creating a throughput bottleneck.

#3092 distributes ActorSelection messages across lanes based on their target path hash instead, eliminating the structural bottleneck.

Issue #3089, opened 2026-06-19
AbstractRemoteSendConsistencySpec actorSelection preserving order test now flaky


pjfanning:
Not sure if this was always flaky but it now fails on a very regular basis - especially on scala 2.13

```
[06-18 22:07:31.843] [info] - must be able to send messages with actorSelection concurrently preserving order *** FAILED *** (1 minute)
[06-18 22:07:31.843] [info]   java.lang.AssertionError: assertion failed: timeout (59999986805 nanoseconds) during expectMsg while waiting for success2
[06-18 22:07:31.843] [info]   at scala.Predef$.assert(Predef.scala:279)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.expectMsg_internal(TestKit.scala:470)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.expectMsg(TestKit.scala:447)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.expectMsg$(TestKit.scala:447)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKit.expectMsg(TestKit.scala:963)
[06-18 22:07:31.843] [info]   at org.apache.pekko.remote.artery.AbstractRemoteSendConsistencySpec.$anonfun$new$10(RemoteSendConsistencySpec.scala:220)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.within(TestKit.scala:427)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.within$(TestKit.scala:414)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKit.within(TestKit.scala:963)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.within(TestKit.scala:442)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKitBase.within$(TestKit.scala:442)
[06-18 22:07:31.843] [info]   at org.apache.pekko.testkit.TestKit.within(TestKit.scala:963)
[06-18 22:07:31.844] [info]   at org.apache.pekko.remote.artery.AbstractRemoteSendConsistencySpec.$anonfun$new$8(RemoteSendConsistencySpec.scala:219)
[...]

PR #3092, opened 2026-06-19
fix(artery): distribute ActorSelection messages across outbound lanes by target path

He-Pin:
### Motivation

The "must be able to send messages with actorSelection concurrently preserving order" test flakes on CI: 4 sender actors each drive 1000 round-trips via `ActorSelection` (4004 messages total). With multi-lane artery config (`outbound-lanes > 1`), **all ActorSelection messages are routed to the same outbound queue** because `selectQueue` uses the anchor's UID as the distribution key:

```scala
OrdinaryQueueIndex + (math.abs(r.path.uid % outboundLanes))
```

The anchor for `ActorSelection` is the root guardian (`RootActorPath`), whose UID is always `0` (`ActorCell.undefinedUid`). So `math.abs(0 % N) = 0` for any N — all ActorSelection traffic concentrates on lane 0 while other lanes sit idle.

In contrast, the ActorRef variant distributes across lanes because each echo actor has a distinct non-zero UID.

This is **not a recent regression** — the `selectQueue` logic has been unchanged since the Pekko fork from Akka. PR #3090 (timeout widening) masks the symptom but doesn't fix the structural bottleneck.

### Modification

Add a dedicated `case sel: ActorSelectionMessage` (non-PriorityMessage) in `Association.send` that computes the queue index from the selection's **target path elements hash** instead of the anchor's UID:

```scala
case sel: ActorSelectionMessage =>
  val queueIndex =
    if (outboundLanes == 1) OrdinaryQueueIndex
    else OrdinaryQueueIndex + (math.abs(sel.elements.hashCode()) % outboundLanes)
  val queue = queues(queueIndex)
  if (!queue.offer(outboundEnvelope))
    dropped(queueIndex, queueSize, outboundEnvelope)
```

This distributes ActorSelection messages across lanes by their target path while preserving **per-path message ordering** (same target path → same hash → same lane).

PriorityMessage ActorSelection (used by cluster heartbeats) continues to go through the control queue unchanged.

### Result

ActorSelection messages are now distributed across all outbound lanes based on target path, eliminating the single-lane throughput bottleneck. The existing ActorRef-based test and PriorityMessage routing are unaffected.

### Tests

- `sbt "remote / Test / compile"` — compile check
- CI will exercise the artery variants (1-lane and 3-lane configs)

### References

Refs #3041 (previous timeout widening), supersedes #3090

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/artery/Association.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/artery/Association.scala
@@ -424,8 +424,26 @@ private[remote] class Association(
             }(materializer.executionContext)
           case _: SystemMessage =>
             sendSystemMessage(outboundEnvelope)
-          case ActorSelectionMessage(_: PriorityMessage, _, _) | _: ControlMessage | _: ClearSystemMessageDelivery =>
-            // ActorSelectionMessage with PriorityMessage is used by cluster and remote failure detector heartbeating
+          case sel: ActorSelectionMessage =>
+            sel.msg match {
+              case _: PriorityMessage =>
+                // ActorSelectionMessage with PriorityMessage is used by cluster and remote failure detector heartbeating
+                if (!controlQueue.offer(outboundEnvelope)) {
+                  dropped(ControlQueueIndex, controlQueueSize, outboundEnvelope)
+                }
+              case _ =>
+                // Distribute ActorSelection messages across lanes based on the target
+                // path hash rather than the anchor's (typically root guardian, UID=0).
+                // Without this, all ActorSelection messages concentrate on a single
+                // outbound lane, creating a throughput bottleneck.
+                val queueIndex =
+                  if (outboundLanes == 1) OrdinaryQueueIndex
+                  else OrdinaryQueueIndex + ((sel.elements.hashCode() & Int.MaxValue) % outboundLanes)
+                val queue = queues(queueIndex)
+                if (!queue.offer(outboundEnvelope))
+                  dropped(queueIndex, queueSize, outboundEnvelope)
+            }
+          case _: ControlMessage | _: ClearSystemMessageDelivery =>
             if (!controlQueue.offer(outboundEnvelope)) {
               dropped(ControlQueueIndex, controlQueueSize, outboundEnvelope)
             }
--- a/remote/src/test/scala/org/apache/pekko/remote/artery/ActorSelectionQueueDistributionSpec.scala
+++ b/remote/src/test/scala/org/apache/pekko/remote/artery/ActorSelectionQueueDistributionSpec.scala
@@ -23,6 +23,7 @@ import org.apache.pekko
 import pekko.actor.SelectChildName
 import pekko.actor.SelectionPathElement
 import pekko.protobufv3.internal.ByteString
+import pekko.protobufv3.internal.UnknownFieldSet
 import pekko.remote.ContainerFormats
 import pekko.remote.ContainerFormats.PatternType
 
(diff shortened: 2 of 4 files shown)


remote/src/main/scala/org/apache/pekko/remote/artery/Association.scala
```diff
@@ -436,6 +436,17 @@ private[remote] class Association(
             // so that the first message can be delivered after the remote actor has been created.
             if (!controlQueue.offer(outboundEnvelope))
               dropped(ControlQueueIndex, controlQueueSize, outboundEnvelope)
+          case sel: ActorSelectionMessage =>
```
pjfanning, line 427: can you group this with https://github.com/apache/pekko/blob/0fb44b4ba7985ea552b310ae582328967f42ce8a/remote/src/main/scala/org/apache/pekko/remote/artery/Association.scala#L427 ?

remote/src/main/scala/org/apache/pekko/remote/artery/Association.scala
```diff
@@ -424,8 +424,25 @@ private[remote] class Association(
...
+                if (!controlQueue.offer(outboundEnvelope))
+                  dropped(ControlQueueIndex, controlQueueSize, outboundEnvelope)
+              case _ =>
+                // Distribute ActorSelection messages across lanes based on the target
+                // path hash rather than the anchor's (typically root guardian, UID=0).
+                // Without this, all ActorSelection messages concentrate on a single
+                // outbound lane, creating a throughput bottleneck.
+                val queueIndex =
+                  if (outboundLanes == 1) OrdinaryQueueIndex
+                  else OrdinaryQueueIndex + (math.abs(sel.elements.hashCode()) % outboundLanes)
```
He-Pin: Should we limit the outbound lanes to a power of 2? then we can do a quic mode

remote/src/test/scala/org/apache/pekko/remote/artery/ActorSelectionQueueDistributionSpec.scala
```diff
@@ -0,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
```
pjfanning: wrong license header?

pjfanning:
#3089 test is failing in this PR - we appear to have to broken this with some change over the last week or 2. This test is now very flaky and I don't recall it being an issue before

remote/src/test/scala/org/apache/pekko/remote/artery/RemoteSendConsistencySpec.scala
```diff
@@ -195,7 +195,7 @@ abstract class AbstractRemoteSendConsistencySpec(config: Config)
 
       def senderProps(sel: ActorSelection) =
         Props(new Actor {
-          var counter = 1000
+          var counter = 100
```
pjfanning: I'd like to delay merging this, especially anything that modifies the test until we found out what commit caused #3089. Because, it definitely seems like some recent change has made something that worked before, not work reliably any more.

pjfanning:
#3089 is happening for tests that don't use actor selection but maybe not quite as often.

pjfanning:
@He-Pin the CI in this repo is now badly affected by the #3090 test issues

He-Pin:
@pjfanning I still think the problem is the CI， not our code ：（

pjfanning (approved):
lgtm

PR #3096, opened 2026-06-19
Clean up stopped FSM transition listeners

fallintoplace:
## Summary

Classic FSM and PersistentFSM keep transition listeners in a listener set until they explicitly unsubscribe. This changes the subscription path to DeathWatch listeners with an internal cleanup message, so stopped listener actors are removed automatically.

The explicit unsubscribe/deafen paths still remove listeners, and they now also unwatch only DeathWatch registrations created by the listener subscription path. If an FSM was already watching the same actor for its own logic, this change does not replace that existing watch.

## Tests

- `actor-tests/testOnly org.apache.pekko.actor.FSMTransitionSpec`
- `persistence/testOnly org.apache.pekko.persistence.fsm.InmemPersistentFSMSpec`
- `docs/paradox`

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/FSM.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/FSM.scala
@@ -99,6 +99,7 @@ object FSM {
    * INTERNAL API
    */
   private final case class TimeoutMarker(generation: Long)
+  private final case class ListenerTerminated(actorRef: ActorRef)
 
   /** INTERNAL API */
   @InternalApi
@@ -406,9 +407,8 @@ object FSM {
  *
  * Another feature is that other actors may subscribe for transition events by
  * sending a <code>SubscribeTransitionCallback</code> message to this actor.
- * Stopping a listener without unregistering will not remove the listener from the
- * subscription list; use <code>UnsubscribeTransitionCallback</code> before stopping
- * the listener.
+ * Stopping a listener will remove it from the subscription list. Subscriptions
+ * may also be removed explicitly with <code>UnsubscribeTransitionCallback</code>.
  *
  * State timeouts set an upper bound to the time which may pass before another
  * message is received in the current state. If no external message is
@@ -779,9 +779,32 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
    * transition handling
    */
   private var transitionEvent: List[TransitionHandler] = Nil
+  private var watchedListeners: Set[ActorRef] = Set.empty
   private def handleTransition(prev: S, next: S): Unit = {
     val tuple = (prev, next)
     for (te <- transitionEvent) { if (te.isDefinedAt(tuple)) te(tuple) }
+  }
+
+  private def addListener(actorRef: ActorRef): Unit = {
+    if (listeners.add(actorRef)) {
+      context.asInstanceOf[ActorCell].isWatching(actorRef) match {
+        case false =>
+          context.watchWith(actorRef, ListenerTerminated(actorRef))
+          watchedListeners += actorRef
+        case true =>
+          log.warning(
+            "Listener [{}] is already watched by this FSM; it will not be automatically removed on termination. " +
+            "Use UnsubscribeTransitionCallBack explicitly.",
+            actorRef)
+      }
+    }
+  }
+
+  private def removeListener(actorRef: ActorRef): Unit = {
+    if (listeners.remove(actorRef) && watchedListeners.contains(actorRef)) {
+      watchedListeners -= actorRef
+      context.unwatch(actorRef)
+    }
   }
 
   /*
@@ -807,19 +830,20 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
         processMsg(msg, t)
       }
     case SubscribeTransitionCallBack(actorRef) =>
-      // TODO Use context.watch(actor) and receive Terminated(actor) to clean up list
-      listeners.add(actorRef)
+      addListener(actorRef)
       // send current state back as reference point
       actorRef ! CurrentState(self, currentState.stateName)
     case Listen(actorRef) =>
-      // TODO Use context.watch(actor) and receive Terminated(actor) to clean up list
-      listeners.add(actorRef)
+      addListener(actorRef)
       // send current state back as reference point
       actorRef ! CurrentState(self, currentState.stateName)
     case UnsubscribeTransitionCallBack(actorRef) =>
-      listeners.remove(actorRef)
+      removeListener(actorRef)
     case Deafen(actorRef) =>
+      removeListener(actorRef)
+    case ListenerTerminated(actorRef) =>
       listeners.remove(actorRef)
+      watchedListeners -= actorRef
     case value =>
       if (timeoutFuture.isDefined) {
         timeoutFuture.get.cancel()
--- a/persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
+++ b/persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
@@ -85,9 +85,8 @@ import pekko.routing.{ Deafen, Listen, Listeners }
  *
  * Another feature is that other actors may subscribe for transition events by
  * sending a <code>SubscribeTransitionCallback</code> message to this actor.
- * Stopping a listener without unregistering will not remove the listener from the
- * subscription list; use <code>UnsubscribeTransitionCallback</code> before stopping
- * the listener.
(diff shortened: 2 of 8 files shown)


He-Pin (approved):
### Review: LGTM with minor suggestions

The approach is clean and correct for the common case. Replacing the old `listeners.add()` with DeathWatch-based cleanup properly addresses the TODO comments and fixes the listener leak on actor stop. The `watchedListeners` tracking to distinguish subscription-path watches from user-path watches is a good design choice.

Two improvement suggestions posted as inline comments — neither blocks merge.

**Verified:**
- `addListener`/`removeListener`/`ListenerTerminated` handler maintain `listeners` and `watchedListeners` consistently
- `removeListener` correctly avoids `unwatch` for actors watched by user logic (not in `watchedListeners`)
- `isWatching` guard prevents `IllegalStateException` from mixing `watch` and `watchWith`
- DeathWatch automatically removes from `watching` on termination, so `ListenerTerminated` handler doesn't need `unwatch`
- Tests cover both Classic FSM and PersistentFSM cleanup paths

actor/src/main/scala/org/apache/pekko/actor/FSM.scala
```diff
@@ -779,11 +779,25 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
    * transition handling
    */
   private var transitionEvent: List[TransitionHandler] = Nil
+  private val watchedListeners: mutable.Set[ActorRef] = mutable.Set.empty
   private def handleTransition(prev: S, next: S): Unit = {
```
He-Pin, line 783: 🟡 **Suggestion: listener leak when FSM already watches the listener actor**

When `isWatching(actorRef)` returns `true` (the FSM's business logic already called `context.watch()` on this actor), `watchWith` is skipped and the actor is not added to `watchedListeners`. If that actor later terminates, the FSM receives a plain `Terminated` instead of `ListenerTerminated`, and the listener is never removed from `listeners`.

Adding a `case Terminated(actorRef) => listeners.remove(actorRef)` handler to the `receive` block would be a **breaking change** (it would intercept `Terminated` before it reaches user state functions via `processMsg`). So this edge case may be acceptable as-is.

Consider adding a log warning when `isWatching` is true at subscription time, so users know the listener won't be auto-cleaned:

```scala
private def addListener(actorRef: ActorRef): Unit = {
  if (listeners.add(actorRef)) {
    if (!context.asInstanceOf[ActorCell].isWatching(actorRef)) {
      context.watchWith(actorRef, ListenerTerminated(actorRef))
      watchedListeners += actorRef
    } else {
      log.warning(
        "Listener [{}] is already watched by this FSM; it will not be " +
        "automatically removed on termination. Use UnsubscribeTransitionCallBack explicitly.",
        actorRef)
    }
  }
}
```

Not blocking — the common case is handled correctly.

persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
```diff
@@ -482,11 +481,25 @@ trait PersistentFSMBase[S, D, E] extends Actor with Listeners with ActorLogging
    * transition handling
```
He-Pin, line 481: 🟡 **Suggestion: code duplication with FSM.scala**

The `watchedListeners` field, `addListener`, and `removeListener` are duplicated nearly identically between `FSM.scala` and `PersistentFSMBase.scala`. This is understandable given they share the `Listeners` trait but not a common base beyond that.

Not blocking for this PR — but consider extracting this into a shared mixin (e.g. `WatchedListeners`) in a follow-up, to avoid divergent fixes if this logic needs to change later.

Also, `ListenerTerminated` visibility differs: `private` in `FSM` companion vs `private[persistence]` in `PersistentFSM` companion. This is fine since each module only accesses its own companion's case class, but worth aligning if the shared mixin approach is taken.

persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
```diff
@@ -482,11 +481,32 @@ trait PersistentFSMBase[S, D, E] extends Actor with Listeners with ActorLogging
...
    */
   private var transitionEvent: List[TransitionHandler] = Nil
+  private val watchedListeners: mutable.Set[ActorRef] = mutable.Set.empty
```
pjfanning: Why not a var with an immutable Set? Like the `transitionEvent` list above.

actor/src/main/scala/org/apache/pekko/actor/FSM.scala
```diff
@@ -779,11 +779,32 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
...
     val tuple = (prev, next)
     for (te <- transitionEvent) { if (te.isDefinedAt(tuple)) te(tuple) }
   }
 
+  private def addListener(actorRef: ActorRef): Unit = {
+    if (listeners.add(actorRef)) {
+      if (!context.asInstanceOf[ActorCell].isWatching(actorRef)) {
```
He-Pin: better with pattern matching

persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
He-Pin: Save ygc, I think? But I agree with you.

persistence/src/main/scala/org/apache/pekko/persistence/fsm/PersistentFSMBase.scala
```diff
@@ -482,11 +481,32 @@ trait PersistentFSMBase[S, D, E] extends Actor with Listeners with ActorLogging
...
   private def handleTransition(prev: S, next: S): Unit = {
     val tuple = (prev, next)
     for (te <- transitionEvent) { if (te.isDefinedAt(tuple)) te(tuple) }
   }
 
+  private def addListener(actorRef: ActorRef): Unit = {
+    if (listeners.add(actorRef)) {
+      if (!context.asInstanceOf[ActorCell].isWatching(actorRef)) {
```
He-Pin: use pattern matching

pjfanning (dismissed):
mima binary compatibility issue

needs an excludes file to include the problem exclusions

```
[error] pekko-actor: Failed binary compatibility check against org.apache.pekko:pekko-actor_2.13:1.0.0! Found 2 potential problems (filtered 351)
[error]  * abstract synthetic method org$apache$pekko$actor$FSM$$watchedListeners()scala.collection.immutable.Set in interface org.apache.pekko.actor.FSM is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.actor.FSM.org$apache$pekko$actor$FSM$$watchedListeners")
[error]  * abstract synthetic method org$apache$pekko$actor$FSM$$watchedListeners_=(scala.collection.immutable.Set)Unit in interface org.apache.pekko.actor.FSM is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.actor.FSM.org$apache$pekko$actor$FSM$$watchedListeners_=")
[error] pekko-persistence: Failed binary compatibility check against org.apache.pekko:pekko-persistence_2.13:1.0.0! Found 2 potential problems (filtered 134)
[error]  * abstract synthetic method org$apache$pekko$persistence$fsm$PersistentFSMBase$$watchedListeners()scala.collection.immutable.Set in interface org.apache.pekko.persistence.fsm.PersistentFSMBase is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.persistence.fsm.PersistentFSMBase.org$apache$pekko$persistence$fsm$PersistentFSMBase$$watchedListeners")
[error]  * abstract synthetic method org$apache$pekko$persistence$fsm$PersistentFSMBase$$watchedListeners_=(scala.collection.immutable.Set)Unit in interface org.apache.pekko.persistence.fsm.PersistentFSMBase is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("org.apache.pekko.persistence.fsm.PersistentFSMBase.org$apache$pekko$persistence$fsm$PersistentFSMBase$$watchedListeners_=")
[error] java.lang.RuntimeException: Failed binary compatibility check against org.apache.pekko:pekko-persistence_2.13:1.0.0! Found 2 potential problems (filtered 134)
```

pjfanning:
@He-Pin do you think this is ready?

He-Pin:
@pjfanning Checking

He-Pin:
@fallintoplace Thank you

PR #3097, opened 2026-06-19
Preserve throttler tokens on failed writes

fallintoplace:
## Summary

`ThrottlerHandle` used to consume token-bucket capacity before delegating to the wrapped transport handle. If the wrapped handle returned `false`, no write happened but the bucket stayed depleted.

This restores or refunds the reserved token state when the wrapped write fails, while leaving the successful write and blackhole paths unchanged.

## Tests

- `remote/testOnly org.apache.pekko.remote.transport.ThrottlerHandleSpec org.apache.pekko.remote.classic.transport.ThrottleModeSpec org.apache.pekko.remote.classic.transport.ThrottlerTransportAdapterSpec`
- `remote/headerCheck`
- `remote/scalafmtAll`

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
@@ -146,6 +146,9 @@ object ThrottlerTransportAdapter {
 
     private def tokensGenerated(nanoTimeOfSend: Long): Int =
       (TimeUnit.NANOSECONDS.toMillis(nanoTimeOfSend - nanoTimeOfLastSend) * tokensPerSecond / 1000.0).toInt
+
+    private[transport] def hasSameSettings(other: TokenBucket): Boolean =
+      capacity == other.capacity && tokensPerSecond == other.tokensPerSecond
   }
 
   @SerialVersionUID(1L)
@@ -616,6 +619,43 @@ private[transport] class ThrottledAssociation(
 /**
  * INTERNAL API
  */
+private[transport] object ThrottlerHandle {
+
+  @tailrec private[transport] def tryConsume(
+      throttleRef: AtomicReference[ThrottleMode],
+      currentBucket: ThrottleMode,
+      tokens: Int): ThrottleMode = {
+    val timeOfSend = System.nanoTime()
+    val (newBucket, allow) = currentBucket.tryConsumeTokens(timeOfSend, tokens)
+    if (allow) {
+      if (throttleRef.compareAndSet(currentBucket, newBucket)) newBucket
+      else tryConsume(throttleRef, throttleRef.get(), tokens)
+    } else null
+  }
+
+  @tailrec private[transport] def refundTokens(
+      throttleRef: AtomicReference[ThrottleMode],
+      previousBucket: ThrottleMode,
+      consumedBucket: ThrottleMode,
+      tokens: Int): Unit = {
+    val currentBucket = throttleRef.get()
+    val refundedBucket =
+      if (currentBucket eq consumedBucket) previousBucket
+      else
+        (currentBucket, previousBucket) match {
+          case (bucket: TokenBucket, previous: TokenBucket) if bucket.hasSameSettings(previous) =>
+            bucket.copy(availableTokens = min(bucket.availableTokens + tokens, bucket.capacity))
+          case _ => currentBucket
+        }
+
+    if ((refundedBucket ne currentBucket) && !throttleRef.compareAndSet(currentBucket, refundedBucket))
+      refundTokens(throttleRef, previousBucket, consumedBucket, tokens)
+  }
+}
+
+/**
+ * INTERNAL API
+ */
 @nowarn("msg=deprecated")
 private[transport] final case class ThrottlerHandle(_wrappedHandle: AssociationHandle, throttlerActor: ActorRef)
     extends AbstractTransportAdapterHandle(_wrappedHandle, SchemeIdentifier) {
@@ -627,21 +667,16 @@ private[transport] final case class ThrottlerHandle(_wrappedHandle: AssociationH
   override def write(payload: ByteString): Boolean = {
     val tokens = payload.length
 
-    @tailrec def tryConsume(currentBucket: ThrottleMode): Boolean = {
-      val timeOfSend = System.nanoTime()
-      val (newBucket, allow) = currentBucket.tryConsumeTokens(timeOfSend, tokens)
-      if (allow) {
-        if (outboundThrottleMode.compareAndSet(currentBucket, newBucket)) true
-        else tryConsume(outboundThrottleMode.get())
-      } else false
-    }
-
     outboundThrottleMode.get match {
-      case Blackhole  => true
-      case bucket @ _ =>
-        val success = tryConsume(outboundThrottleMode.get())
-        if (success) wrappedHandle.write(payload) else false
-      // FIXME: this depletes the token bucket even when no write happened!! See #2825
+      case Blackhole   => true
+      case Unthrottled => wrappedHandle.write(payload)
+      case bucket      =>
+        val consumedBucket = ThrottlerHandle.tryConsume(outboundThrottleMode, bucket, tokens)
+        if (consumedBucket ne null) {
+          val written = wrappedHandle.write(payload)
+          if (!written) ThrottlerHandle.refundTokens(outboundThrottleMode, bucket, consumedBucket, tokens)
+          written
+        } else false
     }
 
   }
--- a/remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
+++ b/remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.remote.transport
+
+import scala.annotation.nowarn
+import scala.concurrent.Promise
+
+import org.apache.pekko
+import pekko.actor.Address
+import pekko.remote.transport.AssociationHandle.HandleEventListener
+import pekko.remote.transport.ThrottlerTransportAdapter.{ Blackhole, TokenBucket, Unthrottled }
+import pekko.testkit.PekkoSpec
+import pekko.util.ByteString
+
+@nowarn("msg=deprecated")
+class ThrottlerHandleSpec extends PekkoSpec {
+
+  class TestHandle(writeResult: Boolean, onWrite: () => Unit = () => ()) extends AssociationHandle {
+    var writeAttempts = 0
+
+    override val localAddress: Address = Address("pekko", "local")
+    override val remoteAddress: Address = Address("pekko", "remote")
+    override val readHandlerPromise: Promise[HandleEventListener] = Promise()
+
+    override def write(payload: ByteString): Boolean = {
+      writeAttempts += 1
+      onWrite()
+      writeResult
+    }
+  }
+
+  "A ThrottlerHandle" must {
+    "preserve token bucket state when the wrapped handle does not write" in {
+      val wrappedHandle = new TestHandle(writeResult = false)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+      val bucket = TokenBucket(capacity = 100, tokensPerSecond = 0, nanoTimeOfLastSend = 0L, availableTokens = 100)
+
+      handle.outboundThrottleMode.set(bucket)
+
+      handle.write(ByteString("1234567890")) should ===(false)
+
+      wrappedHandle.writeAttempts should ===(1)
+      handle.outboundThrottleMode.get() should ===(bucket)
+    }
+
+    "consume tokens when the wrapped handle writes" in {
+      val wrappedHandle = new TestHandle(writeResult = true)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+      val bucket = TokenBucket(capacity = 100, tokensPerSecond = 0, nanoTimeOfLastSend = 0L, availableTokens = 100)
+
+      handle.outboundThrottleMode.set(bucket)
+
+      handle.write(ByteString("1234567890")) should ===(true)
+
+      wrappedHandle.writeAttempts should ===(1)
+      handle.outboundThrottleMode.get().asInstanceOf[TokenBucket].availableTokens should ===(90)
+    }
+
+    "pass through writes when unthrottled" in {
+      Seq(true, false).foreach { writeResult =>
+        val wrappedHandle = new TestHandle(writeResult)
+        val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+
+        handle.write(ByteString("data")) should ===(writeResult)
+
+        wrappedHandle.writeAttempts should ===(1)
+        handle.outboundThrottleMode.get() should ===(Unthrottled)
+      }
+    }
+
+    "return true without calling wrapped handle when blackholed" in {
+      val wrappedHandle = new TestHandle(writeResult = true)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+
+      handle.outboundThrottleMode.set(Blackhole)
+
+      handle.write(ByteString("data")) should ===(true)
+
+      wrappedHandle.writeAttempts should ===(0)
+      handle.outboundThrottleMode.get() should ===(Blackhole)
+    }
+
+    "return false when token bucket has insufficient tokens" in {
+      val wrappedHandle = new TestHandle(writeResult = true)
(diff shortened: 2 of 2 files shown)


remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
```diff
@@ -0,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * license agreements; and to You under the Apache License, version 2.0:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
```
He-Pin, line 16: header is not right, use `sbt headerCreateAll`

remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
```diff
@@ -627,21 +627,41 @@ private[transport] final case class ThrottlerHandle(_wrappedHandle: AssociationH
   override def write(payload: ByteString): Boolean = {
     val tokens = payload.length
 
-    @tailrec def tryConsume(currentBucket: ThrottleMode): Boolean = {
+    @tailrec def tryConsume(currentBucket: ThrottleMode): Option[(ThrottleMode, ThrottleMode)] = {
```
He-Pin: Use a dedicated class instead of tuple2

He-Pin (approved):
## Review Summary

**Verdict: Approve** ✅

The fix correctly addresses the documented FIXME (#2825) where `ThrottlerHandle.write` consumed token bucket capacity even when the wrapped write returned `false`.

### Correctness Analysis

1. **`tryConsume` returning `Option[(ThrottleMode, ThrottleMode)]`** — Sound. Returns the `(previousBucket, consumedBucket)` pair needed for a precise CAS-based refund when no concurrent mutation occurred.

2. **`refundTokens` CAS loop** — Correct three-way dispatch:
   - `currentBucket == consumedBucket` → full restore (no concurrent mutation)
   - Same `TokenBucket` config → add back `tokens`, cap at `capacity`
   - Mode changed (e.g. `SetThrottle` to different mode) → no-op fallback

3. **`@tailrec` retry on CAS failure** — Correct for lock-free concurrent safety.

4. **`Unthrottled` / `Blackhole` paths** — `Blackhole` short-circuits before consume (unchanged). `Unthrottled` flows through `case bucket @ _` → `tryConsume` succeeds → refund's `case _` falls through to no-op (nothing to refund). Both correct.

5. **TOCTOU window** between `tryConsume` and `refundTokens` — Accepted tradeoff: if throttle mode changes (via `SetThrottle`) during the window, the refund is lost. This is rare and benign.

### Test Verification

- `ThrottlerHandleSpec` — 1 test, passes ✅
- `ThrottleModeSpec` — 5 tests, passes ✅  
- `ThrottlerTransportAdapterSpec` — 2 tests, passes ✅

### Suggestions (non-blocking)

See inline comments for test coverage improvements.

He-Pin (changes requested):
welcome and please fix license header

remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+      val bucket = TokenBucket(capacity = 100, tokensPerSecond = 0, nanoTimeOfLastSend = 0L, availableTokens = 100)
+
+      handle.outboundThrottleMode.set(bucket)
+
+      handle.write(ByteString("1234567890")) should ===(false)
+
+      wrappedHandle.writeAttempts should ===(1)
+      handle.outboundThrottleMode.get() should ===(bucket)
+    }
```
He-Pin, line 59: Consider adding two more test cases to strengthen the regression coverage:

1. **Successful write still consumes tokens** — the success path changed from `Boolean` to `Option[(ThrottleMode, ThrottleMode)]`. A test asserting that tokens are consumed and NOT refunded on a successful write would guard against a future regression that accidentally always-refunds.

2. **`Unthrottled` mode is unaffected** — with `Unthrottled` as the default throttle mode, a test that writes with `Unthrottled` set and verifies `wrappedHandle.write()` is called (and returns the correct boolean) would cover the most common code path.

Example sketch:
```scala
"consume tokens on successful write" in {
  val wrappedHandle = new TestHandle(writeResult = true)
  val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
  val bucket = TokenBucket(capacity = 100, tokensPerSecond = 0, nanoTimeOfLastSend = 0L, availableTokens = 100)
  handle.outboundThrottleMode.set(bucket)

  handle.write(ByteString("1234567890")) should ===(true)

  wrappedHandle.writeAttempts should ===(1)
  handle.outboundThrottleMode.get().asInstanceOf[TokenBucket].availableTokens should ===(90)
}

"pass through writes when Unthrottled" in {
  val wrappedHandle = new TestHandle(writeResult = true)
  val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)

  handle.write(ByteString("data")) should ===(true)
  wrappedHandle.writeAttempts should ===(1)
}
```

Non-blocking — the fix itself is correct. These tests would further harden the regression net.

remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
```diff
@@ -627,21 +627,41 @@ private[transport] final case class ThrottlerHandle(_wrappedHandle: AssociationH
...
+    }
+
+    @tailrec def refundTokens(previousBucket: ThrottleMode, consumedBucket: ThrottleMode): Unit = {
+      val currentBucket = outboundThrottleMode.get()
+      val refundedBucket =
+        if (currentBucket == consumedBucket) previousBucket
+        else
+          (currentBucket, previousBucket) match {
+            case (bucket: TokenBucket, previous: TokenBucket)
+                if bucket.capacity == previous.capacity && bucket.tokensPerSecond == previous.tokensPerSecond =>
```
He-Pin: Minor observation: `bucket.tokensPerSecond == previous.tokensPerSecond` uses exact `Double` equality. This is safe here because both values originate from the same `SetThrottle` configuration (same `TokenBucket` instance lineage), so no floating-point drift is possible between them. If a future change were to compute `tokensPerSecond` independently, consider switching to a tolerance-based comparison.

The CAS retry loop and the three-way dispatch logic are sound. Nice fix for a long-standing FIXME.

He-Pin (approved):

pjfanning (dismissed):
blocking until we have a CI run

pjfanning (commented):
build passed

pjfanning:
@He-Pin should we merge this. You have open comments.

He-Pin:
I think this pr can improve a little.

fallintoplace:
@He-Pin @pjfanning let me know if you request any further changes. Thank you for your attention.

He-Pin:
I would like to push directly on this, stay turn.

He-Pin:
## Hot-Path Performance Optimizations for `ThrottlerHandle.write`

Thanks for the PR! The token refund logic is correct. Since `ThrottlerHandle.write` is on the network I/O hot path (potentially called millions of times per second), I've applied a few performance optimizations on top of your work. Here's a summary:

### Optimizations Applied

1. **Extract local functions to companion object** — `tryConsume` and `refundTokens` defined inside `write` create closure allocations on every call. Moving them to `ThrottlerHandle` companion object makes them static methods, eliminating per-call allocation.

2. **Unthrottled fast-path** — `Unthrottled` is the default and most common mode in production. Previously it fell through to the catch-all branch, triggering an unnecessary `System.nanoTime()` and CAS. Now it explicitly bypasses both and delegates directly to `wrappedHandle.write`.

3. **Eliminate `Option` allocation** — `tryConsume` returned `Option[(ThrottleMode, ThrottleMode)]`, creating a `Some` on every successful write. Replaced with null-based signaling (`null` = failure) to avoid heap allocation in the hot path.

4. **Simplify float comparison** — Since `TokenBucket` configurations from the same `SetThrottle` lineage share identical `Double` bit patterns (no arithmetic drift), tolerance-based comparison is unnecessary. Exact `==` is simpler and correct. Added `hasSameSettings` method to `TokenBucket` to encapsulate this.

5. **Use `eq`/`ne` for reference identity** — `ThrottleMode` instances are compared by identity in the CAS loop (`eq`/`ne` instead of `==`/`!=`), avoiding the overhead of case class equality.

### Patch

Apply with `git am`:

```bash
# On your fix/throttler-token-refund branch:
curl -sL "PATCH_URL" | git am
```

Or apply the diff below directly:

<details>
<summary>Click to expand patch</summary>

```diff
diff --git a/remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala b/remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
index dba9cc13e6..4433981692 100644
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/ThrottlerTransportAdapter.scala
@@ -146,6 +146,9 @@ object ThrottlerTransportAdapter {
 
     private def tokensGenerated(nanoTimeOfSend: Long): Int =
       (TimeUnit.NANOSECONDS.toMillis(nanoTimeOfSend - nanoTimeOfLastSend) * tokensPerSecond / 1000.0).toInt
+
+    private[transport] def hasSameSettings(other: TokenBucket): Boolean =
+      capacity == other.capacity && tokensPerSecond == other.tokensPerSecond
   }
 
   @SerialVersionUID(1L)
@@ -613,6 +616,50 @@ private[transport] class ThrottledAssociation(
 
 }
 
+/**
+ * INTERNAL API
+ */
+private[transport] object ThrottlerHandle {
+
+  /**
+   * Attempts to reserve tokens from the token bucket via CAS loop.
+   * Returns the consumed ThrottleMode on success, null on failure (avoids Option allocation).
+   */
+  @tailrec private[transport] def tryConsume(
+      throttleRef: AtomicReference[ThrottleMode],
+      currentBucket: ThrottleMode,
+      tokens: Int): ThrottleMode = {
+    val timeOfSend = System.nanoTime()
+    val (newBucket, allow) = currentBucket.tryConsumeTokens(timeOfSend, tokens)
+    if (allow) {
+      if (throttleRef.compareAndSet(currentBucket, newBucket)) newBucket
+      else tryConsume(throttleRef, throttleRef.get(), tokens)
+    } else null
+  }
+
+  /**
+   * Refunds tokens when the wrapped write fails. Uses CAS loop to handle concurrent mode changes.
+   */
+  @tailrec private[transport] def refundTokens(
+      throttleRef: AtomicReference[ThrottleMode],
+      previousBucket: ThrottleMode,
+      consumedBucket: ThrottleMode,
+      tokens: Int): Unit = {
+    val currentBucket = throttleRef.get()
+    val refundedBucket =
+      if (currentBucket eq consumedBucket) previousBucket
+      else
+        (currentBucket, previousBucket) match {
+          case (bucket: TokenBucket, previous: TokenBucket) if bucket.hasSameSettings(previous) =>
+            bucket.copy(availableTokens = min(bucket.availableTokens + tokens, bucket.capacity))
+          case _ => currentBucket
+        }
+
+    if ((refundedBucket ne currentBucket) && !throttleRef.compareAndSet(currentBucket, refundedBucket))
+      refundTokens(throttleRef, previousBucket, consumedBucket, tokens)
+  }
+}
+
 /**
  * INTERNAL API
  */
@@ -627,43 +674,17 @@ private[transport] final case class ThrottlerHandle(_wrappedHandle: AssociationH
   override def write(payload: ByteString): Boolean = {
     val tokens = payload.length
 
-    @tailrec def tryConsume(currentBucket: ThrottleMode): Option[(ThrottleMode, ThrottleMode)] = {
-      val timeOfSend = System.nanoTime()
-      val (newBucket, allow) = currentBucket.tryConsumeTokens(timeOfSend, tokens)
-      if (allow) {
-        if (outboundThrottleMode.compareAndSet(currentBucket, newBucket)) Some((currentBucket, newBucket))
-        else tryConsume(outboundThrottleMode.get())
-      } else None
+    outboundThrottleMode.get() match {
+      case Blackhole   => true
+      case Unthrottled => wrappedHandle.write(payload)
+      case bucket      =>
+        val consumedBucket = ThrottlerHandle.tryConsume(outboundThrottleMode, bucket, tokens)
+        if (consumedBucket ne null) {
+          val written = wrappedHandle.write(payload)
+          if (!written) ThrottlerHandle.refundTokens(outboundThrottleMode, bucket, consumedBucket, tokens)
+          written
+        } else false
     }
-
-    @tailrec def refundTokens(previousBucket: ThrottleMode, consumedBucket: ThrottleMode): Unit = {
-      val currentBucket = outboundThrottleMode.get()
-      val refundedBucket =
-        if (currentBucket == consumedBucket) previousBucket
-        else
-          (currentBucket, previousBucket) match {
-            case (bucket: TokenBucket, previous: TokenBucket)
-                if bucket.capacity == previous.capacity && bucket.tokensPerSecond == previous.tokensPerSecond =>
-              bucket.copy(availableTokens = min(bucket.availableTokens + tokens, bucket.capacity))
-            case _ => currentBucket
-          }
-
-      if (refundedBucket != currentBucket && !outboundThrottleMode.compareAndSet(currentBucket, refundedBucket))
-        refundTokens(previousBucket, consumedBucket)
-    }
-
-    outboundThrottleMode.get match {
-      case Blackhole  => true
-      case bucket @ _ =>
-        tryConsume(bucket) match {
-          case Some((previousBucket, consumedBucket)) =>
-            val written = wrappedHandle.write(payload)
-            if (!written) refundTokens(previousBucket, consumedBucket)
-            written
-          case None => false
-        }
-    }
-
   }
 
   override def disassociate(reason: String, log: LoggingAdapter): Unit = {
diff --git a/remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala b/remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
index 056934f113..e6f07196a5 100644
--- a/remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
+++ b/remote/src/test/scala/org/apache/pekko/remote/transport/ThrottlerHandleSpec.scala
@@ -15,7 +15,7 @@ import scala.concurrent.Promise
 import org.apache.pekko
 import pekko.actor.Address
 import pekko.remote.transport.AssociationHandle.HandleEventListener
-import pekko.remote.transport.ThrottlerTransportAdapter.TokenBucket
+import pekko.remote.transport.ThrottlerTransportAdapter.{ Blackhole, TokenBucket, Unthrottled }
 import pekko.testkit.PekkoSpec
 import pekko.util.ByteString
 
@@ -48,5 +48,71 @@ class ThrottlerHandleSpec extends PekkoSpec {
       wrappedHandle.writeAttempts should ===(1)
       handle.outboundThrottleMode.get() should ===(bucket)
     }
+
+    "consume tokens when the wrapped handle writes" in {
+      val wrappedHandle = new TestHandle(writeResult = true)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+      val bucket = TokenBucket(capacity = 100, tokensPerSecond = 0, nanoTimeOfLastSend = 0L, availableTokens = 100)
+
+      handle.outboundThrottleMode.set(bucket)
+
+      handle.write(ByteString("1234567890")) should ===(true)
+
+      wrappedHandle.writeAttempts should ===(1)
+      handle.outboundThrottleMode.get().asInstanceOf[TokenBucket].availableTokens should ===(90)
+    }
+
+    "pass through writes when unthrottled" in {
+      Seq(true, false).foreach { writeResult =>
+        val wrappedHandle = new TestHandle(writeResult)
+        val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+
+        handle.write(ByteString("data")) should ===(writeResult)
+
+        wrappedHandle.writeAttempts should ===(1)
+        handle.outboundThrottleMode.get() should ===(Unthrottled)
+      }
+    }
+
+    "return true without calling wrapped handle when blackholed" in {
+      val wrappedHandle = new TestHandle(writeResult = true)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+
+      handle.outboundThrottleMode.set(Blackhole)
+
+      handle.write(ByteString("data")) should ===(true)
+
+      wrappedHandle.writeAttempts should ===(0)
+      handle.outboundThrottleMode.get() should ===(Blackhole)
+    }
+
+    "return false when token bucket has insufficient tokens" in {
+      val wrappedHandle = new TestHandle(writeResult = true)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+      val bucket = TokenBucket(capacity = 100, tokensPerSecond = 0, nanoTimeOfLastSend = 0L, availableTokens = 5)
+
+      handle.outboundThrottleMode.set(bucket)
+
+      handle.write(ByteString("1234567890")) should ===(false)
+
+      wrappedHandle.writeAttempts should ===(0)
+      handle.outboundThrottleMode.get() should ===(bucket)
+    }
+
+    "refund tokens when write fails with concurrent throttle mode change" in {
+      val wrappedHandle = new TestHandle(writeResult = false)
+      val handle = ThrottlerHandle(wrappedHandle, system.deadLetters)
+      val bucket = TokenBucket(capacity = 100, tokensPerSecond = 1000.0, nanoTimeOfLastSend = 0L, availableTokens = 100)
+
+      handle.outboundThrottleMode.set(bucket)
+
+      handle.write(ByteString("1234567890")) should ===(false)
+
+      wrappedHandle.writeAttempts should ===(1)
+      val resultBucket = handle.outboundThrottleMode.get().asInstanceOf[TokenBucket]
+      resultBucket.availableTokens should ===(100)
+      resultBucket.capacity should ===(100)
+      resultBucket.tokensPerSecond should ===(1000.0)
+    }
   }
 }
```

</details>

### Test Results

All 13 tests pass (6 ThrottlerHandleSpec + 5 ThrottleModeSpec + 2 ThrottlerTransportAdapterSpec).

Feel free to incorporate these changes into your branch. Happy to discuss any of the optimization choices!

He-Pin (changes requested):
please apply the patch or come up something better

He-Pin (approved):
thank you

PR #3072, opened 2026-06-15
refactor: replace private[this] with private and = _ with explicit defaults for Scala 3.8

He-Pin:
### Motivation
Scala 3.8 deprecates `private[this]` in favor of `private`, and deprecates uninitialized vars (`= _`) in favor of explicit default values. These deprecations cause compilation warnings under Scala 3.8.

### Modification
- Replace `private[this]` and `protected[this]` with `private` and `protected` across 94 files in `actor`, `actor-typed`, `bench-jmh`, `cluster-sharding-typed`, `persistence`, `remote`, and `stream` modules
- Replace `= _` initializers with explicit defaults (`= null`, `= None`, `= 0`, `= 0L`, `= false`, etc.) in affected files
- No behavioral changes; these are purely mechanical deprecation fixes

### Result
Code compiles cleanly under Scala 3.8 without `private[this]` or uninitialized var deprecation warnings. No runtime behavior changes.

### Tests
- Not run - mechanical refactoring, compilation test sufficient

### References
None - proactive Scala 3.8 forward compatibility

PR #3063, opened 2026-06-13
perf: replace ArrayList consumer wheel with LongMap for O(1) keyed removal

He-Pin:
### Motivation

BroadcastHub's `findAndRemoveConsumer` used `ArrayList.removeIf` which is O(k) per event with lambda allocation on every call. In high-fan-out scenarios (thousands of consumers clustered in the same wheel slot), this creates a producer backpressure bottleneck: the head can only advance after the head slot is empty, and draining a large slot requires k linear scans each of O(k) cost.

Inspired by [akkadotnet/akka.net#8264](https://github.com/akkadotnet/akka.net/pull/8264) which replaced `ImmutableList<Consumer>[]` with `Dictionary<long, Consumer>[]`.

### Modification

Replace `Array[java.util.ArrayList[Consumer]]` with `Array[LongMap[Consumer]]` keyed by `Consumer.id`:

- **O(1) keyed removal**: `LongMap.getOrNull` + `-=` (two primitive hash lookups) replaces `ArrayList.removeIf` (O(k) linear scan with lambda allocation)
- **Zero heap allocation per cycle**: `getOrNull` returns raw `V` (not `Option[V]`), avoiding `Some`/`None` allocation on the hot path
- **Zero Long boxing**: `LongMap` stores primitive `long` keys in contiguous open-addressing arrays, unlike `HashMap[Long, _]` which boxes to `java.lang.Long`
- **Lazy slot allocation**: empty slots are `null` (no backing map), reducing baseline memory from ~40 bytes × `bufferSize×2` to 0
- **GC-friendly release**: when a slot drains to zero consumers, the `LongMap` reference is nulled for immediate GC
- **Null guards**: `Advance`/`NeedWakeup` handlers now guard against `findAndRemoveConsumer` returning null, preventing latent NPE
- **Skipped empty slots**: `onUpstreamFailure` and `wakeupIdx` skip null (empty) slots during iteration

### Result

| Operation | Before (`ArrayList`) | After (`LongMap`) |
|-----------|---------------------|-------------------|
| Remove by ID | O(k) linear scan + lambda alloc | **O(1)**, zero alloc |
| Add | O(1) amortized | O(1) amortized |
| Slot empty check | O(1) | O(1) (null short-circuit) |
| Empty slot memory | ~40 bytes × bufferSize×2 | **0** (null) |
| GC pressure per cycle | Long boxing + lambda + Entry | **0** |

No public API changes. Binary compatibility preserved.

### Benchmark Results

[...]

PR #3113, opened 2026-06-20
Revert "perf: replace ArrayList consumer wheel with LongMap for O(1) keyed removal (#3063)"

He-Pin:
### Motivation

PR #3063 replaced the `BroadcastHub` consumer wheel's `ArrayList` with `scala.collection.mutable.LongMap` for O(1) keyed removal. The PR description credited [akkadotnet/akka.net#8264](https://github.com/akkadotnet/akka.net/pull/8264) as its inspiration.

Although the Pekko implementation was written from scratch (using `LongMap` rather than `Dictionary`/`HashMap`), there is concern that the original idea and design framing — replacing the per-slot consumer collection with a keyed map for O(1) removal — came from reading the akka.net PR description rather than being derived independently from the Pekko codebase. Until the licensing and provenance question is fully resolved, revert the change so it can be re-done cleanly from scratch, with independent motivation and design.

### Modification

Revert the squash-merged commit `28d61c53684dcd22302560220698d74a7c5da789` (PR #3063):

- `BroadcastHub` consumer wheel: `Array[LongMap[Consumer]]` back to `Array[java.util.ArrayList[Consumer]]`; keyed lookup uses `ArrayList.removeIf` again.
- `findAndRemoveConsumer`, `addConsumer`, `wakeupIdx`, `onUpstreamFailure`, and the `Advance`/`NeedWakeup` event handlers are restored to their pre-#3063 shape.
- `BroadcastHubBenchRunner` (standalone benchmark runner added by #3063) is removed.
- `BroadcastHubBenchmark` is restored to its prior `@Param` configuration (`64`, `256`) and single `broadcast` method.
- The two high-consumer `HubSpec` tests added by #3063 (`deliver all elements in order to many consumers`, `handle many consumers when some cancel mid-stream`) are removed.

The post-#3072 `private` field style (no `private[this]`) is preserved — only the LongMap data-structure change from #3063 is undone.

### Result

`BroadcastHub` behavior returns to its state before #3063. No public API or binary compatibility change vs. the current `main` branch — only the internal LongMap-based wheel introduced by #3063 is undone. The optimization can be re-done from scratch in a follow-up PR once the provenance question is settled.

### Tests

- `sbt "stream/compile"` → success
- `sbt "stream-tests/Test/testOnly *HubSpec"` → 48 passed, 0 failed
- `sbt "stream/mimaReportBinaryIssues"` → no issues
- `git diff --check origin/main..HEAD` → clean

### References

Refs #3063, Refs [akkadotnet/akka.net#8264](https://github.com/akkadotnet/akka.net/pull/8264) — provenance review pending.


He-Pin:
@pjfanning let's merge this to reduce the maintain burden.

He-Pin:
He is fine

PR #3114, opened 2026-06-20
fix(stream): prevent double materialization of SourceRef and SinkRef

### LINKED ISSUE #3106: SourceRef.source and SinkRef.sink() create new instances on each access
### Motivation

`SourceRefImpl.source` and `SinkRefImpl.sink()` create a new `SourceRefStageImpl` / `SinkRefStageImpl` instance on every access. If a user accidentally materializes the returned Source/Sink more than once, multiple stage instances compete for the same partner handshake, causing the second materialization to fail with a confusing error.

### Current behavior

`SourceRefImpl.scala:32-33`:
```scala
def source: Source[T, NotUsed] =
  Source.fromGraph(new SourceRefStageImpl(OptionVal.Some(initialPartnerRef)))
```

`SinkRefImpl.scala:33-34`:
```scala
override def sink(): Sink[In, NotUsed] =
  Sink.fromGraph(new SinkRefStageImpl[In](OptionVal.Some(initialPartnerRef)))
```

### Proposed fix

Option A: Cache with `lazy val` (requires changing from `case class` to `class`):
```scala
private[stream] final class SourceRefImpl[T](initialPartnerRef: ActorRef) extends SourceRef[T] {
  lazy val source: Source[T, NotUsed] =
    Source.fromGraph(new SourceRefStageImpl(OptionVal.Some(initialPartnerRef))).mapMaterializedValue(_ => NotUsed)
  // ... equals/hashCode/toString to preserve case class semantics
}
```

Option B: Add a clear error message when double-materialization is attempted, rather than a confusing handshake failure.

He-Pin:
### Motivation

`SourceRefImpl.source` and `SinkRefImpl.sink()` create a new `SourceRefStageImpl` / `SinkRefStageImpl` instance on every access. If a user accidentally materializes the returned `Source`/`Sink` more than once, multiple stage instances compete for the same partner handshake, causing the second materialization to fail with a confusing error message (typically a handshake timeout or `IllegalStateException in state AwaitingSubscription`).

Fixes #3106

### Modification

Add an `AtomicBoolean` guard per `SourceRefImpl` / `SinkRefImpl` instance. When the backing stage is materialized, the guard is atomically claimed via `compareAndSet(false, true)`. A second materialization attempt throws a clear `IllegalStateException`:

```
This SourceRef has already been materialized. SourceRef is single-use: calling .source and materializing it more than once is not supported.
```

The guard is passed as an optional `owner` parameter to `SourceRefStageImpl` / `SinkRefStageImpl` (defaulting to `null` for the initial stage created by `StreamRefs.sourceRef()` / `StreamRefs.sinkRef()`, which are always single-materialization by design).

**Key design decisions:**
- `AtomicBoolean` per instance (not `AtomicBooleanFieldUpdater`, which does not exist in the JDK; `AtomicIntegerFieldUpdater` from a Scala companion object fails due to JVM-level `private` access restrictions on case class body fields)
- The `owner` constructor parameter defaults to `null`, so `StreamRefs.scala` call sites remain unchanged
- `SourceRefImpl` / `SinkRefImpl` are `private[stream]` and `@InternalApi`, so the new field does not affect public binary compatibility

### Result

Double materialization of `SourceRef` or `SinkRef` now fails fast with a descriptive error, instead of the previous opaque handshake failure.

### Tests

```
sbt "stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.StreamRefsSpec"
# 25/25 passed
```

Two new tests added:
- `A SourceRef must "not allow materializing multiple times"` — verifies `IllegalStateException` on second `.source.run()`
- `A SinkRef must "not allow materializing multiple times"` — updated from previous silent-cancellation behavior to expect the new `IllegalStateException`

### References

Fixes #3106

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/StreamRefsSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/StreamRefsSpec.scala
@@ -17,7 +17,10 @@ import scala.collection.immutable
 import scala.concurrent.{ Await, Future }
 import scala.concurrent.Promise
 import scala.concurrent.duration._
+import scala.util.Try
 import scala.util.control.NoStackTrace
+
+import java.util.concurrent.CountDownLatch
 
 import org.apache.pekko
 import pekko.{ Done, NotUsed }
@@ -26,7 +29,6 @@ import pekko.actor.Status.Failure
 import pekko.pattern._
 import pekko.stream._
 import pekko.stream.impl.streamref.{ SinkRefImpl, SourceRefImpl }
-import pekko.stream.testkit.TestPublisher
 import pekko.stream.testkit.Utils.TE
 import pekko.stream.testkit.scaladsl._
 import pekko.testkit.{ PekkoSpec, TestKit, TestProbe }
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

PR #3116, opened 2026-06-20
fix: keep lazy dispatch state on VarHandle

He-Pin:
### Motivation
PR #3035 introduced a VarHandle-backed `@volatile var state` for LazyDispatch scheduling state. The review discussion raised concern around the field and the access mode used through the VarHandle.

### Modification
Keep the static VarHandle and instance `Int` state, but use explicit `getVolatile` / `setVolatile` accessors with `compareAndSet` instead of plain VarHandle `get` / `set`.

This avoids per-instance `AtomicInteger` allocation while making the state access mode explicit and conservative.

### Result
LazyDispatch keeps the same IDLE/SCHEDULED scheduling protocol and uses volatile state access for the scheduling election and IDLE publish/recheck path.

### Tests
- `scalafmt --mode diff-ref=origin/main --non-interactive`
- `scalafmt --list --mode diff-ref=origin/main --non-interactive`
- `sbt "+stream / compile"`
- `sbt "stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.StageActorRefSpec"` (11/11)
- `git diff --check`

### References
Refs #3035

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/stage/GraphStage.scala
@@ -13,8 +13,8 @@
 
 package org.apache.pekko.stream.stage
 
-import java.util.Spliterator
 import java.lang.invoke.{ MethodHandles, VarHandle }
+import java.util.Spliterator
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #3120, opened 2026-06-22
Fix a few broken Scaladoc links reported by unidoc

henricook:
### Motivation
`sbt unidoc` emits a number of "Could not find any member to link for" Scaladoc warnings (#353). A few of these are plain broken links rather than the trickier private or ambiguous-target cases:

- `Tcp`'s `SimpleWriteCommand` and `CompoundWrite` Scaladoc link `[[WriteFile]]`, but that write command was renamed to `WritePath`.
- `ShardingEnvelope`'s `@throws` tag wraps the exception type in `[[ ]]`. Scaladoc does not parse `[[ ]]` inside `@throws`, so the link is reported as missing.

### Modification
- `Tcp.scala`: update two Scaladoc comments from `[[WriteFile]]` to `[[WritePath]]`.
- `ShardingMessageExtractor.scala`: remove the `[[ ]]` around the `@throws` type so it matches the `@throws <fully.qualified.type> <description>` form used elsewhere in the codebase.

### Result
Three unidoc link warnings are resolved. No API or behaviour change.

### Tests
Not run - docs only. Confirmed locally with `sbt unidoc` that the three warnings no longer appear and no new warnings were introduced; `sbt scalafmtCheck` passes for both touched modules.

### References
Refs #353

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/io/Tcp.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/Tcp.scala
@@ -316,7 +316,7 @@ object Tcp extends ExtensionId[TcpExt] with ExtensionIdProvider {
   }
 
   /**
-   * Common supertype of [[Write]] and [[WriteFile]].
+   * Common supertype of [[Write]] and [[WritePath]].
    */
   sealed abstract class SimpleWriteCommand extends WriteCommand {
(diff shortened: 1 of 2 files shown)


He-Pin (approved):

He-Pin:
@henricook Thanks

Issue #3136, opened 2026-06-22
Modernize codebase with JDK 9-17 APIs？


He-Pin:
### Motivation

Apache Pekko 2.0 raised the minimum JDK version to 17. Some parts of the codebase still use pre-JDK 9 patterns that can be replaced with cleaner, safer, and sometimes more performant JDK 9-17 APIs.

---

### PRs

| PR | Title | JDK | Category |
|----|-------|-----|----------|
| #3137 | Use CompletableFuture.orTimeout in FutureTimeoutSupport | 9 | API |
| #3138 | Simplify Collections.unmodifiableSet(emptySet()) to Set.of() | 9 | API |
| #3139 | Use ByteBuffer fluent API chains with JDK 9 covariant returns | 9 | API |
| #3140 | Update Scaladoc examples from Arrays.asList to List.of | 9 | Docs |
| #3141 | Use StackWalker for stack trace inspection in TestKitUtils | 9 | API |
| #3142 | Remove redundant Float cast in CountMinSketch pattern match | 16 | Fix |
| #3143 | Use Files.readString() in PemManagersProvider | 11 | API |
| #3144 | Use String.formatted() instead of String.format() | 15 | API |
| #3145 | Use HexFormat for hex encoding instead of manual formatting | 17 | API |
| #3146 | Replace manual stream copy loops with transferTo/readAllBytes | 9 | API |
| #3147 | Use StandardCharsets.UTF_8 with URLEncoder/URLDecoder | 10 | API |
| #3149 | Add Thread.onSpinWait() to CAS spin loops | 9 | Perf |
| #3150 | Use Arrays.copyOf/copyOfRange instead of manual arraycopy | — | Perf |
| #3151 | Replace sun.reflect.Reflection with StackWalker in Reflect | 9 | API |
| #3152 | Use ProcessHandle.current().pid() instead of RuntimeMXBean hack | 9 | API |
| #3153 | Use RandomGenerator interface instead of java.util.Random | 17 | API |

### Not changed

- `Collections.emptyList/emptyMap/emptySet` — JVM singletons, zero allocation.
- `ByteBufferCleaner.java` — `Unsafe.invokeCleaner` is the standard approach.
- `AtomicReferenceFieldUpdater` in Dispatcher — maintainer decision.
- `failedCompletionStage` — deprecated in 2.0.0, retained for deprecation period.
- `Objects.checkFromIndexSize` — custom implementation has better error messages.
- Protobuf generated files — auto-generated.

### References

[...]

PR #3137, opened 2026-06-23
refactor: use CompletableFuture.orTimeout in FutureTimeoutSupport

He-Pin:
### Motivation

`FutureTimeoutSupport.timeoutCompletionStage` was `@deprecated("Use CompletableFuture#orTimeout instead.", "2.0.0")` but still used a manual `Scheduler`-based implementation internally with `scheduleOnce` + `completeExceptionally(new TimeoutException(...))`.

### Modification

Rewrite `timeoutCompletionStage` to delegate to `CompletableFuture.orTimeout()`. A `handle()` wrapper provides a consistent timeout message (`"Timeout of $duration expired"`) across JDK versions — JDK 25's `orTimeout` produces a null `TimeoutException` message, unlike JDK 17 which included a descriptive message. The `Scheduler` parameter is retained for API compatibility but no longer used.

### Result

- Simpler implementation (~5 fewer lines)
- No `Scheduler` dependency for timeout scheduling (uses JDK shared `ForkJoinPool`)
- Consistent timeout exception messages across JDK versions
- `orTimeout` still cancels the underlying future on timeout (same behavior as before)

### Tests

- `sbt "actor-tests / Test / testOnly org.apache.pekko.pattern.PatternsTest"` — 33/33 passed

### References

Refs #3136

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
+++ b/actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
@@ -134,23 +134,18 @@ trait FutureTimeoutSupport {
       catch {
         case NonFatal(t) => CompletableFuture.failedStage(t)
       }
-    if (stage.toCompletableFuture.isDone) {
-      stage
-    } else {
-      val p = new CompletableFuture[T]
-      val timeout = using.scheduleOnce(duration,
-        () => {
-          p.completeExceptionally(new TimeoutException(s"Timeout of $duration expired"))
-          stage.toCompletableFuture.cancel(true)
-          ()
-        })
-      stage.handle[Unit]((v: T, ex: Throwable) => {
-        timeout.cancel()
-        if (v != null) p.complete(v)
-        if (ex ne null) p.completeExceptionally(ex)
+    val millis = if (duration.isZero || duration.isNegative) 0L else duration.toMillis
+    val timeoutMessage = s"Timeout of $duration expired"
+    stage.toCompletableFuture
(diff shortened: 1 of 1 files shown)


actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -134,23 +134,18 @@ trait FutureTimeoutSupport {
       catch {
         case NonFatal(t) => CompletableFuture.failedStage(t)
       }
-    if (stage.toCompletableFuture.isDone) {
```
pjfanning, line 137: do we not need the isDone check now?

actor/src/main/scala/org/apache/pekko/pattern/FutureTimeoutSupport.scala
```diff
@@ -134,23 +134,18 @@ trait FutureTimeoutSupport {
...
-          p.completeExceptionally(new TimeoutException(s"Timeout of $duration expired"))
-          stage.toCompletableFuture.cancel(true)
-          ()
-        })
-      stage.handle[Unit]((v: T, ex: Throwable) => {
-        timeout.cancel()
-        if (v != null) p.complete(v)
-        if (ex ne null) p.completeExceptionally(ex)
+    val millis = if (duration.isZero || duration.isNegative) 0L else duration.toMillis
+    val timeoutMessage = s"Timeout of $duration expired"
```
pjfanning: this line to create timeoutMessage should be moved to the `case _: TimeoutException` match below.

pjfanning (approved):
lgtm

PR #3138, opened 2026-06-23
refactor: simplify Collections.unmodifiableSet(emptySet()) to Set.of()

He-Pin:
### Motivation

`StreamOfStreams.scala` uses `Collections.unmodifiableSet(Collections.emptySet[Any])` — two method calls to create an immutable empty set.

### Modification

Replace with `java.util.Set.of[Any]()` — a single JDK 9 factory call that already returns an immutable empty set. Remove the unused `java.util.Collections` import.

### Result

Cleaner code: one call instead of two. Both approaches produce immutable empty sets with identical semantics.

### Tests

- `sbt "stream/compile"` — passed

### References

Refs #3136

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
@@ -13,6 +13,7 @@
 
 package org.apache.pekko.stream.impl.fusing
 
+import java.util.Collections
 import java.util.concurrent.atomic.AtomicReference
 
 import scala.annotation.{ nowarn, tailrec }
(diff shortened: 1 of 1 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -399,7 +398,7 @@ import pekko.util.OptionVal
       lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
       private val activeSubstreamsMap = new java.util.HashMap[Any, SubstreamSource]()
       private val closedSubstreams =
-        if (allowClosedSubstreamRecreation) Collections.unmodifiableSet(Collections.emptySet[Any])
+        if (allowClosedSubstreamRecreation) java.util.Set.of[Any]()
```
pjfanning: -1 from me `java.util.Set.of[Any]()` returns a new immutable set - I think we should keep `Collections.emptySet[Any]` but remove the `Collections.unmodifiableSet` call because `Collections.emptySet[Any]` is immutable.

`Collections.emptySet[Any]` is a singleton.

pjfanning (approved):
lgtm

PR #3140, opened 2026-06-23
docs: update Scaladoc examples from Arrays.asList to List.of

He-Pin:
### Motivation

Scaladoc examples in the Stream Java DSL use `Arrays.asList(...)` which creates a mutable fixed-size list backed by an array.

Also updates production code in `org.apache.pekko.stream.javadsl.Graph` to use `List.of` to create an immutable List.

### Modification

Replace with `List.of(...)` (JDK 9 factory) which returns a truly immutable list, consistent with modern Java best practices. Also updates the executable `Graph.scala` code where `util.Arrays.asList` was used.

### Result

Documentation examples use idiomatic JDK 9+ API.

### Tests

CI

### References

Refs #3136

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
@@ -726,7 +726,7 @@ object GraphDSL extends GraphCreate {
       : Graph[S, java.util.List[M]] = {
     require(!graphs.isEmpty, "The input list must have one or more Graph elements")
     val gbuilder = builder[java.util.List[M]]()
-    val toList = (m1: M) => new util.ArrayList(util.Arrays.asList(m1))
+    val toList = (m1: M) => new util.ArrayList(util.List.of(m1))
     val combine = (s: java.util.List[M], m2: M) => {
       val newList = new util.ArrayList(s)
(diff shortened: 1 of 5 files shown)


stream/src/main/scala/org/apache/pekko/stream/javadsl/Graph.scala
```diff
@@ -726,7 +726,7 @@ object GraphDSL extends GraphCreate {
       : Graph[S, java.util.List[M]] = {
     require(!graphs.isEmpty, "The input list must have one or more Graph elements")
     val gbuilder = builder[java.util.List[M]]()
-    val toList = (m1: M) => new util.ArrayList(util.Arrays.asList(m1))
+    val toList = (m1: M) => new util.ArrayList(util.List.of(m1))
```
pjfanning: do we need the mutable ArrayList wrapper?

He-Pin: no I think.

pjfanning: can we change this then? to just `val toList = (m1: M) => util.List.of(m1)`

pjfanning (approved):
lgtm

PR #3141, opened 2026-06-23
refactor: use StackWalker for stack trace inspection in TestKitUtils

He-Pin:
### Motivation

`TestKitUtils.testNameFromCallStack` uses `Thread.currentThread.getStackTrace` which eagerly constructs the full `StackTraceElement[]` array including method names, file names, and line numbers — but only class names are needed.

### Modification

Replace with `StackWalker` (JDK 9) which avoids creating full `StackTraceElement` objects, only extracting class names via the lighter `StackFrame.getClassName()`. The existing filtering logic (`dropWhile` + `startsWith` matching) is preserved unchanged.

This follows the pattern already established by `LoggerClass.scala` in the same codebase.

### Result

More efficient stack inspection for test infrastructure.

### Tests

- `sbt "actor-tests / Test / testOnly org.apache.pekko.actor.ActorSystemSpec"` — 20/20 passed

### References

Refs #3136

### DIFF
--- a/testkit/src/main/scala/org/apache/pekko/testkit/TestKitUtils.scala
+++ b/testkit/src/main/scala/org/apache/pekko/testkit/TestKitUtils.scala
@@ -13,7 +13,9 @@
 
 package org.apache.pekko.testkit
 
+import java.lang.StackWalker.StackFrame
 import java.lang.reflect.Modifier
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #3142, opened 2026-06-23
fix: remove redundant Float cast in CountMinSketch pattern match

He-Pin:
### Motivation

`CountMinSketch.java` line 181 uses `(Float) o` cast inside an `instanceof Float f` pattern match. The pattern variable `f` is already bound to the Float value, making the explicit cast redundant.

### Modification

Use the pattern variable `f` directly instead of `(Float) o`.

### Result

Cleaner code, no functional change.

### Tests

- `sbt "remote/compile"` — passed

### References

Refs #3136


pjfanning (approved):
lgtm

PR #3144, opened 2026-06-23
refactor: use String.formatted() instead of String.format()

He-Pin:
### Motivation

`String.format(fmt, args)` is a static method call. JDK 15 added `String.formatted(args)` as a fluent instance method on the format string itself.

### Modification

Replace 2 occurrences of `String.format(fmt, args)` with `fmt.formatted(args)`:
- `UnsynchronizedByteArrayInputStream.java`: range check error message
- `AeronErrorLog.java`: Aeron error logging

### Result

More readable fluent API style, functionally identical.

### Tests

- `sbt "actor/compile" "remote/compile"` — passed

### References

Refs #3136


pjfanning (approved):
lgtm

PR #3145, opened 2026-06-23
refactor: use HexFormat for hex encoding instead of manual formatting

He-Pin:
### Motivation

Manual per-byte hex encoding with `"%02x".format()` is verbose and allocates a `String` per byte. JDK 17's `HexFormat` formats entire byte arrays in one call with zero intermediate allocations.

### Modification

- `VectorClock.scala`: Replace `digester.digest.map("%02x".format).mkString` with `HexFormat.of().formatHex(digester.digest())`
- `TcpFraming.scala`: Replace `.map("%02x".format).mkString(" ")` with `HexFormat.ofDelimiter(" ").formatHex(bytes.toArray)`

### Result

More concise code, fewer allocations (one `String` vs N+1 `String`s), standard JDK 17 API for hex encoding.

### Tests

- `sbt "cluster/compile" "remote/compile"` — passed

### References

Refs #3136

### DIFF
--- a/cluster/src/main/scala/org/apache/pekko/cluster/VectorClock.scala
+++ b/cluster/src/main/scala/org/apache/pekko/cluster/VectorClock.scala
@@ -37,10 +37,12 @@ private[cluster] object VectorClock {
 
     def fromHash(hash: String): Node = hash
 
+    private val hexFormat = java.util.HexFormat.of()
+
     private def hash(name: String): String = {
       val digester = MessageDigest.getInstance("MD5")
       digester.update(name.getBytes(StandardCharsets.UTF_8))
-      java.util.HexFormat.of().formatHex(digester.digest())
+      hexFormat.formatHex(digester.digest())
     }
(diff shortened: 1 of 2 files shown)


cluster/src/main/scala/org/apache/pekko/cluster/VectorClock.scala
```diff
@@ -40,9 +40,7 @@ private[cluster] object VectorClock {
     private def hash(name: String): String = {
       val digester = MessageDigest.getInstance("MD5")
       digester.update(name.getBytes(StandardCharsets.UTF_8))
-      digester.digest.map { h =>
-        "%02x".format(0xFF & h)
-      }.mkString
+      java.util.HexFormat.of().formatHex(digester.digest())
```
pjfanning: any benefit to caching a HexFormat instance singleton?

pjfanning (approved):
lgtm

PR #3148, opened 2026-06-23
refactor: remove deprecated failedCompletionStage, use CompletableFuture.failedStage

He-Pin:
### Motivation

`Futures.failedCompletionStage` was `@deprecated("Use CompletableFuture#failedStage instead.", "2.0.0")` and used the dangerous `obtrudeException` API which can corrupt other futures sharing the same underlying `CompletableFuture`. `CompletableFuture.failedStage` (JDK 9) is the correct replacement.

### Modification

- Remove `failedCompletionStage` method from `Futures` object in `Future.scala`
- Replace internal call in `CompletionStages.reduce` with `CompletableFuture.failedStage`
- Replace test call sites with `CompletableFuture.failedStage`
- Remove unnecessary `@SuppressWarnings("deprecation")` from 2 test methods
- Remove unused `CompletableFuture` import from `Future.scala`

### Result

Eliminates `obtrudeException` usage, removes deprecated API.

### Tests

- `sbt "actor-tests / Test / testOnly org.apache.pekko.pattern.PatternsTest org.apache.pekko.dispatch.CompletionStagesTests"` — 40/40 passed

### References

Refs #3136

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/dispatch/CompletionStagesTests.java
+++ b/actor-tests/src/test/java/org/apache/pekko/dispatch/CompletionStagesTests.java
@@ -41,7 +41,7 @@ public class CompletionStagesTests {
         assertEquals(42, Await.result(scalaFuture, timeout).intValue());
         //failed
         assertThrows(RuntimeException.class, () -> {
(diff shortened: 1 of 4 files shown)


He-Pin:
Closing: failedCompletionStage was only deprecated in 2.0.0 and should be retained for the deprecation period.

PR #3150, opened 2026-06-23
perf: use Arrays.copyOf/copyOfRange instead of manual new Array + arraycopy

He-Pin:
### Motivation

Several places allocate a new array then immediately copy from an existing array using `System.arraycopy`. `java.util.Arrays.copyOf` and `copyOfRange` combine allocation and copy in a single call that the JIT can intrinsify more effectively.

### Modification

- `ByteString.scala`: 3 sites (`clearTemp`, `resizeTemp`, `fromArray`) — replace `new Array[Byte]` + `System.arraycopy` with `java.util.Arrays.copyOf` / `copyOfRange`
- `ImmutableLongMap.scala`: replace full array copy with `values.clone()`
- `SnapshotSerializer.scala`: replace partial copy with `java.util.Arrays.copyOfRange`

### Result

Fewer lines, potentially faster due to JIT intrinsification of the combined allocate+copy operation. `ByteString` builder operations benefit most as they are in frequently-called paths.

### Tests

- `sbt "actor/compile" "remote/compile" "persistence/compile"` — passed

### References

Refs #3136

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -2795,7 +2795,7 @@ final class ByteStringBuilder extends Builder[Byte, ByteString] {
   }
 
   private def resizeTemp(size: Int): Unit = {
-    val newtemp = java.util.Arrays.copyOf(_temp, size)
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

pjfanning (changes requested):
tests are broken

pjfanning (approved):
lgtm

PR #3151, opened 2026-06-23
refactor: replace sun.reflect.Reflection with StackWalker in Reflect

He-Pin:
### Motivation

`Reflect.getCallerClass` uses `sun.reflect.Reflection.getCallerClass` via reflection — a JDK 8 internal API that was moved to `jdk.internal.reflect.Reflection` in JDK 9+. On JDK 17+ without `--add-opens` for `jdk.internal.reflect`, this always falls back to `None`, making the `findClassLoader` fallback path dead code.

### Modification

Replace with `java.lang.StackWalker` (JDK 9) with `RETAIN_CLASS_REFERENCE` option — the official, supported API for accessing caller class information. The API contract is preserved: given a frame depth index, returns the `Class` at that depth (or `null` if beyond stack).

### Result

- Official JDK API instead of internal reflection hack
- `StackWalker` was already established in the codebase (`LoggerClass.scala`, `TestKitUtils.scala`)
- The `findClassLoader` fallback now works reliably on JDK 17+

### Tests

- `sbt "actor/compile"` — passed

### References

Refs #3136

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/Reflect.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/Reflect.scala
@@ -36,15 +36,17 @@ private[pekko] object Reflect {
    * This optionally holds a function which looks N levels above itself
    * on the call stack and returns the `Class[_]` object for the code
    * executing in that stack frame. Implemented using
-   * `sun.reflect.Reflection.getCallerClass` if available, None otherwise.
+   * `java.lang.StackWalker` (JDK 9+).
    *
    * Hint: when comparing to Thread.currentThread().getStackTrace, add two levels.
    */
   val getCallerClass: Option[Int => Class[?]] = {
     try {
-      val c = Class.forName("sun.reflect.Reflection")
-      val m = c.getMethod("getCallerClass", Array(classOf[Int]): _*)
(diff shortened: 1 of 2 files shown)


actor/src/main/scala/org/apache/pekko/util/Reflect.scala
```diff
@@ -36,15 +36,17 @@ private[pekko] object Reflect {
...
    *
    * Hint: when comparing to Thread.currentThread().getStackTrace, add two levels.
    */
   val getCallerClass: Option[Int => Class[?]] = {
     try {
-      val c = Class.forName("sun.reflect.Reflection")
-      val m = c.getMethod("getCallerClass", Array(classOf[Int]): _*)
-      Some((i: Int) => m.invoke(null, Array[AnyRef](i.asInstanceOf[java.lang.Integer]): _*).asInstanceOf[Class[?]])
+      val walker = java.lang.StackWalker.getInstance(
+        java.util.Set.of(java.lang.StackWalker.Option.RETAIN_CLASS_REFERENCE))
```
pjfanning: why do we need `java.lang.StackWalker` instead of using `StackWalker`?

pjfanning (approved):
lgtm

PR #3152, opened 2026-06-23
refactor: use ProcessHandle.current().pid() instead of RuntimeMXBean hack

He-Pin:
### Motivation

`PerfFlamesSupport.scala` parses the PID from `RuntimeMXBean.getName()` which returns `"pid@hostname"` — a pre-JDK 9 hack that depends on an undocumented string format.

### Modification

Replace with `ProcessHandle.current().pid()` (JDK 9), the official API for retrieving the current process ID.

### Result

Cleaner, more reliable PID retrieval without string parsing.

### Tests

- `sbt "multi-node-testkit/compile"` — passed

### References

Refs #3136


pjfanning (approved):
lgtm

PR #3154, opened 2026-06-23
Fix comparing release against pre-release versions

### LINKED ISSUE #3125: Version.compareTo breaks comparison symmetry for release vs pre-release versions
### Motivation

`org.apache.pekko.util.Version.compareTo` contains a logic error that breaks comparison symmetry when comparing release versions against their pre-release counterparts (e.g. `1.2.0` vs `1.2.0-M1`).

The `rest` string comparison section uses two independent `if` statements instead of an `if`/`else if` chain. When a release version (`rest = ""`) is compared against a pre-release version (`rest = "M1"`), the first branch correctly sets `diff = 1` (release > pre-release), but the second branch's `else` clause then overwrites `diff` with `"".compareTo("M1")`, which returns a negative value. The result incorrectly indicates that the release version is **less than** the pre-release version.

The reverse comparison (`1.2.0-M1` vs `1.2.0`) returns the correct result, violating the `Comparable` contract: `A.compareTo(B)` and `-B.compareTo(A)` disagree.

### Impact

- `AbstractLeastShardAllocationStrategy` uses `appVersion.compareTo` for shard allocation ordering. During rolling upgrades mixing release and pre-release versions, this can cause inconsistent shard allocation decisions.
- `ManifestInfo.checkSameVersion` uses `max` on `Version` objects, which could return the wrong version.
- Any user code relying on `Version` ordering would get incorrect results in one comparison direction.

### Reproduction

```scala
Version("1.2.0").compareTo(Version("1.2.0-M1")) // returns negative (WRONG: should be positive)
Version("1.2.0-M1").compareTo(Version("1.2.0")) // returns -1 (correct)
```

The existing test suite only covers `pre-release < release` but not `release > pre-release`, which is why this asymmetry was never caught.

### Fix

The second `if` in the rest-string comparison section of `compareTo` should be `else if` so the `else` branch (string comparison of rest values) only executes when neither side has an empty rest string.

### References

This same bug was identified and fixed in the Akka.NET project: https://github.com/akkadotnet/akka.net/issues/8051

Philippus:
fixes #3125

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/util/VersionSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/VersionSpec.scala
@@ -60,6 +60,7 @@ class VersionSpec extends AnyWordSpec with Matchers {
       Version("1.2.3-M1") should !==(Version("1.2.3-M2"))
       Version("1.2-M1") should be < Version("1.2.0")
       Version("1.2.0-M1") should be < Version("1.2.0")
+      Version("1.2.0") should be > Version("1.2.0-M1")
       Version("1.2.3-M2") should be > Version("1.2.3-M1")
     }
 
--- a/actor/src/main/scala/org/apache/pekko/util/Version.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/Version.scala
@@ -169,7 +169,7 @@ final class Version(val version: String) extends Comparable[Version] {
             if (diff == 0) {
               if (rest == "" && other.rest != "")
                 diff = 1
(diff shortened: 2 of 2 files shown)


pjfanning (approved):
lgtm

He-Pin:
Why all [Scalafmt / Code is formatted (pull_request)](https://github.com/apache/pekko/actions/runs/28018389658/job/82928309112?pr=3154) is  Queued

He-Pin (approved):

PR #3155, opened 2026-06-23
use decodeString(StandardCharsets.UTF_8)

pjfanning:


He-Pin (approved):

He-Pin:
I think we need do all these kind of thing in all pekko* repos

PR #3156, opened 2026-06-23
fix: use Timers for finishConnect retry instead of raw scheduler callback

### LINKED ISSUE #3135: TcpOutgoingConnection scheduled connect-retry exceptions are silently swallowed, commander can hang
## Summary

`TcpOutgoingConnection` retries a failed `finishConnect` by scheduling a callback
on the system scheduler that calls `channelRegistry.register(channel, OP_CONNECT)`
directly, outside the actor's message loop. If that callback throws, the exception
is caught and reported by the scheduler / dispatcher and **never re-enters the
actor's supervision**, so `Tcp.CommandFailed` is never delivered and the
commander can stay stuck in the `connecting` state indefinitely.

## Code location

`actor/src/main/scala/org/apache/pekko/io/TcpOutgoingConnection.scala`

```scala
// connecting(registration, remainingFinishConnectRetries)
case ChannelConnectable =>
  reportConnectFailure {
    if (channel.finishConnect()) { ... }
    else {
      if (remainingFinishConnectRetries > 0) {
        context.system.scheduler.scheduleOnce(1.millisecond) {
          channelRegistry.register(channel, SelectionKey.OP_CONNECT)   // <-- off-loop
        }(context.dispatcher)
        context.become(connecting(registration, remainingFinishConnectRetries - 1))
      } else {
        stop(FinishConnectNeverReturnedTrueException)
      }
    }
  }
```

The surrounding `reportConnectFailure { ... }` only covers the body of the
`ChannelConnectable` handler — it does **not** cover the scheduled lambda.

## Why this is a problem

`channelRegistry.register` (the `SelectorBasedChannelRegistry` in
`SelectionHandler.scala`) dispatches a `Task` to the selector's execution
context whose `tryRun` only catches `ClosedChannelException`. Any other
throwable (e.g. `IllegalArgumentException` from an invalid ops mask,
`CancelledKeyException` if the key races with `postStop` shutdown,
`IllegalBlockingModeException`, or a JVM/NIO bug-induced `RuntimeException`)
propagates out of the task, is handed to `executionContext.reportFailure`, and
is logged — but:

- it does **not** enter `reportConnectFailure`,
- it does **not** trigger `stop(e)`,
- it does **not** emit `Tcp.CommandFailed(connect).withCause(e)` to the
  commander,
- the `TcpOutgoingConnection` actor keeps running in the `connecting` state
  with the retry counter already decremented, and the commander can wait
  forever for `Connected` / `CommandFailed`.

In practice this has been observed in the wild on akka.net under an analogous
code shape — see
[akkadotnet/akka.net#8195](https://github.com/akkadotnet/akka.net/issues/8195)
for the original report and the failure sequence. The Pekko code path has the
same structural issue: a raw scheduled callback performs I/O work that is not
wrapped in the actor's failure-reporting envelope.

## Expected behaviour

Any exception thrown by the scheduled retry should surface to the commander as
`Tcp.CommandFailed` and the connection actor should stop cleanly, exactly as
happens for exceptions thrown inside the normal `receive` path.

## Suggested fix

Schedule the retry as a self-message so it runs inside the actor's message
loop and is covered by the existing `reportConnectFailure` envelope — e.g.
introduce an internal `RetryRegister` message and, in `connecting`, handle
`case RetryRegister => reportConnectFailure { channelRegistry.register(channel, SelectionKey.OP_CONNECT) }`.
Using `Timers.StartSingleTimer` (as `TcpListener` does in the same module) is
also preferable to `context.system.scheduler.scheduleOnce`, since the pending
timer is cancelled automatically when the actor stops, removing the latent
"register against a channel that is already being torn down" window.

## Environment

- Pekko version: current `main`
- Module: `actor` (`org.apache.pekko.io.Tcp*`)
- JDK/OS: any — the bug is structural; triggering it depends on when
  `channel.register` throws a non-`ClosedChannelException` throwable.

## Reproduction notes

The akka.net report reproduces the swallowed-exception path by pre-seeding DNS
so the host resolves to both IPv4 and IPv6, forcing an IPv4-only socket, and
letting the IPv6 fallback retry throw a platform-wide `NotSupportedException`.
A symmetric Pekko test would force the scheduled `channelRegistry.register`
call to throw (e.g. by injecting a `ChannelRegistry` whose `register` throws
`IllegalArgumentException`) and assert that the commander still receives
`Tcp.CommandFailed` rather than hanging.

## References

- Original field report and analysis: https://github.com/akkadotnet/akka.net/issues/8195
- Pekko source: `actor/src/main/scala/org/apache/pekko/io/TcpOutgoingConnection.scala`,
  `actor/src/main/scala/org/apache/pekko/io/SelectionHandler.scala`

He-Pin:
### Motivation

`TcpOutgoingConnection` retries a failed `finishConnect` by scheduling a callback via `context.system.scheduler.scheduleOnce` that calls `channelRegistry.register(channel, SelectionKey.OP_CONNECT)` **outside the actor's message loop**. If that callback throws (e.g. `IllegalArgumentException`, `CancelledKeyException` racing with `postStop`, or any `RuntimeException`), the exception is caught by the scheduler/dispatcher error handler and **never re-enters the actor's supervision**. As a result, `Tcp.CommandFailed` is never delivered and the commander can stay stuck in the `connecting` state indefinitely.

See the original field report and analysis in [akkadotnet/akka.net#8195](https://github.com/akkadotnet/akka.net/issues/8195) — the Pekko code has the same structural issue.

### Modification

1. Mix in the `Timers` trait on `TcpOutgoingConnection`.
2. Replace `context.system.scheduler.scheduleOnce(1.millisecond) { channelRegistry.register(...) }` with `timers.startSingleTimer(RetryFinishConnect, RetryFinishConnect, 1.millisecond)`, which sends a private self-message processed inside the actor's message loop.
3. Handle `RetryFinishConnect` in the `connecting` state, wrapped in `reportConnectFailure` so any exception surfaces as `CommandFailed` to the commander.

Benefits over the previous approach:
- **Exceptions are caught**: `reportConnectFailure` wraps the `channelRegistry.register` call, ensuring any `NonFatal` exception is converted to `CommandFailed`.
- **Lifecycle-safe**: `Timers` automatically cancels pending timers when the actor stops, removing the latent window where a stale callback could register against a channel being torn down.
- **Consistent with actor model**: the retry runs as a self-message inside the actor's message loop, not as an external callback.

### Result

Exceptions thrown during the `finishConnect` retry path now correctly surface as `Tcp.CommandFailed` to the commander, preventing indefinite hangs in the `connecting` state.

### Tests

- `sbt 'actor-tests / Test / testOnly org.apache.pekko.io.TcpConnectionSpec'` → 35/35 passed (including new test `report failed connection when finishConnect retries are exhausted`)

### References

Fixes #3135

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/io/TcpConnectionSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/io/TcpConnectionSpec.scala
@@ -697,6 +697,48 @@ class TcpConnectionSpec extends PekkoSpec("""
       }
     }
 
+    "report failed connection when finishConnect retries are exhausted" in
+    new UnacceptedConnectionTest() {
+      override lazy val connectionActor = createConnectionActor(serverAddress = UnboundAddress)
+      run {
+        val sel = SelectorProvider.provider().openSelector()
+        try {
+          val key = clientSideChannel.register(sel, SelectionKey.OP_CONNECT | SelectionKey.OP_READ)
+          sel.select(3000)
+          key.isConnectable should ===(true)
+          connectionActor.toString // force the lazy val
+          Thread.sleep(300)
+
+          // First ChannelConnectable: finishConnect() either throws (connection refused)
+          // or returns false (connection still pending, schedules timer-based retry).
+          selector.send(connectionActor, ChannelConnectable)
+
+          // If finishConnect returned false, wait for the timer-based retry to re-register.
+          // Each retry fires after 1ms via timers.startSingleTimer (self-message).
+          var retriesObserved = 0
+          var outcome: Option[AnyRef] = None
+          val deadline = System.nanoTime() + 5.seconds.toNanos
+          while (outcome.isEmpty && System.nanoTime() < deadline) {
+            registerCallReceiver.expectNoMessage(50.millis)
+            if (registerCallReceiver.msgAvailable) {
+              registerCallReceiver.receiveWhile(100.millis) {
+                case Registration(_, ops) if ops == OP_CONNECT =>
+                  retriesObserved += 1
+                  selector.send(connectionActor, ChannelConnectable)
+              }
+            }
+            if (userHandler.msgAvailable) {
+              outcome = Some(userHandler.expectMsgType[CommandFailed])
+            }
+          }
+
+          outcome shouldBe defined
+          watch(connectionActor)
+          expectTerminated(connectionActor)
+        } finally sel.close()
+      }
+    }
+
     "close the connection when connection handler dies while connected" in new EstablishedConnectionTest() {
       run {
         watch(connectionHandler.ref)
--- a/actor/src/main/scala/org/apache/pekko/io/TcpOutgoingConnection.scala
+++ b/actor/src/main/scala/org/apache/pekko/io/TcpOutgoingConnection.scala
@@ -20,7 +20,7 @@ import scala.concurrent.duration._
 import scala.util.control.{ NoStackTrace, NonFatal }
 
 import org.apache.pekko
-import pekko.actor.{ ActorRef, ReceiveTimeout }
+import pekko.actor.{ ActorRef, ReceiveTimeout, Timers }
 import pekko.actor.Status.Failure
 import pekko.annotation.InternalApi
 import pekko.io.SelectionHandler._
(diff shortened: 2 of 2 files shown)


pjfanning (approved):
lgtm

PR #3157, opened 2026-06-23
Delay unfoldResourceAsync close until pending read completes

### LINKED ISSUE #3134: UnfoldResourceSourceAsync: postStop races with in-flight readData
### Description

`UnfoldResourceSourceAsync` calls `close(resource)` in `postStop()` without waiting for any in-flight `readData(resource)` `Future` to complete. This can cause `close()` and `readData()` to run concurrently on the same resource.

### Location

`stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSourceAsync.scala:115-118`

```scala
override def postStop(): Unit = maybeResource match {
  case OptionVal.Some(resource) => close(resource)
  case _                        => // do nothing
}
```

### Race condition

1. `onPull()` is called → `readData(resource)` returns an incomplete `Future` → `readCallback` is registered (`UnfoldResourceSourceAsync.scala:105`)
2. Downstream cancels → the stage is stopped → `postStop()` is invoked → `close(resource)` is called fire-and-forget (`UnfoldResourceSourceAsync.scala:116`)
3. `close(resource)` and the still-running `readData(resource)` now operate concurrently on the same resource `R`

### Consequences

- **Resource corruption**: if the user's `close` and `readData` implementations are not safe to call concurrently (e.g., closing a database cursor while reading from it), the resource may be corrupted or throw exceptions
- **Data loss**: `readData` may return valid data that is never delivered because the stage has already stopped
- **Double-close**: when `readData` eventually completes with `None`, the normal end-of-stream path (`UnfoldResourceSourceAsync.scala:82-87`) may call `close(resource)` again, on top of the `close` already initiated by `postStop`

### Why `UnfoldAsync` is NOT affected

`UnfoldAsync` (`stream/src/main/scala/org/apache/pekko/stream/impl/Unfold.scala:87-122`) also has an in-flight `Future` from `f(state)` when downstream cancels, but it has no `postStop` and no external resource to clean up. The in-flight `Future` simply completes and its callback is rejected by `getAsyncCallback` (stage already stopped). There is no concurrent `close()` call. `UnfoldAsyncJava` (line 129-179) has the same property.

### Why the sync `UnfoldResourceSource` is NOT affected

`UnfoldResourceSource` (`stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSource.scala`) calls `readData(resource)` synchronously on the interpreter thread. `postStop()` also runs on the interpreter thread. There is no window for concurrent execution.

### Suggested fix

Track the in-flight `readData` `Future` in a `var` field. In `postStop()`, instead of calling `close(resource)` directly, chain `close(resource)` to run after the in-flight read completes (or immediately if no read is in flight). For example:

```scala
private var inFlightRead: Future[Option[T]] = _

override def onPull(): Unit = maybeResource match {
  case OptionVal.Some(resource) =>
    try {
      val future = readData(resource)
      inFlightRead = future
      future.value match {
        case Some(value) =>
          inFlightRead = null
          handle(value)
        case None =>
          future.onComplete { result =>
            inFlightRead = null
            readCallback(result)
          }(parasitic)
      }
    } catch errorHandler
  case OptionVal.None =>
}

override def postStop(): Unit = maybeResource match {
  case OptionVal.Some(resource) =>
    val pending = inFlightRead
    if (pending != null) {
      pending.onComplete(_ => close(resource))(parasitic)
    } else {
      close(resource)
    }
  case _ =>
}
```

The key constraint is: `close(resource)` must not be invoked until any in-flight `readData(resource)` has completed.

### Affected operators

- `Source.unfoldResourceAsync` — directly affected
- `FlattenMerge` / `FlattenConcat` fast paths — not affected (no `UnfoldResourceSourceAsync` fast path exists in these operators)

### References

- None - identified during internal code review

He-Pin:
### Motivation
`Source.unfoldResourceAsync` could call `close(resource)` from `postStop()` while a `readData(resource)` `Future` was still running. That allowed `close` and `readData` to operate on the same resource concurrently after downstream cancellation.

### Modification
Track the current read `Future` with a null-sentinel field. On `postStop`, defer `close(resource)` until that pending read completes; clear the field on synchronous and asynchronous read completion to avoid retaining completed reads and keep the hot path GC/JIT friendly.

Add a regression test that cancels after a read has started, waits for stream termination, asserts `close` has not run while the read is still pending, then completes the read and verifies `close` runs exactly after that point.

### Result
Cancellation no longer closes the resource concurrently with an in-flight `readData` call. Existing `unfoldResourceAsync` completion, restart, resume, and close behavior remains covered by the full spec.

### Tests
- `scalafmt --mode diff-ref=origin/main` - passed (JDK 25 Unsafe deprecation warning only)
- `scalafmt --list --mode diff-ref=origin/main` - passed (JDK 25 Unsafe deprecation warning only)
- `sbt "stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.UnfoldResourceAsyncSourceSpec -- -z \"wait for pending read before closing resource when stream is cancelled\""` - passed
- Same focused test against old `postStop` behavior - failed as expected with `close` observing `readCompleted=false`
- `sbt "stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.UnfoldResourceAsyncSourceSpec"` - passed, 21 tests
- `sbt checkCodeStyle` - passed
- `sbt +headerCheckAll` - passed
- `git diff --check` - passed
- `sbt validatePullRequest` - Not run - focused stream test, style, and header validation run locally

### References
Fixes #3134

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
@@ -129,6 +129,40 @@ class UnfoldResourceAsyncSourceSpec extends StreamSpec(UnboundedMailboxConfig) {
       resource.closed.futureValue
     }
 
+    "wait for pending read before closing resource when stream is cancelled" in {
+      val readStarted = Promise[Done]()
+      val readCompleted = Promise[Option[Int]]()
+      val closeProbe = TestProbe()
+      val out = TestSubscriber.manualProbe[Int]()
+
+      val terminated = Source
+        .unfoldResourceAsync[Int, Unit](
+          () => Future.successful(()),
+          _ => {
+            readStarted.success(Done)
+            readCompleted.future
+          },
+          _ => {
+            closeProbe.ref ! readCompleted.isCompleted
+            Future.successful(Done)
+          })
+        .watchTermination(Keep.right)
+        .toMat(Sink.fromSubscriber(out))(Keep.left)
+        .run()
+
+      val subscription = out.expectSubscription()
+      subscription.request(1)
+      readStarted.future.futureValue
+
+      subscription.cancel()
+      terminated.futureValue
+      closeProbe.expectNoMessage(200.millis)
+
+      readCompleted.success(Some(1))
+      closeProbe.expectMsg(true)
+      closeProbe.expectNoMessage(200.millis)
+    }
+
     "fail when create throws exception" in {
       val probe = TestSubscriber.probe[Unit]()
       Source
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSourceAsync.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSourceAsync.scala
@@ -46,6 +46,7 @@ import pekko.util.OptionVal
       private lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider
       private implicit def ec: ExecutionContext = materializer.executionContext
       private var maybeResource: OptionVal[R] = OptionVal.none
+      private var inFlightRead: Future[Option[T]] = _
 
       private val createdCallback = getAsyncCallback[Try[R]] {
         case Success(resource) =>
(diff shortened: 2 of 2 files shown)


pjfanning (approved):
lgtm

PR #3158, opened 2026-06-23
Ignore stale classic remoting ACKs

### LINKED ISSUE #3126: Classic remoting: stale ACK causes irrecoverable quarantine after transient network disruption
### Motivation

In the classic remoting transport (deprecated but still shipped), `AckedSendBuffer.acknowledge` throws an `IllegalArgumentException` when it receives an ACK with a `cumulativeAck` higher than `maxSeq`. Under normal operation this would indicate a protocol invariant violation. However, a **stale ACK from a previous association** can legitimately trigger this condition after a transient network disruption, leading to an irrecoverable quarantine.

### Scenario

1. Node A sends system messages (seq 0, 1, 2) to Node B. `maxSeq = 2`.
2. Node B receives all messages and sends `ACK(2)`, but the ACK is delayed in the network.
3. A transient network disruption causes the connection to drop.
4. Node B restarts with a new UID.
5. A new association is established. Node A receives the new UID and calls `reset()`, resetting `maxSeq` to `-1`.
6. The delayed `ACK(2)` from the old session arrives.
7. `ack.cumulativeAck (2) > maxSeq (-1)` triggers the `IllegalArgumentException`.
8. The exception is caught and wrapped as `HopelessAssociation`, which triggers quarantine.
9. Node B is quarantined for days (configurable via `quarantine-duration`), requiring manual intervention or full cluster restart to recover.

### Impact

- A transient network disruption followed by a remote node restart can cause an **irrecoverable quarantine** lasting days.
- This affects only **classic remoting** (deprecated). Artery transport handles stale ACKs correctly by silently ignoring them.
- While classic remoting is deprecated, it is still compiled, shipped, and used by some deployments.

### Fix

When `cumulativeAck > maxSeq`, the buffer should treat it as a no-op (return the current buffer unchanged) rather than throwing. A warning log should be emitted in the endpoint when a stale ACK is detected, to aid in debugging.

### References

This same bug was identified and fixed in the Akka.NET project: https://github.com/akkadotnet/akka.net/issues/6414

He-Pin:
### Motivation
Classic remoting can receive a delayed system-message ACK from an old association after reconnect/reset. If that stale ACK has a cumulative sequence number newer than the sender's current resend buffer, the old code treated it as a protocol violation, wrapped it as `HopelessAssociation`, and gated/quarantined the remote system.

This fixes #3126.

### Modification
- Treat `Ack(cumulativeAck > maxSeq)` as stale in `AckedSendBuffer` and return the current buffer unchanged.
- Log and skip stale ACKs in `ReliableDeliverySupervisor` before they can be wrapped into `HopelessAssociation`.
- Keep the existing `ResendUnfulfillableException` path for real unfulfillable NACKs.
- Add regression coverage for empty/non-empty resend buffers, stale ACKs with NACKs, and valid ACK processing after a stale ACK.

### Result
Delayed ACKs from a previous association no longer trigger quarantine for the current association, while normal ACK/NACK processing remains unchanged.

### Tests
- `sbt "remote / Test / testOnly org.apache.pekko.remote.AckedDeliverySpec"` / passed: 13 tests
- `sbt checkCodeStyle` / passed
- `scalafmt --list --mode diff-ref=origin/main` / passed: no files listed
- `git diff --check` / passed: no output
- subAgent review / FINAL PASS
- `qodercli` non-interactive review, log at `/tmp/tmo-qoder-review.log` / PASS

### References
Fixes #3126

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/AckedDelivery.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/AckedDelivery.scala
@@ -116,23 +116,25 @@ final case class AckedSendBuffer[T <: HasSequenceNumber](
 
   /**
    * Processes an incoming acknowledgement and returns a new buffer with only unacknowledged elements remaining.
+   * Acknowledgements for sequence numbers newer than this buffer has seen are stale and return this buffer unchanged.
    * @param ack The received acknowledgement
    * @return An updated buffer containing the remaining unacknowledged messages
    */
   def acknowledge(ack: Ack): AckedSendBuffer[T] = {
-    if (ack.cumulativeAck > maxSeq)
-      throw new IllegalArgumentException(s"Highest SEQ so far was $maxSeq but cumulative ACK is ${ack.cumulativeAck}")
-    val newNacked =
-      if (ack.nacks.isEmpty) Vector.empty
+    if (ack.cumulativeAck > maxSeq) this
+    else {
+      val newNacked =
+        if (ack.nacks.isEmpty) Vector.empty
+        else
+          (nacked ++ nonAcked).filter { m =>
+            ack.nacks(m.seq)
+          }
+      if (newNacked.size < ack.nacks.size) throw new ResendUnfulfillableException
       else
-        (nacked ++ nonAcked).filter { m =>
-          ack.nacks(m.seq)
-        }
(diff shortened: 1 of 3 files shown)


He-Pin:
@pjfanning I think this can be backported.

pjfanning (approved):
lgtm

PR #3164, opened 2026-06-23
feat: add optional TLS hostname verification for classic remoting

### LINKED ISSUE #3161: Classic Remoting: improve TLS hostname verification
### Description

Classic Remoting is deprecated. You should use Artery based Remoting.

We should improve hostname verification nonetheless.

He-Pin:
### Motivation

Relates to #3161. Classic Remoting is deprecated and users should migrate to Artery Remoting.
We strongly recommend that the ports used for remoting and/or cluster comms should be not be behind a firewall.

Artery's `ConfigSSLEngineProvider` already supports hostname verification via a configurable option (default off). Classic remoting has no equivalent.

### Result

Classic remoting can now optionally verify TLS certificate hostnames, matching Artery's behavior. Disabled by default for backward compatibility. Enable with:

```hocon
pekko.remote.classic.netty.ssl.security {
  hostname-verification = on
}
```

### Tests

- `sbt "remote / Test / testOnly org.apache.pekko.remote.Ticket1978ConfigSpec"` — verified hostname-verification defaults to off and can be enabled

### References

Fixes #3161

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/SSLEngineProvider.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/SSLEngineProvider.scala
@@ -27,6 +27,7 @@ import javax.net.ssl.SSLEngine
 import javax.net.ssl.TrustManager
 import javax.net.ssl.TrustManagerFactory
 
+import scala.annotation.nowarn
 import scala.util.Try
 
 import org.apache.pekko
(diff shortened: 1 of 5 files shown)


remote/src/main/scala/org/apache/pekko/remote/transport/netty/SSLEngineProvider.scala
```diff
@@ -44,6 +46,13 @@ trait SSLEngineProvider {
 
   def createClientSSLEngine(): SSLEngine
 
+  /**
+   * Create a client SSLEngine with the target hostname and port for hostname verification.
+   * Default implementation delegates to [[createClientSSLEngine()]].
+   */
```
pjfanning, line 52: Can this be `@since 1.7.0` as this change could be useful in pekko 1.x?
If this is too hard to support backporting then we should put `@since 2.0.0`

He-Pin: Will mark it 1.7.0 I think.

He-Pin: Addressed in a912ffbaad by adding @since 1.7.0 to the new hostname-aware overload.

pjfanning (approved):
lgtm

PR #3165, opened 2026-06-23
fix: improve TLS observability - handshake failure logging and fail-fast SSL init

### LINKED ISSUE #3162: UnfoldResourceSource: resource can be closed twice when close() throws
### Summary

`UnfoldResourceSource` can invoke the user-provided `close` function twice on the same resource when the first `close()` call throws a `NonFatal` exception. This can corrupt or leak external resources that are not idempotent-close-safe (file descriptors, reference-counted handles, distributed locks, etc.).

### Symptom

A stream using `Source.unfoldResource` that completes (normally or via downstream cancel) can end up calling `close(resource)` twice if the first `close` invocation throws.

### Location

`stream/src/main/scala/org/apache/pekko/stream/impl/UnfoldResourceSource.scala`

In `closeStage()`:

```scala
private def closeStage(): Unit =
  try {
    close(resource)     // close called FIRST
    open = false        // guard set AFTER close — skipped if close throws
    completeStage()
  } catch {
    case NonFatal(ex) => failStage(ex)   // `open` remains true here
  }

  if (open) close(resource)   // second close when open was never cleared
}
```

### Reproduction scenario

1. Build a stream that consumes from `Source.unfoldResource(create, read, close)`.
2. Arrange for `close(resource)` to throw a `NonFatal` exception (e.g. `IOException`).
3. Terminate the stream either via:
   - downstream cancel (triggers `onDownstreamFinish` → `closeStage()`), or
   - `read` returning `None` (normal completion → `closeStage()`).
4. Observe that `close` is invoked twice.

### Root cause

In `closeStage()`, the `open = false` guard is assigned AFTER `close(resource)`. If `close` throws, the assignment is skipped, `open` remains `true`, and `postStop()` (which the interpreter always calls after `failStage`) invokes `close(resource)` again.

This is inconsistent with two other code paths in the same stage that correctly clear the guard BEFORE calling `close`:

- Supervision.Stop branch in `onPull` (around line 63): sets `open = false` then calls `close`.
- `restartState()` (around line 79): sets `open = false` then calls `close`.

The async counterpart (`UnfoldResourceSourceAsync`) does not have this problem — it clears `maybeResource = OptionVal.none` immediately after calling close, so `postStop` never double-closes.

### Impact

- User-provided `close` has no idempotency contract; double-close can corrupt state, release external resources twice, or produce confusing error logs.
- The second close's exception is swallowed silently by the interpreter's `finalizeStage` try/catch around `postStop`, masking the bug in production.

### Proposed fix

Set `open = false` before calling `close(resource)` in `closeStage()`, matching the pattern already used in the Supervision.Stop branch and `restartState()`:

```scala
private def closeStage(): Unit = {
  open = false
  try {
    close(resource)
    completeStage()
  } catch {
    case NonFatal(ex) => failStage(ex)
  }
}
```

A directional test should assert that `close` is invoked exactly once when `close` throws during stream completion, covering both the downstream-cancel path and the normal-completion path.

### Environment

- Apache Pekko streams, current `main`
- Affects `Source.unfoldResource` (sync variant only; async variant unaffected)

He-Pin:
### Motivation

Classic remoting TLS handshake failures were silently swallowed: the `NettySSLSupport` handshake listener closed the channel without logging the failure reason or peer address. Additionally, `SSLContext` was lazily initialized, meaning keystore/truststore errors were only discovered on the first connection attempt rather than at startup.

### Modification

- `NettySSLSupport`: add `MarkerLoggingAdapter` parameter, log TLS handshake failures with the Netty channel remote address and cause
- `NettyTransport`: pass log to `NettySSLSupport`
- `ConfigSSLEngineProvider`: change `sslContext` from `lazy val` to `val` for eager initialization (fail-fast at provider construction time)
- Add focused coverage for fail-fast SSL init and TLS handshake log formatting fallbacks

### Result

TLS handshake failures now produce actionable warning logs with the remote socket address and error cause when available, with stable fallback text otherwise. Keystore/truststore configuration errors are detected at startup rather than on first connection.

### Tests

- `sbt "remote / Test / testOnly org.apache.pekko.remote.ConfigSSLEngineProviderSpec org.apache.pekko.remote.transport.netty.NettySSLSupportSpec"` — 4 tests passed
- `sbt "remote / Test / testOnly org.apache.pekko.remote.Ticket1978ConfigSpec"` — 1 test passed
- `sbt "remote / Test / testOnly org.apache.pekko.remote.transport.netty.NettySSLSupportSpec"` — 3 tests passed after final cleanup
- `git diff --check` — passed
- `sbt +headerCheckAll` — passed
- `sbt +mimaReportBinaryIssues` — passed for Scala 2.13 and Scala 3.3
- `qodercli -p --output-format stream-json --cwd "$PWD" --attachment /tmp/project-review.diff` — No must-fix findings

### References

Fixes #3162

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettySSLSupport.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettySSLSupport.scala
@@ -68,11 +68,10 @@ private[pekko] object NettySSLSupport {
     val handler = new SslHandler(sslEngine)
     handler.handshakeFuture().addListener((future: Future[Channel]) => {
       if (!future.isSuccess) {
-        val cause = if (future.cause() != null) future.cause() else null
         log.warning(
-          "TLS handshake failed for remote address [{}]. {}",
+          "TLS handshake failed for remote address [{}]: {}",
           sslEngine.getPeerHost,
-          if (cause != null) cause.getMessage else "unknown cause")
+          if (future.cause() != null) future.cause().getMessage else "unknown cause")
         handler.closeOutbound().channel().close()
       }
     })
--- a/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettySSLSupport.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettySSLSupport.scala
@@ -21,6 +21,7 @@ import io.netty.handler.ssl.SslHandler
 import io.netty.util.concurrent.Future
 
 import com.typesafe.config.Config
+import org.apache.pekko.event.MarkerLoggingAdapter
 
 /**
  * INTERNAL API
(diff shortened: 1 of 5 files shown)


pjfanning (approved):
lgtm

PR #3166, opened 2026-06-23
fix: harden EndpointReader against NonFatal dispatch errors and unwrap WrappedMessage in writer logs

He-Pin:
### Motivation

`EndpointReader` only caught `NotSerializableException` and `IllegalArgumentException` during non-reliable message dispatch, allowing other `NonFatal` deserialization failures to escape and stop the endpoint. The reliable delivery path (`deliverAndAck`) did not protect dispatch at all. Additionally, `EndpointWriter` error logs showed the outer wrapper class instead of the actual message type for wrapped messages.

### Modification

- `EndpointReader`: catch `NonFatal` dispatch failures on both non-reliable and reliable delivery paths and log them as transient deserialization errors
- `EndpointWriter`: use `WrappedMessage.unwrap()` to log the inner message class in oversized payload and serialization error messages
- Change `logTransientSerializationError` parameter from `Exception` to `Throwable`
- Update log message from "Serializer not defined" to "Could not deserialize" to cover all deserialization failure scenarios

### Result

Transient deserialization failures no longer tear down the classic remoting endpoint, and error logs now show the actual message type for wrapped messages.

### Tests

- `sbt "remote / Test / testOnly org.apache.pekko.remote.TransientSerializationErrorSpec"`

### References

None - endpoint robustness hardening is not tracked by an issue.

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
@@ -1177,7 +1177,7 @@ private[remote] class EndpointReader(
   private def logTransientSerializationError(msg: PekkoPduCodec.Message, error: Throwable): Unit = {
     val sm = msg.serializedMessage
     log.warning(
-      "Serializer not defined for message with serializer id [{}] and manifest [{}]. " +
+      "Could not deserialize message with serializer id [{}] and manifest [{}]. " +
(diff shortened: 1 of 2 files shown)


pjfanning:
possible candidate for 1.7.0?

pjfanning (approved):
lgtm

PR #3167, opened 2026-06-23
[WIP] init work on new random generator

pjfanning:
* needs benchmarking
* needs more places where the new generator is used instead of ThreadLocalGenerator.
* reworked version of #1757
* for discussion at the moment
* aim is to allow Virtual Thread users optionally avoid ThreadLocalRandom


pjfanning:
We don't have easy access to the ActorSystem in most of the places where we use ThreadLocalGenerator.

I think we're probably best waiting for a proven use case where random number generation is proving to be a bottleneck.

PR #3172, opened 2026-06-24
Fix double close in unfoldResource

### LINKED ISSUE #3162: UnfoldResourceSource: resource can be closed twice when close() throws

He-Pin:
### Motivation
Source.unfoldResource can close the same resource twice when close(resource) fails. closeStage left the stage marked as open until after close returned, so postStop retried close after the failure.

### Modification
Mark the resource as no longer open before invoking close(resource) in closeStage.

Add regression coverage for close failures during normal completion and downstream cancellation.

### Result
UnfoldResourceSource now propagates the close failure without attempting to close the same resource again from postStop.

### Tests
- `rtk sbt "stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.UnfoldResourceSourceSpec"` / passed
- `rtk scalafmt --mode diff-ref=origin/main` / passed with JDK Unsafe deprecation warnings
- `rtk scalafmt --list --mode diff-ref=origin/main` / passed with JDK Unsafe deprecation warnings
- `rtk git diff --check` / passed
- `rtk sbt headerCreateAll` / passed
- `rtk sbt "+headerCheckAll"` / passed
- `rtk sbt checkCodeStyle` / passed
- `rtk qodercli --help` and qodercli stdout review with `/tmp/project-review.diff` / No must-fix findings
- `rtk sbt validatePullRequest` / not completed locally: arm64 host cannot load `leveldbjni-all:1.8` native library. The extracted binary reports x86_64/i386 while this host needs arm64; command was stopped after that environment failure.

### References
Fixes #3162

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/UnfoldResourceSourceSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/UnfoldResourceSourceSpec.scala
@@ -35,7 +35,7 @@ import pekko.stream.testkit.StreamSpec
 import pekko.stream.testkit.TestSubscriber
 import pekko.stream.testkit.Utils._
 import pekko.stream.testkit.scaladsl.TestSink
-import pekko.testkit.EventFilter
+import pekko.testkit.{ EventFilter, TestProbe }
 import pekko.util.ByteString
 
 class UnfoldResourceSourceSpec extends StreamSpec(UnboundedMailboxConfig) {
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3173, opened 2026-06-24
Ignore stale classic remoting reader ACKs

### LINKED ISSUE #3160: Classic remoting should ignore ACKs from stale EndpointReader UIDs
### Motivation

Classic remoting is deprecated, but some downstream projects still depend on it. For example, Apache Flink has an open migration discussion in FLINK-33505 because it uses Pekko classic remoting and cannot immediately move to Artery.

Issue #3126 fixed one concrete stale ACK failure mode where a delayed ACK with `cumulativeAck > resendBuffer.maxSeq` could make classic remoting quarantine an otherwise recoverable association. That fix makes future/out-of-window ACKs harmless.

There appears to be a remaining stale ACK class around EndpointReader UID/source identity.

### Current behavior

Classic remoting decodes ACKs in `EndpointReader` and forwards the bare `Ack` to `ReliableDeliverySupervisor`:

- `EndpointReader` is constructed with the remote UID from the associated handle.
- `EndpointReader.receive` decodes `(Option[Ack], Option[Message])` and sends `reliableDelivery ! ack`.
- `EndpointReader.notReading` also continues to decode inbound payloads and forward ACKs.
- `ReliableDeliverySupervisor` only gates ACK processing on `uidConfirmed`; the `Ack` message itself carries no reader UID/source identity.

The wire ACK only contains `cumulativeAck` and `nacks`; it does not contain UID/incarnation information.

This means a delayed ACK from an older reader/connection can be processed after the supervisor has confirmed a newer UID for the same remote address. If the stale ACK has `cumulativeAck <= resendBuffer.maxSeq`, the #3126 guard will not drop it. Depending on the ACK contents, it can either:

- acknowledge and remove entries from the current resend buffer, even though the ACK belongs to an older UID/incarnation, or
- trigger `ResendUnfulfillableException` when stale NACKs cannot be matched, which is still escalated to `HopelessAssociation`.

### Why this matters

Artery avoids this class of problem by carrying source identity and incarnation state in the system message delivery path:

- `SystemMessageDelivery.Ack(seqNo, from: UniqueAddress)` and `Nack(seqNo, from: UniqueAddress)` include the remote unique address.
- System message delivery also clears state when the association incarnation changes via `ClearSystemMessageDelivery` / incarnation checks.

Classic remoting has partial protection through `uidConfirmed` and `GotUid` reset, but the ACK actor message is not tied to the `EndpointReader` UID that decoded it. That leaves a gap after UID confirmation and during handoff/read-only reader transitions.

### Proposed change

Keep the classic wire protocol unchanged, but tag ACKs internally with the UID of the `EndpointReader` that decoded them.

One possible approach:

1. Add an internal message such as `AckFromReader(uid: Int, ack: Ack)` in `ReliableDeliverySupervisor` or a nearby internal scope.
2. Change `EndpointReader` to send `AckFromReader(uid, ack)` instead of a bare `Ack` when `reliableDeliverySupervisor` is defined.
3. In `ReliableDeliverySupervisor`, process the ACK only if `uidConfirmed` and `uid.contains(readerUid)`.
4. Drop ACKs from stale reader UIDs with a bounded/debug log.
5. Preserve the existing bare `Ack` handling only if needed for tests/backward internal compatibility, or migrate tests to the new internal message.

This would be wire-compatible and scoped to classic remoting internals.

### Suggested tests

- A supervisor-level regression test where:
  - old UID is confirmed and system messages are buffered/sent,
  - a new UID is confirmed, resetting the resend buffer,
  - a stale `AckFromReader(oldUid, Ack(...))` arrives with a sequence number that would otherwise fit the new buffer,
  - the stale ACK is ignored and does not clear the new resend buffer.
- A matching valid `AckFromReader(currentUid, Ack(...))` should still acknowledge the current buffer.
- A handoff/read-only reader test, if practical, covering that `notReading` ACKs from stale UIDs do not affect the current writer/supervisor.

### Related work

- #3126: stale future ACKs should not quarantine classic remoting.
- FLINK-33505: downstream context where classic remoting remains relevant.

He-Pin:
### Motivation
Classic remoting ACKs forwarded by EndpointReader were not tied to the reader UID, so a delayed ACK from an old reader could affect the resend buffer after a new UID was confirmed.

### Modification
Tag decoded ACKs with the EndpointReader UID and have ReliableDeliverySupervisor only process ACKs from the currently confirmed UID.

Add a deterministic supervisor-level regression test covering stale and current reader ACKs. This keeps the regression focused on the actor/UID contract rather than relying on multi-node timing races.

### Result
Stale reader ACKs are ignored while current-reader ACKs still acknowledge the active resend buffer.

### Tests
- sbt "remote / Test / testOnly org.apache.pekko.remote.ReliableDeliverySupervisorSpec"
- sbt headerCreateAll
- sbt +headerCheckAll
- scalafmt --mode diff-ref=origin/main
- git diff --check
- qodercli -p --output-format stream-json --attachment /tmp/pekko-3160-qoder-review.diff (No must-fix findings)
- sbt sortImports (failed: scalafix/scala.meta NoSuchMethodError in stream-tests and multi-node-testkit)

### References
Fixes #3160

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
@@ -226,6 +226,8 @@ private[remote] object ReliableDeliverySupervisor {
   case object Ungate
   case object AttemptSysMsgRedelivery
   final case class GotUid(uid: Int, remoteAddres: Address)
+  @nowarn("msg=deprecated")
+  final case class AckFromReader(readerUid: Int, ack: Ack) extends NoSerializationVerificationNeeded
 
   case object IsIdle
   case object Idle
@@ -363,33 +365,9 @@ private[remote] class ReliableDeliverySupervisor(
     case IsIdle  => // Do not reply, we will Terminate soon, or send a GotUid
     case s: Send =>
       handleSend(s)
-    case ack: Ack =>
-      // If we are not sure about the UID just ignore the ack. Ignoring is fine.
-      if (uidConfirmed) {
-        if (ack.cumulativeAck > resendBuffer.maxSeq)
-          log.warning(
-            "Ignoring stale system message acknowledgement from remote system [{}]. Highest SEQ so far was [{}] but cumulative ACK is [{}].",
-            remoteAddress,
-            resendBuffer.maxSeq,
-            ack.cumulativeAck)
-        else {
-          try resendBuffer = resendBuffer.acknowledge(ack)
-          catch {
-            case NonFatal(e) =>
-              throw new HopelessAssociation(
-                localAddress,
-                remoteAddress,
-                uid,
-                new IllegalStateException(
-                  s"Error encountered while processing system message " +
-                  s"acknowledgement buffer: $resendBuffer ack: $ack",
-                  e))
-          }
-
-          resendNacked()
-        }
-      }
-    case AttemptSysMsgRedelivery =>
+    case AckFromReader(readerUid, ack) => handleAck(readerUid, ack)
+    case _: Ack                        => () // Ignore untagged ACKs. EndpointReader attaches its UID before forwarding ACKs.
+    case AttemptSysMsgRedelivery       =>
       if (uidConfirmed) resendAll()
     case Terminated(_) =>
       currentHandle = None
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

Issue #3110, opened 2026-06-20
Meta: Add supervisor strategy support to stream operators that accept user functions


He-Pin:
### Motivation

Per the [stream error handling docs](https://pekko.apache.org/docs/pekko/current/stream/stream-error.html#supervision-strategies):

> The operators that support supervision strategies are explicitly documented to do so, if there is nothing in the documentation of an operator saying that it adheres to the supervision strategy it means it fails rather than applies supervision.

Several stream operators accept user-provided functions that can throw exceptions, yet do not consult the `SupervisionStrategy` decider. This means that if the user function throws, the stream fails unconditionally, regardless of the configured supervision strategy. For consistency and usability, all operators that apply user functions to stream elements should support supervisor strategy.

Related: #3101 (Throttle-specific issue)

### Operators missing supervisor strategy support

The following operators accept user-provided functions but do **not** implement supervision. Each entry includes the function parameter, the source location, and the specific code path where the function is called without try-catch.

#### 1. `throttle` (with `costCalculation`)

- **Function**: `costCalculation: Out => Int`
- **Implementation**: `Throttle.scala:73` — `costCalculation(elem)` in `onPush()` without try-catch
- **Impact**: If `costCalculation` throws (e.g. for malformed input), the stream fails even with `Supervision.Resume`
- **Tracked in**: #3101

#### 2. `expand`

- **Function**: `expander: Out => Iterator[U]`
- **Implementation**: `Ops.scala:1210` — `extrapolate(grab(in))` in `onPush()` without try-catch
- **Doc**: Explicitly states "Expand does not support `Supervision.Restart` and `Supervision.Resume`. Exceptions from the `seed` function will complete the stream with failure."
- **Impact**: Should be enhanced to support supervision, or the limitation should remain clearly documented if semantics prevent it

#### 3. `groupedWeighted`

- **Function**: `costFn: Out => Long`
- **Implementation**: `Ops.scala:816` — `costFn(elem)` in `GroupedWeighted.onPush()` without try-catch
- **Impact**: Negative cost is handled, but a throwing `costFn` is not supervised

#### 4. `groupedWeightedWithin`

- **Function**: `costFn: Out => Long`
- **Implementation**: `Ops.scala:1797` — `costFn(elem)` in `GroupedWeightedWithin.nextElement()` without try-catch
- **Impact**: Same as `groupedWeighted`

#### 5. `groupedAdjacentBy` / `groupedAdjacentByWeighted`

[...]

PR #3179, opened 2026-06-25
feat: add supervision strategy support for doOnFirst

He-Pin:
### Motivation

Per the [stream error handling docs](https://pekko.apache.org/docs/pekko/current/stream/stream-error.html#supervision-strategies), operators that apply user-provided functions should consult the configured `SupervisionStrategy`. The `doOnFirst` operator runs a user side-effect function `f` on the **first** element, but it was called without a try-catch, so any exception failed the stream **unconditionally** — even under `Supervision.Resume`/`Restart`.

### Modification

- Wrap `f(elem)` in `DoOnFirst` with a `try`/`catch` (`NonFatal`) that consults the `SupervisionStrategy` decider:
  - **Resume** → drop the failing first element and switch to pass-through; `f` is **not** retried (the one-shot is consumed)
  - **Restart** → re-arm first-element detection so the **next** element is treated as the first (`f` is retried on it)
- The first-element `InHandler` was extracted into a named `val firstHandler` so it can be re-armed on Restart. On success, the switch to pass-through now happens after `f` (behaviorally identical on the happy path — `push` does not synchronously re-enter `onPush`).
- Fix a fully-qualified type reference in the `createLogic` signature to use the project's two-line import style.
- Document supervision adherence (including the Resume/Restart semantics) in the Scala/Java DSL scaladoc and the operator reference page.

### Result

`doOnFirst` now adheres to the `SupervisionStrategy` attribute with clear, distinguishable Resume (drop first, no retry) and Restart (next becomes first) semantics.

### Tests

- `sbt "stream-tests/Test/testOnly org.apache.pekko.stream.scaladsl.FlowDoOnFirstSpec"` — **6/6 passed**, including 4 new directional tests:
  - fails on `f` throw with `Supervision.Stop`
  - fails with no supervision attribute set (default Stop) — back-compat guard
  - `Resume` drops the first element, does not retry `f`, passes the rest through
  - `Restart` re-arms so `f` is retried on each subsequent "first" element

### References

Refs #3110

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/doOnFirst.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/doOnFirst.md
@@ -13,6 +13,8 @@ Run the given function when the first element is received.
 
 Run the given function when the first element is received.
 
+The `doOnFirst` operator adheres to the ActorAttributes.SupervisionStrategy attribute. On `Supervision.Resume` the failing first element is dropped and the function is not retried; on `Supervision.Restart` the next element is treated as the first.
+
 ## Examples
 
 Scala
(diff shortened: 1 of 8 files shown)


pjfanning (approved):
lgtm

PR #3180, opened 2026-06-25
feat: add supervision strategy support for groupedWeightedWithin costFn

He-Pin:
### Motivation

Per the [stream error handling docs](https://pekko.apache.org/docs/pekko/current/stream/stream-error.html#supervision-strategies), operators that apply user-provided functions should consult the configured `SupervisionStrategy`. The `groupedWeightedWithin` operator's `costFn` was called without a try-catch, so any exception failed the stream **unconditionally** — even under `Supervision.Resume`/`Restart`.

### Modification

- Wrap `costFn(elem)` inside `GroupedWeightedWithin.nextElement()` with a `try`/`catch` (`NonFatal`) that consults the `SupervisionStrategy` decider. The cost is computed **before** any state mutation, so the happy path is unchanged:
  - **Resume** → skip the offending element, **keep** the current group (weight/number accounting is preserved)
  - **Restart** → **drop** the in-progress group via a new `resetGroupState()` helper, and **reschedule** the time-window timer so the fresh group gets a full window
- Document supervision adherence on **both** `groupedWeightedWithin` overloads in the Scala/Java DSL scaladoc and the operator reference page.

### Result

`groupedWeightedWithin` now adheres to the `SupervisionStrategy` attribute with well-defined Resume/Restart semantics.

### Tests

- `sbt "stream-tests/Test/testOnly org.apache.pekko.stream.scaladsl.FlowGroupedWithinSpec"` — **21/21 passed**, including 5 new deterministic directional tests (Stop, default Stop, Resume keeps group, Resume preserves weight accounting, Restart drops group). They use a 30s window that never fires, so they are deterministic and run in PR validation (intentionally not tagged `TimingTest`).

### References

Refs #3110

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/groupedWeightedWithin.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/groupedWeightedWithin.md
@@ -15,6 +15,7 @@ Chunk up this stream into groups of elements received within a time window, or l
 Chunk up this stream into groups of elements received within a time window, or limited by the weight and number of 
 the elements, whatever happens first. Empty groups will not be emitted if no elements are received from upstream.
 The last group before end-of-stream will contain the buffered elements since the previously emitted group.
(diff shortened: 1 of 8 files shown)


pjfanning (approved):
lgtm

PR #3181, opened 2026-06-25
feat: add supervision strategy support for groupedAdjacentBy/Weighted

He-Pin:
### Motivation

Per the [stream error handling docs](https://pekko.apache.org/docs/pekko/current/stream/stream-error.html#supervision-strategies), operators that apply user-provided functions should consult the configured `SupervisionStrategy`. The `groupedAdjacentBy` and `groupedAdjacentByWeighted` operators apply **two** user functions — a key function `f` and (for the weighted variant) a `costFn` — but both were called without a try-catch, so any exception failed the stream **unconditionally**, even under `Supervision.Resume`/`Restart`.

This is part of the meta-issue #3110 (add supervisor strategy support to stream operators that accept user functions). One operator (group) per PR.

### Modification

- Wrap **both** the `costFn` and the key function calls in `GroupedAdjacentByWeighted` with a `try`/`catch` (`NonFatal`) that consults the `SupervisionStrategy` decider via a shared `superviseUserFn` helper:
  - **Resume** → skip the offending element, **keep** the current group
  - **Restart** → **drop** the in-progress group (an already-completed group awaiting downstream demand is preserved)
- The negative-cost and null-key checks are intentionally left **unchanged** — they remain unconditional failures, because they are *output-contract violations* (the function returned an invalid value) rather than *user-function exceptions*. This matches the sibling `groupedWeighted`/`groupedWeightedWithin` operators.
- Document supervision adherence on both operators in the Scala/Java DSL scaladoc and the operator reference pages.

### Result

`groupedAdjacentBy` and `groupedAdjacentByWeighted` now adhere to the `SupervisionStrategy` attribute for both user functions.

### Tests

- `sbt "stream-tests/Test/testOnly org.apache.pekko.stream.scaladsl.FlowGroupedAdjacentByWeightedSpec"` — **14/14 passed**, including 8 new directional tests covering both functions: `costFn` throws (Stop, default Stop, Resume, Restart), key function throws (Stop, Resume, Restart, and Resume-on-last-element flushed at completion).

### References

Refs #3110

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/groupedAdjacentBy.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/groupedAdjacentBy.md
@@ -14,7 +14,7 @@ Partitions this stream into chunks by a delimiter function.
 
 Partitions this stream into chunks by a delimiter function.
 
-Adheres to the ActorAttributes.SupervisionStrategy attribute (applied to the key function).
+Adheres to the ActorAttributes.SupervisionStrategy attribute (applied to the key function). On `Supervision.Resume` the offending element is skipped; on `Supervision.Restart` the current group is dropped.
 
 See also:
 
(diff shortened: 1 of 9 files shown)


pjfanning (approved):
lgtm

PR #3182, opened 2026-06-25
feat: add supervision strategy support for delayWith

He-Pin:
### Motivation

Per the [stream error handling docs](https://pekko.apache.org/docs/pekko/current/stream/stream-error.html#supervision-strategies), operators that apply user-provided functions should consult the configured `SupervisionStrategy`. The `delayWith` operator computes each element's delay via a user-provided `DelayStrategy.nextDelay`, but it was called without a try-catch, so any exception failed the stream **unconditionally**, even under `Supervision.Resume`/`Restart`.

### Modification

- Wrap `delayStrategy.nextDelay(element)` in `Delay.grabAndPull()` with a `try`/`catch` (`NonFatal`) that consults the `SupervisionStrategy` decider:
  - **Resume** → drop the offending element, keep already-buffered elements and the current strategy
  - **Restart** → recreate the (potentially stateful) delay strategy from the supplier; already-buffered elements keep their computed delays
- `delayStrategy` became a `var` so Restart can recreate it.
- **Buffer-underflow guard**: because a dropped element no longer enqueues, `onPush` and `onTimer` now guard their buffer access with a non-empty check. In particular, `onTimer` could previously hit a `NullPointerException` (the delay buffer is an unchecked fixed-size buffer) when an overflow handler emptied the buffer while a timer was still armed and a `nextDelay` exception was then skipped. A dedicated regression test reproduces this (it fails without the guard).
- Document supervision adherence on `delayWith` in the Scala/Java DSL scaladoc and the operator reference page (plain `delay` is unaffected — it uses a fixed internal strategy that never throws).

### Result

`delayWith` now adheres to the `SupervisionStrategy` attribute with well-defined Resume/Restart semantics and no buffer underflow on the supervised paths.

### Tests

- `sbt "stream-tests/Test/testOnly org.apache.pekko.stream.scaladsl.FlowDelaySpec"` — **26/26 passed**, including 6 new tests: Stop, default Stop, Resume (drops element), Restart (recreates the stateful strategy — verified via an instance counter), Resume (keeps the stateful strategy — inverted mirror), and an overflow-handler + throw regression that guards the `onTimer` underflow fix.

### References

Refs #3110

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/delayWith.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/delayWith.md
@@ -14,6 +14,8 @@ Delay every element passed through with a duration that can be controlled dynami
 
 Delay every element passed through with a duration that can be controlled dynamically, individually for each elements (via the `DelayStrategy`).
 
+This operator adheres to the ActorAttributes.SupervisionStrategy attribute. On `Supervision.Resume` the offending element is dropped; on `Supervision.Restart` the delay strategy is recreated from the supplier (already-buffered elements keep their delays).
+
 
 @@@div { .callout }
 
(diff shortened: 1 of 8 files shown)


pjfanning (approved):
lgtm

Issue #3101, opened 2026-06-20
Throttle stage lacks supervisor strategy support for costCalculation exceptions


He-Pin:
### Motivation

The `Throttle` stage's `costCalculation` function is called without any exception handling. If the user-provided cost function throws, the exception propagates directly and fails the stage unconditionally. Unlike other stages (e.g., `MapAsync`, `Log`, `Filter`), Throttle does not consult the `SupervisionStrategy` decider, preventing users from configuring Resume or Restart behavior for bad elements.

### Current behavior

In `Throttle.scala:71-84`, `costCalculation(elem)` is called without try-catch:

```scala
override def onPush(): Unit = {
  val elem = grab(in)
  val cost = costCalculation(elem)  // No error handling
  val delayNanos = tokenBucket.offer(cost)
  ...
}
```

### Proposed fix

1. Add `lazy val decider = inheritedAttributes.mandatoryAttribute[SupervisionStrategy].decider` in `createLogic`
2. Mix in `StageLogging` trait for error logging
3. Wrap `costCalculation(elem)` in try-catch, consult decider on exception:
   - `Stop` → `failStage(ex)` (current behavior)
   - `Resume`/`Restart` → log error and `pull(in)` to skip the element
4. Add tests verifying supervision directives work with Throttle

### References

- Inspired by Akka.NET [#6406](https://github.com/akkadotnet/akka.net/issues/6406)
- Similar pattern exists in Pekko's `MapAsync` (`Ops.scala:1301`) and `Log` (`Ops.scala:1542`)

PR #3183, opened 2026-06-25
feat: add supervision strategy support for throttle costCalculation

He-Pin:
### Motivation

Per the [stream error handling docs](https://pekko.apache.org/docs/pekko/current/stream/stream-error.html#supervision-strategies), operators that apply user-provided functions should consult the configured `SupervisionStrategy`. The `throttle` operator's `costCalculation` function is user-provided and may throw, but it was called without a try-catch, so any exception failed the stream **unconditionally**, even under `Supervision.Resume`/`Restart`.

### Modification

- Wrap `costCalculation(elem)` in `Throttle.onPush()` with a `try`/`catch` (`NonFatal`) that consults the `SupervisionStrategy` decider:
  - **Resume** → drop the offending element, keep the token bucket state
  - **Restart** → behave the same as Resume (throttle keeps no accumulated per-element state, so there is nothing element-specific to reset)
- The cost calculation is evaluated **before** `tokenBucket.offer(cost)`, so a throwing function never consumes rate budget.
- Add a guard in the timer path so a supervised skip cannot dequeue from an empty buffer after an overflow handler has already mutated it.
- Document supervision adherence on the two `costCalculation` throttle overloads in the Scala/Java DSL scaladoc and the operator reference page. The element-count throttle overloads are intentionally left undocumented because they use a constant cost and never throw.

### Result

`throttle` now adheres to the `SupervisionStrategy` attribute for its costCalculation function, with no buffer-underflow on supervised skip paths.

### Tests

- `sbt "stream-tests/Test/testOnly org.apache.pekko.stream.scaladsl.FlowThrottleSpec"` — **28/28 passed**, including 6 new tests: explicit Stop, default Stop, Resume (Shaping), Resume (Enforcing), Restart (same as Resume for throttle), and a last-element Resume regression.

### References

Refs #3110, Refs #3101

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/throttle.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/throttle.md
@@ -16,6 +16,8 @@ a function has to be provided to calculate the individual cost of each element.
 
 The throttle operator combines well with the @ref[`queue`](./../Source/queue.md) operator to adapt the speeds on both ends of the `queue`-`throttle` pair.
 
+When a cost-calculation function is supplied, this operator adheres to the ActorAttributes.SupervisionStrategy attribute. On Supervision.Resume the offending element is dropped; Supervision.Restart behaves the same as Supervision.Resume because throttle keeps no accumulated per-element state.
+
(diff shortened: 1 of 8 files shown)


pjfanning (approved):
lgtm

PR #3184, opened 2026-06-25
fix: add supervision strategy support for batch costFn

He-Pin:
### Motivation

Issue #3110 tracks stream operators that should honor `SupervisionStrategy` for user functions.

`batchWeighted` already supervised `seed` and `aggregate`, but `costFn` in `Batch.onPush()` still ran outside that boundary, so `costFn` exceptions failed the stream unconditionally.

While adding `costFn` supervision, additional coupled exceptional-path issues were found around pending-element handling in `flush()`/EOS handling (state leakage and invalid pull-after-fail behavior). Those are fixed in the same PR to keep supervision behavior consistent and safe.

### Modification

- Added supervision handling around `costFn(elem)` in `Batch.onPush()`:
  - `Stop` → fail stage
  - `Resume` → drop offending element and continue
  - `Restart` → reset batch state and continue
- Hardened pending exceptional paths in `Batch`:
  - clear `agg` immediately after push before processing `pending`
  - evaluate `costFn(pending)` before `seed(pending)` in `flush()`
  - guard follow-up pulls with `!isClosed(in)`
  - complete stage at EOS when pending is dropped by `Resume`/`Restart` after seed failure
- Expanded directional tests in `FlowBatchSpec`:
  - explicit `Stop` and default `Stop`
  - Resume vs Restart semantics
  - pending/flush second `costFn` evaluation failures under Resume/Restart/Stop
  - EOS pending-seed failure under Restart

### Result

`batch` / `batchWeighted` now consistently apply supervision to `costFn`, `seed`, and `aggregate`, and maintain correct state/pull behavior on exceptional pending paths.

which is now Apache licensed

### Tests

- `sbt "stream-tests/Test/testOnly org.apache.pekko.stream.scaladsl.FlowBatchSpec"` (14/14 passed)
- `sbt "stream/mimaReportBinaryIssues"` (clean)

### References

Refs #3110

### DIFF
--- a/docs/src/main/paradox/stream/operators/Source-or-Flow/batch.md
+++ b/docs/src/main/paradox/stream/operators/Source-or-Flow/batch.md
@@ -23,6 +23,11 @@ to the summary type.
 Will eagerly pull elements, this behavior may result in a single pending (i.e. buffered) element which cannot be
 aggregated to the batched value.
 
+This operator adheres to the `ActorAttributes.SupervisionStrategy` attribute for exceptions thrown by the `costFn`,
+`seed`, or `aggregate` functions. On `Supervision.Stop` (the default) the stream fails. On `Supervision.Resume` the
+offending element is dropped and the accumulated batch (if any) is preserved. On `Supervision.Restart` the offending
+element is dropped, the accumulated batch is discarded, and the operator starts fresh.
+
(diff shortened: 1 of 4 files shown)


pjfanning (approved):
lgtm

PR #3192, opened 2026-06-26
refactor: use pattern matching instanceof (Java 16+)

He-Pin:
## Summary
- Migrate remaining old-style `instanceof` checks with explicit casts to Java 16+ pattern matching instanceof
- 2 files updated: PingSerializerExampleTest (1 cast pattern), ActorDocTest (3 cast patterns)
- Most of the codebase already uses pattern matching instanceof; remaining instanceof checks are pure boolean tests (no cast needed)

## Test plan
- [ ] Existing tests pass (behavior-preserving refactoring)
- [ ] `sbt javafmtAll` passes with JDK 17


pjfanning (approved):
lgtm

PR #3200, opened 2026-06-27
fix(remote): eager SSLContext init for artery ConfigSSLEngineProvider fail-fast

He-Pin:
### Motivation

Artery's `ConfigSSLEngineProvider` has used `lazy val sslContext` since its introduction. As a result, keystore and truststore misconfiguration is only discovered on the first connection attempt rather than at provider construction. Classic remoting's `ConfigSSLEngineProvider` was given the same eager-init fail-fast behavior in #3165; artery should match so that SSL misconfiguration surfaces at `ActorSystem` startup with a clear error.

### Modification

- Change `private lazy val sslContext` to `private val sslContext` in `org.apache.pekko.remote.artery.tcp.ConfigSSLEngineProvider` so that `SslTransportException` is raised during construction when the keystore or truststore cannot be loaded.
- Add a directional regression test `ArteryConfigSSLEngineProviderSpec` that constructs the provider against a config pointing at nonexistent keystore/truststore paths and asserts `SslTransportException` with the "could not be established" message.

### Result

Artery SSL misconfiguration now fails fast at `ActorSystem` startup with a clear error, instead of silently succeeding construction and failing later on the first connection attempt.

### Tests

- `sbt "remote / Test / testOnly org.apache.pekko.remote.artery.tcp.ArteryConfigSSLEngineProviderSpec"`
  - Tests: succeeded 1, failed 0
- Companion classic-side regression test `ConfigSSLEngineProviderSpec` (added in #3165) still passes.

### References

None - companion to the classic remoting fail-fast fix (#3165).

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/artery/tcp/ConfigSSLEngineProvider.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/artery/tcp/ConfigSSLEngineProvider.scala
@@ -66,7 +66,7 @@ class ConfigSSLEngineProvider(protected val config: Config, protected val log: M
   val SSLRequireMutualAuthentication: Boolean = sslEngineConfig.SSLRequireMutualAuthentication
   val HostnameVerification: Boolean = sslEngineConfig.HostnameVerification
 
-  private lazy val sslContext: SSLContext = {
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3205, opened 2026-06-27
Resume classic remoting reader after failed handoff

### LINKED ISSUE #3206: Classic remoting passive read handoff can leave outbound reader stopped
### Problem

Classic remoting passive read handoff can leave the original outbound reader permanently stopped when the replacement read-only endpoint fails.

With `pekko.remote.classic.use-passive-connections = on`, an inbound association for an already-known remote UID causes the existing writable endpoint to stop reading from the outbound association. If the new read-only endpoint then fails during initialization or later terminates, there is no recovery path that resumes the original reader.

### Trigger Path

1. `Remoting.handleInboundAssociation` sees an existing writable endpoint with matching UID.
2. The endpoint manager sends `EndpointWriter.StopReading` to the writable endpoint.
3. `ReliableDeliverySupervisor` forwards `StopReading` to `EndpointWriter`, which forwards it to `EndpointReader`.
4. `EndpointReader` enters `notReading`.
5. In `notReading`, inbound payloads are discarded while ACKs are still handled.
6. If the read-only endpoint terminates, the original outbound reader remains stopped.

```text
Inbound association, same remote UID
              |
              v
+----------------------------------+
| Remoting.handleInboundAssociation |
| existing writable endpoint found  |
+----------------------------------+
              |
              | StopReading
              v
+----------------------------------+
| ReliableDeliverySupervisor        |
| forwards StopReading to writer    |
+----------------------------------+
              |
              v
+----------------------------------+
| EndpointWriter                    |
| forwards StopReading to reader    |
+----------------------------------+
              |
              v
+----------------------------------+
| EndpointReader.notReading         |
| ACKs are handled                  |
| inbound payloads are discarded    |
+----------------------------------+
              |
              | replacement read-only endpoint fails
              v
+----------------------------------+
| EndpointManager.Terminated         |
| no resume path before this fix     |
+----------------------------------+
              |
              v
Original outbound reader remains stopped
```

### Two Node View

Before the passive handoff, Node A already has a writable endpoint for Node B. The same TCP association can still carry inbound payloads from Node B back to Node A.

```text
+--------------------------------------------+        existing writable association        +--------------------------------------------+
| Node A                                     |===========================================>| Node B                                     |
|                                            |                                            |                                            |
| EndpointManager                            |<===========================================| Remote system UID = B1                    |
|  policy: Pass(writer A->B, uid = B1)       |          payloads can arrive on             |                                            |
|                                            |          the A->B association read side     |                                            |
| Writable endpoint A->B                     |                                            |                                            |
|  reader state: reading                     |                                            |                                            |
+--------------------------------------------+                                            +--------------------------------------------+
```

When Node B opens a passive inbound association to Node A with the same UID, Node A stops the read side of the existing writable endpoint and tries to use the new read-only endpoint instead.

```text
+--------------------------------------------+          new passive inbound association    +--------------------------------------------+
| Node A                                     |<===========================================| Node B                                     |
|                                            |                                            |                                            |
| Existing writable endpoint A->B            |                                            | Remote system UID = B1                    |
|  StopReading                               |                                            |                                            |
|  reader state: notReading                  |                                            |                                            |
|                                            |                                            |                                            |
| Replacement read-only endpoint B->A        |                                            |                                            |
|  initialization fails or terminates         |                                            |                                            |
|                                            |                                            |                                            |
| Before this fix:                           |        future payloads on the old           |                                            |
|  no ResumeReading path                     |<-------------------------------------------|                                            |
|  inbound payloads are discarded            |        A->B association are dropped          |                                            |
+--------------------------------------------+                                            +--------------------------------------------+
```

With this fix, Node A keeps enough fallback state to resume the original reader only when the current endpoint policy still points to the same writer and UID.

```text
+--------------------------------------------+                                            +--------------------------------------------+
| Node A                                     |                                            | Node B                                     |
|                                            |                                            |                                            |
| Read-only endpoint B->A terminates          |                                            | Remote system UID = B1                    |
|              |                             |                                            |                                            |
|              v                             |                                            |                                            |
| resumeReadingIfNeeded                      |                                            |                                            |
|  verifies Pass(writer A->B, uid = B1)       |                                            |                                            |
|              |                             |                                            |                                            |
|              v                             |                                            |                                            |
| ResumeReading -> RDS -> Writer -> Reader    |                                            |                                            |
|              |                             |        future payloads on the old           |                                            |
|              v                             |<===========================================|                                            |
| Writable endpoint A->B                     |        A->B association are read again      |                                            |
|  reader state: reading                     |                                            |                                            |
+--------------------------------------------+                                            +--------------------------------------------+
```

Expected recovery flow:

```text
Read-only endpoint terminates
              |
              v
+----------------------------------+
| EndpointManager.resumeReadingIfNeeded |
| verify current writer + UID match     |
+----------------------------------+
              |
              | ResumeReading
              v
+----------------------------------+
| ReliableDeliverySupervisor        |
| active: forward to writer         |
| idle/gated: recreate or forward   |
+----------------------------------+
              |
              v
+----------------------------------+
| EndpointWriter -> EndpointReader  |
| reader leaves notReading          |
+----------------------------------+
              |
              v
Future inbound payloads are read again
```

### Impact

A restarting or reconnecting remote system can fail to rejoin because messages arriving on the original outbound association are dropped after the passive read handoff fails. This can affect classic remoting users that still rely on passive connections.

### Expected Behavior

If the passive read-only endpoint terminates while the original writable endpoint and UID still match, the endpoint manager should resume reading on the original outbound reader.

### Notes

The recovery must be guarded by the current endpoint policy and UID so a stale reader is not resumed after the remote system has restarted with a different UID. Replacement read-only endpoints should also clear or carry fallback state only when their address and UID still match.

### Workaround

Set `pekko.remote.classic.use-passive-connections = off`, or migrate away from classic remoting where possible.

He-Pin:
### Motivation
Classic remoting passive read handoff can leave the original outbound reader stuck in `notReading` if the read-only handoff endpoint fails.

PR validation also exposed an unrelated `BindCanonicalAddressSpec` failure in the `tls-tcp` cases, where the spec fell back to the `reference.conf` default `keystore` and `truststore` relative paths.

### Modification
Track passive read handoff fallback endpoints and resume the original writable reader when the read-only endpoint terminates while the original endpoint and UID still match.

Carry fallback state across replacement read-only endpoints, clear stale fallback state on UID/address changes, handle `ResumeReading` in supervisor states, add logging, and add regression coverage. Also correct a stale classic remoting passive-connections config reference.

Also include the TLS test configuration fix from #3210, reusing `ArterySpecSupport.tlsConfig` in `BindCanonicalAddressSpec.commonConfig` so the existing PR validation can complete.

### Result
Classic remoting can recover future inbound messages on the original association after a failed passive read handoff.

The TLS bind canonical address cases no longer fail with `NoSuchFileException: keystore` during PR validation.

### Tests
- `rtk sbt "remote / Test / testOnly org.apache.pekko.remote.ReliableDeliverySupervisorSpec"` / passed (4 tests)
- `rtk sbt "remote / Test / testOnly org.apache.pekko.remote.classic.RemotingSpec -- -z \"resume the outbound reader\""` / passed
- `rtk sbt "remote / Test / testOnly org.apache.pekko.remote.classic.RemotingSpec"` / passed (22 tests)
- JDK 17: `rtk bash -lc "export JAVA_HOME=$(/usr/libexec/java_home -v 17); export PATH=\"$JAVA_HOME/bin:$PATH\"; java -version; sbt \"remote / Test / testOnly org.apache.pekko.remote.artery.BindCanonicalAddressSpec\""` / passed (15 tests)
- `rtk scalafmt --list --mode diff-ref=origin/main` / passed (JVM Unsafe warnings only)
- `rtk bash -lc "git diff --check"` / passed
- `rtk sbt sortImports` / failed: Scalafix/scala-meta `NoSuchMethodError` in local tooling; unrelated generated change was reverted
- `qodercli` stdout review / No must-fix findings
- independent subAgent review / No must-fix findings

### References
Fixes #3206
Refs #3210

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
@@ -387,6 +387,9 @@ private[remote] class ReliableDeliverySupervisor(
     case s: EndpointWriter.StopReading =>
       writer.forward(s)
 
+    case EndpointWriter.ResumeReading =>
+      writer ! EndpointWriter.ResumeReading
+
     case Ungate => // ok, not gated
   }
 
@@ -418,10 +421,18 @@ private[remote] class ReliableDeliverySupervisor(
         // Resending will be triggered by the incoming GotUid message after the connection finished
         goToActive()
       } else goToIdle()
-    case AttemptSysMsgRedelivery                => // Ignore
-    case s @ Send(_: SystemMessage, _, _, _)    => tryBuffer(s.copy(seqOpt = Some(nextSeq())))
-    case s: Send                                => context.system.deadLetters ! s
-    case EndpointWriter.FlushAndStop            => context.stop(self)
+    case AttemptSysMsgRedelivery             => // Ignore
+    case s @ Send(_: SystemMessage, _, _, _) => tryBuffer(s.copy(seqOpt = Some(nextSeq())))
+    case s: Send                             => context.system.deadLetters ! s
+    case EndpointWriter.FlushAndStop         => context.stop(self)
+    case EndpointWriter.ResumeReading        =>
+      if (!writerTerminated) {
+        writer ! EndpointWriter.ResumeReading
+        context.become(gated(writerTerminated = false, earlyUngateRequested = true))
+      } else {
+        writer = createWriter()
+        goToActive()
+      }
     case EndpointWriter.StopReading(w, replyTo) =>
       replyTo ! EndpointWriter.StoppedReading(w)
       sender() ! EndpointWriter.StoppedReading(w)
@@ -448,7 +459,10 @@ private[remote] class ReliableDeliverySupervisor(
         new TimeoutException(
           "Remote system has been silent for too long. " +
           s"(more than ${settings.QuarantineSilentSystemTimeout.toCoarsest})"))
-    case EndpointWriter.FlushAndStop            => context.stop(self)
+    case EndpointWriter.FlushAndStop  => context.stop(self)
+    case EndpointWriter.ResumeReading =>
+      writer = createWriter()
+      goToActive()
     case EndpointWriter.StopReading(w, replyTo) =>
       replyTo ! EndpointWriter.StoppedReading(w)
     case Ungate => // ok, not gated
@@ -482,7 +496,8 @@ private[remote] class ReliableDeliverySupervisor(
       // and don't really know if they were properly delivered or not.
       resendBuffer = new AckedSendBuffer[Send](0)
       context.stop(self)
-    case _ => // Ignore
+    case EndpointWriter.ResumeReading => // Ignore while shutting down
+    case _                            => // Ignore
   }
 
   private def handleSend(send: Send): Unit =
@@ -625,6 +640,7 @@ private[remote] object EndpointWriter {
   private case object FlushAndStopTimeout
   case object AckIdleCheckTimer
   final case class StopReading(writer: ActorRef, replyTo: ActorRef)
+  case object ResumeReading
   final case class StoppedReading(writer: ActorRef)
 
   final case class Handle(handle: PekkoProtocolHandle) extends NoSerializationVerificationNeeded
@@ -751,8 +767,10 @@ private[remote] class EndpointWriter(
   }
 
   val buffering: Receive = {
-    case s: Send      => enqueueInBuffer(s)
-    case BackoffTimer => sendBufferedMessages()
+    case s: Send       => enqueueInBuffer(s)
+    case BackoffTimer  => sendBufferedMessages()
+    case ResumeReading =>
+      reader.foreach(_ ! ResumeReading)
     case FlushAndStop =>
       // Flushing is postponed after the pending writes
       buffer.offer(FlushAndStop)
@@ -803,6 +821,9 @@ private[remote] class EndpointWriter(
         false
       case s @ StopReading(_, replyTo) =>
         reader.foreach(_.tell(s, replyTo))
+        true
+      case ResumeReading =>
+        reader.foreach(_ ! ResumeReading)
         true
       case unexpected =>
         throw new IllegalArgumentException(s"Unexpected message type: ${unexpected.getClass}")
@@ -996,6 +1017,8 @@ private[remote] class EndpointWriter(
           // initializing, buffer and take care of it later when buffer is sent
           enqueueInBuffer(s)
       }
+    case ResumeReading =>
+      reader.foreach(_ ! ResumeReading)
     case TakeOver(newHandle, replyTo) =>
       // Shutdown old reader
       handle.foreach { _.disassociate("the association was replaced by a new one", log) }
@@ -1113,7 +1136,7 @@ private[remote] class EndpointReader(
     val receiveBuffers: ConcurrentHashMap[Link, ResendState])
     extends EndpointActor(localAddress, remoteAddress, transport, settings, codec) {
 
-  import EndpointWriter.{ OutboundAck, StopReading, StoppedReading }
+  import EndpointWriter.{ OutboundAck, ResumeReading, StopReading, StoppedReading }
 
   val provider = RARP(context.system).provider
   var ackedReceiveBuffer = new AckedReceiveBuffer[Message]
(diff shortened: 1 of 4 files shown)


cluster/src/multi-jvm/scala/org/apache/pekko/cluster/RestartNodeSpec.scala
```diff
@@ -46,7 +46,7 @@ object RestartNodeMultiJvmSpec extends MultiNodeConfig {
       pekko.cluster.downing-provider-class = org.apache.pekko.cluster.testkit.AutoDowning
       pekko.cluster.testkit.auto-down-unreachable-after = 5s
       pekko.cluster.allow-weakly-up-members = off
-      #pekko.remote.use-passive-connections = off
+      #pekko.remote.classic.use-passive-connections = off
```
pjfanning, line 49: please revert this line

He-Pin: This is right, I checked the reference.conf

pjfanning: if the original line is incorrect, please fix it in a standalone PR - this PR may possibly not be merged and a fix for incorrect config names will be approved as a standalone fix

He-Pin: but yes, maybe better revert

remote/src/main/scala/org/apache/pekko/remote/Remoting.scala
```diff
@@ -973,12 +979,77 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter)
...
+      resumableReader: ResumableReader,
+      readOnlyEndpoint: ActorRef,
+      handle: PekkoProtocolHandle): Boolean = {
+    val ResumableReader(writer, remoteAddress, uid) = resumableReader
+    val matchesCurrentHandle = remoteAddress == handle.remoteAddress && uid == handle.handshakeInfo.uid
+    if (!matchesCurrentHandle)
+      log.debug(
+        s"Dropping passive read handoff fallback for [$remoteAddress] with UID [$uid] from writable endpoint " +
+        s"[$writer] because read-only endpoint [$readOnlyEndpoint] now handles [${handle.remoteAddress}] " +
+        s"with UID [${handle.handshakeInfo.uid}]")
```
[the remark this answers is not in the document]
He-Pin: won't compile

remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
```diff
@@ -418,10 +421,18 @@ private[remote] class ReliableDeliverySupervisor(
...
       } else goToIdle()
-    case AttemptSysMsgRedelivery                => // Ignore
-    case s @ Send(_: SystemMessage, _, _, _)    => tryBuffer(s.copy(seqOpt = Some(nextSeq())))
-    case s: Send                                => context.system.deadLetters ! s
-    case EndpointWriter.FlushAndStop            => context.stop(self)
+    case AttemptSysMsgRedelivery             => // Ignore
+    case s @ Send(_: SystemMessage, _, _, _) => tryBuffer(s.copy(seqOpt = Some(nextSeq())))
+    case s: Send                             => context.system.deadLetters ! s
+    case EndpointWriter.FlushAndStop         => context.stop(self)
+    case EndpointWriter.ResumeReading        =>
```
He-Pin, line 428: There was no such case, so it can never resume.

remote/src/main/scala/org/apache/pekko/remote/Remoting.scala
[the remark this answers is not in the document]
He-Pin: Fixed in a29905fb66. The debug message is now guarded with log.isDebugEnabled so the interpolated string is only built when debug logging is enabled. I kept it as a triple-quoted string because LoggingAdapter.debug only has overloads up to four arguments here.

cluster/src/multi-jvm/scala/org/apache/pekko/cluster/RestartNodeSpec.scala
He-Pin: Reverted in a29905fb66. RestartNodeSpec is no longer part of the PR diff.

pjfanning (approved):
lgtm

PR #3262, opened 2026-06-28
fix: suppress dead-letter logging during a terminating shutdown when disabled (#3256)

### LINKED ISSUE #3256: log-dead-letters-during-shutdown=off not respected during CoordinatedShutdown
### Motivation

When `log-dead-letters-during-shutdown = off`, dead letters are still logged throughout the entire CoordinatedShutdown process. The `DeadLetterListener` is only stopped at the very end in `finalTerminate()`, but during all shutdown phases (potentially seconds to minutes), it remains active and logs unconditionally.

### Current Behavior

**Shutdown sequence** (`ActorSystem.scala:1110-1138`):

1. Line 1111: `terminating = true` — set immediately
2. Line 1120: `CoordinatedShutdown(this).run(...)` — runs all shutdown phases
3. **During all phases**: actors stop, messages become dead letters, `DeadLetterListener` is **still subscribed and logging**
4. Phase `actor-system-terminate`: finally calls `finalTerminate()`
5. Line 1136: listener stopped — **only now**, at the very tail end

**`DeadLetterListener` has no shutdown awareness** (`DeadLetterListener.scala:73-79`):

```scala
private def receiveWithAlwaysLogging: Receive = {
  case d: AllDeadLetters =>
    if (!isWrappedSuppressed(d)) {
      incrementCount()
      logDeadLetter(d, doneMsg = "")
    }
}
```

No check for `isTerminating()` or the `LogDeadLettersDuringShutdown` setting. All receive variants log unconditionally until the actor is stopped.

### Expected Behavior

The `DeadLetterListener` should check `context.system.isTerminating()` (which returns `true` from the moment `terminate()` is called) and suppress logging when `LogDeadLettersDuringShutdown = off`, throughout the entire shutdown process — not just after `finalTerminate`.

### Environment

- Pekko Actor (any version)
- `ActorSystem.scala:1133-1138`, `DeadLetterListener.scala:73-79`

### References

None

He-Pin:
### Motivation

When `pekko.log-dead-letters-during-shutdown = off` (which is the **default**,
`reference.conf:63`), dead letters were still logged throughout the entire
`CoordinatedShutdown` sequence. `DeadLetterListener` had no shutdown awareness
and was only stopped at the very end of shutdown (`finalTerminate`), so every
dead letter produced while the system was terminating — potentially for the
whole (and sometimes lengthy) shutdown — was logged unconditionally. The
setting was effectively ignored until the very last moment.

Fixes #3256.

### Modification

`DeadLetterListener` now suppresses dead-letter logging while a coordinated
shutdown **that actually terminates this `ActorSystem`** is in progress, when
`log-dead-letters-during-shutdown` is disabled.

Crucially, whether the shutdown terminates the system is taken from the
**effective per-reason** configuration, using the exact same resolution that
`CoordinatedShutdown` itself uses
(`CoordinatedShutdown.confWithOverrides(conf, shutdownReason()).getBoolean("terminate-actor-system")`).
This matters because `terminate-actor-system` can be overridden per shutdown
reason. A coordinated shutdown that does **not** terminate the system (globally
via `terminate-actor-system = off`, or only for a specific reason via
`reason-overrides`) leaves a shutdown reason set on a still-running system —
that must **not** suppress logging, otherwise dead-letter logging would be
silently disabled for the remaining lifetime of the running system.

The `CoordinatedShutdown` extension reference is cached and the effective
decision is memoized (computed at most once — the shutdown reason is set once
and never changes), so the per-dead-letter path stays allocation-free and cheap
(a volatile read of the shutdown reason during normal operation, no extension
lookup).

### Result

- Dead letters are no longer logged during a **terminating** shutdown when
  `log-dead-letters-during-shutdown = off` — the setting is now honored for the
  whole shutdown, not just after the listener is finally stopped.
- Logging continues normally for a **non-terminating** coordinated shutdown
  (globally or per shutdown reason) and during normal operation.
- All new members are `private`; **no public API or binary-compatibility
  change** — `actor / mimaReportBinaryIssues` is clean.

### Tests

- `sbt "actor-tests/testOnly org.apache.pekko.event.DeadLetterListenerShutdownSpec"`
  → 4 passed, plus `DeadLetterSupressionSpec` and `DeadLetterSuspensionSpec`
  (no regression).
- Directional tests (a paused first `CoordinatedShutdown` phase keeps the
  listener alive mid-shutdown) that **fail before the fix**:
  - a terminating shutdown must suppress the dead letter;
  - a non-terminating shutdown (global `terminate-actor-system = off`) must keep
    logging;
  - a non-terminating shutdown via a **per-reason override** must keep logging
    (this one fails against an earlier base-setting-only gate);
  - logging must continue even after a non-terminating run completes.

### References

Fixes #3256

---

This is an original contribution to Apache Pekko, made under the Apache License
2.0 (i.e. the changes are now Apache licensed). No third-party or Akka-derived
code is included.

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/event/DeadLetterListenerShutdownSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/event/DeadLetterListenerShutdownSpec.scala
@@ -1,0 +1,185 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.event
+
+import scala.concurrent.Await
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.concurrent.duration._
+
+import com.typesafe.config.ConfigFactory
+
+import org.apache.pekko
+import pekko.Done
+import pekko.actor.Actor
+import pekko.actor.ActorRef
+import pekko.actor.ActorSystem
+import pekko.actor.CoordinatedShutdown
+import pekko.actor.DeadLetter
+import pekko.actor.Props
+import pekko.testkit.PekkoSpec
+import pekko.testkit.TestDuration
+import pekko.testkit.TestKit
+import pekko.testkit.TestProbe
+
+object DeadLetterListenerShutdownSpec {
+  case object RunningMsg
+  case object ShutdownMsg
+
+  // A shutdown reason whose `terminate-actor-system` is overridden to `off` in the test config, so a
+  // CoordinatedShutdown run for this reason executes its phases without terminating the system.
+  case object NonTerminatingReason extends CoordinatedShutdown.Reason
+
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3263, opened 2026-06-28
fix: expand Array arguments in MarkerLoggingAdapter log templates (#3257)

### LINKED ISSUE #3257: MarkerLoggingAdapter.format1 does not expand Array arguments into template placeholders
### Motivation

`MarkerLoggingAdapter.format1` calls `format(t, a.toIndexedSeq)` without varargs expansion (`: _*`), causing Array arguments to be treated as a single template argument instead of being expanded into individual placeholders.

### Current Behavior

`actor/src/main/scala/org/apache/pekko/event/Logging.scala:2004-2008`:

```scala
private def format1(t: String, arg: Any): String = arg match {
    case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq)
    case a: Array[?]                                             => format(t, a.toIndexedSeq.asInstanceOf[IndexedSeq[AnyRef]])
    case x                                                       => format(t, x)
}
```

This calls the varargs `format(t: String, arg: Any*)` (line 1538), which wraps the `IndexedSeq` as a **single element**: `arg = Seq(IndexedSeq("a","b","c"))`.

**Comparison with base class** (`LoggingAdapter.format1`, line 1532-1536):

```scala
private def format1(t: String, arg: Any): String = arg match {
    case a: Array[?] if !a.getClass.getComponentType.isPrimitive => formatImpl(t, a.toSeq)
    case a: Array[?]                                             => formatImpl(t, a.toSeq.asInstanceOf[Seq[AnyRef]])
    case x                                                       => format(t, x)
}
```

The base class calls `formatImpl` directly (not through varargs), so each array element becomes a separate placeholder.

### Concrete Example

For `markerLog.info(marker, "{} {} {}", Array("a", "b", "c"))`:

| | Base LoggingAdapter | MarkerLoggingAdapter |
|---|---|---|
| Output | `"a b c"` | `"Vector(a, b, c) {} {}"` |

### Root Cause

Line 2003 comments: *"Copy of LoggingAdapter.format1 due to binary compatibility restrictions"*. The base class's `formatImpl` is `private` to the trait, so `MarkerLoggingAdapter` substituted `format(t, ...)` but forgot `: _*` for varargs expansion.

### Fix

```scala
case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq: _*)
case a: Array[?]                                             => format(t, a.toIndexedSeq.asInstanceOf[IndexedSeq[AnyRef]]: _*)
```

### Scope

All 8 one-argument template methods in `MarkerLoggingAdapter` are affected (each log level × with-cause and without-cause variants). `DiagnosticMarkerBusLoggingAdapter` inherits the same bug. No test covers Array argument handling.

### Environment

- Pekko Actor (any version)
- `Logging.scala:2004-2008`

### References

None

He-Pin:
### Motivation

`MarkerLoggingAdapter.format1` passed an `Array` argument to the varargs
`format(t, arg: Any*)` **without** spreading it (`: _*`), so the whole array was
treated as a single template argument instead of being expanded into the
individual placeholders.

For example:

```scala
markerLog.info(LogMarker.Security, "{} {} {}", Array("a", "b", "c"))
```

| | base `LoggingAdapter` | `MarkerLoggingAdapter` (before) |
|---|---|---|
| Output | `a b c` | `ArraySeq(a, b, c) {} {}` |

The base `LoggingAdapter.format1` expands arrays correctly (it calls
`formatImpl(t, a.toSeq)` directly). The marker copy exists for binary
compatibility reasons (it cannot reach the base trait's `private formatImpl`),
and when it was written the varargs spread (`: _*`) was forgotten. The scaladoc
promising that an `Array` argument "will be expanded into replacement
arguments" was therefore not honored by the marker adapter.

Fixes #3257.

### Modification

Spread the array elements as varargs in both `Array` branches of
`MarkerLoggingAdapter.format1`, so each element fills its own placeholder, and
documented the reason so it is not regressed:

```scala
case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq: _*)
case a: Array[?]                                             => format(t, a.toIndexedSeq.asInstanceOf[IndexedSeq[AnyRef]]: _*)
case x                                                       => format(t, x)
```

`DiagnosticMarkerBusLoggingAdapter` inherits this method and is fixed by the
same change.

### Result

- `Array` arguments to the single-argument marker log template methods are now
  expanded into separate placeholders, matching `LoggingAdapter` and the
  documented behavior (e.g. `"a b c"` instead of `"ArraySeq(a, b, c) {} {}"`).
- `format1` is `private` and only its body changed — **no public API or
  binary-compatibility change** (`actor / mimaReportBinaryIssues` is clean).

### Tests

- `sbt "actor-tests/testOnly org.apache.pekko.event.MarkerLoggingSpec"` →
  all passed.
- Added directional tests for a non-primitive `Array[String]` and a primitive
  `Array[Int]` (the two `format1` branches). Both **fail before the fix**
  (rendering `"ArraySeq(a, b, c) {} {}"` / `"ArraySeq(1, 2, 3) {} {}"`) and pass
  after.

### References

Fixes #3257

---

This is an original contribution to Apache Pekko, made under the Apache License
2.0 (i.e. the changes are now Apache licensed). No third-party or Akka-derived
code is included.

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/event/Logging.scala
+++ b/actor/src/main/scala/org/apache/pekko/event/Logging.scala
@@ -2000,10 +2000,13 @@ class MarkerLoggingAdapter(
     }
   }
 
-  // Copy of LoggingAdapter.format1 due to binary compatibility restrictions
+  // Copy of LoggingAdapter.format1 due to binary compatibility restrictions.
+  // The array branches must spread the elements as varargs (`: _*`) so that each element is
+  // expanded into its own template placeholder; without it the whole IndexedSeq would be passed
+  // as a single argument and rendered as e.g. "ArraySeq(a, b, c)" (issue #3257).
   private def format1(t: String, arg: Any): String = arg match {
-    case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq)
-    case a: Array[?]                                             => format(t, a.toIndexedSeq.asInstanceOf[IndexedSeq[AnyRef]])
+    case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq: _*)
+    case a: Array[?]                                             => format(t, a.toIndexedSeq.asInstanceOf[IndexedSeq[AnyRef]]: _*)
     case x                                                       => format(t, x)
   }
 }
--- a/actor-tests/src/test/scala/org/apache/pekko/event/MarkerLoggingSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/event/MarkerLoggingSpec.scala
@@ -44,5 +44,23 @@ class MarkerLoggingSpec extends PekkoSpec with ImplicitSender {
 
       expectMsgType[Error].cause.getMessage should be("Security Exception")
     }
+
+    "expand a non-primitive Array argument into separate template placeholders (#3257)" in {
+      system.eventStream.subscribe(self, classOf[Info])
+      system.eventStream.publish(TestEvent.Mute(EventFilter.info()))
+
+      markerLogging.info(LogMarker.Security, "{} {} {}", Array("a", "b", "c"))
+
+      expectMsgType[Info3].message should be("a b c")
+    }
+
+    "expand a primitive Array argument into separate template placeholders (#3257)" in {
+      system.eventStream.subscribe(self, classOf[Info])
+      system.eventStream.publish(TestEvent.Mute(EventFilter.info()))
+
+      markerLogging.info(LogMarker.Security, "{} {} {}", Array(1, 2, 3))
+
+      expectMsgType[Info3].message should be("1 2 3")
+    }
   }
 }


pjfanning:
Is there any chance that since the existing behaviour has been like that for a long time, that we just accept that that is the behaviour? Where in the docs does it say that this change is what users should expect?

He-Pin:
The problem is this ,it change the default behavior.

actor/src/main/scala/org/apache/pekko/event/Logging.scala
```diff
@@ -2000,10 +2000,13 @@ class MarkerLoggingAdapter(
...
 
-  // Copy of LoggingAdapter.format1 due to binary compatibility restrictions
+  // Copy of LoggingAdapter.format1 due to binary compatibility restrictions.
+  // The array branches must spread the elements as varargs (`: _*`) so that each element is
+  // expanded into its own template placeholder; without it the whole IndexedSeq would be passed
+  // as a single argument and rendered as e.g. "ArraySeq(a, b, c)" (issue #3257).
   private def format1(t: String, arg: Any): String = arg match {
-    case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq)
-    case a: Array[?]                                             => format(t, a.toIndexedSeq.asInstanceOf[IndexedSeq[AnyRef]])
+    case a: Array[?] if !a.getClass.getComponentType.isPrimitive => format(t, a.toIndexedSeq: _*)
```
He-Pin, line 2008: Not sure if it's safe to change ,let's defer this.

pjfanning:
Actually, the javadoc says
```
 /**
   * Message template with 1 replacement argument.
   * The marker argument can be picked up by various logging frameworks such as slf4j to mark this log statement as "special".
   *
   * If `arg1` is an `Array` it will be expanded into replacement arguments, which is useful when
   * there are more than four arguments.
   * @see [[LoggingAdapter]]
   */
  def error(marker: LogMarker, cause: Throwable, template: String, arg1: Any): Unit =
    if (isErrorEnabled(marker)) bus.publish(Error(cause, logSource, logClass, format1(template, arg1), mdc, marker))
```

So your change is probably correct.

pjfanning (approved):
lgtm

PR #3268, opened 2026-06-28
fix: return proper non-cancelled Cancellable from StubbedActorContext.scheduleOnce

### LINKED ISSUE #3251: StubbedActorContext.scheduleOnce returns pre-cancelled Cancellable
### Motivation

`StubbedActorContext.scheduleOnce` returns a `Cancellable` with `isCancelled = true` and `cancel()` returning `false`, i.e., it is pre-cancelled. This violates the `Cancellable` contract and causes test/production divergence.

### Current Behavior

`StubbedActorContext.scala:162-166`:

```scala
override def scheduleOnce[U](delay: FiniteDuration, target: ActorRef[U], message: U): classic.Cancellable =
  new classic.Cancellable {
    override def cancel() = false
    override def isCancelled = true
  }
```

The `Cancellable` contract (`Scheduler.scala:458-472`) states:
- `isCancelled`: "Returns true **if and only if** this Cancellable has been successfully cancelled"
- `cancel()`: "Returns true if cancellation was successful; false if already cancelled"

A just-scheduled task has NOT been cancelled, so `isCancelled` should be `false`.

### Production vs Test Divergence

The real implementation (`ActorContextAdapter.scala:139-141`) delegates to the real scheduler and returns a genuine `Cancellable` where `isCancelled` starts as `false`. An actor that checks `cancellable.isCancelled` will take a different code path in tests vs. production.

### Expected Behavior

Return a `Cancellable` that starts non-cancelled and can be properly cancelled:

```scala
new classic.Cancellable {
  private var _cancelled = false
  override def cancel(): Boolean = {
    if (!_cancelled) { _cancelled = true; true } else false
  }
  override def isCancelled: Boolean = _cancelled
}
```

### Environment

- Pekko Actor TestKit Typed (any version)
- `StubbedActorContext.scala:162-166`

### References

None

He-Pin:
### Motivation

`StubbedActorContext.scheduleOnce` returned a `Cancellable` with `isCancelled = true` and `cancel() = false`, i.e., it was pre-cancelled. This violates the `Cancellable` contract (`Scheduler.scala:458-472`):

- `isCancelled`: "Returns true **if and only if** this Cancellable has been successfully cancelled"
- `cancel()`: "Returns true if cancellation was successful; false if already cancelled"

A freshly scheduled task has NOT been cancelled, so `isCancelled` should start as `false`. The real implementation in `ActorContextAdapter` delegates to the real scheduler and returns a genuine `Cancellable` where `isCancelled` starts as `false`, causing test/production divergence.

### Modification

Replace the pre-cancelled `Cancellable` stub in `StubbedActorContext.scheduleOnce` with a thread-safe implementation that mixes `AtomicBoolean` directly into the `Cancellable`:

```scala
new AtomicBoolean(false) with classic.Cancellable {
  override def cancel(): Boolean = compareAndSet(false, true)
  override def isCancelled: Boolean = get()
}
```

- `compareAndSet(false, true)` ensures at most one successful cancellation even under concurrent access, matching the thread-safety best practice noted in the `Cancellable` trait Scaladoc
- Mixing `AtomicBoolean` directly avoids an extra field allocation — the object itself IS the atomic boolean

### Result

`StubbedActorContext.scheduleOnce` now returns a `Cancellable` that matches the `Cancellable` contract and the real production implementation behavior. Actors that check `cancellable.isCancelled` will now behave consistently in tests and production.

### Tests

- `sbt "actor-testkit-typed / Test / testOnly *BehaviorTestKitSpec *BehaviorTestKitTest"` — 52 passed, 1 ignored

### References

Fixes #3251

### DIFF
--- a/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/StubbedActorContext.scala
+++ b/actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/StubbedActorContext.scala
@@ -14,6 +14,7 @@
 package org.apache.pekko.actor.testkit.typed.internal
 
 import java.util.concurrent.ThreadLocalRandom.{ current => rnd }
+import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.collection.immutable.TreeMap
 import scala.concurrent.ExecutionContextExecutor
@@ -160,9 +161,9 @@ private[pekko] final class FunctionRef[-T](override val path: ActorPath, send: (
   }
 
   override def scheduleOnce[U](delay: FiniteDuration, target: ActorRef[U], message: U): classic.Cancellable =
-    new classic.Cancellable {
-      override def cancel() = false
-      override def isCancelled = true
+    new AtomicBoolean(false) with classic.Cancellable {
+      override def cancel(): Boolean = compareAndSet(false, true)
+      override def isCancelled: Boolean = get()
     }
 
   // TODO allow overriding of this
--- a/actor-testkit-typed/src/test/java/org/apache/pekko/actor/testkit/typed/javadsl/BehaviorTestKitTest.java
+++ b/actor-testkit-typed/src/test/java/org/apache/pekko/actor/testkit/typed/javadsl/BehaviorTestKitTest.java
@@ -61,6 +61,13 @@ public class BehaviorTestKitTest {
   public record AskForCookiesFrom(ActorRef<CookieDistributorCommand> distributor)
       implements Command {}
 
+  public record ScheduleOnceCommand(
+      Duration delay,
+      ActorRef<String> target,
+      String message,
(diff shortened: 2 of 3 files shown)


actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/StubbedActorContext.scala
```diff
@@ -161,8 +161,12 @@ private[pekko] final class FunctionRef[-T](override val path: ActorPath, send: (
 
   override def scheduleOnce[U](delay: FiniteDuration, target: ActorRef[U], message: U): classic.Cancellable =
     new classic.Cancellable {
-      override def cancel() = false
-      override def isCancelled = true
+      private var _cancelled = false
```
He-Pin: shuold use AtomicBoolean

pjfanning (approved):
lgtm

PR #3270, opened 2026-06-29
consistent-hashing shard allocation strategy

pjfanning:
Copy over some Akka 2.8.3 changes that are now available under Apache License, version 2.0.

https://github.com/akka/akka-core/releases/tag/v2.8.3

The changes in #3277 relate to these changes too

### ClusterShardAllocationMixin (PR 31967)
- Extracts shared allocation logic (`RegionEntry`, `ShardSuitabilityOrdering`, `AllocationMap`, `JoiningCluster`, `LeavingClusterStatuses`, `isAGoodTimeToRebalance`, `regionEntriesFor`) from `AbstractLeastShardAllocationStrategy` into a new `ClusterShardAllocationMixin` trait
- `AbstractLeastShardAllocationStrategy` companion object retains backward-compat type aliases delegating to the mixin
- Fixes two test cases in `LeastShardAllocationStrategySpec` / `DeprecatedLeastShardAllocationStrategySpec`: rolling-update members now use distinct IPs (was incorrectly reusing `127.0.0.1`); unreachable/joining-dc tests now include a third member and use `Set.empty` for unreachable where appropriate
- https://github.com/akka/akka-core/pull/31967

### ConsistentHashingShardAllocationStrategy (PR 31967)
- New `ConsistentHashingShardAllocationStrategy` that uses `pekko.routing.ConsistentHash` to assign shards, enabling co-location of related entity types sharing the same shard-ID scheme
- Extends `ActorSystemDependentAllocationStrategy with ClusterShardAllocationMixin` (does not inherit `AbstractLeastShardAllocationStrategy`)
- Compile-only docs tests added for both Scala and Java APIs in `cluster-sharding-typed`
- https://github.com/pjfanning/incubator-pekko/pull/58 describes the fix needed for this test based on the fact that the `pekko` protocol affects the hash algorithm compared with the `akka` protocol that the tests were originally written to support

```scala
ClusterSharding(system).init(
  Entity(Building.TypeKey)(createBehavior = entityContext => Building(entityContext.entityId))
    .withMessageExtractor(new Building.MessageExtractor)
    .withAllocationStrategy(new ConsistentHashingShardAllocationStrategy(rebalanceLimit = 10)))
```

### Translation notes
- All `akka.*` package/config references → `org.apache.pekko.*` / `pekko.*`
- `scala.compat.java8.FutureConverters` → `scala.jdk.FutureConverters` (`.toJava`/`.asJava`)
- `akka.util.JavaDurationConverters` → `scala.jdk.DurationConverters` (`.toScala`)
- `akka.util.ccompat.JavaConverters` → `scala.jdk.CollectionConverters`
- `@unused` (`akka.util.unused`) → `@nowarn("msg=never used")` (`scala.annotation`)


pjfanning:
Problem Analysis
The test is failing at line 84:

Code
Actor[pekko://myapp@127.0.0.1:252525/system/fake/regionA] did not equal Actor[pekko://myapp@127.0.0.2:252525/system/fake/regionB]
This indicates the consistent hashing is returning regionB when the test expects regionC for shard "0".

Root Cause
Looking at the test setup in ConsistentHashingShardAllocationStrategySpec:

Scala
val memberA = newUpMember("127.0.0.1")
val memberB = newUpMember("127.0.0.2")
val memberC = newUpMember("127.0.0.3")

val regionA = newFakeRegion("regionA", memberA)
val regionB = newFakeRegion("regionB", memberB)
val regionC = newFakeRegion("regionC", memberC)

private def strategy(rebalanceLimit: Int = 0) =
  new ConsistentHashingShardAllocationStrategy(rebalanceLimit) {
    override protected def clusterState: CurrentClusterState =
      CurrentClusterState(SortedSet(memberA, memberB, memberC))
    ...
  }

"allocate to regions" in {
  val allocationStrategy = strategy()
  val allocations = emptyAllocationsABC
  allocationStrategy.allocateShard(regionA, "0", allocations).futureValue should ===(regionC)
The issue is in how the consistent hash is computed. In ConsistentHashingShardAllocationStrategy.scala:

Scala
private def nodesForRegions(...): Vector[Address] = {
  currentShardAllocations.keysIterator.map(nodeForRegion).toVector
}

private def updateHashing(nodes: Vector[Address]): Unit = {
  val sortedNodes = nodes.sorted  // ← This is the issue
  ...
  consistentHashing = ConsistentHash(sortedNodes, virtualNodesFactor)
}
The problem is that nodes.sorted returns the nodes in an unpredictable order when passed to the consistent hash, and it depends on the Address ordering. The test expectations were written based on specific hash behavior that isn't guaranteed by the current implementation.

Solution
The ConsistentHash should be initialized with nodes in a deterministic order. The nodes should be sorted by their Address in a consistent way. Looking at the code, it uses Address.addressOrdering, but the sorted call without an explicit ordering may not be using it correctly.

The fix is to ensure the nodes are sorted using the same ordering consistently:

Scala
private def updateHashing(nodes: Vector[Address]): Unit = {
  val sortedNodes = nodes.sorted(Address.addressOrdering)  // Explicit ordering
  if (sortedNodes != hashedByNodes) {
    ...
However, there may be a deeper issue: the test expectations themselves may need to be adjusted to match the actual consistent hash output for these node IPs and shard IDs.

pjfanning:
I'm not yet sure why these tests fail. I don't see anything obvious in terms of diffs in the code with akka 2.8.3. I haven't had a lot of time to compare it and won't over the next few days. My AI budget is now very limited but the comment above is a basic Haiku take on the issue.

He-Pin:
ConsistentHashingShardAllocationStrategySpec.scala:84 is actual regionA but expecting regionB

He-Pin:
I will try to debug this with Ai agent first and send a PR to your repo

He-Pin:
On chat.qwen.ai we can use qwen 3.7 max to review the pr for free

He-Pin (approved):
LGTM.

cluster-sharding/src/main/mima-filters/2.0.x.backwards.excludes/consistent-hashing-allocation.excludes
```diff
@@ -0,0 +1,22 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Internal classes affected by refactor related to Collocating shards with consistent hashing
+ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.cluster.sharding.internal.AbstractLeastShardAllocationStrategy$")
+ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.cluster.sharding.internal.AbstractLeastShardAllocationStrategy$RegionEntry")
+ProblemFilters.exclude[MissingClassProblem]("org.apache.pekko.cluster.sharding.internal.AbstractLeastShardAllocationStrategy$RegionEntry$")
```
He-Pin, line 21: Nit: "Collocating" → "Colocating" — matches the spelling used in the docs added by this PR ("Colocate Shards", "Colocate processes").

cluster-sharding-typed/src/test/java/jdocs/org/apache/pekko/cluster/sharding/typed/ConsistentHashingShardAllocationCompileOnlyTest.java
```diff
@@ -0,0 +1,141 @@
...
+      @Override
+      public Command unwrapMessage(ShardingEnvelope<Command> envelope) {
+        return envelope.message();
+      }
+    }
+
+    static EntityTypeKey<Building.Command> typeKey =
+        EntityTypeKey.create(Building.Command.class, "Building");
+
+    public interface Command {}
```
He-Pin, line 57: Nit: `static EntityTypeKey<...> typeKey` could be `static final` for immutability. Same for `Device.typeKey` below. (Matches Akka source verbatim so non-blocking.)

He-Pin (approved):

He-Pin (approved):
LGTM. I did not find any must-fix issues in this diff. The only residual risk is the defensive IllegalStateException paths in the new allocation strategy, but they do not block merge.

PR #3272, opened 2026-06-30
docs: document preMaterialize error-propagation caveat

He-Pin:
### Motivation

`preMaterialize` crosses a reactive-streams `Publisher`/`Subscriber` boundary, which introduces a buffer and turns upstream failures into cancellations without error details. This surprising behaviour was previously undocumented, so users reasonably expected error propagation through a pre-materialized `Source`/`Sink`/`Flow`.

### Modification

Add the caveat to the scaladoc of every `preMaterialize` overload on `Source`/`Sink`/`Flow` in both the Scala and Java DSLs, and to the `Sink` and `Source-or-Flow` paradox operator pages:

- Scala DSL: `Source`, `Sink`, `Flow`
- Java DSL: `Source` (×2), `Sink` (×2), `Flow` (×2)
- Paradox: `Sink/preMaterialize.md`, `Source-or-Flow/preMaterialize.md`

The wording follows the DSL primitive used by each operator: `Source` → `Publisher`, `Sink` → `Subscriber`, `Flow` → `Publisher` and `Subscriber` pair. For consistency the caveat is also added to the Java DSL `materializer` overloads of `Sink`/`Source`, which had no direct upstream counterpart (Akka documented only the `ActorSystem` overloads there); both overloads delegate to the same `scaladsl.preMaterialize()`, so the caveat applies equally.

### Result

Users are clearly warned that `preMaterialize` buffers elements and does not propagate upstream errors, avoiding surprises when a pre-materialized graph fails.

which is now Apache licensed

### Tests

Not run - docs only (scaladoc + paradox). The change is mechanically comment-only and was validated with native `scalafmt` (3.11.1).

### References

Ports [akka/akka-core@b7e625fb](https://github.com/akka/akka-core/commit/b7e625fb1274323afb9e8c5f0c2f4c9560cc4da1) (akka#31982), which is now Apache licensed.

### DIFF
--- a/docs/src/main/paradox/stream/operators/Sink/preMaterialize.md
+++ b/docs/src/main/paradox/stream/operators/Sink/preMaterialize.md
@@ -13,3 +13,6 @@ Materializes this Sink, immediately returning (1) its materialized value, and (2
 
 Materializes this Sink, immediately returning (1) its materialized value, and (2) a new Sink that can be consume elements 'into' the pre-materialized one. Useful for when you need a materialized value of a Sink when handing it out to someone to materialize it for you.
 
+Note that `preMaterialize` is implemented through a reactive streams `Subscriber` which means that a buffer is introduced
+and that errors are not propagated upstream but are turned into cancellations without error details.
+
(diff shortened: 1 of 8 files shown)


pjfanning (approved):
Lgtm

PR #3273, opened 2026-06-30
Close unregistered TCP channels on setup failure

### LINKED ISSUE #3246: TCP connection file descriptor leak when setup fails before ChannelRegistration
### Description

`TcpOutgoingConnection` leaks the `SocketChannel` file descriptor when the connection setup fails before the `ChannelRegistration` message is received from the selector.

### Root cause

In `TcpOutgoingConnection.scala` (lines 54-72), the `SocketChannel` is opened in the constructor (line 44) before the actor starts processing messages. If any operation before `channelRegistry.register()` throws (e.g., a user-provided `SocketOption.beforeConnect`), `reportConnectFailure` catches the exception and calls `stop(e)`.

At this point, `registration` is still `None` because the `ChannelRegistration` message has not been received yet.

In `TcpConnection.stopWith` (lines 390-404):

```scala
registration match {
  case None =>
    context.stop(self)   // channel NOT closed!
  case Some(reg) =>
    context.become(unregistering)
    reg.cancelAndClose(() => self ! Unregistered)  // channel IS closed
}
```

When `registration` is `None`, `stopWith` only calls `context.stop(self)` without closing the channel.

`postStop` (lines 406-429) also fails to close the channel because the `isCommandFailed` condition is true, routing to the `notifyInterested()` branch instead of the `cancelAndClose` branch.

### Impact

The `SocketChannel` file descriptor is never explicitly closed. It relies on GC finalization (via `AbstractInterruptibleChannel`'s cleaner) to eventually close, which is non-deterministic and can lead to file descriptor exhaustion under high connection-failure rates.

### Existing test coverage gap

`TcpIntegrationSpec.scala` (lines 187-201) tests the "reply with CommandFailed when a connect socket option fails before connect" scenario but does **not** verify that the underlying `SocketChannel` was closed.

### Suggested fix

In `TcpConnection.postStop`, add an explicit `channel.close()` call when `registration` is `None` and the channel is still open:

```scala
  if (channel.isOpen) {
    if (registration.isEmpty)
      channel.close()  // close unregistered channels directly
    else
      prepareAbort()
  }
  // ... rest of cleanup
}
```

goutamadwant:
### Motivation
Outgoing TCP connection setup can fail before the connection actor receives a `ChannelRegistration`. In that path, `TcpConnection.postStop` reported `CommandFailed` but left the `SocketChannel` open because there was no registration callback available to close it.

### Modification
Close still-open channels directly from `postStop` when no `ChannelRegistration` exists. Registered channels continue to use the existing `cancelAndClose` path.

Also extend the existing connect socket option failure test to verify that the socket captured before the failed `beforeConnect` call is closed after `CommandFailed`.

### Result
TCP setup failures before `ChannelRegistration` no longer leave the underlying `SocketChannel` open.

### Tests
- `sbt "actor-tests / Test / testOnly org.apache.pekko.io.TcpIntegrationSpec"`
- `sbt "actor / Compile / scalafmtCheck" "actor-tests / Test / scalafmtCheck"`
- `git diff --check`

### References
Fixes #3246

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/io/TcpIntegrationSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/io/TcpIntegrationSpec.scala
@@ -15,6 +15,7 @@ package org.apache.pekko.io
 
 import java.io.IOException
 import java.net.{ InetSocketAddress, ServerSocket, Socket }
+import java.util.concurrent.atomic.AtomicReference
 
 import scala.concurrent.duration._
 
@@ -187,10 +188,10 @@ class TcpIntegrationSpec extends PekkoSpec("""
     "reply with CommandFailed when a connect socket option fails before connect" in {
       val connectCommander = TestProbe()
       val failure = new UnsupportedOperationException("boom")
-      @volatile var socket: Socket = null
+      val socket = new AtomicReference[Socket]
       val failingOption = new Inet.SocketOption {
         override def beforeConnect(s: Socket): Unit = {
-          socket = s
+          socket.set(s)
           throw failure
         }
       }
@@ -202,8 +203,8 @@ class TcpIntegrationSpec extends PekkoSpec("""
       val commandFailed = connectCommander.expectMsgType[CommandFailed]
       commandFailed.cmd should ===(command)
       commandFailed.cause should ===(Some(failure))
-      socket should not be null
-      awaitCond(socket.isClosed)
+      socket.get should not be null
+      awaitCond(socket.get.isClosed)
     }
 
     "handle tcp connection actor death properly" in new TestSetup(shouldBindServer = false) {
--- a/actor-tests/src/test/scala/org/apache/pekko/io/TcpIntegrationSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/io/TcpIntegrationSpec.scala
@@ -187,8 +187,12 @@ class TcpIntegrationSpec extends PekkoSpec("""
     "reply with CommandFailed when a connect socket option fails before connect" in {
       val connectCommander = TestProbe()
       val failure = new UnsupportedOperationException("boom")
+      @volatile var socket: Socket = null
       val failingOption = new Inet.SocketOption {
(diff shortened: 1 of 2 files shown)


He-Pin:
Thanks

He-Pin (approved):

actor-tests/src/test/scala/org/apache/pekko/io/TcpIntegrationSpec.scala
```diff
@@ -187,8 +187,12 @@ class TcpIntegrationSpec extends PekkoSpec("""
     "reply with CommandFailed when a connect socket option fails before connect" in {
       val connectCommander = TestProbe()
       val failure = new UnsupportedOperationException("boom")
+      @volatile var socket: Socket = null
```
He-Pin: better with atomicRefernece

actor/src/main/scala/org/apache/pekko/io/TcpConnection.scala
```diff
@@ -410,8 +410,15 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
...
       closedMessage.toList.flatMap(_.notificationsTo).toSet
 
-    if (channel.isOpen) // if channel is still open here, we didn't go through stopWith => unexpected actor termination
+    if (channel.isOpen) {
       prepareAbort()
+      // Without a ChannelRegistration there is no selector callback to close through.
+      if (registration.isEmpty)
+        try channel.close()
+        catch {
+          case NonFatal(e) => log.debug("Error closing SocketChannel: {}", e)
```
He-Pin: Error when closing?, and the `{}` should not be there and it should be  an error

He-Pin (commented):
@goutamadwant Please update with my comments ,thank you

He-Pin (changes requested):

goutamadwant:
Thanks @He-Pin addressed both the comments. Let me know if you have any suggestions.

He-Pin (approved):

PR #3274, opened 2026-06-30
docs: how to write a custom Durable State store plugin

He-Pin:
### Motivation

Pekko Persistence supports pluggable Durable State storage backends, but there was no documentation explaining how to implement one. Plugin authors had to read the source to discover the `DurableStateUpdateStore` and `DurableStateStoreProvider` SPI.

### Modification

Add a "Building a storage backend for Durable State" paradox page with Scala and Java examples (plugin API, provider, and minimal activation configuration), backed by a runnable doc spec. Also fix a stale scaladoc on `DurableStateStoreProvider.scaladslDurableStateStore` that referred to `ReadJournal` instead of `DurableStateStore`.

New files:
- `docs/src/main/paradox/durable-state/state-store-plugin.md` (+ index link)
- `docs/src/main/scala/docs/persistence/state/MyStateStore.scala`
- `docs/src/main/java/docs/persistence/state/MyJavaStateStore.java`, `MyJavaStateStoreProvider.java`
- `docs/src/test/scala/docs/persistence/state/PersistenceStatePluginDocSpec.scala`

**Improvements over the upstream example:**
- `MyJavaStateStore.upsertObject` uses the generic type parameter `A` instead of `Object`, so the snippet is a correct, type-safe template when adapted to a concrete state type.
- The provider uses the diamond operator instead of raw types.
- Fixed an upstream grammar typo ("described" → "describes").

**Intentional divergence from upstream:** Pekko's `DurableStateUpdateStore` already documents `@param revision` (richer than the upstream `@param seqNr` fix), and the upstream `//#plugin-api` markers on the real trait are not referenced by the docs (the snippet is taken from the example file), so they are omitted to avoid leaking rendering markers into production sources.

### Result

Durable State plugin authors have a step-by-step guide with compiling, tested Scala and Java examples.

which is now Apache licensed

### Tests

- `sbt "docs/testOnly docs.persistence.state.PersistenceStatePluginDocSpec"` (2/2 passed)
- `sbt "docs/paradox"` (success; all snippets resolve)

### References

Ports [akka/akka-core@1ce8748](https://github.com/akka/akka-core/commit/1ce8748239e968f1fba15d23213130e5be98945b) (akka#31973), which is now Apache licensed.

### DIFF
--- a/docs/src/main/scala/docs/persistence/state/MyStateStore.scala
+++ b/docs/src/main/scala/docs/persistence/state/MyStateStore.scala
@@ -13,12 +13,15 @@
 
 package docs.persistence.state
 
-import org.apache.pekko.Done
-import org.apache.pekko.actor.ExtendedActorSystem
-import org.apache.pekko.persistence.state.{ DurableStateStoreProvider, DurableStateStoreRegistry }
-import org.apache.pekko.persistence.state.scaladsl.{ DurableStateStore, DurableStateUpdateStore, GetObjectResult }
+import org.apache.pekko
+import pekko.Done
+import pekko.actor.ExtendedActorSystem
+import pekko.persistence.state.{ DurableStateStoreProvider, DurableStateStoreRegistry }
+import pekko.persistence.state.javadsl.{ DurableStateStore => JDurableStateStore }
+import pekko.persistence.state.scaladsl.{ DurableStateStore, DurableStateUpdateStore, GetObjectResult }
+
 import com.typesafe.config.Config
-import org.apache.pekko.persistence.state.javadsl.{ DurableStateStore => JDurableStateStore }
+
 import scala.concurrent.Future
 
 //#plugin-provider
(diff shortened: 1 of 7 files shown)


docs/src/main/scala/docs/persistence/state/MyStateStore.scala
```diff
@@ -0,0 +1,68 @@
...
+ * This file is part of the Apache Pekko project, which was derived from Akka.
+ */
+
+/*
+ * Copyright (C) 2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package docs.persistence.state
+
+import org.apache.pekko.Done
```
pjfanning: in the 2 new scala files in this PR, could you import org.apache.pekko and change the others to start with `import pekko.`?

docs/src/main/scala/docs/persistence/state/MyStateStore.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.pekko.actor.ExtendedActorSystem
+import org.apache.pekko.persistence.state.{ DurableStateStoreProvider, DurableStateStoreRegistry }
+import org.apache.pekko.persistence.state.scaladsl.{ DurableStateStore, DurableStateUpdateStore, GetObjectResult }
+import com.typesafe.config.Config
+import org.apache.pekko.persistence.state.javadsl.{ DurableStateStore => JDurableStateStore }
```
pjfanning: could you move the `com.typesafe.config.Config` line so the Pekko imports are together?

pjfanning (approved):
lgtm

PR #3275, opened 2026-06-30
feat: add FilteredPayload placeholder for filtered journal events

He-Pin:
### Motivation

Projections and events-by-slice queries need to keep per-persistence-id sequence numbers gap free, which means filtered events must still be stored in the journal. Pekko Persistence had no shared placeholder payload for this, so query/projection plugins (e.g. r2dbc, projections) had nothing to align on.

### Modification

- Add the public `FilteredPayload` case object, serialized to a 0-byte representation by the new internal `FilteredPayloadSerializer` (manifest `"F"`, serializer id `34`).
- Register the serializer, binding and identifier in persistence `reference.conf`.
- Make typed `EventEnvelope.event`/`getEvent` raise a clearer error when the payload is absent because the event was *filtered* (vs. simply *not loaded*).

**Serializer id:** `34` matches the upstream id and follows pekko's established convention of preserving upstream serializer ids (e.g. `ReplicatedEventSourcingSerializer = 40` matches akka). Verified free within pekko's documented `0-40` reserved range.

**Intentional divergence from upstream:** the upstream `akka-persistence-typed` reference.conf change was a cosmetic HOCON reformat (dotted key → block form) and is skipped; `scripts/show-serializer.sh` does not exist in pekko and is skipped.

**Improvement over upstream:** added an `EventEnvelopeSpec` directional test covering the `filtered` vs `not loaded` error behaviour, which upstream shipped without a test.

### Result

Pekko Persistence has a standard placeholder for filtered journal events, flagged as `filtered` in the typed `EventEnvelope`, enabling query and projection plugins to represent gap-free sequence numbers.

which is now Apache licensed

### Tests

- `sbt "persistence/testOnly org.apache.pekko.persistence.serialization.FilteredPayloadSerializerSpec"` (1/1 passed)
- `sbt "persistence-query/testOnly org.apache.pekko.persistence.query.typed.EventEnvelopeSpec"` (2/2 passed)
- `sbt "persistence/mimaReportBinaryIssues" "persistence-query/mimaReportBinaryIssues"` (clean)

### References

Ports [akka/akka-core@b179bb1](https://github.com/akka/akka-core/commit/b179bb15279962d657b9192f374a98123fabd7ee) (akka#31985), which is now Apache licensed.

### DIFF
--- a/persistence-query/src/test/scala/org/apache/pekko/persistence/query/typed/EventEnvelopeSpec.scala
+++ b/persistence-query/src/test/scala/org/apache/pekko/persistence/query/typed/EventEnvelopeSpec.scala
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.pekko.persistence.query.typed
+
+import org.apache.pekko
+import pekko.persistence.query.NoOffset
+import pekko.testkit.PekkoSpec
+
+class EventEnvelopeSpec extends PekkoSpec {
+
+  private def envelope(filtered: Boolean): EventEnvelope[String] =
+    new EventEnvelope[String](
+      NoOffset,
+      persistenceId = "pid-1",
+      sequenceNr = 1L,
+      eventOption = None,
+      timestamp = 0L,
+      eventMetadata = None,
+      entityType = "entity",
+      slice = 0,
+      filtered = filtered,
+      source = "")
+
+  "EventEnvelope" should {
+
+    "throw a filtered-specific error from event/getEvent when the payload was filtered out" in {
+      val env = envelope(filtered = true)
+      intercept[IllegalStateException](env.event).getMessage should include("filtered")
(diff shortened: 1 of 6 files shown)


persistence/src/main/resources/reference.conf
```diff
@@ -233,14 +233,17 @@ pekko.actor {
...
     }
     serialization-bindings {
         "org.apache.pekko.persistence.serialization.Message" = pekko-persistence-message
         "org.apache.pekko.persistence.serialization.Snapshot" = pekko-persistence-snapshot
+        "org.apache.pekko.persistence.FilteredPayload$" = pekko-persistence-filtered
     }
     serialization-identifiers {
         "org.apache.pekko.persistence.serialization.MessageSerializer" = 7
         "org.apache.pekko.persistence.serialization.SnapshotSerializer" = 8
+        "org.apache.pekko.persistence.serialization.FilteredPayloadSerializer" = 34
```
He-Pin, line 246: Should we keep this the same as akka?

persistence-query/src/test/scala/org/apache/pekko/persistence/query/typed/EventEnvelopeSpec.scala
```diff
@@ -0,0 +1,53 @@
+/*
```
pjfanning, line 1: if this is copied from akka, we will need the Lightbend copyright header and the special Apache header that mentions Akka (the most common source header in Pekko code)

He-Pin: This file is not copied from Akka — it is a new directional test I wrote for this PR (as noted in the PR body under "Improvement over upstream: added an EventEnvelopeSpec directional test covering the filtered vs not loaded error behaviour, which upstream shipped without a test"). The Akka `EventEnvelopeSpec` tests metadata support (single/composite), while this one tests `filtered` vs `not loaded` error semantics. The standard Apache header is correct here.

persistence/src/main/resources/reference.conf
He-Pin: Yes — serializer id 34 matches the upstream Akka id, following pekko's established convention of preserving upstream serializer ids (e.g. `ReplicatedEventSourcingSerializer = 40` also matches akka). Keeping this aligned ensures wire compatibility for any plugin that already hard-codes id 34 for `FilteredPayload`. Noted in the PR body under "Serializer id: 34 matches the upstream id".

pjfanning (approved):
lgtm

PR #3281, opened 2026-07-02
Fix ambiguous scaladoc overload links in persistence-typed #353

Abubakar-Sattar:
### Motivation
  `sbt unidoc` reports 91 scaladoc warnings on `main`. The largest single cluster (19) is in `persistence-typed`: doc comments bare-link a method name that has multiple overloads (`onAnyCommand`, `onAnyEvent`,
  `RetentionCriteria.snapshotEvery`, `EffectBuilder.thenRun`), so Scaladoc can't resolve which overload is meant. Same root cause #3064 fixed for `ShardRegion`. A malformed `[[tagsFor(Event, State)]]` link in the same file is also fixed.

  ### Modification
  Replace the bare `[[method]]` links with plain backtick text in the 8 affected files (20 lines total), matching the approach used in #3064.

  ### Result
  The 19 targeted ambiguous-link warnings plus the `tagsFor` syntax warning are resolved, verified by exact warning-identity diff (not just the aggregate count). The total count moved from 91 to 74, not 71, because 3 unrelated, previously-latent warnings in `actor` and `stream` surfaced as a side effect of unidoc's cross-module link resolution ordering - confirmed via a clean rebuild that these are pre-existing and untouched by this change, not a regression it introduces.

  ### Tests
  - Not run - docs only.
  - `sbt "persistence-typed/doc"` confirms the 19 targeted warnings no longer appear.
  - `sbt -Dpekko.scaladoc.diagrams=false unidoc` (clean rebuild) confirms the exact same 20 warnings are gone; 3 unrelated warnings in `actor`/`stream` appeared independently of this change (explained above).
  - Native scalafmt not installed; ran `sbt persistence-typed/scalafmtAll` instead - made no additional changes beyond the 20 targeted lines.
  - `git diff --check`: clean.

  ### References
  Refs #353

### DIFF
--- a/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/javadsl/CommandHandler.scala
+++ b/persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/javadsl/CommandHandler.scala
@@ -313,7 +313,7 @@ final class CommandHandlerBuilderByState[Command, Event, S <: State, State] @Int
    * and no further lookup is done. Therefore you must make sure that their matching conditions don't overlap,
    * otherwise you risk to 'shadow' part of your command handlers.
    *
-   * Extra care should be taken when using [[onAnyCommand]] as it will match any command.
+   * Extra care should be taken when using `onAnyCommand` as it will match any command.
    * This method builds and returns the command handler since this will not let through any states to subsequent match statements.
    *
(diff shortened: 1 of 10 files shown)


He-Pin (approved):
Lgtm, thamks

PR #3290, opened 2026-07-04
refactor: remove deprecated DurableStateUpdateStore.deleteObject(persistenceId) (since Akka 2.6.20)

He-Pin:
### Motivation
The single-argument `deleteObject(persistenceId)` was deprecated since Akka 2.6.20 in favor of the overload that takes a `revision` parameter for optimistic locking.

### Modification
Remove the deprecated abstract method from both scaladsl and javadsl `DurableStateUpdateStore` traits. Remove the corresponding TCK test for the deprecated overload.

### Result
Only the revision-aware `deleteObject(persistenceId, revision)` remains in the trait.

Note: Implementation projects (pekko-persistence-r2dbc, pekko-persistence-jdbc, pekko-persistence-cassandra, pekko-persistence-dynamodb) will need follow-up PRs to remove their overrides of this deleted method.

### Tests
Not run - trait method removal with TCK test cleanup

### References
None - deprecated API cleanup

### DIFF
--- a/docs/src/main/java/docs/persistence/state/MyJavaStateStore.java
+++ b/docs/src/main/java/docs/persistence/state/MyJavaStateStore.java
@@ -58,13 +58,6 @@ class MyJavaStateStore<A> implements DurableStateUpdateStore<A> {
     return null;
   }
 
-  /** Deprecated. Use the deleteObject overload with revision instead. */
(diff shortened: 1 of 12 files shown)


He-Pin:
Maybe we should keep this a little longer

PR #3294, opened 2026-07-05
Fix timer current message for supervision

### LINKED ISSUE #3265: Timers leaks internal TimerMsg wrapper to preRestart/supervisor for non-stash actors
### Motivation

When an actor mixing in `Timers` throws while handling a **timer-triggered**
message, the internal `TimerSchedulerImpl.TimerMsg` wrapper leaks into the
supervision machinery (`preRestart` / the failed-message reported to the
supervisor) instead of the unwrapped user message that `receive` actually
processed. This only happens for actors that are **not** stash-capable.

### Current Behavior

`Timers.aroundReceive` unwraps the `TimerMsg` and invokes the user's `receive`
with the unwrapped message `m`, but it only rewrites `ActorCell.currentMessage`
to that unwrapped message when the actor is a `StashSupport`:

`actor/src/main/scala/org/apache/pekko/actor/Timers.scala:53-67`:

```scala
case timerMsg: TimerSchedulerImpl.TimerMsg =>
  _timers.interceptTimerMsg(timerMsg) match {
    case OptionVal.Some(m: AutoReceivedMessage) => ...
    case OptionVal.Some(m) =>
      if (this.isInstanceOf[StashSupport]) {
        actorCell.currentMessage = actorCell.currentMessage.copy(message = m)
      }
      super.aroundReceive(receive, m)
    case _ => // discard
  }
```

For a **non-stash** actor the `currentMessage` rewrite is skipped, so
`currentMessage.message` remains the internal `TimerMsg` wrapper.

`ActorCell.invoke` sets `currentMessage = messageHandle` and, on failure, routes
to `handleInvokeFailure`, which reports `currentMessage` to the supervisor and
passes `currentMessage.message` to `preRestart(reason, message)`
(`actor/src/main/scala/org/apache/pekko/actor/ActorCell.scala:552-560`). As a
result, `preRestart` and the supervisor strategy observe a
`TimerSchedulerImpl.TimerMsg` (an internal type) rather than the message the
user's `receive` was actually given.

### Expected Behavior

The failed message surfaced to `preRestart` and to the supervisor for a
timer-triggered failure should be the **unwrapped** user message, consistent
with what `receive` processed — for all actors, not only stash-capable ones.

### Suggested Fix

Rewrite `currentMessage` to the unwrapped message unconditionally (not only for
`StashSupport` actors) before delegating to `super.aroundReceive`, so that both
stash and supervision observe the unwrapped message. This is a superset of the
stash-specific rewrite that already exists.

### Notes

- Typed actors are not affected: timer messages are unwrapped before the
  behavior is interpreted.
- This was noticed while reviewing the stash-side variant of the same root
  cause (the `StashSupport` guard).

### Environment

- Pekko Actor (any version)
- `Timers.scala:53-67`, `ActorCell.scala:552-560`

### References

None

He-Pin:
### Motivation
Timer-triggered failures in non-stash classic actors exposed the internal TimerMsg wrapper to preRestart instead of the user message delivered to receive.

### Modification
Rewrite ActorCell.currentMessage to the unwrapped timer message for all Timers actors before invoking receive, keeping stash and supervision paths consistent.

Add a regression test covering a non-stash Timers actor that throws from a timer-triggered message and verifies preRestart receives the unwrapped user message.

### Result
preRestart observes the same unwrapped message that receive processed, while existing timer and stash behavior remains covered.

### Tests
- sbt "actor-tests / Test / testOnly org.apache.pekko.actor.TimersAndSupervisionSpec org.apache.pekko.actor.TimersAndStashSpec org.apache.pekko.actor.TimerSpec" - passed
- sbt "actor-tests / Test / testOnly org.apache.pekko.actor.FsmTimerSpec" - passed
- sbt "actor-tests / Test / testOnly org.apache.pekko.actor.TimersAndSupervisionSpec org.apache.pekko.actor.TimersAndStashSpec" - passed
- scalafmt --mode diff-ref=origin/main - passed with JDK Unsafe deprecation warnings
- scalafmt --list --mode diff-ref=origin/main - passed with JDK Unsafe deprecation warnings
- git diff --check - passed
- qodercli final diff review - No must-fix findings

### References
Fixes #3265

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/TimerSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/TimerSpec.scala
@@ -321,6 +321,46 @@ object TimersAndStashSpec {
   case object StopStashing
 
 }
+
+object TimersAndSupervisionSpec {
+
+  case object StartTimer
+  case object Scheduled
+
+  class FailingTimerActor(probe: ActorRef) extends Actor with Timers {
+    override def preRestart(reason: Throwable, message: Option[Any]): Unit = {
+      probe ! message
+      super.preRestart(reason, message)
+    }
+
+    def receive: Receive = {
+      case StartTimer => timers.startSingleTimer("key", Scheduled, 10.millis)
+      case Scheduled  => throw new TimerSpec.Exc
+    }
+  }
+
+}
+
+class TimersAndSupervisionSpec extends PekkoSpec {
+  import TimersAndSupervisionSpec._
+
+  "Timers combined with supervision" should {
+
+    "pass the unwrapped timer message to preRestart for non-stash actors" taggedAs TimingTest in {
+      val probe = TestProbe()
+      val actor = system.actorOf(Props(new FailingTimerActor(probe.ref)))
+
+      actor ! StartTimer
+
+      probe.expectMsg(Some(Scheduled))
+      watch(actor)
+      system.stop(actor)
+      expectTerminated(actor)
+    }
+  }
+
+}
+
 class TimersAndStashSpec extends PekkoSpec {
   import TimersAndStashSpec._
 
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3295, opened 2026-07-05
feat: add Scala 3.8.4 cross-compilation target

He-Pin:
### Motivation
Ensure pekko compiles and tests pass on the next Scala 3 release line (3.8.x), preparing for future Scala 3.9.x compatibility.

### Modification
- Add `scala3NextVersion = "3.8.4"` to Dependencies.scala
- Add Scala 3.8.4 to `allScalaVersions` / `crossScalaVersions`
- Add `"3.8.x"` to CI matrix in build-test-prValidation.yml

### Result
CI now compiles and tests against Scala 2.13.x, 3.3.x, and 3.8.x. Publishing continues with Scala 3.3.8.

### Tests
CI will validate

### References
None - proactive compatibility testing

### DIFF
--- a/.github/scripts/resolve-scala-version.sh
+++ b/.github/scripts/resolve-scala-version.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -euo pipefail
+
+scala_version="${1:?scala version is required}"
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+resolve_from_build() {
+  local pattern="$1"
+  local resolved
+  resolved="$(sed -n "$pattern" "$repo_root/project/Dependencies.scala")"
+  if [ -z "$resolved" ]; then
+    echo "Unable to resolve Scala version '$scala_version' from project/Dependencies.scala" >&2
+    exit 1
+  fi
+  printf '%s\n' "$resolved"
+}
+
+case "$scala_version" in
+  2.13 | 2.13.x | scala213)
+    resolve_from_build 's/.*val scala213Version = "\(.*\)".*/\1/p'
+    ;;
+  3.3 | 3.3.x | scala3)
+    resolve_from_build 's/.*val scala3Version = "\(.*\)".*/\1/p'
+    ;;
+  next)
+    resolve_from_build 's/.*val scala3NextVersion = "\(.*\)".*/\1/p'
+    ;;
+  *)
+    printf '%s\n' "$scala_version"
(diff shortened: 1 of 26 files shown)


.github/scripts/resolve-scala-version.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
```
pjfanning, line 2: license header

serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/PekkoJacksonModule.scala
```diff
@@ -13,10 +13,37 @@
...
 /**
  * Complete module with support for all custom serializers.
  */
-class PekkoJacksonModule extends JacksonModule with ActorRefModule with AddressModule with FiniteDurationModule {
+class PekkoJacksonModule
+    extends JacksonModule
+    with ActorRefModule
+    with AddressModule
+    with FiniteDurationModule
+    with ScalaEnumerationModule {
```
pjfanning: why do we need to write our own ScalaEnumerationModule? DefaultScalaModule from jackson-scala-module supports it already.

stream/src/main/scala/org/apache/pekko/stream/impl/io/InputStreamSinkStage.scala
```diff
@@ -13,7 +13,7 @@
 
 package org.apache.pekko.stream.impl.io
 
-import java.io.{ IOException, InputStream }
+import java.io.{ IOException, InputStream => JavaInputStream }
```
pjfanning: why do we need to name InputStream as JavaInputStream?

serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/PekkoJacksonModule.scala
He-Pin: Added a comment explaining why this override is needed: DefaultScalaModule has legacy scala.Enumeration support, but its serializer reads the private `$outer` field reflectively and fails for Scala 3.8+ compiled Enumeration values with IllegalAccessException. The Pekko serializer preserves the same legacy wire format and keeps `@JsonScalaEnumeration` string output working. Verified with the Scala 3.8.4 Jackson JSON/CBOR serializer specs.

.github/scripts/resolve-scala-version.sh
He-Pin: Fixed by adding the standard ASF license header for the new shell script. Verified with `sbt +headerCheckAll`.

stream/src/main/scala/org/apache/pekko/stream/impl/io/InputStreamSinkStage.scala
He-Pin: Fixed by removing the unnecessary alias and using `InputStream` directly. Verified `stream / Compile / compile` with Scala 2.13.18, 3.3.8 and 3.8.4.

serialization-jackson/src/main/scala/org/apache/pekko/serialization/jackson/PekkoJacksonModule.scala
pjfanning: I maintain the jackson lib - I will look into getting that lib to work with Scala 3.8.
Much better than moving the code here.

pjfanning: https://github.com/FasterXML/jackson-module-scala/issues/795 -- is this the issue that you need fixed?

He-Pin: Yeah.

pjfanning:
@He-Pin can you revert the jackson code that you added and disable the tests that are broken when you use Scala 3.8 for the build?

He-Pin:
@pjfanning fixed

pjfanning (approved):
lgtm

PR #3296, opened 2026-07-05
test: cover async boundary failure draining

He-Pin:
### Motivation
Async boundary batching should preserve elements already produced by an upstream stage before that stage fails.

### Modification
Add a focused FusingSpec regression using transformed string elements upstream of an async boundary, then assert the already emitted elements arrive before the failure.

### Result
Async boundary failure-draining behavior has a direct regression test.

### References
None - regression coverage for async boundary batching


pjfanning (approved):
lgtm

PR #3297, opened 2026-07-05
fix: fix compilation warnings across Scala 2.13 and 3.3 cross-builds

He-Pin:
### Motivation
Several compilation warnings were present across the codebase when building with `-Dpekko.allwarnings=true`:
- `@SerialVersionUID does nothing on a trait` (Scala 3.3.8 changed the warning text from "has no effect" to "does nothing", causing existing `@nowarn` filters to miss)
- Unchecked type tests for `IndexedSeq[Byte]` in ByteString and `InternalRecipientRef[Req]` in AskPattern (pattern match type erasure)
- Unreachable case in TlsGraphStage (sealed trait `SslTlsOutbound` exhaustively matched)
- Invalid `@nowarn` filter `cat=unused-params` in bench-jmh benchmark (not a valid Scala 3 category)

### Modification
- Update all `@nowarn` message filters for `@SerialVersionUID` to use broader regex `(has no effect|does nothing)` matching both Scala 3.3 and older warning text (17 source locations + boilerplate template)
- Add `@unchecked` annotation to type tests that can't be checked at runtime due to type erasure (ByteString `addAll`, AskPattern `ask`)
- Change unreachable `case other =>` to `case null =>` in TlsGraphStage since the sealed trait match is exhaustive
- Change invalid `@nowarn("cat=unused-params")` to `@nowarn("msg=never used")` in bench-jmh benchmark
- Update boilerplate template for generated `Functions.scala`

### Result
Clean compilation with zero Scala warnings on both Scala 2.13.18 and 3.3.8 with `-Dpekko.allwarnings=true`.

### Tests
- `sbt -Dpekko.allwarnings=true "+test:compile"` - passes with no Scala warnings

### References
None - code quality improvement

### DIFF
--- a/actor-typed/src/main/scala/org/apache/pekko/actor/typed/scaladsl/AskPattern.scala
+++ b/actor-typed/src/main/scala/org/apache/pekko/actor/typed/scaladsl/AskPattern.scala
@@ -119,8 +119,8 @@ object AskPattern {
       // We do not currently use the implicit scheduler, but want to require it
       // because it might be needed when we move to a 'native' typed runtime, see #24219
       ref match {
-        case a: InternalRecipientRef[Req] => askClassic[Req, Res](a, timeout, replyTo)
-        case a                            =>
+        case a: (InternalRecipientRef[Req] @unchecked) => askClassic[Req, Res](a, timeout, replyTo)
+        case a                                         =>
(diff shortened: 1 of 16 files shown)


pjfanning (approved):
lgtm

PR #3301, opened 2026-07-06
Use StandardCharsets.UTF_8 for UTF-8 encoding

He-Pin:
Summary: Replace UTF-8 charset string arguments with StandardCharsets.UTF_8 in tests and build property loading. Verification: ran formatting, targeted compile, git diff --check, and residual charset literal searches.


pjfanning (approved):
lgtm

PR #3300, opened 2026-07-06
Replace java.lang.reflect with MethodHandle/VarHandle across core modules

He-Pin:
### Motivation
`java.lang.reflect` method, constructor, and field access adds repeated lookup/adaptation overhead in actor creation, serialization, and JVM integration paths.

### Modification
- Cache pre-adapted constructor handles in `ReflectiveDynamicAccess`, `Reflect`, and `IndirectActorProducer`.
- Cache and use MethodHandles/VarHandles in `VirtualThreadSupport`, `ProtobufSerializer`, `LineNumbers`, `DnsSettings`, `ByteBufferCleaner`, and typed `ExtensionsImpl`.
- Preserve JPMS lookup fallback, class initialization, fixed-arity varargs, and reflection-compatible exception behavior with focused regression tests.
- Keep test-only JFR and file-descriptor access on cached MethodHandles.
- Move the independent `HashCode` array-hashing optimization to #3312.

### Result
Hot paths avoid repeated reflective lookup and per-call adaptation while preserving behavior on modular JDKs. Actor creation JMH showed no regression and a small non-conclusive signal from 9.640 +/- 0.314 to 9.341 +/- 0.279 us/op.

### Tests
- `sbt "actor-tests / Test / testOnly org.apache.pekko.util.ReflectSpec org.apache.pekko.actor.DynamicAccessSpec org.apache.pekko.PekkoExceptionSpec"` - passed, 26 tests after the split
- `sbt "remote / Test / testOnly org.apache.pekko.remote.serialization.ProtobufSerializerSpec"` - passed, 9 tests after the split
- `sbt "testkit / Test / testOnly org.apache.pekko.testkit.metrics.FileDescriptorMetricSetSpec"` - passed, 1 test after the split
- Earlier focused Scala 3.3.8, JDK 21, and JDK 25 reflection/virtual-thread validation passed on the same MethodHandle changes
- `sbt +headerCheckAll` - passed
- `scalafmt --list --mode diff-ref=origin/main` - passed
- `git diff --check` - passed
- `qodercli` formal stdout review - no must-fix findings

### References
Refs #3300; independent HashCode optimization split to #3312

PR #3312, opened 2026-07-10
Optimize JVM array hashing without reflection

He-Pin:
### Motivation
`HashCode.hash` iterates JVM arrays through `java.lang.reflect.Array`, adding reflective per-element access and boxing in a shared utility. A focused JDK 17 JMH comparison showed this path is substantially slower than typed array loops.

### Modification
- Dispatch all nine JVM array kinds to typed `while` loops.
- Add a directional test that uses the original `java.lang.reflect.Array` algorithm as the parity oracle for primitive, reference, empty, null-element, and nested arrays.
- Add a reproducible JMH benchmark for primitive and reference arrays.

### Result
Hash values remain compatible while the measured array paths improve by roughly 39x to 135x. JDK 17, 2 forks, 5 warmup iterations, 8 measurement iterations, 250 ms each:

| Benchmark | `main` | optimized | speedup |
|---|---:|---:|---:|
| `Int[16]` | 824.277 ns/op | 6.106 ns/op | 135.0x |
| `Int[256]` | 12749.433 ns/op | 183.033 ns/op | 69.7x |
| `AnyRef[16]` | 635.492 ns/op | 13.945 ns/op | 45.6x |
| `AnyRef[256]` | 9491.604 ns/op | 240.387 ns/op | 39.5x |

### Tests
- `sbt "actor-tests / Test / testOnly org.apache.pekko.util.HashCodeSpec"` - passed, 3 tests
- `sbt "++3.3.8" "actor-tests / Test / testOnly org.apache.pekko.util.HashCodeSpec"` - passed, 3 tests
- `sbt "bench-jmh/Jmh/run -wi 5 -i 8 -f 2 -w 250ms -r 250ms -t 1 -p size=16,256 .*HashCodeArrayBenchmark.*"` - passed; results above
- `sbt +headerCheckAll` - passed
- `sbt checkCodeStyle` - passed
- `scalafmt --list --mode diff-ref=origin/main` - passed
- `git diff --check` - passed
- `qodercli` formal stdout review - no must-fix findings

### References
Refs #3300

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/util/HashCodeSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/util/HashCodeSpec.scala
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

PR #3317, opened 2026-07-10
Bound Artery stream shutdown

### LINKED ISSUE #3247: Artery ActorSystem termination slow for Aeron-UDP transport due to stacked liveness timeouts
### Description

When shutting down an `ActorSystem` using Artery with Aeron-UDP transport, termination can take excessively long due to stacked Aeron liveness timeouts and the embedded media driver drain sequence. This is most pronounced in virtualized environments.

### Evidence from Pekko CI

The issue has been acknowledged through test timeout widening commits:

- Commit `4d17e08c5b`: widened `shutdownAll` await for aeron-udp drain on JDK 25 nightly. The commit message states: *"aeron-udp transport is the slowest to drain (embedded media driver + stacked Aeron liveness timeouts)"*
- Commit `3fcbea76ae`: widened TCK `actor-system-terminate` phase timeout because *"the inner CoordinatedShutdown actor-system-terminate phase keeps its default 10s, which is occasionally too tight"*

Both commits show `CoordinatedShutdown` phase `actor-system-terminate` timing out with `ActorGraphInterpreter` and stream actors still mid-shutdown.

### Root cause

The shutdown flow involves multiple sequential waits that accumulate:

1. **FlushOnShutdown** — waits for all associations to acknowledge termination (up to `shutdown-flush-timeout` = 1s)
2. **ArteryTransport.shutdown()** — `internalShutdown()` waits for `streamsCompleted` (all outbound/inbound stream completions)
3. **shutdownHook** — blocking `Await.result` with up to `DriverTimeout + 3.seconds` (23s total)
4. **Aeron liveness timeouts** stack up:
   - `client-liveness-timeout` = 20s
   - `publication-unblock-timeout` = 40s
   - `image-liveness-timeout` = 10s
   - `driver-timeout` = 20s
5. **give-up-message-after** = 60s — messages can be retried for up to 60 seconds

### Impact

- TCP transport is not affected
- Healthy shutdowns complete in under a second
- The issue manifests with Aeron-UDP in virtualized environments or with embedded media drivers
- Test CI intermittently times out, requiring repeated timeout widening

### Current mitigation

All timeouts are configurable via `reference.conf`, but the default stacking of Aeron liveness timeouts makes the worst-case shutdown duration unnecessarily long.

### Suggested approach

- Decouple the `actor-system-terminate` phase from Aeron liveness timeouts (force-drain with a shorter deadline during shutdown)
- Add a dedicated `artery-shutdown-timeout` that bounds the total Artery drain time independently of individual Aeron liveness settings

He-Pin:
### Motivation
Artery waits without a deadline for remoting stream completion during ActorSystem termination. A stalled stream can therefore prevent transport-specific shutdown from starting and keep the actor system alive indefinitely.

### Modification
- Add `pekko.remote.artery.advanced.shutdown-streams-timeout` with a 10-second default.
- Continue with transport shutdown when graceful stream completion exceeds that deadline.
- Continue with transport shutdown if the system scheduler is already unavailable.
- Preserve the normal streams-before-transport ordering and shutdown idempotency.
- Document the setting and add directional regression coverage for both timeout paths.

### Result
A stalled remoting stream no longer blocks Artery transport shutdown indefinitely.

### References
Fixes #3247

### DIFF
--- a/docs/src/main/paradox/remoting-artery.md
+++ b/docs/src/main/paradox/remoting-artery.md
@@ -93,6 +93,14 @@ listening for connections and handling messages as not to interfere with other a
 The example above only illustrates the bare minimum of properties you have to add to enable remoting.
 All settings are described in @ref:[Remote Configuration](#remote-configuration-artery).
 
+### Shutdown
+
+Artery first flushes outstanding remote messages and then aborts its streams during `ActorSystem` termination.
+The `pekko.remote.artery.advanced.shutdown-streams-timeout` setting limits how long Artery waits for those streams to
+complete. If the timeout expires, Artery proceeds with transport shutdown so a stalled stream cannot indefinitely
+delay transport-specific shutdown handling. The separate
+`pekko.remote.artery.advanced.shutdown-flush-timeout` setting controls the preceding flush.
+
 ## Introduction
 
 We recommend @ref:[Pekko Cluster](cluster-usage.md) over using remoting directly. As remoting is the
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -1024,6 +1024,11 @@ pekko {
         # remote messages has been completed
         shutdown-flush-timeout = 1 second
 
+        # After flushing, remoting aborts its streams and waits this long for them
+        # to complete before proceeding with transport shutdown. This bounds the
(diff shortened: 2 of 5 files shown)


pjfanning (approved):
lgtm

PR #3320, opened 2026-07-10
Clarify JsonFraming object-only behavior

### LINKED ISSUE #3244: JsonObjectParser rejects valid top-level non-object JSON
### Description

`JsonObjectParser` rejects valid JSON input when the top-level structure is not a JSON object. The 256-byte lookup table (`outerChars`) only tolerates `{`, `[`, `]`, `,`, and whitespace at depth 0 outside a string. Any other byte causes a `FramingException` with the misleading message "Invalid JSON encountered".

### Affected code

`stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala` (lines 170-180):

```scala
private val outerChars: Array[Byte] =
  Array.tabulate[Byte](256) { i =>
    i.toByte match {
      case CurlyBraceStart                           => 2 // found object
      case SquareBraceStart | SquareBraceEnd | Comma => 1 // skip
      case b if isWhitespace(b.toByte)               => 1 // skip
      case _                                         => 0 // error
    }
  }
```

At depth 0, encountering a byte with value `0` triggers `1 / outer` (line 142), raising `ArithmeticException`, which is converted to a `FramingException`.

### Valid JSON inputs that are incorrectly rejected

| Input | Why rejected |
|-------|-------------|
| `42` (bare number) | `4` → outer=0 |
| `"hello"` (bare string) | `"` → outer=0 |
| `true` / `null` / `false` | `t`/`n`/`f` → outer=0 |
| `[[1], [2]]` (array of arrays) | digits not in skip set |
| `\xEF\xBB\xBF{"a":1}` (UTF-8 BOM) | 0xEF → outer=0 |

### Documentation inconsistency

`JsonFraming.scala` line 56 states "a top-level JSON Array can also be understood" — but this only works for arrays of objects, not arrays of primitives or arrays of arrays.

### Suggested improvements

1. Make the error message more specific: distinguish between "Invalid JSON" and "Top-level JSON element is not a JSON object"
2. Clarify the documentation to state that `JsonFraming.objectScanner` only frames JSON objects
3. Consider supporting UTF-8 BOM at the start of input (strip `\xEF\xBB\xBF` before parsing)

He-Pin:
### Motivation
`JsonFraming.objectScanner` only frames JSON objects, but its API documentation described valid JSON chunks and top-level arrays without explicitly stating that non-object values and a leading UTF-8 BOM are unsupported. This made the intended contract unclear in #3244.

### Modification
- Clarify that the brace-counting operator emits objects and does not validate their contents.
- Document the supported array-of-objects and object-only JSON Lines/NDJSON inputs, unsupported top-level values, and BOM handling.
- Keep Scala DSL, Java DSL, and internal parser documentation in sync and link nested JSON use cases to Pekko Connectors.

### Result
Users can distinguish object framing from general-purpose JSON parsing and choose the appropriate API without changing runtime behavior.

### References
Fixes #3244

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/JsonObjectParser.scala
@@ -52,8 +52,8 @@ import pekko.util.ByteString
  * INTERNAL API: Use [[pekko.stream.scaladsl.JsonFraming]] instead.
  *
  * **Mutable** framing implementation that given any number of [[pekko.util.ByteString]] chunks, can emit JSON objects contained within them.
- * Typically JSON objects are separated by new-lines or commas, however a top-level JSON Array can also be understood and chunked up
- * into valid JSON objects by this framing implementation.
+ * Typically JSON objects are separated by newlines or commas. A top-level JSON array is supported as a container
+ * of objects, but top-level non-object values are not supported.
  *
  * Leading whitespace between elements will be trimmed.
  */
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

PR #3321, opened 2026-07-10
Make TcpConnectionSpec pull-mode reads deterministic

### LINKED ISSUE #3311: Flaky TcpConnectionSpec pull-mode test times out waiting for Tcp.Received
### Summary

The TcpConnectionSpec test named An outgoing connection must respect pull mode timed out while waiting for Tcp.Received during validatePullRequest.

### Reproduction

Run the full PR validation on macOS with JDK 25:

    sbt validatePullRequest

The failure was observed after the test suite had already exercised several TCP integration tests in the same run.

### Observed failure

    An outgoing connection must respect pull mode *** FAILED ***
    java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.apache.pekko.io.Tcp$Received
      at org.apache.pekko.io.TcpConnectionSpec$$anon$15.$anonfun$new$35(TcpConnectionSpec.scala:419)

### Expected behavior

The test should reliably receive Tcp.Received and pass under normal local PR validation load.

### Environment

- macOS
- Oracle JDK 25.0.1
- sbt 1.12.13
- Apache Pekko main-based branch at be509230319db45589bca9a3b7765e78ba367b09

### Notes

This surfaced while validating PR #3030, but the PR only changes GraphInterpreter stage-finalization bookkeeping. The focused GraphInterpreter, lifecycle, port-state, failure-mode, and actor-shell suites passed 126 of 126 tests before this unrelated TCP timeout. The failure therefore looks like a timing-sensitive test flake rather than a regression from PR #3030.

The full validatePullRequest run was stopped after this failure because an unrelated AGENTS.md worktree change caused the all-build filter to select the complete repository test matrix.

He-Pin:
### Motivation

The `TcpConnectionSpec` pull-mode test assumed data written on the server side was immediately readable by the client. Under load, `ResumeReading` could perform a non-blocking read before the data arrived; it then registered `OP_READ`, but the test uses a probe as the connection selector and never delivered the later `ChannelReadable` notification.

### Modification

Assert that each payload is written in full and use the test NIO selector to wait until the client channel is readable before each `ResumeReading` command.

### Result

The test still verifies that pull mode emits exactly one `Tcp.Received` per pull while no longer depending on loopback delivery timing.

### References

Fixes #3311

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/io/TcpConnectionSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/io/TcpConnectionSpec.scala
@@ -395,15 +395,18 @@ class TcpConnectionSpec extends PekkoSpec("""
           val maxBufferSize = 1 * 1024
           val ts = "t" * maxBufferSize
           val us = "u" * (maxBufferSize / 2)
+          val tsAndUs = ts ++ us
 
           // send a batch that is bigger than the default buffer to make sure we don't recurse and
           // send more than one Received messages
-          serverSideChannel.write(ByteBuffer.wrap((ts ++ us).getBytes("ASCII")))
+          serverSideChannel.write(ByteBuffer.wrap(tsAndUs.getBytes("ASCII"))) should ===(tsAndUs.length)
+          clientSelectionKey should be(selectedAs(OP_READ, remainingOrDefault))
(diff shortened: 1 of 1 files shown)


pjfanning (approved):
lgtm

PR #3322, opened 2026-07-10
Fix OutboundIdleShutdownSpec harmless quarantine setup

He-Pin:
### Motivation
Pekko 2.0 changed `pekko.remote.artery.propagate-harmless-quarantine-events` to default to `off`, but `OutboundIdleShutdownSpec` still expects a harmless quarantine to publish `ThisActorSystemQuarantinedEvent`. The spec therefore waits for an event that the default configuration deliberately suppresses.

### Modification
Explicitly enable harmless quarantine event propagation for `OutboundIdleShutdownSpec`.

`HarmlessQuarantineSpec` continues to use the production default and cover the suppression behavior.

### Result
The two specs again cover both directions:
- legacy propagation enabled: the remote system observes `ThisActorSystemQuarantinedEvent`
- production default disabled: harmless quarantine does not propagate the event

No production configuration or runtime behavior changes.

### References
Refs #2430


pjfanning (approved):
lgtm

Issue #3245, opened 2026-06-27
TLS handshake failure sends close_notify instead of fatal alert


He-Pin:
### Description

When a TLS handshake fails, both the classic remoting and stream TLS implementations send a `close_notify` alert (graceful shutdown) instead of a `fatal` TLS alert as required by RFC 5246 §7.2.1 and RFC 8446 §6.1.

### Affected code

**Classic remoting** — `remote/src/main/scala/org/apache/pekko/remote/transport/netty/NettySSLSupport.scala` (lines 83-92):

```scala
handler.handshakeFuture().addListener((future: Future[Channel]) => {
  if (!future.isSuccess) {
    val channel = handler.closeOutbound().channel()  // sends close_notify
    channel.close()
  }
})
```

**Stream TLS GraphStage** — `stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsGraphStage.scala` (lines 332-340, 599-610):

```scala
case ex: SSLException =>
  failTls(ex, closeTransport = false)  // triggers writeFailureAlert()
  // writeFailureAlert() calls engine.closeOutbound() → close_notify
```

**Legacy stream TLSActor** — `stream/src/main/scala/org/apache/pekko/stream/impl/io/TLSActor.scala` (lines 312-318):

```scala
case ex: SSLException =>
  fail(ex, closeTransport = false)
  engine.closeInbound()
  completeOrFlush()  // no explicit fatal alert
```

### Analysis

- `close_notify` is defined for orderly shutdown of established connections, not for handshake failures
- Per TLS spec, failed handshakes should produce a fatal alert (`handshake_failure`, `bad_certificate`, etc.)
- Sending `close_notify` instead of a fatal alert can cause the peer to wait indefinitely for handshake completion
- The non-TLS TCP code correctly uses `Abort` (TCP reset) for failure scenarios, but the TLS code does not follow this pattern

### Suggested fix

On handshake failure:
1. Skip `closeOutbound()` / `writeFailureAlert()` 
2. Perform an abrupt TCP close (e.g., `channel.close()` directly without TLS close_notify)
3. This aligns with the non-TLS path in `TcpStages.scala` which uses `Abort` for failures

pjfanning:
how does this differ from #3237

PR #3323, opened 2026-07-11
Abort TLS transport on session verification failure

He-Pin:
### Motivation
A failed post-handshake session verifier was handled as if the `SSLEngine` itself had failed. When the verifier returned an `SSLException`, the GraphStage TLS implementation caught it in the engine failure path and called `closeOutbound()`. Because the handshake had already succeeded and there was no pending fatal alert, this generated `close_notify` and completed the transport gracefully instead of aborting it.

Engine-originated handshake failures are intentionally left unchanged: those paths must continue wrapping and flushing pending fatal alerts such as `certificate_unknown`.

### Modification
- Handle a failed `verifySession` result as an explicit TLS stage failure.
- Stop wrap/unwrap handshake processing immediately after the verifier rejects the session.
- Use structured `if`/`else` control flow rather than a non-local return.
- Add a directional regression test that observes the cipher-side transport termination and checks that the original verifier exception fails it.
- Run the shared test for both the GraphStage and legacy actor implementations, covering TLS 1.2 and TLS 1.3.

### Result
Post-handshake session verification failures now abort the transport with the original exception. The GraphStage behavior matches the legacy TLS actor, while engine-originated TLS failures continue to flush their fatal alerts.

### Tests
- `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageSpec` — 113 passed
- `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsSpec` — 113 passed
- `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageDeferredCloseSpec` — 3 passed
- `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageIsolatedSpec` — 7 passed
- `+mimaReportBinaryIssues` — passed for Scala 2.13 and Scala 3
- `checkCodeStyle` — passed
- `+headerCheckAll` — passed
- `git diff --check origin/main...HEAD` — passed
- `validatePullRequest` — not run per request; focused TLS, compatibility, style, and header checks above were run
- Qoder stdout blocker review — no must-fix findings

### References
Refs #3245

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsGraphStage.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsGraphStage.scala
@@ -380,26 +380,29 @@ import pekko.util.ByteString
         val result = wrapIntoTransportBuffer()
         lastHandshakeStatus = result.getHandshakeStatus
 
-        if (lastHandshakeStatus == FINISHED) handshakeFinished()
-        runDelegatedTasks()
+        if (lastHandshakeStatus == FINISHED && !handshakeFinished()) {
+          // stop processing after failed session verification
+        } else {
+          runDelegatedTasks()
 
-        result.getStatus match {
-          case OK =>
-            if (transportOutBuffer.position() == 0 && lastHandshakeStatus == NEED_WRAP)
-              throw new IllegalStateException("SSLEngine trying to loop NEED_WRAP without producing output")
+          result.getStatus match {
+            case OK =>
+              if (transportOutBuffer.position() == 0 && lastHandshakeStatus == NEED_WRAP)
+                throw new IllegalStateException("SSLEngine trying to loop NEED_WRAP without producing output")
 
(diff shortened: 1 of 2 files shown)


stream/src/main/scala/org/apache/pekko/stream/impl/io/TlsGraphStage.scala
```diff
@@ -376,7 +376,7 @@ import pekko.util.ByteString
         val result = wrapIntoTransportBuffer()
         lastHandshakeStatus = result.getHandshakeStatus
 
-        if (lastHandshakeStatus == FINISHED) handshakeFinished()
+        if (lastHandshakeStatus == FINISHED && !handshakeFinished()) return
```
pjfanning: could you use an if-else to avoid the return?

pjfanning (approved):
lgtm

PR #3324, opened 2026-07-11
Fix deferred TLS output handling on inbound close

He-Pin:
### Motivation
`TlsGraphStage` migrated the TLS pump phases from `TlsActor` but also added deferred plaintext batching and deferred transport flushing. When transport input completes before either timer runs, the copied close transition can enter a non-executable `FlushingOutboundPhase` or complete before buffered ciphertext is emitted.

### Modification
Flush buffered transport output before selecting the close phase, and use the `SSLEngine`'s `NEED_WRAP` state to decide whether protocol output still has to be generated. Add deterministic tests that hold zero-delay timers and cover pending plaintext, buffered ciphertext, and engine wrap output.

### Result
Inbound close no longer strands the stage, already-wrapped ciphertext is delivered before completion, and pending engine output is wrapped before shutdown.

### Tests
- `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageDeferredCloseSpec` (3 passed)
- `stream-tests / Test / testOnly org.apache.pekko.stream.io.TlsGraphStageIsolatedSpec` (7 passed)
- `headerCreateAll` and `+headerCheckAll`
- `git diff --check`
- Full test suite not run; focused validation only as requested.

### References
Refs #2860

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/io/TlsGraphStageIsolatedSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/io/TlsGraphStageIsolatedSpec.scala
@@ -17,21 +17,31 @@
 
 package org.apache.pekko.stream.io
 
+import java.nio.ByteBuffer
+import java.util.concurrent.ThreadFactory
 import javax.net.ssl._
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3325, opened 2026-07-11
Bound persistence event replay batches

### LINKED ISSUE #3241: asyncReplayMessages can cause mailbox OOM for large event logs
### Description

The event replay mechanism in `AsyncWriteJournal` sends `ReplayedMessage` events directly to the persistent actor's mailbox via `tell`, with no backpressure or batch size limit. For persistent actors with millions of events, this can overflow the mailbox and cause OOM.

### Affected code

`persistence/src/main/scala/org/apache/pekko/persistence/journal/AsyncWriteJournal.scala` (lines 190-195):

```scala
asyncReplayMessages(persistenceId, fromSequenceNr, toSeqNr, max) { p =>
  if (!p.deleted)
    adaptFromJournal(p).foreach { adaptedPersistentRepr =>
      replyTo.tell(ReplayedMessage(adaptedPersistentRepr), Actor.noSender)
    }
}
```

### Analysis

- The replay callback invokes `tell` for every event with no rate limiting
- The replay runs on a separate dispatcher thread, pushing into the actor's mailbox faster than the actor can process events
- No `replay-batch-size` or memory-bounding configuration exists in `reference.conf`
- Even streaming journal plugins (e.g., LevelDB's `DBIterator`) cannot prevent this because the bottleneck is the mailbox, not the journal
- The `InmemJournal` (test-only) additionally loads all matching events into memory via `immutable.Seq`

### Impact

Persistent actors with millions of events risk OOM during recovery. This is a production concern for any deployment using large event logs.

### Suggested approach

Introduce bounded streaming between the replay callback and the persistent actor — either via a batched `ReplayedMessages` protocol with flow control, or a bounded buffer between the journal callback and the actor's mailbox.

He-Pin:
### Motivation
`AsyncWriteJournal` can emit an entire recovery stream without waiting for the recovering actor to process it. A large journal can therefore enqueue an unbounded number of replay messages in the persistent actor mailbox and exhaust the heap.

### Modification
- Add configurable, acknowledged replay batching with a default batch size of 1000.
- Let classic and typed persistent actors opt in while retaining the existing `ReplayMessages` request for compatibility with journal plugins that do not understand batching.
- Preserve replay limits, ordering, sender semantics, replay-filter state, cancellation, and recovery isolation across actor or journal restarts.
- Forward the batching protocol through the persistence plugin proxy.
- Add focused classic, typed, replay-filter, protocol, and remote-proxy regression coverage.
- Document the replay batching behavior and configuration.

### Result
Recovery bounds the number of replayed events in flight without truncating recovered history. Existing journal plugins continue to use the legacy replay path unless they support the opt-in batching marker.

### Tests
- Focused classic protocol, replay filter, and replay batching tests: 24 passed
- Focused typed replay batching tests: 2 passed
- `sbt "persistence / Test / test"`: 63 passed; LevelDB JNI emitted an Apple ARM64/x86 architecture warning
- `sbt "persistence-typed-tests / Test / test"`: 204 passed
- `sbt "persistence-shared / Test / testOnly org.apache.pekko.persistence.journal.leveldb.PersistencePluginProxySpec"`: 1 passed
- `sbt headerCreateAll`, `sbt +headerCheckAll`, `sbt checkCodeStyle`, `sbt docs/paradox`, and `sbt +mimaReportBinaryIssues`: passed; Paradox reported duplicate-anchor warnings
- `scalafmt --list --mode diff-ref=origin/main` and `git diff --check`: passed
- `sbt sortImports`: environment/task failure; the project-scoped key was unavailable and root scalafix raised `NoSuchMethodError`
- `sbt validatePullRequest`: ran for 28m26s and failed on this macOS ARM64 host because LevelDB JNI 1.8 contains x86 binaries; unrelated docs tests also failed from a missing `pekko.dispatch.UnboundedMailbox` class and UDP multicast timeouts on all local interfaces

### References
Fixes #3241

### DIFF
--- a/docs/src/main/paradox/persistence.md
+++ b/docs/src/main/paradox/persistence.md
@@ -136,6 +136,17 @@ until other recoveries have been completed. This is configured by:
 pekko.persistence.max-concurrent-recoveries = 50
 ```
 
+Journal plugins based on `AsyncWriteJournal` also bound each individual recovery. They replay at most
+`replay-batch-size` event sequence numbers and wait until the recovering actor has processed that batch before
+requesting the next one. The default is `1000`; it can be changed in the journal plugin configuration, for example:
+
+```
+pekko.persistence.journal.my-plugin.replay-batch-size = 500
+```
+
+This setting limits replay pressure on the persistent actor's mailbox without limiting the total number of events
+that are recovered.
+
 @@@ note
 
 Accessing the @scala[`sender()`]@java[sender with `getSender()`] for replayed messages will always result in a `deadLetters` reference,
--- a/docs/src/main/paradox/typed/persistence.md
+++ b/docs/src/main/paradox/typed/persistence.md
@@ -468,6 +468,17 @@ until other recoveries have been completed. This is configured by:
 pekko.persistence.max-concurrent-recoveries = 50
 ```
 
+Journal plugins based on `AsyncWriteJournal` also bound each individual recovery. They replay at most
+`replay-batch-size` event sequence numbers and wait until the recovering actor has processed that batch before
+requesting the next one. The default is `1000`; it can be changed in the journal plugin configuration, for example:
+
+```
+pekko.persistence.journal.my-plugin.replay-batch-size = 500
+```
+
+This setting limits replay pressure on the persistent actor's mailbox without limiting the total number of events
+that are recovered.
+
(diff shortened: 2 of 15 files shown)


He-Pin:
@pjfanning I think this is the last one for me in 2.0.0 ,quite busy at work

pjfanning (approved):
lgtm

PR #3328, opened 2026-07-11
fix: validate legal files in packaged jars

### LINKED ISSUE #1470: META-INF/LICENSE and META-INF/NOTICE are not consistently added to jars
During the 1.1.0 release, we noticed some artifacts were missing the `META-INF/LICENSE` and `META-INF/NOTICE` files. They do not seem to consistently be missing, so we should figure out what's going on.

He-Pin:
### Motivation
Published jars can silently omit required `META-INF/LICENSE` and `META-INF/NOTICE` files. The `protobuf-v3` assembly also needs explicit merge handling for legal resources.

### Modification
- Merge LICENSE and NOTICE resources in the `protobuf-v3` assembly while retaining the protobuf COPYING file.
- Add a package validation task that checks both `packageBin` and the published artifact for required META-INF legal files.
- Include the validation task in pull request validation for published projects.

### Result
Normal package jars and assembly-backed published jars fail validation when either required legal file is missing.

### References
Fixes #1470

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -381,6 +381,11 @@ lazy val protobufV3 = pekkoModule("protobuf-v3")
         // https://github.com/sbt/sbt-assembly/issues/400
         .inLibrary(Dependencies.Compile.Provided.protobufRuntime)
         .inProject),
+    assembly / assemblyMergeStrategy := {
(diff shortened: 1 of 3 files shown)


pjfanning (approved):
lgtm

Issue #1668, opened 2025-01-03
Check for type pollution - can have a major performance impact


pjfanning:
Check for type pollution - can have a major performance impact. Basically, Java Runtime up to v23 has an issue in a secondary cache used internally to track results of instanceof checks.

https://bugs.openjdk.org/browse/JDK-8180450

https://github.com/FasterXML/jackson-databind/pull/4848

See the micronaut utility mentioned in the issue description.

It may also be useful to check pekko-http.

He-Pin:
Tools: https://github.com/RedHatPerf/type-pollution-agent

PR #3332, opened 2026-07-11
Avoid receive timeout type pollution

He-Pin:
### Motivation

`Identify` implements both `AutoReceivedMessage` and `NotInfluenceReceiveTimeout`. On JVMs affected by [JDK-8180450](https://bugs.openjdk.org/browse/JDK-8180450), alternating successful checks against those two secondary supertypes makes the same concrete `Klass` secondary-super cache ping-pong between marker types. The checks occur on the Actor receive-timeout path and can contend across actor-dispatcher threads.

The receive path also classified the same message against `NotInfluenceReceiveTimeout` twice.

### Modification

- Classify final `Identify` with a concrete-class guard before the `NotInfluenceReceiveTimeout` marker check.
- Return the pre-cancellation timeout state and infer the post-receive reschedule decision from the state transition, removing the second marker check.
- Add focused classification coverage.
- Add a 12-thread JMH benchmark for the polluted, concrete-guarded, and final production-shaped paths.

### Result

Receive-timeout behavior remains unchanged, while `Identify` no longer alternates the concrete class secondary-super cache between its two marker interfaces. The final path also performs one classification instead of two.

JMH throughput (`ops/us`, 12 threads, 3 forks, 5 x 1 s warmup and measurement iterations):

| Runtime | Polluted path | Concrete guard | Final path | Final / polluted |
| --- | ---: | ---: | ---: | ---: |
| JDK 17.0.17 | 824.700 +/- 10.066 | 1633.766 +/- 53.719 | 2205.529 +/- 92.313 | 2.67x |
| JDK 21.0.8 | 1125.827 +/- 16.496 | 1609.084 +/- 31.694 | 2158.924 +/- 36.895 | 1.92x |
| JDK 25.0.2 | 1717.703 +/- 77.349 | 1841.141 +/- 53.279 | 2467.700 +/- 262.562 | 1.44x |

JDK 21.0.8 and JDK 25 contain the JVM fix, so their remaining gain primarily measures the removed classification/call rather than secondary-super-cache contention. The JDK 25 result was noisier than the other runs.

### References

Refs #1668

- [JDK-8180450](https://bugs.openjdk.org/browse/JDK-8180450)

### DIFF
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/ReceiveTimeoutSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/ReceiveTimeoutSpec.scala
@@ -21,6 +21,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 
 import org.apache.pekko
+import pekko.actor.dungeon.{ ReceiveTimeout => ReceiveTimeoutSupport }
 import pekko.testkit._
 
 object ReceiveTimeoutSpec {
@@ -79,6 +80,12 @@ class ReceiveTimeoutSpec extends PekkoSpec() {
   import ReceiveTimeoutSpec._
 
   "An actor with receive timeout" must {
+
+    "classify messages that do not influence the timeout" in {
+      ReceiveTimeoutSupport.isNotInfluenceReceiveTimeout(Identify(None)) should ===(true)
+      ReceiveTimeoutSupport.isNotInfluenceReceiveTimeout(TransparentTick) should ===(true)
+      ReceiveTimeoutSupport.isNotInfluenceReceiveTimeout(Tick) should ===(false)
+    }
 
     "get timeout" taggedAs TimingTest in {
       val timeoutLatch = TestLatch()
--- a/actor-tests/src/test/scala/org/apache/pekko/actor/ReceiveTimeoutSpec.scala
+++ b/actor-tests/src/test/scala/org/apache/pekko/actor/ReceiveTimeoutSpec.scala
@@ -21,7 +21,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 
 import org.apache.pekko
-import pekko.actor.dungeon.{ ReceiveTimeout => ReceiveTimeoutSupport }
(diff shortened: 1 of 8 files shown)


actor-tests/src/test/scala/org/apache/pekko/actor/ReceiveTimeoutSpec.scala
```diff
@@ -20,6 +20,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import scala.concurrent.Await
 import scala.concurrent.duration._
 
+import org.apache.pekko.actor.dungeon.{ ReceiveTimeout => ReceiveTimeoutSupport }
```
pjfanning: can you rebase and fix the imports to be `pekko.`?

actor/src/main/scala/org/apache/pekko/actor/dungeon/ReceiveTimeout.scala
```diff
@@ -19,10 +19,16 @@ import scala.concurrent.duration.FiniteDuration
...
 import pekko.actor.Cancellable
+import pekko.actor.Identify
 import pekko.actor.NotInfluenceReceiveTimeout
 
 private[pekko] object ReceiveTimeout {
   final val emptyReceiveTimeoutData: (Duration, Cancellable) = (Duration.Undefined, ActorCell.emptyCancellable)
+
+  // Identify is also an AutoReceivedMessage. Checking its concrete class first avoids polluting its secondary
+  // supertype cache on JDKs affected by JDK-8180450 when actor runtime code also checks the AutoReceivedMessage marker.
+  @inline def isNotInfluenceReceiveTimeout(message: Any): Boolean =
```
pjfanning: `@inline` doesn't work for Scala 3. You might need to create a Scala 2 and Scala 3 file with just this method. Better than copy/pasting this whole scala file.

actor-tests/src/test/scala/org/apache/pekko/actor/ReceiveTimeoutSpec.scala
pjfanning: I pushed a fix for this

pjfanning (approved):
lgtm

PR #3334, opened 2026-07-12
Reduce receive-timeout state allocations

He-Pin:
### Motivation

An actor with receive timeout enabled currently recreates immutable `(Duration, Cancellable)` tuples while cancelling and rescheduling the timeout for every influencing message. This puts two short-lived `Tuple2` objects on the `ActorCell.invoke` hot path even though receive-timeout state is mailbox-thread confined.

### Modification

- Replace tuple replacement with one lazily allocated mutable receive-timeout state object.
- Track state changes with an `Int` counter compared only across one message invocation, preserving the previous tuple-reference identity semantics without per-message state allocation.
- Extend `ReceiveTimeoutSpec` with a 100-message reset burst and an explicit assertion that the timeout is measured from the final processed message.
- Add an end-to-end Actor JMH benchmark that processes real mailbox messages while receive timeout is active and supports the JMH GC profiler.

### Result

JDK 17.0.17, G1, one Actor/dispatcher thread, 10,000 messages per invocation, 3 forks, 5 x 1 second warmup and measurement:

| Implementation | Throughput | Steady allocation |
| --- | ---: | ---: |
| Immutable tuple state | 16.049 ± 1.323 messages/µs | 168.018 B/message |
| Reused mutable state | 16.476 ± 1.825 messages/µs | 120.018 B/message |

The allocation result is stable across all non-teardown measurement iterations: exactly 48 B/message less, a 28.6% reduction in total allocation for this real Actor path. Throughput confidence intervals overlap, so this PR claims allocation/GC improvement rather than a statistically significant CPU improvement.

The state object is allocated only when an actor first configures receive timeout; actors that never enable it retain the existing null-reference footprint. Public Actor APIs and behavior are unchanged.

### References

Refs #1668

- Related type-pollution fix: #3332

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/actor/dungeon/ReceiveTimeout.scala
+++ b/actor/src/main/scala/org/apache/pekko/actor/dungeon/ReceiveTimeout.scala
@@ -22,7 +22,8 @@ import pekko.actor.Cancellable
 import pekko.actor.dungeon.ReceiveTimeoutCompat
 
 private[pekko] object ReceiveTimeout {
-  final val emptyReceiveTimeoutData: (Duration, Cancellable) = (Duration.Undefined, ActorCell.emptyCancellable)
+  // Compared only across one message invocation, so wrapping over an actor's lifetime is harmless.
+  private final class State(var timeout: Duration, var task: Cancellable, var version: Int)
 }
 
 private[pekko] trait ReceiveTimeout { this: ActorCell =>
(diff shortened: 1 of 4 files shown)


actor/src/main/scala/org/apache/pekko/actor/dungeon/ReceiveTimeout.scala
```diff
@@ -22,63 +22,83 @@ import pekko.actor.Cancellable
...
+  private final class State(var timeout: Duration, var task: Cancellable, var version: Int)
 }
 
 private[pekko] trait ReceiveTimeout { this: ActorCell =>
 
   import ActorCell._
   import ReceiveTimeout._
 
-  private var receiveTimeoutData: (Duration, Cancellable) = emptyReceiveTimeoutData
+  private var receiveTimeoutData: State = _
```
pjfanning: Is Scala 3.8/3.9 moving us towards using explicit nulls in instances like this?
Or maybe
```
import scala.compiletime.uninitialized
private var receiveTimeoutData: State = uninitialized
```

He-Pin: Used explicit `null` instead. `scala.compiletime.uninitialized` is Scala 3 only and this file lives in the shared `src/main/scala/` directory which must cross-compile with Scala 2.13. Explicit `null` is clearer than `= _` and works on both versions.

pjfanning (approved):
lgtm

PR #3342, opened 2026-07-13
refactor: replace sort-imports plugin with scalafix OrganizeImports

He-Pin:
### Motivation
The `com.nequissimus:sort-imports:0.6.1` plugin has not been updated since December 2021 and is effectively unmaintained. Scalafix's built-in `OrganizeImports` rule (available since 0.11.0, project uses 0.14.7) is more powerful, actively maintained, and requires no external dependency.

### Modification
- Replace `SortImports` rule with `OrganizeImports` in `.scalafix.conf`, configured with 10 import groups matching Pekko's existing style (java/javax, scala, third-party, com.sun, pekko/akka, reactivestreams, netty, scalatest, slf4j, typesafe)
- Remove `SortImports` from `.scalafmt.conf` `rewrite.rules` to avoid conflict with scalafix
- Remove `com.nequissimus:sort-imports:0.6.1` dependency from `ScalaFixExtraRulesPlugin`
- Update `sortImports` command alias in `build.sbt`, `ScalaFixForJdk21Plugin`, and `ScalafixForMultiNodePlugin` to use `OrganizeImports`
- Add `-Wunused:imports` compiler option for Scala 2.13 and Scala 3 to support `OrganizeImports.removeUnused`
- Disable `RemoveUnused.imports` (conflicts with `OrganizeImports`) and enable `OrganizeImports.removeUnused = true` instead
- Apply full project import reformatting across all main and test sources (1991 files)

### Result
Import sorting now uses scalafix's built-in `OrganizeImports` rule with no external dependency. The `sbt sortImports` command continues to work as before, now also automatically removing unused imports. Import grouping matches Pekko's existing style with 10 configured groups.

### Tests
- `sbt "Test / compile"` passed with Java 21

### References
None - build tooling improvement


build.sbt
```diff
@@ -48,7 +48,7 @@ addCommandAlias(
   name = "fixall",
   value = ";scalafixEnable; scalafixAll; scalafmtAll; test:compile; multi-jvm:compile; reload")
 
-addCommandAlias(name = "sortImports", value = ";scalafixEnable; scalafixAll SortImports; scalafmtAll")
+addCommandAlias(name = "sortImports", value = ";scalafixEnable; scalafixAll OrganizeImports; scalafmtAll")
```
He-Pin, line 52: OrganizeImports

actor-testkit-typed/src/main/scala/org/apache/pekko/actor/testkit/typed/internal/BehaviorTestKitImpl.scala
```diff
@@ -23,6 +23,7 @@ import scala.util.control.Exception.Catcher
 import scala.util.control.NonFatal
 
 import org.apache.pekko
+
```
pjfanning, line 26: this is a change that I prefer that we fix - let's keep no gap between the org.apache.pekko import and the pekko imports

He-Pin: Still tweaking

He-Pin: This seems can't be done for now.

pjfanning: I think we should not merge this until we can get the tooling to support our preferred sort order and style.

He-Pin: Need a fix in scala-fix I think

He-Pin: https://github.com/scalacenter/scalafix/pull/2478

Issue #3345, opened 2026-07-14
BroadcastHub: published source not completed when materializing actor dies/stops


ifta-zoy:
# Summary

When materializing a BroadcastHub on an actor's materializer, stopping/crashing the actor does not complete the BroadcastHub Source.

# Analysis

Two independent actors are running in this situation: 
The first ("Worker") is materializing a Source via a `BroadcastHub.sink`, running with a Materializer bound to the actor. The source created by the BroadcastHub is then passed to a different actor ("Listener") that runs it to itself using an ActorSink.

When the Worker stops (or is crashed), the source created by the BroadcastHub is not terminated (i.e. the Listener does not receive a Completed/Error message). 
The source that runs into the BroadcastHub is cancelled, as expected. 

If the source that runs the BroadcastHub is completed/failed manually, the BroadcastHub source completes as expected.

The reproducer has a third actor that sends a Crash/Stop message to the Worker after a couple of seconds.

# Reproducer

```Scala
import org.apache.pekko.NotUsed
import org.apache.pekko.actor.typed.scaladsl.Behaviors
import org.apache.pekko.actor.typed.{ActorRef, ActorSystem, Behavior}
import org.apache.pekko.dispatch.ExecutionContexts
import org.apache.pekko.stream.scaladsl.{BroadcastHub, Keep, Source, SourceQueueWithComplete}
import org.apache.pekko.stream.typed.scaladsl.ActorSink
import org.apache.pekko.stream.{Materializer, OverflowStrategy}

import scala.concurrent.duration.DurationInt
import scala.util.{Failure, Success}

// Actor 1: creates a Source and materializes it to a BroadcastHub. Uses Materializer bound to the actor.
sealed trait WorkerMsg
case object CrashOrStop    extends WorkerMsg
object Worker {
  private def running(queue: SourceQueueWithComplete[Int]): Behavior[WorkerMsg] = Behaviors.receiveMessage {
    case CrashOrStop =>
//      queue.complete()                                // => expected Behavior -- Source is closed
//      queue.fail(RuntimeException("Worker crashed"))  // => expected Behavior -- same
//      throw RuntimeException("Worker has crashed")    // => unexpected Behavior -- Source not closed
      Behaviors.stopped                                 // => unexpected Behavior -- same
  }

  def apply(listenerRef: ActorRef[ListenerMsg]): Behavior[WorkerMsg] = Behaviors.setup { ctx =>
    given Materializer = Materializer(ctx)

[...]

PR #3347, opened 2026-07-14
fix(stream): notify PartitionHub consumers on materializer shutdown

He-Pin:
### Motivation

When the materializer running a `PartitionHub.sink` is shut down (e.g. the hosting actor stops or crashes), consumers materialized on a different materializer may not be notified of the hub's termination. This mirrors the same bug pattern found in `BroadcastHub` (see #3345).

### Modification

- Move registered-consumer notification from `onUpstreamFailure` to `postStop` so that `postStop` is the single notification point for all consumers.
- `postStop` now handles both `Open` (normal termination) and `Closed` (after upstream failure) states, ensuring registered consumers always receive the appropriate signal.

**Before:** `onUpstreamFailure` notified registered consumers directly, but `postStop` did not. This left a gap where consumers could miss notification during materializer shutdown.

**After:** `postStop` is the sole notification point for registered consumers, using the failure reason from the `Closed` state when applicable.

### Result

PartitionHub consumers are reliably notified when the hub's materializer shuts down, preventing consumers from hanging indefinitely.

### Tests

- `sbt "stream-tests / Test / testOnly org.apache.pekko.stream.scaladsl.HubSpec"` — 51 passed, 0 failed
- Added new test: `"notify consumers when hub materializer is shut down"` (PartitionHub section)

### References

Refs #3345

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/HubSpec.scala
@@ -968,6 +968,29 @@ class HubSpec extends StreamSpec {
       result2.futureValue should ===((1 to 9 by 2).filterNot(_ == 3))
     }
 
+    "notify consumers when hub materializer is shut down" in {
+      val hubMat = Materializer(system)
+      val upstream = TestPublisher.probe[Int]()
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3351, opened 2026-07-14
Check if expected messages are produced in FlowLimitSpec

Philippus:
resolves a todo


stream-tests/src/test/scala/org/apache/pekko/stream/scaladsl/FlowLimitSpec.scala
```diff
@@ -16,6 +16,7 @@ package org.apache.pekko.stream.scaladsl
 import scala.concurrent.Await
 
 import org.apache.pekko
+import org.apache.pekko.stream.testkit.scaladsl.TestSink
```
pjfanning: can you fix the import to omit the org.apache bit?

pjfanning (approved):
lgtm

PR #3353, opened 2026-07-15
WIP: avoid initializing classes via forName

pjfanning:
Probably doesn't help much but generally, it is a bad idea to take a class name and load it with its static initializers.
`this.getClass.getClassLoader` is the default classloader used by Class.forName.


pjfanning:
most changes done elsewhere

PR #3359, opened 2026-07-17
Stop PersistentActor by default when RecoveryCompleted handler fails

### LINKED ISSUE #3234: PersistentActor does not stop when RecoveryCompleted handler throws
### Motivation
When a `PersistentActor`'s `RecoveryCompleted` handler throws an exception, the actor enters an infinite restart loop instead of stopping. All other recovery failure paths (replayed events, snapshot loading, journal failures, recovery timeout) explicitly stop the actor via `context.stop(self)` and log the failure via `onRecoveryFailure`. The `RecoveryCompleted` path is the only one that lacks this handling. The typed persistence API (`pekko-persistence-typed`) correctly stops the actor in this scenario.

### Current behavior
In `Eventsourced.scala`, the `recovering` state handles `RecoverySuccess` as follows (lines 786-789):

```scala
try Eventsourced.super.aroundReceive(recoveryBehavior, RecoveryCompleted)
finally transitToProcessingState() // in finally in case exception and resume strategy
// if exception from RecoveryCompleted the permit is returned in below catch
returnRecoveryPermit()
```

When the `RecoveryCompleted` handler throws, the exception propagates to the outer catch (lines 810-813):

```scala
catch {
  case NonFatal(e) =>
    returnRecoveryPermit()
    throw e
}
```

This rethrows to the `ActorCell`, which applies the default supervisor strategy (`Restart` for `Exception`). The actor restarts, re-enters recovery, throws again, and loops indefinitely.

Compare with every other recovery failure path in the same state, e.g. `ReplayedMessage` (lines 772-778):

```scala
catch {
  case NonFatal(t) =>
    timeoutCancellable.cancel()
    try onRecoveryFailure(t, Some(p.payload))
    finally context.stop(self)
    returnRecoveryPermit()
}
```

All other paths call `onRecoveryFailure` and `context.stop(self)`. The `RecoveryCompleted` path calls neither.

### Expected behavior
When the `RecoveryCompleted` handler throws, the actor should:
1. Call `onRecoveryFailure(cause, event = None)` to log the error
2. Call `context.stop(self)` to stop the actor
3. Call `returnRecoveryPermit()` to release the recovery permit

This matches the behavior of all other recovery failure paths in the same file and the behavior of the typed persistence API (where the default `supervisionStrategy` for `JournalFailureException` is `SupervisorStrategy.stop`, see `EventSourcedBehaviorImpl.scala` line 120).

### Code evidence
- `persistence/src/main/scala/org/apache/pekko/persistence/Eventsourced.scala:786-789` -- the `RecoveryCompleted` handler invocation without `onRecoveryFailure` or `context.stop(self)`
- `persistence/src/main/scala/org/apache/pekko/persistence/Eventsourced.scala:810-813` -- the outer catch that returns the permit and rethrows
- `persistence/src/main/scala/org/apache/pekko/persistence/Eventsourced.scala:772-778` -- the `ReplayedMessage` handler showing the correct pattern
- `persistence/src/main/scala/org/apache/pekko/persistence/Eventsourced.scala:676-681` -- the `SnapshotOffer` handler showing the correct pattern
- `persistence/src/test/scala/org/apache/pekko/persistence/RecoveryPermitterSpec.scala:174-197` -- existing test that documents the restart loop behavior (comment: `// it's restarting`), verifying only that the permit is returned, not that the actor lifecycle is correct
- `persistence-typed/src/main/scala/org/apache/pekko/persistence/typed/internal/EventSourcedBehaviorImpl.scala:120` -- typed API default `SupervisorStrategy.stop` for journal failures

### Reproduction
1. Create a `PersistentActor` whose `receiveRecover` throws when it receives `RecoveryCompleted`
2. Persist at least one event and restart the actor
3. Observe: the actor enters an infinite restart loop (repeated `RecoveryCompleted` -> exception -> restart -> `RecoveryCompleted` -> exception -> ...)
4. Expected: the actor should stop after the first failure, logging the error via `onRecoveryFailure`

Suggested test: update `RecoveryPermitterSpec` "return permit when actor throws from RecoveryCompleted" to assert the actor stops (is terminated) after the first `RecoveryCompleted` exception, rather than expecting 5+ restarts.

### Impact
- **Module**: `pekko-persistence` (classic PersistentActor API)
- **Affected users**: Any user whose `receiveRecover` handler for `RecoveryCompleted` can throw (e.g., state validation, external resource initialization). A single exception causes an infinite restart loop, consuming resources and potentially flooding logs.
- **Not affected**: `pekko-persistence-typed` (already handles this correctly via `SupervisorStrategy.stop`)

### LINKED ISSUE #3249: PersistentActor enters infinite restart loop when RecoveryCompleted handler throws
### Motivation

When a classic `PersistentActor`'s `RecoveryCompleted` handler throws an exception, the actor enters an infinite restart loop instead of stopping. This is inconsistent with every other recovery failure path, which all explicitly call `context.stop(self)`.

### Current Behavior

In `Eventsourced.scala:786-814`, when `RecoveryCompleted` handler throws:

1. `finally transitToProcessingState()` runs — changes state
2. Exception propagates to outer `catch` which calls `returnRecoveryPermit()` and rethrows
3. **No `context.stop(self)`** is called
4. **No `onRecoveryFailure()` callback** is invoked
5. Default supervision (`Restart` with unlimited retries) causes infinite restart loop
6. Same journal events replayed → same exception → forever

### Comparison with Other Recovery Failure Paths

| Failure Scenario | `onRecoveryFailure()` | `context.stop(self)` | Result |
|---|---|---|---|
| `ReplayedMessage` throws | Yes | Yes | Actor stops |
| `ReplayMessagesFailure` | Yes | Yes | Actor stops |
| `RecoveryTick` timeout | Yes | Yes | Actor stops |
| **`RecoveryCompleted` throws** | **No** | **No** | **Restart loop** |

### Expected Behavior

The `RecoveryCompleted` failure path should be consistent with other recovery failure paths — call `onRecoveryFailure()` and `context.stop(self)` to prevent the infinite restart loop.

### Reproduction

Confirmed by existing test `RecoveryPermitterSpec.scala:174-197` which observes 5 consecutive restarts and must manually call `system.stop()` to break the loop.

### Environment

- Pekko Persistence (any version)
- `Eventsourced.scala:786-814`

### References

None

He-Pin:
### Motivation
A classic `PersistentActor` can enter an unbounded restart loop when its `RecoveryCompleted` handler throws because the Classic default decider restarts ordinary exceptions. Issues #3234 and #3249 describe the same trigger and root cause.

`RecoveryCompleted` runs after event replay has succeeded, so this failure should remain visible to the parent supervisor instead of being translated into an `onRecoveryFailure` replay failure.

### Modification
- Wrap `NonFatal` exceptions from the `RecoveryCompleted` handler in `ActorInitializationException` and pass them to Classic supervision.
- Preserve the existing processing-state transition and outer recovery-permit cleanup path.
- Add directional regression coverage for both the default Stop decision and a custom parent Restart decision.
- Document the resulting supervision contract.

### Result
The default supervision strategy stops the actor instead of repeatedly restarting it. A custom parent strategy can still Resume or Restart the same `ActorRef`, and the original exception is retained as the cause.

### Tests
- `sbt "persistence / Test / testOnly org.apache.pekko.persistence.RecoveryPermitterSpec"` — restart-direction test failed before the production change; 7 passed after it
- `sbt "persistence / Test / test"` — 62 passed
- `sbt docs/paradox` — passed with existing duplicate-anchor warnings
- `sbt headerCreateAll` — passed
- `sbt +headerCheckAll` — passed for Scala 2.13.18 and 3.3.8
- `sbt checkCodeStyle` — passed
- `sbt validatePullRequest` — ran for 24m44s; affected persistence, typed, query, cluster-sharding, multi-JVM, and docs tests completed. Overall exit 1 is limited to unchanged baseline/environment failures: `LeveldbJournalNativeSpec` cannot load the x86 `leveldbjni-all` 1.8 binary on macOS ARM64, and `IntegrationDocTest` references the existing invalid class `pekko.dispatch.UnboundedMailbox`
- `sbt sortImports` — not completed because the repository Scalafix rule fails against the current scala.meta with `NoSuchMethodError`; no unrelated changes were retained
- `git diff --check` — passed
- Qoder stdout review — No must-fix findings

### References
Fixes #3234
Fixes #3249

### DIFF
--- a/docs/src/main/paradox/persistence.md
+++ b/docs/src/main/paradox/persistence.md
@@ -215,6 +215,10 @@ The actor will always receive a `RecoveryCompleted` message, even if there are n
 in the journal and the snapshot store is empty, or if it's a new persistent actor with a previously
 unused `persistenceId`.
 
+If handling `RecoveryCompleted` throws an exception, the failure is reported to the parent supervisor
+as an `ActorInitializationException`. The default supervision strategy stops the actor, while a custom
+supervision strategy may resume or restart it.
+
 If there is a problem with recovering the state of the actor from the journal, `onRecoveryFailure`
 is called (logging the error by default) and the actor will be stopped.
 
--- a/persistence/src/main/scala/org/apache/pekko/persistence/Eventsourced.scala
+++ b/persistence/src/main/scala/org/apache/pekko/persistence/Eventsourced.scala
@@ -22,7 +22,7 @@ import scala.concurrent.duration.FiniteDuration
 import scala.util.control.NonFatal
 
 import org.apache.pekko
-import pekko.actor.{ Actor, ActorCell, DeadLetter, StashOverflowException }
+import pekko.actor.{ Actor, ActorCell, ActorInitializationException, DeadLetter, StashOverflowException }
 import pekko.annotation.{ InternalApi, InternalStableApi }
 import pekko.dispatch.Envelope
 import pekko.event.{ Logging, LoggingAdapter }
@@ -792,9 +792,15 @@ private[persistence] trait Eventsourced
               sequenceNr = highestSeqNr
               setLastSequenceNr(highestSeqNr)
               _recoveryRunning = false
-              try Eventsourced.super.aroundReceive(recoveryBehavior, RecoveryCompleted)
-              finally transitToProcessingState() // in finally in case exception and resume strategy
-              // if exception from RecoveryCompleted the permit is returned in below catch
+              try {
+                Eventsourced.super.aroundReceive(recoveryBehavior, RecoveryCompleted)
+              } catch {
+                case NonFatal(t) =>
+                  throw ActorInitializationException(
+                    self,
+                    s"Exception in receiveRecover when handling RecoveryCompleted for persistenceId [$persistenceId]",
+                    t)
+              } finally transitToProcessingState()
               returnRecoveryPermit()
             case ReplayMessagesFailure(cause) =>
               timeoutCancellable.cancel()
--- a/persistence/src/test/scala/org/apache/pekko/persistence/RecoveryPermitterSpec.scala
+++ b/persistence/src/test/scala/org/apache/pekko/persistence/RecoveryPermitterSpec.scala
@@ -27,6 +27,7 @@ import com.typesafe.config.ConfigFactory
 object RecoveryPermitterSpec {
 
   class TestExc extends RuntimeException("simulated exc") with NoStackTrace
+  final case class Supervised(cause: Throwable)
 
   def testProps(name: String, probe: ActorRef, throwFromRecoveryCompleted: Boolean = false): Props =
     Props(new TestPersistentActor(name, probe, throwFromRecoveryCompleted))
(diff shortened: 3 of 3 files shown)


pjfanning (approved):
lgtm

PR #3361, opened 2026-07-18
fix(remote): buffer messages during passive handoff

### LINKED ISSUE #3208: fix(remote): buffer inbound payloads during classic remoting notReading state for replay on ResumeReading
### Motivation

PR #3205 fixes the core issue where the outbound reader remains permanently stopped after a passive read handoff failure (akka/akka#24393). However, even with that fix, there is a message loss window between `StopReading` and `ResumeReading` during which inbound payloads arriving on the outbound association are discarded by `EndpointReader.notReading`.

Cluster protocol messages like `InitJoinAck` are **unreliable messages** (`ClusterMessage with DeadLetterSuppression`, not `SystemMessage` — see `ClusterDaemon.scala:155`). They have no `reliableDeliveryEnabled` flag, no seq number, and no retransmission mechanism. When discarded during the notReading window, they are permanently lost. The cluster join protocol must retry from scratch, increasing rejoin latency after node restarts.

**Reliable messages (system messages) do NOT need buffering**: they have seq-based retransmission. If the read-only endpoint dies during the notReading window, the remote system detects the TCP disconnection and re-sends on the outbound connection. After ResumeReading restores the reader, re-sent messages are received and delivered normally.

### Modification

In `EndpointReader.notReading`, buffer only **unreliable** messages instead of discarding them. When `ResumeReading` arrives, synchronously dispatch the buffered messages, then return to the normal `receive` state.

Key design decisions:

1. **Only buffer unreliable messages** (`!msg.reliableDeliveryEnabled`). Reliable messages must NOT be cached: if the read-only endpoint already delivered them before failing, replaying from the writable endpoint would cause cross-reader duplicate delivery — the writable endpoint's `AckedReceiveBuffer` has no record of those messages (they were discarded in `notReading`), so seq-based dedup would not catch the duplicate.

2. **Synchronous dispatch on ResumeReading**, not `self ! InboundPayload(p)`. The `self !` approach queues replayed messages at the mailbox tail, which can invert temporal ordering with already-queued new payloads (buffered msg at T1 delivered after new msg at T2 where T2 > T1). Synchronous dispatch also avoids redundant protobuf re-decoding and duplicate ACK sending.

3. **Buffer upper bound**. In the normal path (read-only endpoint works, no ResumeReading), the writable endpoint stays in `notReading` indefinitely. Without a limit, the buffer grows unbounded. A configurable upper bound (e.g., 128 messages) prevents this.

Proposed change in `EndpointReader`:

```scala
private var suspendedMessages: Vector[PekkoPduCodec.Message] = Vector.empty
private val MaxSuspendedMessages = 128

def notReading: Receive = {
  case ResumeReading =>
    suspendedMessages.foreach { msg =>
      try msgDispatch.dispatch(msg.recipient, msg.recipientAddress,
        msg.serializedMessage, msg.senderOption)
      catch { case NonFatal(e) => logTransientSerializationError(msg, e) }
    }
    suspendedMessages = Vector.empty
    context.become(receive)

  case InboundPayload(p) if p.length <= transport.maximumPayloadBytes =>
    val (ackOption, msgOption) = tryDecodeMessageAndAck(p)
    for (ack <- ackOption; reliableDelivery <- reliableDeliverySupervisor)
      reliableDelivery ! ReliableDeliverySupervisor.AckFromReader(uid, ack)

    msgOption.foreach { msg =>
      if (!msg.reliableDeliveryEnabled &&
          suspendedMessages.size < MaxSuspendedMessages)
        suspendedMessages :+= msg
    }

    if (log.isWarningEnabled)
      log.warning(
        "Discarding inbound message to [{}] in read-only association to [{}]. ...",
        msgOption.map(_.recipient).getOrElse("unknown"),
        remoteAddress)
}
```

### Test plan

Add a test in `RemotingSpec` near the existing handoff regression test:

1. Establish outbound association (A→B)
2. Trigger inbound association → StopReading on writable endpoint
3. Write an unreliable message from B on the outbound connection (A→B)
4. Break the read-only endpoint (triggers ResumeReading)
5. Assert the unreliable message is received after ResumeReading

### Result

- No unreliable message loss during the `notReading` → `ResumeReading` recovery window
- `InitJoinAck` and similar cluster messages are preserved, reducing rejoin latency after node restarts
- No impact on the normal passive connection flow (when read-only endpoint works correctly, buffer stays empty or is never replayed)

### Scope

- Classic remoting only. Artery uses shared global inbound streams and has no `StopReading`/`notReading` mechanism — not affected.

### Dependencies

Depends on #3205 (ResumeReading mechanism). Should be merged after #3205 or as a stacked PR.

### References

- Builds on #3205
- Original Akka issue: https://github.com/akka/akka/issues/24393

He-Pin:
### Motivation

Classic remoting suspends the original read-only reader while a passive connection takes over. Payloads arriving on the suspended reader were discarded, so they were permanently lost when the replacement connection failed and the original reader resumed.

### Modification

- Add a bounded `pekko.remote.classic.passive-connection-buffer-size` setting, defaulting to 128 messages.
- Buffer unreliable payloads while the original reader is suspended; reliable payloads continue to rely on retransmission.
- Track whether a replacement reader dispatched a message. Replay the original buffer only when it is ordering-safe; otherwise reconnect the writer while preserving reliable-delivery sequence and resend state.
- Add focused coverage for replay, overflow, reliable-message handling, cross-connection ordering, reconnect state, configuration, and binary compatibility.

### Result

Failed passive handoffs recover buffered unreliable payloads when safe, without replaying older messages after a replacement has already delivered newer ones or losing buffered system messages during reconnect.

### References

Fixes #3208

### DIFF
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -355,6 +355,12 @@ pekko {
       # Reuse inbound connections for outbound messages
       use-passive-connections = on
 
+      # Maximum number of unreliable inbound messages to buffer on the original
+      # outbound connection while its reader is suspended during a passive
+      # connection handoff. The messages are replayed if the passive reader fails
+      # before delivering a message; additional messages are discarded.
+      passive-connection-buffer-size = 128
+
       # Controls the backoff interval after a refused write is reattempted.
       # (Transports may refuse writes if their internal buffer is full)
       backoff-interval = 5 ms
--- a/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/Endpoint.scala
@@ -15,6 +15,7 @@ package org.apache.pekko.remote
 
 import java.io.NotSerializableException
 import java.util.concurrent.{ ConcurrentHashMap, TimeoutException }
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.locks.LockSupport
 
 import scala.annotation.nowarn
@@ -224,6 +225,7 @@ private[remote] class OversizedPayloadException(msg: String) extends EndpointExc
  */
 private[remote] object ReliableDeliverySupervisor {
   case object Ungate
+  case object Reconnect
   case object AttemptSysMsgRedelivery
   final case class GotUid(uid: Int, remoteAddres: Address)
   @nowarn("msg=deprecated")
@@ -308,6 +310,11 @@ private[remote] class ReliableDeliverySupervisor(
     bailoutAt = None
   }
 
+  private def prepareForReconnect(): Unit = {
+    currentHandle = None
+    uidConfirmed = false
+  }
+
   reset()
 
   def nextSeq(): SeqNo = {
@@ -387,6 +394,11 @@ private[remote] class ReliableDeliverySupervisor(
     case s: EndpointWriter.StopReading =>
       writer.forward(s)
 
+    case Reconnect =>
+      prepareForReconnect()
+      context.stop(writer)
+      context.become(gated(writerTerminated = false, earlyUngateRequested = true))
+
     case EndpointWriter.ResumeReading =>
       writer ! EndpointWriter.ResumeReading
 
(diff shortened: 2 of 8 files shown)


pjfanning (approved):
lgtm

Issue #3253, opened 2026-06-27
AdjustPoolSize bypasses pool configuration bounds


He-Pin:
### Motivation

Sending `AdjustPoolSize` to a pool router can grow or shrink the pool beyond configured limits (`nr-of-instances`, resizer `lowerBound`/`upperBound`). The handler directly creates or removes routees without any bounds checking.

### Current Behavior

`RoutedActorCell.scala:205-213`:

```scala
case AdjustPoolSize(change: Int) =>
  if (change > 0) {
    val newRoutees = Vector.fill(change)(pool.newRoutee(cell.routeeProps, context))
    cell.addRoutees(newRoutees)
  } else if (change < 0) {
    val currentRoutees = cell.router.routees
    val abandon = currentRoutees.drop(currentRoutees.length + change)
    cell.removeRoutees(abandon, stopChild = true)
  }
```

- No check against `pool.nrOfInstances(system)` (initial pool size)
- No check against resizer `lowerBound` or `upperBound`
- `pool.newRoutee()` is just a factory — creates routee with correct dispatcher but no bounds check

### Resizer Bounds Bypassed

`DefaultResizer.capacity()` (`Resizer.scala:178-187`) does enforce bounds, but it's only called from `ResizablePoolCell.resize()` — NOT from the `AdjustPoolSize` handler. Even `ResizablePoolActor` delegates `AdjustPoolSize` to `super.receive` (the unbounded handler).

### Test Confirms Unbounded Behavior

`RoundRobinSpec.scala:117-120` shows `AdjustPoolSize(+4)` growing a pool from 3 to 7 (beyond configured size), with no bounds test asserting this is correct.

### Expected Behavior

`AdjustPoolSize` should respect pool configuration bounds. If a resizer is configured, the handler should delegate to the resizer's capacity logic. If no resizer is configured, the handler should at minimum not shrink below 1 or grow beyond a reasonable limit.

### Environment

- Pekko Actor (any version)
- `RoutedActorCell.scala:205-213`

### References

None

pjfanning:
@He-Pin  is this fixed by #3364 ?

PR #3363, opened 2026-07-18
docs: clarify pool management and resizer bounds

He-Pin:
### Motivation

`AdjustPoolSize` is a direct management operation, but the resizer documentation described lower and upper bounds as values the router should ever have. This made the relationship between initial pool size, explicit management, and automatic resizing unclear in #3253.

### Modification

Clarify that `nr-of-instances` defines the initial pool size, routee-changing management messages may move a pool outside configured resizer bounds, and a later resize may restore the configured range. Document the asynchronous recovery and message-loss behavior when a resizable pool is reduced to zero routees.

### Result

Router management and automatic resizer responsibilities are documented consistently with the existing implementation and API contract.

### Tests

- Not run - docs only
- `sbt "actor / Compile / doc" "docs / paradox"`
- `sbt +headerCheckAll checkCodeStyle`
- `git diff --check origin/main`

### References

Refs #3253

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -240,10 +240,14 @@ pekko {
 
           enabled = off
 
-          # The fewest number of routees the router should ever have.
+          # These bounds are enforced when this resizer is invoked. Router
+          # management messages can move the pool outside them until the
+          # resizer runs again.
+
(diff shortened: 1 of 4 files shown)


pjfanning (approved):
lgtm

PR #3369, opened 2026-07-19
refactor to use StackWalker

pjfanning:
Not a big improvement but I think this is a bit tidier and we use StackWalker in a few places in this project now


raboof (approved):
neat!

PR #3371, opened 2026-07-20
Clean up todos in traversal builder

Philippus:


stream-tests/src/test/scala/org/apache/pekko/stream/impl/TraversalBuilderSpec.scala
```diff
@@ -14,8 +14,8 @@
 package org.apache.pekko.stream.impl
 
 import scala.concurrent.Promise
-
 import org.apache.pekko
+import org.apache.pekko.stream.impl.StreamLayout.AtomicModule
```
pjfanning: can you remove the org.apache from this?

He-Pin:
better with jmh numbers

He-Pin (approved):

He-Pin:
Not sure about whether an additional map operation will reduce performance.

He-Pin (changes requested):
I think this will reduce some performance ? the code is correct.

Philippus:
Not sure why this would reduce performance? There are (almost) no code changes in this PR, also no additional map operations.

pjfanning:
I agree with Philippus. This change doesn't look like it contains anything that affects performance.

He-Pin (approved):

PR #3377, opened 2026-07-25
Deprecate org.apache.pekko.dispatch.Futures

### LINKED ISSUE #1417: Deprecate `org.apache.pekko.dispatch.Futures`

He-Pin:
### Motivation

The `org.apache.pekko.dispatch.Futures` class was introduced when Akka still supported Java versions before Java 8, which introduced `CompletionStage`. Java users should not need to touch the Scala `Future` at all — we should make sure that where they currently must use `Future`, we provide equivalent APIs using the more Java-native `CompletionStage`.

### Modification

- Deprecate the entire `org.apache.pekko.dispatch.Futures` object (since 2.0.0) with migration guidance:
  - `Futures.asJava` → `CompletionStages.fromScala`
  - `Futures.future(body, executor)` → `CompletableFuture.supplyAsync(body, executor)`
  - `Futures.promise()` → `new CompletableFuture<>()`
  - `Futures.failed(ex)` → `CompletableFuture.failedFuture(ex)`
  - `Futures.successful(result)` → `CompletableFuture.completedFuture(result)`
  - `Futures.failedCompletionStage(ex)` → `CompletableFuture.failedStage(ex)`
- Add `CompletionStages.fromScala` (since 2.0.0) as the replacement for `Futures.asJava`
- Replace internal usage of `Futures.failedCompletionStage` in `CompletionStages.reduce` with `CompletableFuture.failedStage`
- Migrate Java tests to use standard `java.util.concurrent.CompletableFuture` APIs (Java 9+/17):
  - `CompletableFuture.failedStage(ex)` instead of `Futures.failedCompletionStage(ex)`
  - `CompletableFuture.completedFuture(result)` instead of `Futures.successful(result)`
  - `CompletionStages.fromScala(future)` instead of `Futures.asJava(future)`
  - `scala.concurrent.Promise$.MODULE$.apply()` instead of `Futures.promise()`

### Result

Java users are guided toward standard `java.util.concurrent.CompletableFuture` APIs (Java 9+/17) and `CompletionStages` utilities. The deprecated `Futures` object will be removed in a future major version.

### Tests

- `sbt "actor-tests / Test / testOnly org.apache.pekko.dispatch.JavaFutureTests org.apache.pekko.dispatch.CompletionStagesTests org.apache.pekko.pattern.PatternsTest"` — 44 passed, 0 failed
- `sbt "actor/mimaReportBinaryIssues"` — passed (deprecation annotations are binary compatible)

### References

Fixes #1417

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
@@ -158,7 +158,7 @@ object CompletionStages {
     if (iterator.hasNext) {
       iterator.next().thenCompose { v => foldWithNext[T, R](iterator, v, function) }
     } else {
-      CompletableFuture.failedFuture(new NoSuchElementException("reduce of an empty iterable of CompletionStages"))
+      CompletableFuture.failedStage(new NoSuchElementException("reduce of an empty iterable of CompletionStages"))
     }
   }
 
--- a/actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
+++ b/actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
@@ -22,8 +22,6 @@ import java.util.concurrent.{ CompletableFuture, CompletionStage, Executor }
 import java.util.concurrent.atomic.AtomicReference
 import java.util.function.{ BiConsumer, BiFunction }
 
-import scala.annotation.nowarn
-
 import org.apache.pekko
 
 /**
@@ -37,6 +35,16 @@ object CompletionStages {
   def asScala[T](stage: CompletionStage[T]): scala.concurrent.Future[T] = {
     import scala.jdk.FutureConverters._
     stage.asScala
+  }
+
+  /**
+   * Convert a Scala `Future` to a `CompletionStage`.
+   *
+   * @since 2.0.0
+   */
+  def fromScala[T](future: scala.concurrent.Future[T]): CompletionStage[T] = {
+    import scala.jdk.FutureConverters._
+    future.asJava
   }
 
   /**
@@ -143,7 +151,6 @@ object CompletionStages {
    *         or a `NoSuchElementException` if the given iterable is empty
    * @since 1.2.0
    */
-  @nowarn("msg=deprecated")
   def reduce[T <: AnyRef, R >: T](
       stages: java.lang.Iterable[? <: CompletionStage[? <: T]],
       function: pekko.japi.function.Function2[R, T, R]): CompletionStage[R] = {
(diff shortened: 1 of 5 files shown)


actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
```diff
@@ -39,6 +37,16 @@ object CompletionStages {
...
 
+  /**
+   * Convert a Scala `Future` to a `CompletionStage`.
+   *
+   * @since 2.0.0
+   */
+  def fromScala[T](future: scala.concurrent.Future[T]): CompletionStage[T] = {
+    import scala.jdk.FutureConverters._
+    future.asJava
+  }
```
raboof: shouldn't we just recommend people to use FutureConverters (for Scala code) or https://www.scala-lang.org/api/2.13.0/scala/jdk/javaapi/FutureConverters$.html (for Java code) instead?

raboof (commented):
This generally seems sensible to me.

I'm not sure we should introduce the `CompletionStages.fromScala` - shouldn't people rather use the regular Scala classes for that?

Should we do a pass over all API's to verify we're not exposing Future's to Java API's anymore for 2.x?

actor/src/main/scala/org/apache/pekko/dispatch/CompletionStages.scala
pjfanning: there is an existing asScala method in this class - maybe this needs to be deprecated if we don't add fromScala - I'm reasonably supported of using scala.jdk.javaapi for these conversions

actor/src/main/scala/org/apache/pekko/dispatch/Future.scala
```diff
@@ -101,29 +105,33 @@ object Futures {
...
    * Creates an already completed Promise with the specified result
    */
+  @deprecated("Use CompletableFuture.completedFuture instead.", since = "2.0.0")
   def successful[T](result: T): Future[T] = Future.successful(result)
 
   /**
    * Creates an already completed CompletionStage with the specified exception
    */
-  @deprecated("Use `CompletableFuture#failedStage` instead.", since = "2.0.0")
+  @deprecated("Use CompletableFuture.failedStage instead.", since = "2.0.0")
```
pjfanning, line 134: it might be useful to add Java `@Deprecated` annotations too because Java compilers may not know what Scala deprecated annotation means

He-Pin: Yes, but otherwise we need to add all of them back. I would like to leave that to a dedicated PR to do it all in once.

pjfanning (approved):
lgtm

PR #3382, opened 2026-07-25
fix: support java.math.BigInteger/BigDecimal in MetricNumericConverter

He-Pin:
### Motivation
`MetricNumericConverter.convertNumber` only handled `scala.math.BigInt` and `scala.math.BigDecimal` but not `java.math.BigInteger` and `java.math.BigDecimal`. Since `Metric.value` is typed as `java.lang.Number`, Java users passing `java.math.BigInteger` or `java.math.BigDecimal` would get an `IllegalArgumentException`. The `NumberInputStream` deserializer also explicitly allows `java.math` classes through, so deserialized values could also fail.

### Modification
Add pattern match cases for `java.math.BigInteger` and `java.math.BigDecimal` in `convertNumber`, placed before the Scala `BigInt`/`BigDecimal` cases.

### Result
Java users can now use `java.math.BigInteger` and `java.math.BigDecimal` as `Metric` values without errors.

### Tests
- `sbt "cluster-metrics / Test / testOnly org.apache.pekko.cluster.metrics.MetricNumericConverterSpec"` — All 7 tests passed
- `sbt "cluster-metrics / Test / testOnly org.apache.pekko.cluster.metrics.protobuf.MessageSerializerSpec"` — All 2 tests passed

### References
None - discovered during Java API audit for Scala number type usage

### DIFF
--- a/cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/Metric.scala
+++ b/cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/Metric.scala
@@ -271,13 +271,15 @@ private[metrics] trait MetricNumericConverter {
    * May involve rounding or truncation.
    */
   def convertNumber(from: Any): Either[Long, Double] = from match {
-    case n: Int        => Left(n)
-    case n: Long       => Left(n)
-    case n: Double     => Right(n)
-    case n: Float      => Right(n)
-    case n: BigInt     => Left(n.longValue)
-    case n: BigDecimal => Right(n.doubleValue)
-    case x             => throw new IllegalArgumentException(s"Not a number [$x]")
+    case n: Int                  => Left(n)
+    case n: Long                 => Left(n)
+    case n: Double               => Right(n)
+    case n: Float                => Right(n)
+    case n: java.math.BigInteger => Left(n.longValue)
+    case n: java.math.BigDecimal => Right(n.doubleValue)
+    case n: BigInt               => Left(n.longValue)
+    case n: BigDecimal           => Right(n.doubleValue)
+    case x                       => throw new IllegalArgumentException(s"Not a number [$x]")
   }
 
 }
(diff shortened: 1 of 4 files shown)


cluster-metrics/src/main/scala/org/apache/pekko/cluster/metrics/Metric.scala
```diff
@@ -271,13 +271,15 @@ private[metrics] trait MetricNumericConverter {
...
-    case n: Float      => Right(n)
-    case n: BigInt     => Left(n.longValue)
-    case n: BigDecimal => Right(n.doubleValue)
-    case x             => throw new IllegalArgumentException(s"Not a number [$x]")
+    case n: Int                  => Left(n)
+    case n: Long                 => Left(n)
+    case n: Double               => Right(n)
+    case n: Float                => Right(n)
+    case n: java.math.BigInteger => Left(n.longValue)
+    case n: java.math.BigDecimal => Right(n.doubleValue)
```
pjfanning, line 281: Short and Byte are numbers - just match against java.lang.Number and call longValue or doubleValue on it - do this before the `case x` and leave the existing specialised matches

He-Pin: OK，but seems those are not been used in code.

He-Pin: The internal producers (JMX MXBeans, Sigar, protobuf deserialization) only emit Int/Long/Double/Float + the now-added java.math types. Short/Byte never appear in any internal path. A generic Number catch-all would also face the longValue vs doubleValue ambiguity — we'd have to guess the user's intent. The current fail-fast for unrecognized types is intentional to surface misuse early

pjfanning: I don't understand. The PR description says that the API allows any Number. We've never had complaints about not supporting java.math numbers. If we add support for them, why not support other numbers?

He-Pin: I just found ,we need change the serialization code too.

pjfanning (approved):
lgtm - but the code format needs to be fixed

PR #3397, opened 2026-07-29
introduce keystore-password config

pjfanning:
* the affected keystore is in memory so this is just a minor hardening
* adds logging if users don't override the passwords for keystores (including other ones)

### DIFF
--- a/remote/src/main/resources/reference.conf
+++ b/remote/src/main/resources/reference.conf
@@ -1265,6 +1265,10 @@ pekko {
           # This is a convention that people may follow if they wish to save themselves some configuration
           secret-mount-point = /var/run/secrets/pekko-tls/rotating-keys-engine
 
+          # Password for the in-memory keystore used to wrap PEM-loaded keys.
+          # Override this in production to avoid using the well-known default.
+          keystore-password = "changeit"
+
           # The absolute path the PEM file with the private key.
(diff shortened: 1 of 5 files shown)


remote/src/main/resources/reference.conf
```diff
@@ -1265,6 +1265,10 @@ pekko {
           # This is a convention that people may follow if they wish to save themselves some configuration
           secret-mount-point = /var/run/secrets/pekko-tls/rotating-keys-engine
 
+          # Password for the in-memory keystore used to wrap PEM-loaded keys.
+          # Override this in production to avoid using the well-known default.
+          keystore-password = "changeit"
```
He-Pin, line 1270: changeme?

He-Pin (approved):

remote/src/main/resources/reference.conf
pjfanning: this one was set to changeit in the existing code while changeme was hardcoded for other keystore passwords

mkurz:
It would great if on Linux we could support the [Secret Service API](https://specifications.freedesktop.org/secret-service/latest) (KDE also finally is removing kwallet and now supports the API already).

Quick online research found two Java project to access the API (which in the end are "just" d-bus calls):
- https://github.com/swiesend/secret-service
- https://github.com/purejava/secret-service

I also opened https://github.com/swiesend/secret-service/discussions/70

PR #3399, opened 2026-07-29
feat: add timeout duration to ReceiveTimeout message

### LINKED ISSUE #2569: ReceiveTimeout contains no information
The `ReceiveTimeout` contains no duration and then you can't do some logging around it:(

```java
            case ReceiveTimeout timeout -> {
                log.info("SessionEntity [{}] receive timeout :[{}], will stop itself", timeout, getSelf().path());
                getContext().stop(getSelf());
            }
```

He-Pin:
## Summary
- Convert `ReceiveTimeout` from a `case object` singleton to a `final case class ReceiveTimeout(timeout: FiniteDuration)` so the message carries the configured timeout duration
- Users can now access the timeout directly from the message handler: `case timeout: ReceiveTimeout => log.info("timeout: {}", timeout.timeout)`
- Closes #2569

## Motivation

Previously `ReceiveTimeout` was a case object with no information, making it impossible to log or inspect the timeout value without separately accessing `context.receiveTimeout`. As noted in the issue, this made logging awkward:

```java
case ReceiveTimeout timeout -> {
    log.info("SessionEntity [{}] receive timeout :[{}], will stop itself", timeout, getSelf().path());
    getContext().stop(getSelf());
}
```

## Changes

- **`actor/src/main/scala/.../Actor.scala`**: Convert `ReceiveTimeout` from `case object` to `final case class` with `timeout: FiniteDuration` field and Java API `getTimeout` method
- **`actor/src/main/scala/.../dungeon/ReceiveTimeout.scala`**: Scheduler now sends `ReceiveTimeout(timeout)` instead of the singleton
- **All pattern matches**: Updated from stable identifier patterns (`case ReceiveTimeout =>`) to type patterns (`case _: ReceiveTimeout =>`)
- **Java API**: Updated from `matchEquals(ReceiveTimeout.getInstance(), ...)` to `match(ReceiveTimeout.class, ...)`
- **MiMa filters**: Added binary compatibility exclusion for 2.0.x
- **Tests**: Added new test verifying the timeout duration is carried correctly

## Breaking changes (2.0)

- `ReceiveTimeout` is no longer a singleton; pattern matches must use type patterns
- `ReceiveTimeout.getInstance()` removed; Java users should use `.match(ReceiveTimeout.class, ...)`

## Test plan
- [x] `actor-tests/testOnly *.ReceiveTimeoutSpec` — 15 tests pass (including new "carry the configured timeout duration" test)
- [x] `actor-typed-tests/testOnly *.CancelReceiveTimeoutSpec` — passes
- [x] All affected modules compile: actor, actor-typed, cluster, cluster-tools, cluster-sharding, distributed-data, persistence, docs


actor/src/main/scala/org/apache/pekko/actor/Actor.scala
```diff
@@ -144,19 +144,17 @@ private[pekko] final case class AddressTerminated(address: Address)
...
 @SerialVersionUID(1L)
-case object ReceiveTimeout extends ReceiveTimeout {
+final case class ReceiveTimeout(timeout: scala.concurrent.duration.FiniteDuration) extends PossiblyHarmful {
 
   /**
-   * Java API: get the singleton instance
+   * Java API: get the timeout duration
    */
-  def getInstance = this
+  def getTimeout: java.time.Duration = java.time.Duration.ofNanos(timeout.toNanos)
```
pjfanning: use scala.jdk.DuartionConverters

actor/src/main/mima-filters/2.0.x.backwards.excludes/receive-timeout-with-duration.excludes
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# ReceiveTimeout changed from case object to case class with a timeout field
+# https://github.com/apache/pekko/issues/2569
+ProblemFilters.exclude[FinalClassProblem]("org.apache.pekko.actor.ReceiveTimeout")
```
pjfanning, line 20: I presume that this is really an internal class and that we don't need to upgrade the migration guide markdown file for 2.0.0 release

pjfanning (approved):
lgtm

PR #3404, opened 2026-07-30
fix(stream): undeprecate getFirst/getFirstAttribute methods

### LINKED ISSUE #345: fix use of deprecated code
```
[warn] /Users/pj.fanning/code/incubator-pekko-http/http-core/src/main/scala/org/apache/pekko/http/scaladsl/model/HttpEntity.scala:666:23: method getFirst in class Attributes is deprecated (since Akka 2.5.7): Attributes should always be most specific, use get[T]
[warn]           _attributes.getFirst[SizeLimit] match {
```

c4mdax:
This PR removes the @deprecated annotation from getFirst and getFirstAttribute in the Attributes class.

As discussed in apache/pekko-http#345, migrating away from getFirst to get reverses the attribute precedence logic (since getFirst uses reverseIterator to find the least specific attribute, while get finds the most specific one). ¡¡Keeping these methods active and undeprecated is necessary to maintain correct precedence in the HTTP layer!!.

Related Issue:
Resolves apache/pekko-http#345

### DIFF
--- a/stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
+++ b/stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
@@ -304,28 +304,28 @@ final class Attributes private[pekko] (
   }
 
   /**
-    *  INTERNAL API
-    * 
-    * Java API: Get the least specific attribute (added first) of a given `Class` or subclass thereof.
-    * If no such attribute exists the `default` value is returned.
-    */
+   *  INTERNAL API
+   *
+   * Java API: Get the least specific attribute (added first) of a given `Class` or subclass thereof.
+   * If no such attribute exists the `default` value is returned.
+   */
   @InternalStableApi
   def getFirstAttribute[T <: Attribute](c: Class[T], default: T): T =
     getFirstAttribute(c).orElse(default)
 
   /**
-    *  INTERNAL API
-    * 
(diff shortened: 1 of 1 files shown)


pjfanning:
I'd prefer to just leave it deprecated to discourage its use. Or if we must change it, add an @InternalStableApi annotation to it. And change scaladoc to say its an internal method.

stream/src/main/scala/org/apache/pekko/stream/Attributes.scala
```diff
@@ -306,22 +306,22 @@ final class Attributes private[pekko] (
    * Java API: Get the least specific attribute (added first) of a given `Class` or subclass thereof.
    * If no such attribute exists the `default` value is returned.
    */
-  @deprecated("Attributes should always be most specific, use getAttribute[T]", "Akka 2.5.7")
+  // Maintained active: required to retrieve the least specific attribute without breaking precedence. See apache/pekko-http#345
```
pjfanning: this type of comment is useless - should be in the scaladoc or not at all

c4mdax: Thanks for the review! I've just pushed an update that replaces the @deprecated annotations with @InternalStableApi, and prepended INTERNAL API to the scaladocs

pjfanning (approved):
lgtm - this should work, thanks

Issue #3410, opened 2026-08-05
RememberEntitiesShardStore$UpdateDone is leaking over ExtractorAdapter


mehmetsalgar:
Hi,

I am using

HashCodeNoEnvelopeMessageExtractor

which is only expected my sealed Events and crashes by receiving 

RememberEntitiesShardStore$UpdateDone

with Pekko 1.6.0

java.lang.ClassCastException: class org.apache.pekko.cluster.sharding.internal.RememberEntitiesShardStore$UpdateDone cannot be cast to class xxx.TravelSM$TravelSMEvent (org.apache.pekko.cluster.sharding.internal.RememberEntitiesShardStore$UpdateDone and xxx.TravelSM$TravelSMEvent are in unnamed module of loader 'app')
    	at xxx.TravelSMGuardian$$anon$2.entityId(TravelSMGuardian.scala:364)
    	at org.apache.pekko.cluster.sharding.typed.internal.ExtractorAdapter.entityId(ClusterShardingImpl.scala:67)
    	at org.apache.pekko.cluster.sharding.typed.internal.ClusterShardingImpl$$anonfun$1.isDefinedAt(ClusterShardingImpl.scala:177)
    	at org.apache.pekko.cluster.sharding.Shard$$anonfun$idle$1.applyOrElse(Shard.scala:637)

I think the problem lies in 

ExtractorAdapter

override def entityId(message: Any): String = {
  message match {
    case ShardingEnvelope(entityId, _) => entityId
    case ClassicStartEntity(entityId)  => entityId
    case msg                           => delegate.entityId(msg.asInstanceOf[E])
  }
}

does not forsee that it will receivce events from

RememberEntitiesShardStore

definetely HashCodeNoEnvelopeMessageExtractor does also not.

Is this event leaking to ExtractorAdapter or it is wanted the ExtractorAdapter receive this...

with this configuration

pekko {
  ....

  cluster {
    shutdown-after-unsuccessful-join-seed-nodes = 60s
    sharding {
      remember-entities = on
      ...
  }
....
}

pjfanning:
I created #3411

Philippus:
Looks like these unsolved (>3 years old) issues in akka: https://github.com/akka/akka-core/issues/31100 (also reported by @mehmetsalgar) and https://github.com/akka/akka-core/issues/30641.

mehmetsalgar:
[...]

PR #3411, opened 2026-08-05
filter messages from remember-entities store

pjfanning:
Relates to #3410 

I'm not sure why 3410 is happening and any issue is likely to have been there for a while.


pjfanning:
I'll get back to this in the next few days. The test code doesn't compile.

He-Pin (approved):

PR #3413, opened 2026-08-06
Remove Test Reports step from workflow

pjfanning:
Removed Test Reports step from timing-tests workflow.

Our timing tests plan is failing every night. https://github.com/apache/pekko/actions/workflows/timing-tests.yml

This action is now not allowed due to https://github.com/apache/infrastructure-actions/pull/828

Can't find any of the replacements in this allowed list:
https://github.com/apache/infrastructure-actions/blob/main/actions.yml

I don't think this action adds much anyway. When the tests fail, I just look at the logs to find the stacktraces and test reports.

### DIFF
--- a/.github/workflows/timing-tests.yml
+++ b/.github/workflows/timing-tests.yml
@@ -62,12 +62,3 @@ jobs:
             -Dmultinode.Xmx256M \
             -Dmultinode.XX:+AlwaysActAsServerClassMachine \
             clean test
-
-      - name: Test Reports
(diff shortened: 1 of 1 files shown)


raboof (approved):
> I don't think this action adds much anyway. When the tests fail, I just look at the logs to find the stacktraces and test reports.

I'd like seeing test failure details more easily in theory, but in practice I agree I haven't found much use for this particular step

PR #3423, opened 2026-08-08
Add LeaseHealthCheck

Philippus:
Applying this PR will add a `LeaseHealthCheck` class folks can use as one of their startup/liveness/readiness probes or as part of their home grown status check setup.
Rationale: a lease is often used for SBR (split brain resolution), but the lease is only created when a split-brain situation occurs. Finding out that it is actually impossible to acquire a lease because of, for example, a misconfiguration only then is undesirable.

### DIFF
--- a/coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
+++ b/coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
@@ -17,15 +17,17 @@
 
 package org.apache.pekko.coordination.lease
 
-import org.apache.pekko.actor.ActorSystem
-import org.apache.pekko.coordination.lease.scaladsl.{ Lease, LeaseProvider }
-import org.apache.pekko.event.Logging
-import org.apache.pekko.pattern.AskTimeoutException
+import java.util.UUID
 
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.{ Failure, Success }
-import java.util.UUID
 import scala.annotation.nowarn
+
+import org.apache.pekko
+import pekko.actor.ActorSystem
+import pekko.coordination.lease.scaladsl.{ Lease, LeaseProvider }
+import pekko.event.Logging
+import pekko.pattern.AskTimeoutException
 
 /**
  * Performs a lease health check by attempting to acquire and immediately release a test lease.
--- a/coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
+++ b/coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.apache.pekko.coordination.lease
+
+import org.apache.pekko.actor.ActorSystem
+import org.apache.pekko.coordination.lease.scaladsl.{ Lease, LeaseProvider }
+import org.apache.pekko.event.Logging
+import org.apache.pekko.pattern.AskTimeoutException
+
+import scala.concurrent.{ ExecutionContext, Future }
```
pjfanning, line 22: import org.apache.pekko and modify the other pekko imports to omit the org.apache bit

Sort the scala and java imports.

coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
```diff
@@ -0,0 +1,85 @@
...
+ * Returns true if:
+ * - Lease is successfully acquired (true)
+ * - Lease cannot be acquired due to conflict (another owner has it, but the API is reachable) (false)
+ *
+ * Returns false if:
+ * - Any exception occurs (LeaseException, AskTimeoutException, etc.)
+ */
+class LeaseHealthCheck(system: ActorSystem, leaseProviderName: String) extends (() => Future[Boolean]) {
+
+  implicit val executionContext: ExecutionContext = system.dispatcher
```
pjfanning: make this private

coordination/src/main/scala/org/apache/pekko/coordination/lease/LeaseHealthCheck.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  private val log = Logging(system, classOf[LeaseHealthCheck])
+
+  @volatile private var healthCheckPassed = false
+
+  private val leaseName = s"lease-${UUID.randomUUID().toString}"
```
pjfanning: can we share the same uuid instance between the lease and owner name?

coordination/src/test/scala/org/apache/pekko/coordination/lease/LeaseHealthCheckSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+
+    "return false on generic Exception" in {
+      LeaseHealthCheckSpec.nextAcquire = () => Future.failed(new RuntimeException("generic error"))
+      new LeaseHealthCheck(system, "mock-lease").check().futureValue shouldEqual false
+    }
+  }
+}
+
+object LeaseHealthCheckSpec {
+  @volatile var nextAcquire: () => Future[Boolean] = () => Future.successful(true)
```
pjfanning: Would it be possible to make these vars part of an abstract class or trait and then create anonymous classes in the tests that extend the abstract class/trait? This object approach means that tests run concurrently can affect other tests.

pjfanning (approved):
lgtm

PR #3429, opened 2026-08-13
try to improve perf on ByteStrings.apply

pjfanning:
The change adds a pre-computed fragmentStartOffsets array (one Int per fragment, built once in the constructor) and uses java.util.Arrays.binarySearch in apply(idx) to locate the fragment in O(log n) instead of the previous O(n) linear scan. The extra memory is one Int per fragment (typically 3-10 fragments).


He-Pin (approved):

pjfanning:
I'm going to hold this change back until I can benchmark it. I suspect that it hurts us for scenarios where we have 10 or fewer fragments.
We may need something fancier - that is only applied if we have above a certain number of fragments.
I suspect that the right approach is to find a Pekko use case where we need to call apply on a ByteStrings instance and to see if we can optimise for these use cases.

PR #3432, opened 2026-08-14
add changes from Akka 2.8.4

pjfanning:
https://github.com/akka/akka-core/releases/tag/v2.8.4 is now available under the Apache License, Version 2.0.

The bulk import. This brings across several Akka 2.8.4 PRs - specifically for Persistence. Key additions:
- EventWriter (EventWriter.scala) — a new actor for batched event writes in typed persistence (port of Akka PR #32010).
- EventWriterSpec — tests for the above.
- EventSourcedBehaviorSpec — expanded test coverage.
- Durable State Effect.delete() javadsl (port of Akka PR #32027).
- InmemJournal updates and reference.conf config additions to support the new writer.
- ReplayingEvents and Eventsourced — minor wiring changes.
- doc changes from https://github.com/akka/akka-core/pull/32019

Many of the other changes were already handled independently in Pekko.
- PR #32030 - Harden PersistentActorRecoveryTimeoutSpec - Already adequately hardened in Pekko
- PR #32016: Already implemented in Pekko (ActorAdapter.scala lines 102-107)
- PR #32031: Not applicable - Pekko uses different buffer architecture (no PartitionedBuffer, no perPartition)
- PR #32035: Already implemented - Pekko has shouldRetry: (T, Throwable) => Boolean which is more general than Akka's Throwable => Boolean
- PR #31624 reports a TimeoutException at SteppingInmemJournal.step() — the problem is that waiting for the second stepped operation can race with the recovery timeout under heavy CI load. Pekko's fix avoids this by releasing both recovery journal tokens up front instead of stepping them one at a time, eliminating the race.

### DIFF
--- a/docs/src/main/paradox/io-dns.md
+++ b/docs/src/main/paradox/io-dns.md
@@ -27,7 +27,7 @@ Users should pick one of the built in extensions.
 
 @@@
 
-Pekko DNS is a pluggable way to interact with DNS. Implementations much implement `org.apache.pekko.io.DnsProvider` and provide a configuration
+Pekko DNS is a pluggable way to interact with DNS. Implementations must implement `org.apache.pekko.io.DnsProvider` and provide a configuration
 block that specifies the implementation via `provider-object`.
 
 @@@ note { title="DNS via Pekko Discovery" }
--- a/docs/src/main/paradox/split-brain-resolver.md
+++ b/docs/src/main/paradox/split-brain-resolver.md
@@ -227,22 +227,22 @@ of 4 and 5 nodes the side with 5 nodes will survive and the other 4 nodes will b
 in the 5 node cluster, no more failures can be handled, because the remaining cluster size would be
 less than 5. In the case of another failure in that 5 node cluster all nodes will be downed.
 
-Therefore it is important that you join new nodes when old nodes have been removed.
+Therefore, it is important that you join new nodes when old nodes have been removed.
 
 Another consequence of this is that if there are unreachable nodes when starting up the cluster,
 before reaching this limit, the cluster may shut itself down immediately. This is not an issue
 if you start all nodes at approximately the same time or use the `pekko.cluster.min-nr-of-members`
-to define required number of members before the leader changes member status of 'Joining' members to 'Up'
+to define required number of members before the leader changes member status of 'Joining' members to 'Up'.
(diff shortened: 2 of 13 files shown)


He-Pin:
Would you like to split these PR to dedicated ones?

diff.txt
```diff
@@ -0,0 +1,29 @@
+⏺ All changes are in place. Here's a summary of what was implemented from Akka v2.8.4:
```
He-Pin: this is not needed

pjfanning:
Some of the code in the PRs overlap - so there is at least some advantage in not moving one PR at a time.
For time purposes, I think it is as easy to review this size of change in one PR than having it to split up. If it was bigger, then it would make more sense to split it.

diff.txt
pjfanning: removed - I just kept that around for doc purposes and forgot to remove it

He-Pin:
## Review Notes

Thanks for putting this together! A few issues I noticed:

### Must fix

1. **`diff.txt` should be removed** — this file shouldn't be committed to the repository.

2. **`split-brain-resolver.md` missing one fix** — line 384: "links but they" should be "links, but they" (comma after "links"). This was part of the upstream akka/akka-core#32019 fix.

### Missing from v2.8.3...v2.8.4 range

3. **Retry short-circuit Java test** (akka/akka-core#32035) — Pekko already has the `shouldRetry: (T, Throwable) => Boolean` API (since 1.1.0), but lacks a Java DSL directional test for `Patterns.retry` with `Predicate2`. See apache/pekko#3435.

4. **Harden PersistentActorRecoveryTimeoutSpec** (akka/akka-core#32030) — The "should not interfere with receive timeouts" test can be hardened to verify that a short receive timeout doesn't trigger recovery timeout, and that recovery timeout still fires when journal is unresponsive. See apache/pekko#3436.

### Minor observation

The `InmemJournal` uses `scala.jdk.DurationConverters._` with `.toScala` — this is the correct approach for Scala 3 cross-compilation (vs `pekko.util.JavaDurationConverters._` which fails under `-Xfatal-warnings` in Scala 3).

He-Pin (approved):
lgtm

pjfanning:
As a native English speaker, I tend to find adding a comma before 'but' to be somewhat unusual so that Akka change was omitted.

PR #3433, opened 2026-08-15
fix: ignore FilteredPayload in replay of persisted events

He-Pin:
### Motivation
When `FilteredPayload` is stored in the journal (e.g. via an `EventAdapter` that maps certain events to `FilteredPayload`), replaying those events would pass `FilteredPayload` to the user's `eventAdapter.fromJournal` or `receiveRecover` handler, which is unexpected and can cause failures.

### Modification
- In `ReplayingEvents` (typed persistence), skip `FilteredPayload` by returning `EmptyEventSeq` before calling `eventAdapter.fromJournal`.
- In `Eventsourced` (classic persistence), add a case to ignore `FilteredPayload` before it reaches the user's `receiveRecover` handler.
- Add directional tests for both classic and typed persistence that fail before the fix.

### Result
`FilteredPayload` entries in the journal are silently skipped during recovery. Sequence numbers are still correctly advanced.

### License
This PR ports a bug fix from Akka (https://github.com/akka/akka-core/pull/32000). The Apache License 2.0 is applicable.

### Tests
- `sbt "persistence / Test / testOnly org.apache.pekko.persistence.PersistentActorSpec"`
- `sbt "persistence-typed-tests / Test / testOnly org.apache.pekko.persistence.typed.scaladsl.EventSourcedBehaviorSpec"`

### References
Refs akka/akka-core#32000

### DIFF
--- a/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedBehaviorSpec.scala
+++ b/persistence-typed-tests/src/test/scala/org/apache/pekko/persistence/typed/scaladsl/EventSourcedBehaviorSpec.scala
@@ -46,6 +46,9 @@ import pekko.persistence.query.Sequence
 import pekko.persistence.snapshot.SnapshotStore
 import pekko.persistence.testkit.PersistenceTestKitPlugin
 import pekko.persistence.testkit.query.scaladsl.PersistenceTestKitReadJournal
+import pekko.persistence.FilteredPayload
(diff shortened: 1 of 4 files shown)


pjfanning:
#3432 has this and with Akka derived PRs, there should always be a declaration that the Apache License is applicable.

He-Pin:
Superseded by #3432 which covers the same FilteredPayload fix with better Scala 3 compatibility.

PR #3435, opened 2026-08-15
test: add Java test for retry short-circuit with Predicate2

He-Pin:
### Motivation
The Java DSL `Patterns.retry` with `Predicate2` (shouldRetry) has been available since 1.1.0 but lacks a directional test exercising the short-circuit behavior from Java. The Scala DSL has coverage in `RetrySpec` but the Java path is untested.

### Modification
Add `testShortCircuitRetry` to `PatternsTest.java` that verifies:
- Retry continues when predicate returns true (`IllegalStateException`)
- Retry stops immediately when predicate returns false (`IllegalArgumentException`)
- Exactly 3 attempts are made (2 retriable + 1 non-retriable)

### Result
Java DSL retry short-circuit behavior is now covered by a directional test that would fail if the `Predicate2` wiring were broken.

### Tests
- `sbt "actor-tests / Test / testOnly org.apache.pekko.pattern.PatternsTest"`

### References
Refs akka/akka-core#32035

### DIFF
--- a/actor-tests/src/test/java/org/apache/pekko/pattern/PatternsTest.java
+++ b/actor-tests/src/test/java/org/apache/pekko/pattern/PatternsTest.java
@@ -339,6 +339,38 @@ public class PatternsTest {
   }
 
   @Test
+  public void testShortCircuitRetry() throws Exception {
+    AtomicInteger failureCounter = new AtomicInteger();
+
+    final CompletableFuture<Object> ise = new CompletableFuture<>();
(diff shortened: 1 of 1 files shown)


pjfanning:
with Akka derived PRs, there should always be a declaration that the Apache License is applicable.

He-Pin:
The license is OK.

pjfanning (approved):
lgtm

PR #3436, opened 2026-08-15
test: harden PersistentActorRecoveryTimeoutSpec receive timeout test

He-Pin:
### Motivation
The "should not interfere with receive timeouts" test previously only verified that the receive timeout value is preserved after recovery. It did not verify that a short receive timeout doesn't accidentally trigger the recovery timeout, nor that the recovery timeout still fires when the journal is unresponsive.

### Modification
- Use separate probes for persist and replay phases to avoid cross-contamination.
- Use a short receive timeout (20ms) during replay to verify it does not trigger recovery timeout (`expectNoMessage`).
- Verify that recovery timeout (500ms) still fires when the journal does not respond (`expectMsgType[Failure]` with `RecoveryTimedOut`).
- Reduce `recovery-event-timeout` from 30s to 500ms for the receive-timeout journal to keep the test fast.
- Remove unused `TestDuration` import.

### Result
The test now encodes the intended contract: receive timeout and recovery timeout are independent mechanisms. The test is deterministic and fast (~700ms total for the replay phase).

### Tests
- `sbt "persistence / Test / testOnly org.apache.pekko.persistence.PersistentActorRecoveryTimeoutSpec"`

### References
Refs akka/akka-core#32030

### DIFF
--- a/persistence/src/test/scala/org/apache/pekko/persistence/PersistentActorRecoveryTimeoutSpec.scala
+++ b/persistence/src/test/scala/org/apache/pekko/persistence/PersistentActorRecoveryTimeoutSpec.scala
@@ -19,7 +19,7 @@ import org.apache.pekko
 import pekko.actor.{ Actor, ActorLogging, ActorRef, Props }
 import pekko.actor.Status.Failure
 import pekko.persistence.journal.SteppingInmemJournal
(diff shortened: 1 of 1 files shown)


pjfanning:
with Akka derived PRs, there should always be a declaration that the Apache License is applicable.

He-Pin:
The license is OK.

pjfanning (approved):
lgtm

PR #3461, opened 2026-08-24
fix: preserve StageActor demand across fused streams

### LINKED ISSUE #3459: Stream: demand across async boundary replenished per batch, not per element, since 2.0.0-M4
## Summary

Between 2.0.0-M3 and 2.0.0-M4, demand replenishment across an internal async boundary changed from **per element** to **roughly per input-buffer batch**. A stage that produces elements from outside the stream (a stage actor / async callback) and pushes only when `isAvailable(out)` is true now sees demand about once every 16 elements instead of after every push.

This silently breaks `pekko-connectors`' UDP connector, which drops received datagrams when there is no instantaneous demand. On M4 it delivers roughly 1 datagram in 16.

## Reproduction

`pekko-connectors` `udp` module, unmodified, against the two milestones (only `PekkoCoreDependency.currentVersion` changed):

```
sbt "udp/test"
```

| pekko | result |
|---|---|
| 2.0.0-M3 | 5/5 pass (`UdpSpec` 3, `UdpTest` 2) |
| 2.0.0-M4 | `UdpSpec` "must send and receive messages" + "... with options" fail, `UdpTest` both fail, with `Expected OnNext(_), yet no element signaled during 3 seconds` |

The test binds `Udp.bindFlow` to `localhost:0`, requests 100 elements from a `TestSink` probe, then sends 100 datagrams to the bound address via `Udp.sendSink()`.

## Evidence

The send side is fine — with `pekko.io.udp.trace-logging=on`, all 100 datagrams are written (`Wrote [9] bytes to channel` x100), no `CommandFailed`, no "Dropping write because queue is full". All 100 also arrive at the listener and are delivered to the stage as `Udp.Received`.

The receive stage (`UdpBindLogic`) does:

```scala
case (_, Udp.Received(data, sender)) =>
  if (isAvailable(out)) push(out, Datagram(data, sender))   // else silently dropped

setHandler(out, new OutHandler { override def onPull(): Unit = () })
```

Instrumenting that stage with `isAvailable(out)` and `System.nanoTime()`:

**2.0.0-M3** — `availableOut=true` for every one of the 100 datagrams (a pull returns after each push).

**2.0.0-M4** — a pull returns only once per ~16 datagrams:

```
onPull at ...655633171
received Message 1  availableOut=true  at ...678674953
received Message 2  availableOut=false at ...679066879
...
received Message 16 availableOut=false at ...680086695
onPull at ...680861861
received Message 17 availableOut=true  at ...681002103
received Message 18 availableOut=false at ...681068777
```

The ~16 grouping matches the default `Attributes.InputBuffer` max.

## Suspected cause

`2610cda2af perf: batch async boundary elements (#3288)`, which is in the M3..M4 range and changes internal boundary elements to be batched and flushed "when the interpreter parks" rather than sent per element, deferring the demand round trip to once per flush cycle.

I have matched the symptom to that commit by tag range and timing, not by bisecting.

## Impact

Demand *bounds* are still respected, so this is arguably within the letter of the contract, but it is a large behavioural change for any stage that treats "no demand right now" as a signal to drop — external-source stages built on `getStageActor` or `getAsyncCallback` are the common shape. The loss is silent.

`pekko-connectors` can defend itself by buffering in the stage, but it seems worth deciding upstream whether the reduced demand-replenishment frequency is intended.

He-Pin:
### Motivation

`StageActor` lazy dispatch coalesces a burst into one async callback. A callback can enqueue interpreter work that replenishes downstream demand; dispatching the next message before that work runs makes it observe stale demand and can drop data, as reported in #3459.

Fixed nested calls such as `interpreter.execute(2)` or `execute(16)` are not a general fix: the required work depends on the fused topology depth, and the nested work is not charged to the actor shell's processing quota. Restoring a saved `activeStage` is also unsafe when interpreter execution may have completed and released that stage.

### Modification

- Add an internal event-limit-aware async handler contract so lazy `StageActor` dispatch and the graph interpreter share the enclosing shell's event budget.
- Drain callback-generated interpreter work before dispatching the next `StageActor` message. If the budget is exhausted, park the dispatcher and resume it only after the interpreter queue becomes idle; this has no topology-dependent event limit.
- Avoid entering `GraphInterpreter.execute` when a callback enqueues no interpreter work, preserving the common no-graph-work hot path.
- Set `activeStage` before every callback without restoring a potentially finalized stage.
- Stop and clear queued dispatch after stage completion or callback failure.
- Add directional tests for per-element demand replenishment, a 256-stage fused topology, `sync-processing-limit = 1`, stage completion, and callback failure.

### Result

Demand is replenished before the next `StageActor` message independently of fused topology depth. Actor/interpreter fairness remains bounded by the existing shell quota, lazy dispatch still coalesces mailbox traffic, and completed or failed stages are not dispatched again.

The original `pekko-connectors` UDP reproduction was run from connector commit `1c15f3b6f`; only `PekkoCoreDependency.currentVersion` changed:

| Pekko Core | `sbt "udp/test"` |
| --- | --- |
| `2.0.0-M4` | 1/5 passed; all four Scala/Java 100-datagram send/receive tests timed out |
| `2.0.0-M4+14-3f07c449-SNAPSHOT` | 5/5 passed |

This directly verifies the connector regression, not only the reduced Core reproducer.

The direct `StageActorRefBenchmark.lazy_stage_actor_ref_tell_10k` JMH benchmark was run on the same machine and OpenJDK 25.0.2 with `-wi 3 -i 5 -w 2s -r 2s -f 3`:

| Revision | Throughput (ops/s, JMH 99.9% error) |
| --- | ---: |
| `origin/main` (`6be39c658f`) | 26,285,866 ± 6,002,205 |
| This PR | 27,987,687 ± 3,124,268 |

The PR point estimate is 6.5% higher, but the confidence intervals overlap. The supported conclusion is that this run found no measurable throughput regression, not that the change is statistically faster.

### Tests

- `pekko-connectors`: `sbt "udp/test"` — 2.0.0-M4 failed 4/5; PR snapshot passed 5/5.
- `sbt "stream-tests / Test / testOnly org.apache.pekko.stream.FusingSpec org.apache.pekko.stream.scaladsl.StageActorRefSpec org.apache.pekko.stream.scaladsl.ActorRefSourceSpec org.apache.pekko.stream.scaladsl.StreamRefsSpec"` — 80/80 passed.
- `sbt -Dpekko.stream.fuzzing-mode=on "stream-tests / Test / testOnly org.apache.pekko.stream.FusingSpec"` — 24/24 passed with randomized event ordering and chasing disabled.
- `sbt "stream-tests / Test / testOnly org.apache.pekko.stream.impl.fusing.LifecycleInterpreterSpec"` — 13/13 passed.
- `sbt validatePullRequest` — Streams suite passed 3,418/3,418. The aggregate command failed outside this change: LevelDB JNI is unavailable on this ARM64 host; two Remote tests failed on local bind/TLS timing; `IntegrationDocTest` failed while instantiating `UnboundedMailbox`.
- `sbt "stream / mimaReportBinaryIssues" "++3.3.8" "stream / mimaReportBinaryIssues"` — passed.
- `sbt headerCreateAll "+headerCheckAll" checkCodeStyle`, native scalafmt, and `git diff --check` — passed.
- `sbt "bench-jmh / Jmh / run -wi 3 -i 5 -w 2s -r 2s -f 3 .*StageActorRefBenchmark.lazy_stage_actor_ref_tell_10k.*"` — results shown above.

### References

Fixes #3459

### DIFF
--- a/stream-tests/src/test/scala/org/apache/pekko/stream/FusingSpec.scala
+++ b/stream-tests/src/test/scala/org/apache/pekko/stream/FusingSpec.scala
@@ -13,7 +13,7 @@
 
 package org.apache.pekko.stream
 
-import scala.concurrent.{ duration, Await, Promise }
+import scala.concurrent.{ duration, Await, Future, Promise }
 
 import duration._
 
@@ -23,7 +23,7 @@ import pekko.stream.QueueOfferResult
 import pekko.stream.impl.UnfoldResourceSource
 import pekko.stream.impl.fusing.GraphInterpreter
 import pekko.stream.scaladsl._
-import pekko.stream.stage.GraphStage
+import pekko.stream.stage.{ GraphStage, GraphStageLogic, GraphStageWithMaterializedValue, OutHandler }
 import pekko.stream.testkit.TestPublisher
 import pekko.stream.testkit.StreamSpec
 import pekko.stream.testkit.Utils.TE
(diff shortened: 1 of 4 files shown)


pjfanning (approved):
LGTM as it fixes the issue (or seems likely to).

Claude AI raised some concerns (below).

Two things I'd want addressed

1. activeStage is clobbered by the nested execute().

runAsyncInput sets activeStage = logic (GraphInterpreter.scala:577) and calls afterStageHasRun(logic) at the end (line 592). The new execute(2) runs processEvent, which does activeStage = null (line 602) and then reassigns it to whichever connection's owner it processes (lines 616/636/655/666). So when handler(item) returns for message 2, activeStage is no longer this logic.

This is benign for finalization — afterStageHasRun takes the logic explicitly, so it finalizes the right stage. But reportStageError (line 442) uses the mutable activeStage field, not a parameter. If handler(item) throws for message 3, the error is attributed to, and failStage is called on, whatever stage the previous nested execute last touched. That's a misattributed failure — wrong stage failed, wrong stage named in the log.

I have not constructed a failing case for this, so I'd flag it as a question rather than a defect: is it worth saving and restoring activeStage around the execute(2) call? It's a two-line change if so.

2. The 2 deserves a firmer justification than "sufficient".

The PR says "1 event slot for the push + chase budget for the pull". Reading execute (line 486), eventsRemaining is decremented per dequeued event and chaseCounter = math.min(ChaseLimit, eventsRemaining) — so with a limit of 2, the chase budget after the first dequeue is 1. That's enough for the simple push→pull chain the test exercises.

What I can't tell from the diff is what happens with a longer downstream chain inside the same shell, where the pull needs more than one chase hop to reach back. The events aren't lost — they stay queued and get drained by the runBatch that follows — but demand may still not be replenished by the time message N+1 is handled, which is exactly the bug. That would make the fix correct for the tested topology and partial for others.

A test with two or three map stages between the source and the async boundary would settle it. Worth asking the author whether that was considered.

He-Pin:
Let's me check this again.

He-Pin:
Thanks @pjfanning for the careful review — both concerns are addressed in 56783632b8.

**1. activeStage save/restore — done, with a test.**

You were right to flag it. One interesting empirical detail: in the minimal topology the clobbering self-heals, because the last chased event of execute() is the pull back to the source port, and processPull sets activeStage to that port's owner — i.e. this logic again. But with a longer fused upstream chain and an event limit that is too small, execute() stops mid-chain and activeStage is left pointing at an intermediate stage, so any activeStage-dependent helper used by the next handler invocation (e.g. the default InHandler/OutHandler termination methods) would act on the wrong stage. Fixed by saving/restoring activeStage around execute(), and both the longer-upstream-chain test and a dedicated attribution test assert `currentInterpreter.activeStage eq logic` for every dispatched message.

**2. The limit is now justified — and you were right that 2 was not enough.**

I added exactly the test you suggested (three map stages), in both directions:

- *Longer upstream chain* (maps between source and boundary): **fails with limit 2, passes with the new limit** — this is the case execute() can and must cover, since the whole push/pull round trip stays inside one shell.
- *Longer downstream chain* (maps after the boundary): **fails with any limit** (verified up to 1024). The reason is structural, not a matter of budget: demand crosses the async boundary via BatchingActorInputBoundary, which requests upstream in asynchronous batches of size/2. While the drain loop occupies the source shell's actor thread, no RequestMore can be processed, so bursts larger than the outstanding demand batch can drop — which is the documented at-most-once semantics of the getStageActor + isAvailable pattern. The downstream-chain test therefore bursts within the initial demand batch (8 < 16) and documents this boundary.

The magic 2 is now `LazyDispatch.ReplenishEventLimit = 16` with a comment explaining the cost model (a few event hops per fused stage per direction; leftovers stay queued for the interpreter's outer run loop).

I also fixed a latent startup race in the original test: messages were sent right after the stage actor materialized, but before initial demand had necessarily propagated to the source port. The tests now wait for the first onPull before bursting.

Tests: FusingSpec (4 new/updated tests, 34/34 with the whole suite), StageActorRefSpec, ActorRefSourceSpec, StreamRefsSpec all green; stream MiMa clean.

pjfanning:
@He-Pin this is still ok to merge for me - the new commit does seem to address the highlighted issues

PR #3478, opened 2026-08-27
Add a draft threat model and wire it for discoverability

potiuk:
### Motivation

The Pekko PMC asked the ASF Security team for a draft threat model to refine, as
preparation for a Claude security scan. Pekko already documents its security posture
well — `remote-security.md`, `serialization.md` and `remoting.md` between them state
most of the trust model — but that material is spread across three files and states it
implicitly, rather than in a form a triager (or a scanning agent) can cite when
classifying a finding.

Separately, the `AGENTS.md` → `SECURITY.md` chain that a scanning agent follows to
locate a project's model does not resolve today: there is no `SECURITY.md` in the repo.

### Modification

- **`THREAT_MODEL.md`** — a v0 draft. Every claim is tagged `(documented)` with a
  citation into Pekko's own docs, or `(inferred)` with a matching open question in §14.
  **40 documented / 0 maintainer / 14 inferred.**
- **`SECURITY.md`** — reporting policy (`security@apache.org`; Pekko does not operate a
  separate project security list), a link to the threat model, and the two assumptions
  that catch most reporters.
- **`AGENTS.md`** — a `## Security` section pointing at both.

Nothing the existing docs assert has been dropped or weakened. **§15 is a back-map**
from each existing statement to the section that now carries it, so you can diff the two.

### Result

The chain `AGENTS.md` → `SECURITY.md` → `THREAT_MODEL.md` resolves on `main`, and
triagers get a citable document with a closed set of dispositions (§13).

**§14 holds 13 questions, each stated as a proposed answer** — confirming or correcting
is enough, no prose needed. The two that most change the document:

- **Q1** — `pekko.remote.artery.transport` ships `tcp`, so a stock cluster has no peer
  authentication. Is that the supported production posture *given the documented
  network-isolation assumption*, so a report of "unauthenticated peer can associate" is
  out of model when it assumes internet exposure — but valid if it shows harm reachable
  **pre-association** from an adjacent-network host?
- **Q4** — are `untrusted-mode` and the remote deployment allow list **security
  boundaries**, or **hardening**? The draft says hardening, on the strength of the
  documented *"untrusted mode does not give full protection against attacks by itself"*.

One conclusion is worth flagging explicitly because it disposes of a whole class of
scan findings: the draft reads §7 as having **no Byzantine-peer model**, on the strength
of *"you'll have to trust all cluster nodes the same in a Pekko cluster anyway"* and
*"as soon as an actor system can connect to another remotely, it may in principle send
any possible message to any actor contained within that remote system."* If that
generalises from remoting to sharding, singleton and `distributed-data` (Q9), then any
finding whose precondition is "a cluster member misbehaves" is out of model.

**This is a draft and should not be treated as canonical until at least Q1 and Q4 are
answered.** Happy to fold the answers in and push, or hand it over entirely — the PMC
owns the document.

### Tests

Not run - docs only

### References

None - requested by the Pekko PMC chair on the ASF security scan thread


pjfanning:
I have https://github.com/pjfanning/incubator-pekko/pull/59 - I'm not sure what the merge order is but feel free to comment on either PR>

potiuk:
Thanks @pjfanning — this is exactly what the draft was for, and considerably more
than I hoped for. Thirteen rulings, and the provenance moves from
40 documented / 0 maintainer / 14 inferred to 43 / 48 / 3.

Two of your commits correct me rather than extend me, and both are fair:
`191e00d` drops a scope ruling about app-bound sockets that I inferred and
shouldn't have, and `02066b0` restores the upstream-coordination wording in
`SECURITY.md` that I had paraphrased — I should have preserved that verbatim
rather than reworded it.

**On merge order** — your branch fast-forwarded cleanly from this one, so I've
pushed it here. All 19 of your commits are now on this PR with your authorship
intact; no rebase, no force push, nothing resolved by hand. That leaves a single
PR against `apache/pekko` carrying the full history, so
pjfanning/incubator-pekko#59 can be closed whenever suits you.

If you'd rather it went the other way round — yours opened against
`apache/pekko` and this one closed — that's equally fine and I'll close this.
It's your document now; I only wanted to avoid ending up with two.

I also pushed one small follow-up commit (`9ed2a94`): the header still read
`Status: DRAFT — awaiting Pekko PMC review. Not yet maintainer-ratified`, which
stopped being true once you answered the questions. A triager citing §9 to close
a report shouldn't be citing a document that calls itself unratified. It now
reads "Reviewed by the Apache Pekko PMC, 2026-08-27" and names the one item still
open. Two lines; no ruling, citation or disposition touched.

That open item is the last loose end: §2's module in/out split is still the ASF
Security team's proposal rather than a maintainer call, as Q6 notes. Not
blocking — worth a line from you at some point so the document has no inferred
claims left at all.

pjfanning:
Thanks @potiuk - I'll leave it open for review by others.

THREAT_MODEL.md
```diff
@@ -0,0 +1,420 @@
...
+
+- **"Remote deployment permits arbitrary class instantiation."** Remote deployment is *not* remote code loading — the class must already be on the target's classpath *(P3, documented)*. Absent a way to introduce a class, this is not RCE.
+- **"Remoting transport defaults to plaintext."** Correct, and by design under the §4 network-isolation assumption. In-model only if the finding shows harm reachable **pre-association**; a request to change the default is `BY-DESIGN: default-configuration` per §5b.
+- **"`PoisonPill` can be sent remotely to shut down a system."** Documented behaviour, gated by `untrusted-mode` *(documented — `remote-security.md`)*.
+- **"`hostname-verification` is disabled by default."** Known and documented, with the trade-off spelled out for dynamic-hostname deployments, and warned at runtime under `LogMarker.Security`. A request to change the default is `BY-DESIGN: default-configuration` per §5b.
+- **Findings in `*-tests`, `*-testkit`, `bench-jmh`, `docs`** — `OUT-OF-MODEL: unsupported-component` per §3.
+- **"SHA1PRNG with `/dev/urandom` reuses the seed."** Documented trade-off, deliberately recommended to avoid blocking *(documented)*.
```
samueleresca, line 284: I think this is a very out-of-date recommendation (2014). Should we consider removing that from the doc entirely and from the security model?

pjfanning: This new doc was generated using AI in the main and it just found stuff in our own Pekko docs.
https://github.com/search?q=repo%3Aapache%2Fpekko+SHA1PRNG&type=code

No problem reviewing this but this threat model entry basically means we won't accept someone telling us to change the docs or default configs as a CVE issue. This line is in the 'Known non-findings' section.

THREAT_MODEL.md
```diff
@@ -0,0 +1,420 @@
...
+- **Entropy.** On Linux with SHA1PRNG, the docs recommend `-Djava.security.egd=file:/dev/urandom` to avoid blocking, noting it *"is NOT as secure because it reuses the seed."* *(documented — `remote-security.md`)*
+
+### What Pekko does not do to its host
+
+These are negative claims, rarely written down anywhere. Each is confirmed and carries the citation, or the exception, behind it *(maintainer — §14 Q7)*:
+
+- Installs no signal handlers and spawns no child processes — no `sun.misc.Signal`/`SignalHandler`, `Runtime.exec` or `ProcessBuilder` in the main sources. It does register **one JVM shutdown hook**, via `CoordinatedShutdown` (`actor/.../actor/CoordinatedShutdown.scala:381`).
```
samueleresca: > One JVM shutdown hook is registered by `CoordinatedShutdown`, as disclosed in §5; a shutdown hook is not a signal handler.


AI is also pointing me to the one below. Does it need to be mentioned?
 https://github.com/apache/pekko/blob/6bbd5d041de06ae43bea2a8e9dbc74bb5cad6d1d/remote/src/main/scala/org/apache/pekko/remote/artery/ArteryTransport.scala#L391-L392

samueleresca:
Should we advertise this PR in the committee mailing list for review?

pjfanning:
@samueleresca Can you send an email to the Dev mailing list? There are more people watching the GitHub repo but there might be some people only watching the list.

THREAT_MODEL.md
```diff
@@ -0,0 +1,420 @@
...
+| --- | --- |
+| **Project** | Apache Pekko (core toolkit) |
+| **Written against** | commit `90b02d60`, `main` |
+| **Date** | 2026-08-27 |
+| **Authors** | ASF Security team, at the request of the Pekko PMC |
+| **Version binding** | This model is versioned alongside the project. A report against Pekko version *N* is triaged against the model as it stood at *N*, not at `main`. |
+| **Reporting** | Findings that violate a §8 property should be reported per [`SECURITY.md`](SECURITY.md). Findings that fall under §3 or §9 will be closed citing this document. |
+
+**Provenance legend.** Every non-trivial claim is tagged:
+*(documented)* — stated in Pekko's own docs, cited.
```
samueleresca, line 15: We need to double-check that the claims based on documentation are correct (e.g., docs are up to date). I will take a deeper look over the weekend.

THREAT_MODEL.md
```diff
@@ -0,0 +1,420 @@
...
+*(maintainer)* — stated by a Pekko maintainer in review of this document.
+*(inferred)* — reasoned from code or config defaults, **not yet confirmed**; each has a matching question in §14.
+
+**Provenance mix:** 43 documented / 48 maintainer / 3 inferred. The documented share is unusually high because Pekko's own remoting and serialization docs already state much of the trust model explicitly. The §5a default rulings — previously the largest inferred block — are now answered by the §5b posture statement and §14 Q1 to Q5. What remains inferred is the residual module in/out split (§14 Q6).
```
samueleresca: > The documented share is unusually high 

Feels unnecessary

THREAT_MODEL.md
potiuk: Good catch — you're right, and §5 was wrong as written.

There are two `addShutdownHook` sites in the main sources, not one:

- `actor/.../CoordinatedShutdown.scala:381` — the one §5 named
- `remote/.../artery/ArteryTransport.scala:392` — the one you found

The second is gated on `system.settings.JvmShutdownHooks`, and
`actor/src/main/resources/reference.conf:98` ships `jvm-shutdown-hooks = on`,
so it registers under default configuration as soon as remoting starts. That
made "one JVM shutdown hook" wrong for any node running Artery, which is most
of them.

The other two hits (`NodeChurnSpec`, `StressSpec`) are multi-jvm test sources
and out of model per §3.

Fixed in `ac092c8`: §5 and the Q7 answer both say **two** now, name both sites,
and record that the Artery one is conditional on `pekko.jvm-shutdown-hooks`.
The conditionality is the part worth having — an integrator who turns that
setting off gets different behaviour from remoting than from
`CoordinatedShutdown`.

Worth noting this is exactly the section the original draft flagged as its
highest-priority confirmation target, on the grounds that negative claims
("the project never does X") are rarely written down and almost impossible to
cite. It took a maintainer reading the code to find the gap, which is the
process working.

THREAT_MODEL.md
potiuk: Agreed — cut in `ac092c8`.

It was a note to the reviewer about how much of the v0 draft rested on
inference, which stopped being relevant the moment you and @pjfanning answered
the questions. The counts stay; the editorialising goes.

potiuk: Yeah - it will be removed - it's more that the document is both the "draft" + "status" (what's left/where we are now). Final threat model will be.

THREAT_MODEL.md
```diff
@@ -0,0 +1,424 @@
+# Apache Pekko — Threat Model
+
+**Status:** Reviewed by the Apache Pekko PMC, 2026-08-27. The thirteen questions this document was drafted around are answered in §14 and carry *(maintainer)* provenance; the residual module in/out split noted in §2 and Q6 is the one open item.
+
+| | |
+| --- | --- |
+| **Project** | Apache Pekko (core toolkit) |
+| **Written against** | commit `90b02d60`, `main` |
```
samueleresca, line 8: This commit needs to be updated if we want to include the recent PRs: 

https://github.com/apache/pekko/pull/3490
https://github.com/apache/pekko/pull/3491
https://github.com/apache/pekko/pull/3492
https://github.com/apache/pekko/pull/3493

pjfanning: I asked Claude AI and it didn't think these have a material impact on the file.

THREAT_MODEL.md
```diff
@@ -0,0 +1,424 @@
...
+| OSGi | `osgi` | bundle activator | classloading | **yes** *(maintainer — §14 Q6)* |
+| Test kits | `*-testkit`, `multi-node-testkit`, `*-tests`, `persistence-tck`, `stream-tests-tck` | — | — | **no** — §3 |
+| Benchmarks | `bench-jmh` | — | — | **no** — §3 |
+| Build / docs | `docs`, `project`, `scripts`, `legal`, `kubernetes` (test-cluster provisioning scripts), `plugins`, `bill-of-materials`, `scala-nightly` | — | — | **no** — §3 |
+
+*(inferred — the in/out split is the ASF Security team's proposal, except `osgi`, which is confirmed in model; see §14 Q6)*
+
+---
+
+## §3 Out of scope (explicit non-goals)
```
samueleresca, line 61: Should we mention a supply chain attack (e.g., a malicious package) as out of scope?

pjfanning: If we have a supply chain attack and something malicious gets injected into our source release or published jars, I would regard that as a very serious CVE worthy episode.

samueleresca: Yeah agree with it being serious concern. Is that an attack vector that should be handled at org level (i.e. ASF)?

pjfanning: https://github.com/apache/infrastructure-actions is about the ASF being able to approve CI actions and block the use of certain actions or specific versions.

Otherwise, the onus is with the Pekko PMC is to ensure that our build is secure. We need to promptly fix any issues that are reported if we assess that there is a risk to our users.

pjfanning:
@samueleresca is it ok to merge this? It doesn't have to be perfect and will remain a living document that we can adjust. When we merge this, we can look to move onto the scan stage.

samueleresca:
> @samueleresca is it ok to merge this? It doesn't have to be perfect and will remain a living document that we can adjust. When we merge this, we can look to move onto the scan stage.

Yeah I'm ok with this. Let's move to the scan stage.

pjfanning (approved):
lgtm - can be iterated on from here

PR #3487, opened 2026-08-30
docs: remove stale Java 8/11 framing now that JDK 17 is the minimum

samueleresca:
Pekko 2.0.x requires JDK 17 as the minimum supported runtime (see release-notes/releases-2.0.md and project/JdkOptions.scala).

Modification:
- `remoting-artery.md`: reworded the Aeron `--add-opens` note and the Flight Recorder section, which claimed JFR is enabled by "detecting JDK 11" even though there is no such runtime detection in `FlightRecorderLoader.scala` - it is controlled purely by the `pekko.java-flight-recorder.enabled` config flag.
- `remote-security.md`: dropped a TLS cipher caveat about Java 8 runtimes, which are no longer supported.
- `stream/reactive-streams-interop.md`: removed Java-8-specific Reactive Streams artifact/JavaFlowSupport caveats.
- `persistence.md`, `typed/cluster-sharding.md`, `typed/distributed-data.md`: reworded the LevelDB/LMDB `--add-opens` notes so they no longer frame JDK 17 as an optional/special case.
- `actor/src/main/resources/reference.conf`: fixed the same stale "JDK 11 or later" framing in the java-flight-recorder comment, which is snip-included into docs/src/main/paradox/general/configuration-reference.md.

Result:
Doc pages and the flight-recorder config comment reflect the current JDK 17+ baseline instead of pre-2.0 Java 8/11 assumptions.

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -102,8 +102,8 @@ pekko {
   # by setting this property to `off`.
   fail-mixed-versions = on
 
-  # Some modules (remoting only right now) can emit custom events to the Java Flight Recorder if running
-  # on JDK 11 or later. If you for some reason do not want that, it can be disabled and switched to no-ops
+  # Some modules (remoting only right now) can emit custom events to the Java Flight Recorder.
+  # If you for some reason do not want that, it can be disabled and switched to no-ops
   # with this toggle.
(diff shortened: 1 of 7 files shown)


pjfanning (approved):
lgtm

PR #3489, opened 2026-08-30
docs: refresh stale TLS security recommendations in remoting docs

samueleresca:
Motivation:
- `remote-security.md` and `remoting.md` both cited RFC 7525 (2015) as the current source for recommended TLS cipher suites, but RFC 7525 was obsoleted by RFC 9325 (Nov 2022).
- Both docs also linked to the Java 8-specific JSSE Reference Guide as "the official" documentation, even though Pekko 2.0.x requires JDK 17+ as a minimum.
- `remote-security.md` additionally carried a dead caveat about Java 8 runtime support for a TLS 1.3 cipher suite. The classic (legacy) `remoting.md` TLS section had also drifted out of sync with `remote-security.md`, which recently gained TLS 1.3 cipher guidance, post-handshake session verification, a random-number-generator recommendation, and mTLS-rotation-in- Kubernetes guidance that remoting.md never received.

Modification:
- `remote-security.md`, remoting.md: cite RFC 9325 instead of the obsoleted RFC 7525, noting that it obsoletes RFC 7525.
- `remote-security.md`, remoting.md: point the JSSE documentation link at the current (Java SE 25) JSSE Reference Guide instead of the Java 8 version.
- `remote-security.md`: drop the "(may not be supported on Java 8 runtimes)" caveat on `TLS_CHACHA20_POLY1305_SHA256`, which no longer applies now that Java 8 isn't a supported runtime.
- `remoting.md`: add a note pointing to remote-security.md for the additional current TLS guidance that isn't duplicated into the deprecated classic-remoting doc, instead of letting the two pages keep drifting out of sync.

Result:
The remoting security docs cite current IETF guidance and JDK documentation instead of a decade-old RFC and Java 8-era links, and classic remoting users are pointed at the fuller, current guidance instead of being left with silently weaker advice.

Tests:
- Not run - docs only.

References:
None - identified while auditing docs/src/main/paradox for content out of date relative to the current Pekko version and JDK support.

### DIFF
--- a/docs/src/main/paradox/remote-security.md
+++ b/docs/src/main/paradox/remote-security.md
@@ -53,8 +53,8 @@ pekko.remote.artery {
 Always use [substitution from environment variables](https://github.com/lightbend/config#optional-system-or-env-variable-overrides)
 for passwords. Don't define real passwords in config files.
 
-According to [RFC 7525](https://www.rfc-editor.org/rfc/rfc7525.html), the recommended algorithms to use with TLS 1.2
-are:
+According to [RFC 9325](https://www.rfc-editor.org/rfc/rfc9325.html) (which obsoletes RFC 7525), the recommended
+algorithms to use with TLS 1.2 are:
 
  * TLS_DHE_RSA_WITH_AES_128_GCM_SHA256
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #3491, opened 2026-08-30
fix: bound Jackson payload decompression size

pjfanning:
### Motivation
`JacksonSerializer.decompress` inflated gzip payloads with an unbounded `transferTo`, and passed the lz4 decompressed length declared on the wire straight to the decompressor as the allocation size. A small, well-formed compressed message could therefore declare (lz4) or expand to (gzip) an arbitrarily large size and drive an `OutOfMemoryError` on deserialization — reachable wherever the Jackson serializer deserializes attacker-influenced payloads.

### Modification
- Add a `compression.max-decompressed-size` setting (default `256 MiB`) to both the `serialization-jackson` and `serialization-jackson3` reference configs.
- On deserialization the gzip path now copies through a bounded loop (`gunzip`) that fails once the output exceeds the cap; the lz4 path rejects a declared length that is negative or over the cap **before** allocating.
- The bound applies regardless of the `algorithm` setting, because decompression is selected by the payload's magic bytes, not by local config.

All changed symbols are `@InternalApi`/private, so no MiMa impact.

### Result
A payload that would decompress beyond the cap is rejected with an `IllegalArgumentException` instead of exhausting the heap. Payloads within the cap round-trip as before.

### Tests
- `sbt "serialization-jackson/testOnly *JacksonJsonSerializerSpec" "serialization-jackson3/testOnly *JacksonJsonSerializerSpec"` — 71 passed each, including new gzip and lz4 cap-rejection tests
- `sbt "serialization-jackson/mimaReportBinaryIssues"` — no issues
- `sbt scalafmt` for the changed main and test sources

### References
None - found while reviewing the draft threat model in #3478

### DIFF
--- a/serialization-jackson/src/main/resources/reference.conf
+++ b/serialization-jackson/src/main/resources/reference.conf
@@ -205,6 +205,12 @@ pekko.serialization.jackson {
     # If compression is enabled with the `algorithm` setting the payload is compressed
     # when it's larger than this value.
     compress-larger-than = 0 KiB
+
+    # Maximum size of a payload after decompression. A compressed (gzip or lz4)
+    # payload that decompresses to more than this is rejected rather than
+    # allocated, guarding against a small message that inflates without bound.
+    # This applies on deserialization regardless of the `algorithm` setting above.
+    max-decompressed-size = 256 MiB
   }
 
   # Whether the type should be written to the manifest.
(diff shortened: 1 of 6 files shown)


samueleresca (approved):
This might break if anyone is using pekko with large uncompressed messages. That said, the limits are configurable and the change is resonable to put bound decompression and avoid zip bombs.

pjfanning:
We have a migration page in the docs and I'll add a note about this there. I've labelled this as a release note item as a reminder.

PR #3501, opened 2026-09-01
fix: bound nesting depth when deserializing enclosed payloads (#3500)

pjfanning:
cherry pick 3def3978a79e3a07db2703d40a0c0ac1789752ee #3500

### DIFF
--- a/actor/src/main/resources/reference.conf
+++ b/actor/src/main/resources/reference.conf
@@ -775,14 +775,14 @@ pekko {
     #
     # Please note that this option does not stop you from manually invoking java serialization
     #
+    allow-java-serialization = off
+
     # Maximum nesting depth when a serialized payload encloses another serialized
     # payload (for example Some, Optional, Status.Failure, StatusReply, a Throwable
     # cause, or an ActorSelectionMessage). Each level is parsed separately, so this is
     # the only bound on the chain. A message nested deeper than this is rejected with a
     # NotSerializableException rather than failing with a StackOverflowError.
     serialization-max-nesting-depth = 32
-
-    allow-java-serialization = off
 
     # Log warnings when the Java serialization is used to serialize messages.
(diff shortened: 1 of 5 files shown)


actor/src/main/resources/reference.conf
```diff
@@ -775,6 +775,13 @@ pekko {
     #
     # Please note that this option does not stop you from manually invoking java serialization
     #
+    # Maximum nesting depth when a serialized payload encloses another serialized
```
nvollmar, line 780: cuts off the comment of the following `allow-java-serialization` setting

actor/src/main/scala/org/apache/pekko/serialization/NestedDeserialization.scala
```diff
@@ -0,0 +1,63 @@
...
+ * rather than the size of the message, and a sufficiently deep chain fails with a
+ * `StackOverflowError` instead of a serialization error.
+ *
+ * The depth is per thread because one message is deserialized synchronously on one thread.
+ */
+@InternalApi
+private[pekko] object NestedDeserialization {
+
+  private val depth = new ThreadLocal[Array[Int]] {
+    override def initialValue(): Array[Int] = new Array[Int](1)
```
nvollmar, line 44: Maybe add a comment that Array[Int](1) is used to avoid boxing/unboxing int

nvollmar (approved):

actor/src/main/scala/org/apache/pekko/serialization/NestedDeserialization.scala
pjfanning: done

actor/src/main/resources/reference.conf
pjfanning: fixed

nvollmar (approved):

PR #3507, opened 2026-09-01
fix: reject messages whose parallel repeated fields disagree in length

pjfanning:
### Motivation
Three places read repeated protobuf fields that `toBinary` writes in lockstep, but read them as
if the lengths were guaranteed to agree.

**`DaemonMsgCreateSerializer:191-201`** drives its loop over the constructor arguments by
`getSerializerIdsCount` and then indexes `getArgs(idx)`, `getManifests(idx)` and
`getHasManifest(idx)` with it. A message where those disagree raises
`IndexOutOfBoundsException` out of the generated accessor. The pre-2.4 branch below it zips
args with manifests, which silently drops the tail of the longer list instead.

**`ArteryMessageSerializer:185-190`** zips the keys and values of a compression table
advertisement. A mismatch silently builds a table the sender did not advertise — and this table
becomes the node's outbound compression table (`ArteryTransport.scala:523`), which is then
acknowledged back to the sender as accepted at the advertised version.

**`ArteryMessageSerializer:190` and `:205`** narrow the advertised table version, and the ack's
version, from `int` to `byte` with `byteValue` / `toByte`. `CompressionTable.version` is a
`Byte` on both sides, so anything outside that range is not a version a peer advertised, and
narrowing makes values 256 apart indistinguishable. I only spotted the ack site while writing
the fix; it has the same problem.

### Modification
Check the lengths agree before indexing, and check the version fits in a byte before narrowing,
reporting either as `NotSerializableException`. Both are conditions no `toBinary` produces:
`DaemonMsgCreateSerializer.toBinary` appends to all four lists per argument, and
`serializeCompressionAdvertisement` adds a key and a value per entry from a `Byte` version.

### Result
A malformed message is reported as a serialization failure rather than raising
`IndexOutOfBoundsException` or being silently accepted as something other than what it said.

### Tests
- `sbt "remote/testOnly org.apache.pekko.remote.serialization.*"` — 199 passed, 1 pending

Five new tests, four of which were checked to discriminate by reverting the two production files
and re-running. The pre-fix failures are worth recording because they confirm each diagnosis:

| Test | Without the fix |
| --- | --- |
| reject a compression table advertisement whose keys and values disagree in length | **no exception** — the truncated table was accepted |
| reject a compression table version that does not fit in a byte (advertisement and ack) | **no exception** — silently narrowed |
| reject props whose repeated fields disagree in length | `IndexOutOfBoundsException` |
| reject old format props whose args and manifests disagree in length | `ClassNotFoundException` — it zipped down to one pair, then tried to load the string manifest as a class name |

The fifth, `accept the whole byte range of compression table versions`, passes either way by
design: it is the no-regression guard for `Byte.MinValue`, `-1`, `0`, `1` and `Byte.MaxValue`,
since `-1` means "disabled" and the range is documented on `CompressionTable`.

- `sbt "remote/mimaReportBinaryIssues"` — no issues
- `sbt "remote/scalafmtCheckAll" headerCreateAll` — clean

### References
None.

### DIFF
--- a/remote/src/main/scala/org/apache/pekko/remote/serialization/ArteryMessageSerializer.scala
+++ b/remote/src/main/scala/org/apache/pekko/remote/serialization/ArteryMessageSerializer.scala
@@ -213,13 +213,34 @@ private[pekko] final class ArteryMessageSerializer(val system: ExtendedActorSyst
         s"Compression table advertisement carries [${protoAdv.getKeysCount}] entries, more than " +
         s"the configured maximum of [$maxEntries]")
 
+    // the two lists are parallel; `zip` on its own would drop the tail of the longer one and
+    // build a table the sender did not advertise, which is then acknowledged as accepted
+    if (protoAdv.getKeysCount != protoAdv.getValuesCount)
+      throw new NotSerializableException(
+        s"Compression table advertisement carries [${protoAdv.getKeysCount}] keys and " +
+        s"[${protoAdv.getValuesCount}] values, which must match")
+
     val kvs =
       protoAdv.getKeysList.asScala
         .map(keyDeserializer)
         .zip(protoAdv.getValuesList.asScala.asInstanceOf[Iterable[Int]] /* to avoid having to call toInt explicitly */ )
 
-    val table = CompressionTable[T](protoAdv.getOriginUid, protoAdv.getTableVersion.byteValue, kvs.toMap)
+    val table =
+      CompressionTable[T](protoAdv.getOriginUid, tableVersion(protoAdv.getTableVersion), kvs.toMap)
     create(deserializeUniqueAddress(protoAdv.getFrom), table)
+  }
+
+  /**
+   * A compression table version is a `Byte` on both sides, so a value that does not survive the
+   * narrowing is not one any peer advertised. Narrowing it silently would make versions 256
+   * apart indistinguishable, and the version is echoed back to the sender in an ack.
+   */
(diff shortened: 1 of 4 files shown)


remote/src/main/scala/org/apache/pekko/remote/serialization/ArteryMessageSerializer.scala
```diff
@@ -182,15 +182,36 @@ private[pekko] final class ArteryMessageSerializer(val system: ExtendedActorSyst
       create: (UniqueAddress, CompressionTable[T]) => U): U = {
     val protoAdv = ArteryControlFormats.CompressionTableAdvertisement.parseFrom(bytes)
 
+    // the two lists are parallel; `zip` on its own would drop the tail of the longer one and
+    // build a table the sender did not advertise, which is then acknowledged as accepted
+    if (protoAdv.getKeysCount != protoAdv.getValuesCount)
```
He-Pin, line 218: This branch conflicts with `main`: #3521 landed the `maxEntries` bound at exactly this point in `deserializeCompressionAdvertisement`, and `ArteryMessageSerializerSpec.scala` conflicts too. It needs a rebase before it can merge. The two checks coexist fine — put this one after the `maxEntries` check so the cheap entry-count bound runs first.

pjfanning: fixed

He-Pin (approved):

PR #3539, opened 2026-09-12
chore: remove leftover Scala 2.12 idioms

pjfanning:
### Motivation
The codebase still carried a number of Scala 2.12-era idioms that are redundant or deprecated on Scala 2.13/3: `immutable.Seq` qualifiers (`scala.Seq` has been `immutable.Seq` since 2.13), the deprecated `mutable.WrappedArray` alias, `filterKeys`/`mapValues` on `Map`, `Either` left/right projections, and `.toIterator`/`Stream` in user-facing doc snippets.

### Modification
- Replace `immutable.Seq` with `Seq` across ~210 files and drop the now-unused `scala.collection.immutable` imports. `scala.Seq` is a type alias for `scala.collection.immutable.Seq` on 2.13 and 3, so this is source- and binary-compatible (same erasure). `DslFactoriesConsistencySpec` is left as-is since it deliberately distinguishes `collection.Seq` from `immutable.Seq`.
- `ByteStringBuilder.addAll`: match `mutable.ArraySeq.ofByte` and `immutable.ArraySeq.ofByte` instead of the deprecated `WrappedArray.ofByte`. `Array.toSeq`/`toIndexedSeq` produce `immutable.ArraySeq.ofByte` on 2.13, which previously fell through to the generic path.
- Remove a dead commented-out `CanBuildFrom` block in `ByteString`.
- `ManifestInfo`: replace deprecated `filterKeys` (and its `@nowarn`) with `filter`.
- `ShardRegion` scaladoc example: `filterKeys` -> `filter`.
- Rendered doc snippets: `Source(Stream.from(1))` -> `Source(LazyList.from(1))` (`KillSwitchDocSpec`), `buffer.toIterator` -> `buffer.iterator` (`GraphStageDocSpec`).
- Tests: `Shrink.withLazyList` instead of `Shrink(_ => Stream.empty)`, `view.mapValues` instead of a `@nowarn` wrapper, `.iterator` instead of `.toIterator`, and plain `Either` equality instead of `.left.get`/`.right.get` (dropping the `@nowarn` on `MetricNumericConverterSpec`).
- Paradox docs: prose and `@apidoc` scala anchors updated from `immutable.Seq` to `Seq` to match the regenerated scaladoc ids (all 25 `Seq`-bearing anchors verified against `stream/doc` output); four anchors that were already broken on `main` were corrected while touching them.
- MiMa: one `IncompatibleSignatureProblem` filter for `Shape.hasOnePort` (`@InternalApi private[pekko]`): erased descriptor is unchanged, only the Java generic signature moved from `Seq<?>` to `Seq<Object>`.

The 2.12 branches in `project/PekkoDisciplinePlugin.scala` and the `filterKeys`/`mapValues` uses under `project/` (sbt still runs on 2.12) are intentionally left unchanged.

### Result
No remaining Scala 2.12 collection idioms in main, test or doc sources.

### Tests
- native `scalafmt` run on all changed files
- `sbt stream/mimaReportBinaryIssues` passes locally with the filter
- `sbt stream/doc` generated and used to verify the paradox anchors
- full `Test/compile` and tests: relying on CI

### References
None - Scala 2.12 support was dropped previously

### DIFF
--- a/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
+++ b/actor/src/main/scala/org/apache/pekko/util/ByteString.scala
@@ -21,8 +21,8 @@ import java.util.Base64
 
 import scala.annotation.{ nowarn, tailrec, varargs }
 import scala.collection.{ immutable, mutable }
-import scala.collection.immutable.{ IndexedSeq, IndexedSeqOps, StrictOptimizedSeqOps, VectorBuilder }
-import scala.collection.mutable.{ Builder, WrappedArray }
+import scala.collection.immutable.{ ArraySeq, IndexedSeq, IndexedSeqOps, StrictOptimizedSeqOps, VectorBuilder }
+import scala.collection.mutable.Builder
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
 
@@ -161,11 +161,6 @@ object ByteString {
 
   /** Java API */
   def createBuilder: ByteStringBuilder = new ByteStringBuilder
-
-  // implicit val canBuildFrom: CanBuildFrom[TraversableOnce[Byte], Byte, ByteString] =
-  //   new CanBuildFrom[TraversableOnce[Byte], Byte, ByteString] {
-  //     override def apply(ignore: TraversableOnce[Byte]): ByteStringBuilder = new ByteStringBuilder
-  //   }
 
   private[pekko] object ByteString1C extends Companion {
     val empty = new ByteString1C(Array.emptyByteArray)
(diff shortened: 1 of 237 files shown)


actor/src/main/scala/org/apache/pekko/util/ByteString.scala
```diff
@@ -3269,9 +3264,11 @@ final class ByteStringBuilder extends Builder[Byte, ByteString] {
 
   override def addAll(xs: IterableOnce[Byte]): this.type = {
     xs match {
-      case bs: ByteString          => addAll(bs)
-      case xs: WrappedArray.ofByte =>
+      case bs: ByteString              => addAll(bs)
+      case xs: mutable.ArraySeq.ofByte =>
         if (xs.nonEmpty) putByteArrayUnsafe(xs.array.clone)
+      case xs: ArraySeq.ofByte =>
```
He-Pin, line 3270: This new immutable.ArraySeq.ofByte fast path is the PR's only behavior-path change and is currently covered only indirectly by the ++= property test. Consider an explicit case (e.g. in ByteStringBuilderScala213PlusSpec): addAll(array.toSeq), then mutate the source array and assert the built ByteString is unchanged, to lock in the clone/no-aliasing contract.

He-Pin (commented):
Triaged all 237 files: the sweep is ~95% immutable.Seq -> Seq (identical type and erasure on 2.13 and 3), with one functional change (the ByteStringBuilder fast path, inline below), one justified MiMa filter (Shape.hasOnePort internal generic signature), and no weakened tests (MetricSpec assertions are actually strengthened). CI is green on both Scala versions. One suggestion inline.

actor/src/main/scala/org/apache/pekko/util/ByteString.scala
pjfanning: Added explicit tests in `ByteStringBuilderScala213PlusSpec` in 7a7ce63 (branch rebased on main): both `immutable.ArraySeq.ofByte` (via `unsafeWrapArray`, which shares the array) and `mutable.ArraySeq.ofByte` are added, the source array is mutated before `result()`, and the built `ByteString` is asserted unchanged. Dropping the `clone` makes the test fail. Empty-array cases are covered too.

PR #3541, opened 2026-09-13
perf: optimise UnsynchronizedByteArrayInputStream bulk operations

pjfanning:
### Motivation
`UnsynchronizedByteArrayInputStream` (used by `ByteString.asInputStream` and the Java/Jackson serializers) only overrode the basic `read`/`skip` methods. `readAllBytes`, `readNBytes`, `skipNBytes` and `transferTo` fell through to the generic `InputStream` defaults, which allocate 16 KiB scratch buffers in a loop and then concatenate — wasteful when the whole payload already sits in a `byte[]`. `skip` also advanced `offset` unconditionally via `Math.addExact(offset, Math.toIntExact(n))`, so a large `n` could throw `ArithmeticException` and `offset` could end up past `eod`.

### Modification
- Override `readAllBytes`, `readNBytes(int)`, `readNBytes(byte[],int,int)`, `skipNBytes` and `transferTo` to operate directly on the backing array (one `Arrays.copyOfRange` / `OutputStream.write` each) — the same set of overrides `java.io.ByteArrayInputStream` has.
- Clamp `skip` to the remaining bytes so `offset <= eod` always holds and large skips cannot overflow; simplify `available()` and `readLocal()` accordingly.
- Simplify the `(data, offset, length)` constructor so `offset`/`eod` are clamped to the array using long arithmetic (no int overflow), and use `Objects.checkFromIndexSize` for the `read(byte[],int,int)` bounds check.
- Expand `UnsynchronizedByteArrayInputStreamSpec` to cover every method and edge case (EOF, zero-length reads, clamping, overflow, negative args).
- Add `readAllBytes` and `transferTo` cases to `ByteString_asInputStream_Benchmark`.

### Result
`readAllBytes`/`readNBytes`/`transferTo` on `ByteString.asInputStream` copy the data once instead of chunking through temporary buffers. Observable behaviour is otherwise unchanged (negative `skip` still throws, as before).

Local JMH (`-f 1 -wi 2 -i 3`, short run so error bars are wide):

| Benchmark | kb | before (ops/s) | after (ops/s) |
|---|---|---|---|
| `single_bs_as_input_stream_read_all_bytes` | 10 | 201,138 | 502,768 |
| `single_bs_as_input_stream_read_all_bytes` | 1000 | 3,307 | 5,460 |
| `single_bs_as_input_stream_transfer_to` | 10 | 238,211 | 267,642 |
| `single_bs_as_input_stream_transfer_to` | 1000 | 2,883 | 2,747 |

`transferTo` is dominated by the `ByteArrayOutputStream` copy on the receiving side, so the difference there is within noise.

### Tests
- `sbt "actor-tests/testOnly org.apache.pekko.util.UnsynchronizedByteArrayInputStreamSpec org.apache.pekko.util.ByteStringSpec"` — 226 passed
- `sbt "actor/mimaReportBinaryIssues"` — no issues (class is `@InternalApi`; only methods added)
- `sbt "bench-jmh/Jmh/compile"` and `Jmh/run` of the new benchmarks
- `scalafmt` on changed Scala files, `sbt actor/javafmtAll` on JDK 17
- `git diff --check`

### References
None - follow-up to #2300 which introduced this class


actor-tests/src/test/scala/org/apache/pekko/io/UnsynchronizedByteArrayInputStreamSpec.scala
```diff
@@ -26,9 +27,13 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
 class UnsynchronizedByteArrayInputStreamSpec extends AnyWordSpec with Matchers {
+
+  private def bytes(s: String): Array[Byte] = s.getBytes(StandardCharsets.UTF_8)
+  private def str(b: Array[Byte]): String = new String(b, StandardCharsets.UTF_8)
```
He-Pin, line 34: Consider adding a differential test that runs randomized operation sequences against both this stream and a ByteArrayInputStream, asserting identical outputs except the two documented deviations (zero-length read at EOF returns 0, not -1; negative skip throws IllegalArgumentException). It is the strongest guard for a serialization-path class.

He-Pin (commented):
Verified the new bulk overrides match java.io.ByteArrayInputStream's JDK 17 behavior and never read past eod, so the shared-array ByteString1C path cannot leak sibling bytes; the skip-clamping and constructor-overflow fixes are correct. One test suggestion inline.

actor-tests/src/test/scala/org/apache/pekko/io/UnsynchronizedByteArrayInputStreamSpec.scala
pjfanning: Added a randomized differential test against `java.io.ByteArrayInputStream` in 7339b14 (rebased on main as well). It compares return values, exceptions, buffer contents and `available()` after every operation, special-casing only the two documented deviations.

PR #3542, opened 2026-09-13
perf: validate actor path elements without String.charAt

pjfanning:
### Motivation
`ActorPath.findInvalidPathElementCharPosition` runs for every named `actorOf`. It scanned the name with `String.charAt` inside a pattern match with guards, plus a `ValidSymbols.indexOf` for every non-alphanumeric character.

`String.charAt` inlines to `isLatin1() ? StringLatin1.charAt : StringUTF16.charAt`, and that branch is profiled once, JVM-wide, in `String.charAt`'s own bytecode. Any non-ASCII string handled anywhere in the process (a Jackson payload, a log line) pollutes that profile, after which C2 compiles both coders into every `charAt` loop. Measured locally, the validator halves in throughput once the profile is polluted.

### Modification
- Copy the element out once with `getBytes(ISO_8859_1)` — an intrinsic array copy for Latin-1 strings — and scan the `byte[]` with a 128-entry flag table (valid-char bit, hex-digit bit) instead of calling `charAt` per character. Characters outside Latin-1 encode as `?` and 0x80–0xFF as negative bytes; both are invalid, so the accepted set and the reported position are unchanged.
- Add directional `ActorPathSpec` tests for accepted names, rejected names and the reported error position, including Latin-1, non-Latin-1 and surrogate-pair input.
- Extend `ActorPathValidationBenchmark` with a `polluted` param that pre-warms `String.charAt` with UTF-16 strings, and keep the previous `charAt`-based validator there as `charAtLoop*` for comparison.

### Result
Name validation no longer depends on the `String.charAt` coder profile. Local JMH (`-f 1`, short run, noisy machine — the trend is what matters):

| Benchmark | polluted=false | polluted=true |
|---|---|---|
| `charAtLoopActor_1` (previous code) | 101 ops/µs | 49 ops/µs |
| `handLoopActor_1` (this PR) | 139 ops/µs | 162 ops/µs |
| `charAtLoop7000` (previous code) | 0.197 ops/µs | 0.107 ops/µs |
| `handLoop7000` (this PR) | 0.176 ops/µs | 0.142 ops/µs |

I tried the same treatment on the other hot `charAt` users, `MurmurHash.stringHash` and `Helpers.base64`, and measured no benefit (hash arithmetic and `StringBuilder.append` dominate; `toCharArray` only adds an allocation), so those are left as they are.

### Tests
- `sbt "actor-tests/testOnly org.apache.pekko.actor.ActorPathSpec org.apache.pekko.actor.LocalActorRefProviderSpec"` — 25 passed
- `sbt "actor-tests/testOnly org.apache.pekko.routing.ConsistentHashingRouterSpec"` — passed
- `sbt "actor/mimaReportBinaryIssues"` — no issues
- `sbt "bench-jmh/Jmh/compile"` and `Jmh/run` of `ActorPathValidationBenchmark`
- `scalafmt` on changed Scala files, `git diff --check`

### References
None - performance follow-up to the hand-written actor name validator


bench-jmh/src/main/scala/org/apache/pekko/actor/ActorPathValidationBenchmark.scala
```diff
@@ -32,6 +34,21 @@ import org.openjdk.jmh.annotations.Warmup
...
+their loop; the byte-table scanner (isValidPathElement) is unaffected:
+
+[info] Benchmark                                                  (polluted)   Mode  Cnt    Score    Error   Units
+[info] ActorPathValidationBenchmark.charAtLoop7000                     false  thrpt    5    0.197 ±  0.028  ops/us
+[info] ActorPathValidationBenchmark.charAtLoop7000                      true  thrpt    5    0.107 ±  0.017  ops/us
+[info] ActorPathValidationBenchmark.charAtLoopActor_1                  false  thrpt    5  100.954 ± 14.868  ops/us
+[info] ActorPathValidationBenchmark.charAtLoopActor_1                   true  thrpt    5   49.306 ±  8.706  ops/us
+[info] ActorPathValidationBenchmark.handLoop7000                       false  thrpt    5    0.176 ±  0.012  ops/us
+[info] ActorPathValidationBenchmark.handLoop7000                        true  thrpt    5    0.142 ±  0.094  ops/us
+[info] ActorPathValidationBenchmark.handLoopActor_1                    false  thrpt    5  138.841 ± 111.054 ops/us
```
He-Pin: The handLoopActor_1 confidence intervals here (+/-111 and +/-281 ops/us) are wider than the scores themselves, so this run cannot support the unpolluted speedup claim for short names. Could you rerun with more forks/iterations (e.g. -f 3 -i 10) before merge?

actor-tests/src/test/scala/org/apache/pekko/actor/ActorPathSpec.scala
```diff
@@ -81,6 +82,57 @@ class ActorPathSpec extends AnyWordSpec with Matchers {
         "must not be empty")
     }
 
+    "accept valid path elements" in {
```
He-Pin, line 85: Consider adding an exhaustive sweep over all 65,536 chars asserting isValidPathElement matches an independently computed predicate; it runs in milliseconds and would permanently lock in equivalence with the previous charAt implementation.

He-Pin (commented):
Verified the new byte-table validator is semantically equivalent to the old charAt scan (exhaustive differential check across all 65,536 code units, 0 divergences) and binary-compatible (only a private method body changed). Two suggestions inline.

actor-tests/src/test/scala/org/apache/pekko/actor/ActorPathSpec.scala
pjfanning: Added an exhaustive sweep over all 65,536 chars in 5199810 (branch rebased on main too). It checks `isValidPathElement` against an independently defined valid-char/hex-digit predicate for each char at position 0, after another char, and in both hex positions of a `%XX` escape. Runs in under a second.

bench-jmh/src/main/scala/org/apache/pekko/actor/ActorPathValidationBenchmark.scala
pjfanning: I need to come back to this - I may have access to another machine later this week where I can run benchmarks with less noise

pjfanning: Reran with `-f 3 -wi 5 -i 10` (2s iterations) and updated the benchmark comment and PR description:

| Benchmark | polluted=false | polluted=true |
|---|---|---|
| `charAtLoopActor_1` (previous code) | 108.3 ± 4.3 ops/µs | 41.8 ± 0.5 ops/µs |
| `handLoopActor_1` (this PR) | 161.0 ± 6.8 ops/µs | 162.0 ± 3.5 ops/µs |
| `charAtLoop7000` (previous code) | 0.152 ± 0.003 ops/µs | 0.086 ± 0.004 ops/µs |
| `handLoop7000` (this PR) | 0.101 ± 0.014 ops/µs | 0.096 ± 0.007 ops/µs |

The short-name speedup holds with tight intervals (~1.5× unpolluted, ~3.9× polluted). The longer run also shows that for the 7000-char input the byte-table version is ~34% slower than `charAt` when unpolluted (the `getBytes` copy dominates) and roughly equal when polluted; the earlier noisy run hid that, and the description now says so. Actor names are short in practice, so I think the trade-off is worth it.

PR #3546, opened 2026-09-17
build: apply the Develocity Common Custom User Data sbt plugin

clayburn:
### Motivation
Pekko publishes Build Scans to develocity.apache.org, but does not apply the [Common Custom User Data (CCUD) sbt plugin](https://github.com/gradle/common-custom-user-data-sbt-plugin). Without it, scans lack the shared tags, links and custom values that make scans searchable and comparable across builds:

- a tag for the operating system
- a tag for how the build was invoked (IDE or command line)
- a tag for CI builds, plus tags, links and custom values specific to the CI provider
- for CI builds of a Git repository, the commit id, branch name, status, and whether the checkout was dirty

### Modification
Add `sbt-develocity-common-custom-user-data` 1.5.0 to `project/plugins.sbt`, next to the existing `sbt-develocity` plugin.

No change to `PekkoDevelocityPlugin` is needed. CCUD 1.5.0 registers itself through `internalDevelocityConfigurationTransformers`, a key introduced in sbt-develocity 1.4.5 — the version Pekko already uses. On that path CCUD composes *after* the build's own `develocityConfiguration` rather than redefining the setting, so Pekko's server URL, publishing predicate, IP obfuscation, disabled local build cache and project id are all left intact. (On sbt-develocity older than 1.4.5 CCUD would instead redefine `develocityConfiguration`, which is the case worth worrying about here — Pekko is past it.)

The plugin is released by Gradle under the Apache 2.0 licence.

### Result
Build Scans published from Pekko builds carry the standard CCUD tags, links and custom values.

### References
https://github.com/gradle/common-custom-user-data-sbt-plugin


pjfanning:
We don't even use the existing develocity results. It is probably better just to take all of this stuff out of Pekko.

He-Pin (commented):
The plugin addition itself looks correct and license-compatible (Apache-2.0, build-time only, scans go to the ASF Develocity server). One directional concern before merge: pjfanning's comment on this PR questions whether Pekko should keep the Develocity integration at all. Could a committer confirm the project still wants build-scan enrichment before this merges?

pjfanning (approved):

pjfanning:
Merged for now but I never have found https://develocity.apache.org/ to give me info that I need. GitHub CI usually has what I need and if something is broken, running local builds fills the gaps.

PR #3547, opened 2026-09-18
docs: remove unused warnOldDocs.js and projectSpecificFooter template

pjfanning:
### Motivation
`docs/src/main/paradox/assets/js/warnOldDocs.js` is the Akka-era "you are reading an old version of the docs" banner script. It is dead code:

- Its only reference is `docs/src/main/paradox/_template/projectSpecificFooter.st`, a footer hook from Akka's own Paradox theme. The `pekko-theme-paradox` `page.st`/`footer.st` templates never include `projectSpecificFooter`, so no generated page loads the script. The published site confirms this: `https://pekko.apache.org/docs/pekko/current/index.html` only loads the theme's scripts, though the unused `.js` file is still copied into the site as a static asset.
- Even if loaded, it fetches `//pekko.io/versions.json`, which does not exist (`pekko.io` does not resolve; `pekko.apache.org/versions.json` is a 404), so the banner logic is a no-op.
- Its URL parsing assumes Akka's `<project>/<version>/` path layout and does not match `/docs/pekko/<version>/`.

None of the other Pekko repos (pekko-http, pekko-connectors, pekko-grpc, pekko-persistence-jdbc, pekko-management) carry this file.

### Modification
Delete `warnOldDocs.js` and `_template/projectSpecificFooter.st` (both directories are now empty and removed).

### Result
Unused docs assets are gone. Generated docs are unchanged apart from no longer shipping the unused script.

### Tests
- Not run locally - docs only; relying on the CI docs build.

### References
None - cleanup of unused Akka-era docs assets.


raboof (approved):
a bit sad (we're still publishing old versions of the docs, e.g. https://pekko.apache.org/docs/pekko/1.6.0/ , and it's nice to point people to the latest version), but better to clean it up (and potentially resurrect later) than to leave it broken

